PR #4, opened 2018-05-07
[WIP] Make container image creation maintainable

rmohr:
* switch to bazel
* distroless static-qemu, to allow easily mix-and-match of container layers on updates


rmohr:
@cynepco3hahue FYI

rmohr:
It is  one of the small steps which will allow us mix-and-match updates in our containers. E.g. just update qemu, leave the layers with cluster disk untouched. If every layer is basically dependency free things get much simpler.

PR #9, opened 2018-05-16
Logging

petrkotas:
This PR add centrilized logging to the CI infastrucure.

Please review @rmohr @cynepco3hahue

Add logs:
- [x] auditd
- [x] journal
- [x] kubelet
- [x] etcd
- [x] kubernetes logs (kubectl logs)

Filter logs:
- [x] define what to filter out

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirtci/9)


fluentd/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM fluent/fluentd:v1.1.3-onbuild
+MAINTAINER Petr Kotas <petr.kotas@gmail.com>
+
+EXPOSE 24284/udp
```
cynepco3hahue: Where do you use this port?

petrkotas: Ha good catch. This is typo that somehow slipped here.

petrkotas:
@cynepco3hahue It logs both syslog and auditd. 
I can either configure the collecting from the file or let the kubernetes push to the syslog directly.

What exactly do we want to collect?

fluentd/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM fluent/fluentd:v1.1.3-onbuild
+MAINTAINER Petr Kotas <petr.kotas@gmail.com>
+
+EXPOSE 24224
```
cynepco3hahue: I see `fluent/fluentd:v1.1.3-onbuild` already expose this port
https://github.com/fluent/fluentd-docker-image/blob/94ab981b056210cdd331413215724445faf6ee71/v1.1/alpine-onbuild/Dockerfile#L54
so no need for separate custom `Dockerfile`

cynepco3hahue:
In general, I want to collect all container logs(that we have from `kubectl logs` command), in additional logs from relevant services
- k8s: `kubelet, docker`
- openshift: `origin-master-api, origin-master-controllers, origin-node`
maybe also `etcd` logs

petrkotas:
Fine with me, these logs are useful. On it

base/scripts/logging.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+echo "*.* @@192.168.66.2:24224" >> /etc/rsyslog.conf
+
```
rmohr: I would prefer if we simply skip logging configuration on the nodes completely in this repo. I would prefer if the consumer of our gocli infra and our clusters can tweak the log-collection in the dev-env or on CI to their needs with e.g. a daemonset: https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/fluentd-daemonset-syslog.yaml

gocli/cmd/run.go
```diff
@@ -241,6 +240,42 @@ func run(cmd *cobra.Command, args []string) (err error) {
...
+		}
+
+		// Pull the fluent-bit image
+		reader, err = cli.ImagePull(ctx, "docker.io/pkotas/fluentd", types.ImagePullOptions{})
+		if err != nil {
+			panic(err)
+		}
+		docker.PrintProgress(reader, os.Stdout)
+
+		// Start the ganesha image
```
rmohr: outdated comment.

cli/cli
```diff
@@ -181,6 +181,7 @@ if [ "$COMMAND" == "provision" ] ; then
   docker cp ${SCRIPTS}/. ${VM_CID}:/scripts/
   docker exec ${VM_CID} /bin/bash -c "while [ ! -f /usr/local/bin/ssh.sh ] ; do sleep 1; done"
   docker exec ${VM_CID} /bin/bash -c "ssh.sh sudo /bin/bash -s ${CRIO} < /scripts/provision.sh"
+  docker exec ${VM_CID} /bin/bash -c "ssh.sh sudo /bin/bash < /logging.sh"
```
rmohr: As said above, I would prefer if we don't do anything regarding to logging here, I would want it to be an explicit decision by the users. For instance if we don't expose a log-dir, we may pile up gigabytes of logs quickly and no one knows what is going on.

rmohr: > The solution I think could work, is to provide after run deployment of a fluent daemon set, which would be default, but can be easily overridden.

@petrkotas ok, let's see where that leads us.

petrkotas:
@rmohr the general idea is here. I am still missing one moving bit (the logs are sent broken), but this one is farily simple to fix, once found :).

What do you think?

petrkotas:
Fixed a logging to the degree that it send a logs. The logs are still a little broken. However now the logging has access to the cluster. Next moving bit done.

petrkotas:
I have hit the dead end. Deploying the logging in a form of daemonset will not work. This solution is challenging the chicken egg problem. Do you first setup cluster logging deployed on the cluster, or cluster and start logging on to the logging deployed on the cluster. This simply will not work. 

I have created a few workarounds that solve a few issues:
* added a systemd-journal group to the containers so it can access the cluster
* created a persistent journal so the daemonset can access it properly (looking inside /var/run/journal did not worked)
* changed the default group permissions for /var/lib/containers/ to allow non root read, this does not work on cluster setup. The dir is not there, so it cannot be modified.
* I have also modified the origin daemonset docker images to work properly. The original ones, were using old versions of configuration files, and were not working properly.
* I have modified the yaml deployment to provide functional RBAC roles. Without these roles, the daemonset cannot access the underlying host properly.
* Also I have changed the setup from fluentd -> rsyslog to fluentd -> fluentd. Redirecting fluentd to rsyslog is hitting the known drawback of syslog protocol. It does not work with long complicated logs formatted in json. We are collecting logs in json format from kubernetes and its pods. The logs are malformed and solving this will result in more work than benefit.

Next step is to create host deploy and do the setup that way.

rmohr:
> Also I have changed the setup from fluentd -> rsyslog to fluentd -> fluentd. Redirecting fluentd to rsyslog is hitting the known drawback of syslog protocol. It does not work with long complicated logs formatted in json. We are collecting logs in json format from kubernetes and its pods. The logs are malformed and solving this will result in more work than benefit.

You will have to  use json-over-syslog for that (cee extensions).

> Next step is to create host deploy and do the setup that way.

How does that improve the "malformed" logs situation?

> First setup cluster logging deployed on the cluster, or cluster and start logging onto the logging deployed on the cluster. This simply will not work.

Can you elaborate that?

> changed the default group permissions for /var/lib/containers/ to allow non root read, this does not work on cluster setup. The dir is not there, so it cannot be modified.

Why can't you run the daemonset as root?

> created a persistent journal so the daemonset can access it properly (looking inside /var/run/journal did not worked)

Which logs are not going to files there and are only in the journal?

petrkotas:
I have given a json-over-syslog a thought earlier, but in my opinion it adds extra layer that can break.

Deploying on the host does not solve the malformed logs. It solves the access issue. It would no longer require doing workarounds.

The chicken egg problem goes like this:
- When you install first kubernetes and than logging, than you have to solve the issue of linking the logging to the cluster. Since you want to log the cluster. This can be done by tailing the docker logs. However, this requires to change the container access to became root, which is not default in the container. And for a good reason, the security than goes away. You can also change the logging driver to fluentd in kubernetes. However, than Kubernetes have to be restarted. Not a good idea.
- When you try to install first the logging onto the cluster, you do not have the cluster.

The kubelet and audit logs are stored in the journal. The fluentd-systemd plugin can parse the data. Unfortunately when I experimented with it, it did not worked when for `/var/run/journal`. It worked only for `/var/log/journal`.

Overall, the logging infrastructure deployed on the host as advised in official [kubernetes doc](https://kubernetes.io/docs/concepts/cluster-administration/logging/) is to deploy logging agent on the node to log cluster. IT provides more robustness, can be run as root and does not require any workarounds.

What do you think of it? Can I proceed?

rmohr:
> When you install first kubernetes and than logging, than you have to solve the issue of linking the logging to the cluster. Since you want to log the cluster. This can be done by tailing the docker logs. However, this requires to change the container access to became root, which is not default in the container. And for a good reason, the security than goes away. You can also change the logging driver to fluentd in kubernetes. However, than Kubernetes have to be restarted. Not a good idea.

Don't worry about privileges here. Being privileged is perfectly fine for infra pods in our develop- and CI clusters.

> Overall, the logging infrastructure deployed on the host as advised in official kubernetes doc is to deploy logging agent on the node to log cluster.

All examples there run the agent in a pod. I like the approach where they supply the fluentd config for the logging location via a ConfigMap (https://github.com/kubernetes/website/blob/master/content/en/docs/concepts/cluster-administration/fluentd-sidecar-config.yaml, you can use the DownwardsAPI). That is the pattern I would also want to adopt. Then you have the easy experience which you also mentioned. By default we log the most relevant things (but not too much), but you can change the config map and it will log exactly what you want.

> Deploying on the host does not solve the malformed logs. It solves the access issue. It would no longer require doing workarounds.

We can solve the "malformed" logs issue later on.

> When you try to install first the logging onto the cluster, you do not have the cluster.

That does not concern me that much. If the cluster never becomes ready, we can't do anything with it anyway. If all components get so much ready that the logger manifests start working we will be informed about issues.

> The fluentd-systemd plugin can parse the data. Unfortunately when I experimented with it, it did not worked when for /var/run/journal. It worked only for /var/log/journal.

Regarding the persistent journal: Ensuring that we have a persistent journal, that can be the part which belongs to the provisioner step.

Regarding the issue with the location of the journal: Are you sure that you set up everything the right way?

rmohr:
> The kubelet and audit logs are stored in the journal. The fluentd-systemd plugin can parse the data. Unfortunately when I experimented with it, it did not worked when for /var/run/journal. It worked only for /var/log/journal.

Btw. why do you have to customize the fluentd image? Can't you simply use e.g. https://github.com/fluent/fluentd-kubernetes-daemonset/tree/05509df37d1b90499f9bc7fbe792908d64ba1c76/docker-image/v0.12/debian-syslog?

It already contains the journal plugin. We should be able to point it to our own fluentd config files by overriding FLUENTD_CONF.

cli/cli
```diff
@@ -233,6 +234,9 @@ else
     elif docker exec ${VM_CID} /bin/bash -c "test -f /scripts/nodes.sh" ; then
       docker exec ${VM_CID} /bin/bash -c "ssh.sh sudo /bin/bash < /scripts/nodes.sh"
     fi
+    if docker exec ${VM_CID} /bin/bash -c "test -f /scripts/logging.sh" ; then
```
cynepco3hahue: no need we use `gocli` to run containers, so I think this part of `cli` will be removed soon

fluentd/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM fluent/fluentd:v1.2.1-onbuild
```
cynepco3hahue: Why do you not use the image from https://github.com/cynepco3hahue/fluentd-docker-image/blob/master/v1.2?

base/scripts/logging.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+echo "*.* @@192.168.66.2:24224" >> /etc/rsyslog.conf
```
cynepco3hahue: IIUC you pass all logs via fluent, so you do not need this line

cli/cli
```diff
@@ -233,6 +234,9 @@ else
...
+      docker exec ${VM_CID} /bin/bash -c "ssh.sh sudo /bin/bash < /logging.sh"
```
cynepco3hahue: and you do not use the correct file I believe it must be `/scripts/logging.sh`

k8s-1.9.3/scripts/logging.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
```
cynepco3hahue: I think you first need to export kubeconfig directory

fluentd/Dockerfile
rmohr: Why can't we still just use the official ones?

base/scripts/logging.sh
petrkotas: Ah, this is early forgotten artefact. Thanks :)

k8s-1.9.3/scripts/provision.sh
```diff
@@ -36,7 +36,20 @@ yum install -y docker
...
+     "fluentd-address": "fluentdhost:24224"
+   }
+}' 
+EOF > /etc/docker/daemon.json
+
+# Enable the permanent logging
+mkdir -p /var/log/journal
+
+# Add vagrant user to journal group, so he can read it and process it
+usermod -a -G systemd-journal vagrant
```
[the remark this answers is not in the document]
petrkotas: No longer needed, it is WIP artifact. Can be removed.

k8s-1.9.3/scripts/provision.sh
```diff
@@ -36,7 +36,20 @@ yum install -y docker
 
 # Log to json files instead of journald
 sed -i 's/--log-driver=journald //g' /etc/sysconfig/docker
-echo '{ "insecure-registries" : ["registry:5000"] }' > /etc/docker/daemon.json
+echo <<EOF '{ 
+  "insecure-registries" : ["registry:5000"],
+  "log-driver": "fluentd",
```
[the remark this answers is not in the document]
petrkotas: Again WIP artifact. No longer needed. Remove it :)

petrkotas:
The custom image is used since the official support is only for: elasticsearch, gce, papertrail or syslog.
This design is using fluentd -> fluentd approach which does not have pre-existing image available.

The benefit of fluentd -> fluentd is in direct message transport. The messages does not need to be converted between the systems.

gocli/cmd/run.go
```diff
@@ -99,6 +102,9 @@ func run(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
+	log_dir, err := cmd.Flags().GetString("log-dir")
```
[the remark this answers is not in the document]
petrkotas: ok will rename it

fluentd/Dockerfile
petrkotas: I prefer using the official image and customize the configuration as needed.

petrkotas:
I have pushed reactions to the comments.

petrkotas:
@rmohr @cynepco3hahue There is one more thing I do inside of the docker image I use. The image is
stripped from the vagrant/fluentd user. It uses root.
The official image uses `exec gosu fluentd` to switch to user `fluentd` which this image cannot use.
The fluentd requires access to the system logs and therefore requires root access.

fluentd-daemonset/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM fluent/fluentd:v1.2.1-debian-onbuild
+MAINTAINER Petr Kotas <petr.kotas@gmail.com>
+
+RUN buildDeps="sudo make gcc g++ libc-dev ruby-dev" \
+ && apt-get update \
+ && apt-get install -y --no-install-recommends $buildDeps \
+ && sudo gem install \
+        fluent-plugin-secure-forward \
```
rmohr: Secure forward should work out of the box, no need for installing it.

fluentd-daemonset/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+        fluent-plugin-record-reformer \
```
rmohr: Where is that needed?

fluentd-daemonset/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+        fluent-plugin-kubernetes_metadata_filter \
+        ffi \
+        fluent-plugin-systemd \
```
rmohr: This is installed in all images in the fluentd repo which I checked. E.g.: https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/docker-image/v1.2/debian-syslog/Gemfile

fluentd-daemonset/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM fluent/fluentd:v1.2.1-debian-onbuild
+MAINTAINER Petr Kotas <petr.kotas@gmail.com>
+
+RUN buildDeps="sudo make gcc g++ libc-dev ruby-dev" \
+ && apt-get update \
+ && apt-get install -y --no-install-recommends $buildDeps \
+ && sudo gem install \
+        fluent-plugin-secure-forward \
+        fluent-plugin-record-reformer \
+        fluent-plugin-kubernetes_metadata_filter \
```
rmohr: This is installed in all images in the fluentd repo which I checked. E.g.: https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/docker-image/v1.2/debian-syslog/Gemfile

fluentd-daemonset/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+        ffi \
```
rmohr: This is installed in all images in the fluentd repo which I checked. E.g.: https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/docker-image/v1.2/debian-syslog/Gemfile

fluentd-daemonset/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ && sudo gem sources --clear-all \
+ && SUDO_FORCE_REMOVE=yes \
+    apt-get purge -y --auto-remove \
+                  -o APT::AutoRemove::RecommendsImportant=false \
+                  $buildDeps \
+ && rm -rf /var/lib/apt/lists/* \
+           /home/fluent/.gem/ruby/2.3.0/cache/*.gem
+
+
+COPY kubernetes.conf /fluentd/etc/kubernetes.conf
```
rmohr: You can provide the configs via a configMap and the downstream API.

fluentd/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+MAINTAINER Petr Kotas <petr.kotas@gmail.com>
+
+EXPOSE 24224/udp
+
+CMD exec fluentd -c /fluentd/etc/${FLUENTD_CONF} -p /fluentd/plugins $FLUENTD_OPT -v
```
rmohr: This line seems to support the same environment variables like `fluent/fluentd:v1.2.1-onbuild` itself. Why do we need this extra Dockerfile?

rmohr:
> @rmohr @cynepco3hahue There is one more thing I do inside of the docker image I use. The image is
> stripped from the vagrant/fluentd user. It uses root.

I see two different behavours here: 
 * The entrypoint in the daemonsets brings it own entrypoint.sh. This one does not seem to change the user: https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/docker-image/v1.2/debian-syslog/entrypoint.sh
 * The entrypoint in https://github.com/fluent/fluentd-docker-image/blob/master/v1.2/debian/entrypoint.sh which you use as a receiver. There the entrypoint changes to the fluentd user, but simply overriding the entrypoint with the command `exec fluentd "$@"` should be good enough, if you need to run as root here too.

slintes:
Hi, I wasn't aware of this WIP, and added some logging "manually" to my local cluster. I wanted to add a document about it, but I guess that's not needed, but I'll share it here. At least for the logs of the kubevirt pods I didn't need to change anything in the cluster itself. See https://gist.github.com/slintes/63de873cfda2d9610622946b91bea5fd

fabiand:
@slintes thanks for sharing - because you thought about the same topic I just added you as a reviewer :)

petrkotas:
@rmohr @slintes  @cynepco3hahue @booxter I have finalized the pr. Please do a round of review.
I tried to clarify as much as i can inside the pr and tried hard to not forget about any comment.
Also I no longer this pr is wip.

fluentd-daemonset/Dockerfile
rmohr: @petrkotas see here: https://github.com/tagomoris/fluent-plugin-secure-forward on why it is not needed.

fabiand:
@petrkotas is this fixing https://github.com/kubevirt/kubevirt/issues/865 somewhat?

petrkotas:
@fabiand Hi, no it is prerequisite to that. I will build on the logs to detect the denials.

petrkotas:
@rmohr Oh yes, I tried to answer inside the [PR](https://github.com/kubevirt/kubevirtci/pull/9/files#diff-b81809a9a78e2f9fff2c669e07253554).

So to recap. 
 
- I am dropping everything that is not required
- I am preparing the grounds for additional plugins required for logging auditd (currently it works to some degree, but I am not sure if it is enough)
- Also the dockerfile opens udp port. The forwarding was not working properly without it.
- The entrypoint, sure I can override it in the Dockerfile. 
- The secure forward, Thanks for pointing it, will remove it.

So basically I am building the custom image to drop the non-needed parts and to have more control on the image.

@rmohr did I answered all the questions? Sorry if I have missed some of them. It is not intentional. The github keeps hiding them so they are easier to miss.

logging/fluentd-daemonset/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN buildDeps="sudo make gcc g++ libc-dev ruby-dev" \
+ && apt-get update \
+ && apt-get install -y --no-install-recommends $buildDeps \
+ && sudo gem install \
+        fluent-plugin-secure-forward \
+        fluent-plugin-record-reformer \
+        fluent-plugin-kubernetes_metadata_filter \
+        ffi \
+        fluent-plugin-systemd \
+        fluent-plugin-record-reformer \
```
slintes: - Roman mentioned this earlier: what is `fluent-plugin-secure-forward` for? Seems it's not used, I don't see configuration for it. And I think we just don't need it, why would we want to encrypt logs in dev/ci?
- `fluent-plugin-record-reformer` also isn't used (and it's mentioned twice)
- the other plugins are in the `fluent/fluentd-kubernetes-daemonset:v1.2-debian-syslog` docker image, can't we just use that one and override config?

logging/fluentd/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM fluent/fluentd:v1.2.1-onbuild
+MAINTAINER Petr Kotas <petr.kotas@gmail.com>
+
+# Needed for the forwaring to work properly
+EXPOSE 24224/udp
```
slintes: Is `udp` only reason for our own Docker image? If so, we can probably get around this by setting `heartbeat_type` config of the forward plugin to tcp, see https://docs.fluentd.org/v1.0/articles/out_forward#heartbeat_type

gocli/cmd/run.go
```diff
@@ -41,6 +41,8 @@ func NewRunCommand() *cobra.Command {
 	run.Flags().Uint("k8s-port", 0, "port on localhost for the k8s cluster")
 	run.Flags().Uint("ssh-port", 0, "port on localhost for ssh server")
 	run.Flags().String("nfs-data", "", "path to data which should be exposed via nfs to the nodes")
+	run.Flags().String("log-dir", "", "directory where to store the aggregated logs, deploys the fluent endpoint")
+	run.Flags().Bool("deploy-logging", false, "deploy the default fluent logging onto the cluster")
```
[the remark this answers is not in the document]
petrkotas: The custom logging is designed to allow overriding. This means, you can deploy aggregated endpoint (log-dir) and have your logging deployed by your way in the test cluster. 
What I can do. Let the deploy logging automaticaly deploy the aggregated endpoint and let it default to `logs` subfolder. If the log-dir is present, the subfolder will be set accordingly.

@slintes what do you think?

cli/cli
```diff
@@ -183,6 +183,7 @@ if [ "$COMMAND" == "provision" ] ; then
   docker cp ${SCRIPTS}/. ${VM_CID}:/scripts/
   docker exec ${VM_CID} /bin/bash -c "while [ ! -f /usr/local/bin/ssh.sh ] ; do sleep 1; done"
   docker exec ${VM_CID} /bin/bash -c "ssh.sh sudo version=${K8S_VERSION} /bin/bash -s ${CRIO} < /scripts/provision.sh"
+  docker exec ${VM_CID} /bin/bash -c "ssh.sh sudo /bin/bash < /scripts/logging.sh"
```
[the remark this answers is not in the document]
petrkotas: Yes, good catch. This somehow survived the refactoring. Removing it.

logging/fluentd-daemonset/Dockerfile
petrkotas: Yep, we can remove the forwarder. Doing that.

As for the `fluent/fluentd-kubernetes-daemonset:v1.2-debian-syslog`, I double checked and there really is no need to use custom build.  Will remove it, use the default image and override configs.

logging/fluentd/Dockerfile
petrkotas: Yes, the forward was sending broken messages without the udp exposed. Thanks for pointing the heartbeat will check on that.

gocli/cmd/run.go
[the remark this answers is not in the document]
slintes: I don't see the point in "have your logging deployed by your way in the cluster". Is there a specific usecase? @rmohr Do you agree that we always want to get the complete logging setup?
I would also be happy with `log-dir` defaulting to `logs`.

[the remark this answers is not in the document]
rmohr: Onifying to one option is the way to go in my oppinion. Let's use `log-dir` here. We can then also add other flags like `graylog`, if engineers want a nice UI during development.

Maybe `--log-to-dir` might be better?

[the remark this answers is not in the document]
rmohr: The flags would then work this way:

 * No logging specific option specified: no logging
 * --log-to-dir will log to a directory on your machine
 * --log-to-graylog would log to a graylog instance which we can spawn in a container
...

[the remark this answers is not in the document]
slintes: good idea!

petrkotas:
@slintes @rmohr updated the PR, would you mind doing review?

slintes:
logging/fluentd/Dockerfile can be deleted now, right?
other than that lgtm

petrkotas:
@slintes not yet. I have not moved the configuration for `logging/fluentd/` yet. But yes that is the goal.

petrkotas:
@slintes @rmohr Asking for brainstorm. There are two possibilities to put the config into the container.

1. Currently used custom image with our config. Pro: zero config, just work. Con: Custom image
2. Mount volume. Pro: No custom image. Con: It will require aditional param to provide the folder with config. I can hardcode the path, making it sensible to gocli location.

My take is to leave the custom image, solution 1.

What do you think?

rmohr:
> Mount volume. Pro: No custom image. Con: It will require aditional param to provide the folder with config. I can hardcode the path, making it sensible to gocli location.

@petrkotas maybe it helps you if you override the manifest fluentd command with `exec fluentd -c /fluentd/etc/${FLUENTD_CONF} -p /fluentd/plugins $FLUENTD_OPT -v`? Then you know exactly where to look for the configs. The you can use option 1.

slintes:
not sure if I understand the problem tbh... is the question how to get `fluent.conf` etc. into the fluentd container? You can use a configmap and map it into `/fluentd/etc/`, it should be picked up automatically then. See https://github.com/fluent/fluentd-docker-image/blob/master/v1.2/debian/Dockerfile#L74

logging/fluentd/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM fluent/fluentd:v1.2.1-onbuild
+MAINTAINER Petr Kotas <petr.kotas@gmail.com>
+
+# This image is here to distribute the config
+# other option is to provide config via mount volume
+# which will make gocli either have:
+#  - one more option ie. `agregator-conf-dir`
+#  - will have the config path hard coded thus resulting
+#    gocli being dependent of particular folder structure
+
```
petrkotas: @slintes @rmohr I am talking about this particular config file. I have provided explanation here.
I vouch for keeping the custom image. It results in much simpler setup and code.

slintes: what about a ConfigMap with the config file(s)? No need for a hardcoded or configured host path then, and no custom image :)

slintes: ups, too late ;)

petrkotas: @rmohr @slintes sorry for the delay in answer. This fluentd is created [here](https://github.com/kubevirt/kubevirtci/pull/9/files#diff-45eddaf18034ddf5a593d826b8b9c1a7R268). I am fairly sure, I can not use the ConfigMap here. This side is outside of Kubernetes cluster.

Does this add more sense to my [question](https://github.com/kubevirt/kubevirtci/pull/9#issuecomment-406192832)?

slintes: Oh sorry! Totally forgot that this container isn't part of the k8s cluster. In that case I'm fine with a custom image. (actually I'd prefer to put this into the cluster, but I guess we would miss some logs you're interested in then, right?)

cynepco3hahue:
@rmohr Do you other objections or can we merge it?

gocli/cmd/run.go
```diff
@@ -241,6 +247,46 @@ func run(cmd *cobra.Command, args []string) (err error) {
...
+		// Pull the fluent image
+		reader, err = cli.ImagePull(ctx, "docker.io/pkotas/fluentd", types.ImagePullOptions{})
+		if err != nil {
+			panic(err)
+		}
+		docker.PrintProgress(reader, os.Stdout)
+
+		// Start the fluent image
+		fluentd, err := cli.ContainerCreate(ctx, &container.Config{
+			Image: "pkotas/fluentd",
```
rmohr: We can't merge it taken from here.

logging/fluentd/Dockerfile
rmohr: @petrkotas but you are not copying in anything. This seems to be really only about the `CMD`? You can override that always. exactly with that line when starting the container.

logging/fluentd/fluent.conf
```diff
@@ -0,0 +1,14 @@
+<system>
```
rmohr: Was your goal to copy that into the custom fluentd container? I guess a `CP` is missing then in the Dockerfile.

logging/fluentd/Dockerfile
rmohr: Could you have a look at the `-i` or `--inline-config` parameter for fluentd? That seems to be exactly what we want. You can then supply any config (like the config file below) via the commandline.

gocli/cmd/run.go
```diff
@@ -328,6 +374,22 @@ func run(cmd *cobra.Command, args []string) (err error) {
 		}(node.ID)
 	}
 
+	// If logging is enabled, deploy the default fluent logging
+	if logDir != "" {
+		nodeName := nodeNameFromIndex(1)
+		success, err := docker.Exec(cli, nodeContainer(prefix, nodeName), []string{
+			"/bin/bash",
+			"-c",
+			"ssh.sh sudo /bin/bash < /scripts/logging.sh",
```
rmohr: Ok, if we do it this way, we would not have precashed fluentd containers in the image. So far we tried to have everything ready to start. Maybe you could put the manifests in the same directlry like  @awels in #26.

petrkotas:
Okay, on it.

logging/fluentd/Dockerfile
petrkotas: I have looked at the parameter, unfortunately it does not solve the issue. It just moves it on different location.
If I run the fluentd like this
```
fluentd -i "<system>\n log_level debug\n</system>\n<source>\n@type  forward\n@log_level error\nport  24224\n</source>\n<match **>\n@type file\npath /fluentd/log/collected\n</match>" -p /fluentd/plugins $FLUENTD_OPT -v
```
I still have to build the custom docker image to host the `CMD`. 

There is a hack I can do. I can create custom entrypoint script that will build the parameters from the environment variables and than run the fluentd command. Something like
```
#!/usr/bin/env bash

fluentd -i "${MY__FLUENTD_CONFIG}" -p /fluentd/plugins $FLUENTD_OPT -v
```
However, even in this case, I still have to build custom image to house the changes.

Therefore, I vouch for the current custom image to stay as the simplest solution.

petrkotas:
@rmohr I figured out what you meant by overriding the `CMD`, sorry for the delay. Takeaway is go home when you are tired. 
I will override the `CMD` in the execution from the code.

petrkotas:
@rmohr I have overridden the `CMD`. Now the config is hard-wired in the go code. I do not agree with the idea of having the config hard-coded in the code, but I can understand the reason.
Just note, that any change in config will be followed by change in code base.

What do you think of the PR now?

petrkotas:
@rmohr @cynepco3hahue @slintes Ping? Can we proceed with this PR if you agree?

gocli/cmd/run.go
```diff
@@ -241,6 +248,51 @@ func run(cmd *cobra.Command, args []string) (err error) {
...
+		if err != nil {
+			return err
+		}
+
+		if _, err = os.Stat(logDir); os.IsNotExist(err) {
+			os.Mkdir(logDir, 0755)
+		}
+
+		// Pull the fluent image
+		reader, err = cli.ImagePull(ctx, "docker.io/fluent/fluentd:v1.2-debian", types.ImagePullOptions{})
```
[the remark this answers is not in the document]
petrkotas: The same as comment above.

k8s-multus/scripts/provision.sh
```diff
@@ -157,4 +157,7 @@ token: abcdef.1234567890123456
 kubernetesVersion: ${version}
 networking:
   podSubnet: 10.244.0.0/16
-EOF
\ No newline at end of file
+EOF
+
+# Pre pull fluentd image used in logging
+docker pull docker.io/fluentd/fluentd
```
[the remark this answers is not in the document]
petrkotas: Pulling in the provision results in precached images. Pulling here results in pulling everytime the CI runs. 
I placed the pull there based on the hint from @rmohr to have tu CI more swift.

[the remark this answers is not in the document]
petrkotas: As for the unified manifest. This is a good idea. Thanks. Done it.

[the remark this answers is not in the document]
petrkotas: Regarding pulling in the `provision.sh`, Thanks to your comment, I figured I was pulling wrong images. Thanks, fixed that.

petrkotas:
@slintes Done with the comments.

k8s-multus/scripts/provision.sh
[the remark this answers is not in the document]
rmohr: Yes we need to pull in the provision.sh script too and the images in the manifests and the pre-pulled image need to be identical.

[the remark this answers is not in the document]
slintes: ok, check
nevertheless the docker pull at this place still needs to be updated

petrkotas:
@slintes thanks for the catch. Fixed.

petrkotas:
Okay this is weird, the commit is in my repository https://github.com/petrkotas/kubevirtci/commit/b6fc20b7f07fa4b2108cd43e2a1d5cea356796ee but does not show here.

petrkotas:
@rmohr since @slintes also agrees, can we merge?

rmohr:
@petrkotas I will bring in #34 first (where I just wait for a final ack that the uploaded images work). Then I will build the images, add the shasums to your PR and merge it.

petrkotas:
Awesome. Thanks!

rmohr:
@petrkotas updates k8s clusters and gocli. openshift will follow tomorrow.

petrkotas:
@rmorh that is great news. Thank you for your effort.

PR #15, opened 2018-06-10
Add pre-allocated 2M hugepages to k8s images

cynepco3hahue:
Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

### DIFF
--- a/k8s/scripts/provision.sh
+++ b/k8s/scripts/provision.sh
@@ -10,7 +10,7 @@ swapoff -a
 sed -i '/ swap / s/^/#/' /etc/fstab
 
 # Disable spectre and meltdown patches
-sed -i 's/quiet"/quiet spectre_v2=off nopti"/' /etc/default/grub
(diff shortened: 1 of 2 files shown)


k8s/scripts/provision.sh
```diff
@@ -10,7 +10,7 @@ swapoff -a
 sed -i '/ swap / s/^/#/' /etc/fstab
 
 # Disable spectre and meltdown patches
-sed -i 's/quiet"/quiet spectre_v2=off nopti"/' /etc/default/grub
+sed -i 's/quiet"/quiet spectre_v2=off nopti hugepagesz=2M hugepages=64"/' /etc/default/grub
```
rmohr: @cynepco3hahue can there be kernels where this does not work? Do we need a switch?

cynepco3hahue: @rmohr If we talk only about x86_64, it will always work, the problem happens only with 1G hugepages(it requires specific CPU flag, two of our CI hosts does not have this flag)

PR #18, opened 2018-06-13
Set selinux to permissive for openshift

petrkotas:
This PR introduces permissive selinux on openshift test cluster.
The permissive setting simplifies the deployment while still logging
every denial.

This PR is preliminary requirement for aggregated logging and will simplify
our life for running tests in the future.

@cynepco3hahue @rmohr Would you kindly review and merge?
Thank you.

Signed-off-by: Petr Kotas <petr.kotas@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirtci/18)

### DIFF
--- a/os-3.9/scripts/provision.sh
+++ b/os-3.9/scripts/provision.sh
@@ -2,6 +2,10 @@
 
 set -ex
 
+# Set SELinux to permissive. Still logging the denials.
+setenforce 0
+sed -i "s/^SELINUX=.*/SELINUX=permissive/" /etc/selinux/config
+
 # Install epel
 yum -y install epel-release
 


cynepco3hahue:
@petrkotas OpenShift cluster must run with enforcing by default, while it enforcing it also write all errors to audit, why do you want to disable it?

petrkotas:
As I have written in the PR comment, without setting it to permissive I cannot run the logging agent. It is blocked. Moreover, since the logging platform is built to allow selinux denials, it is helpful to run in permissive mode and collect all denials and than create selinux politic.

The openshift requires selinux to be enabled. Permissive is also enabled. I have tested the cluster and it installs and runs fine. 

However, if it still is a concern to you, we can sort a different solution. I am open to discussion.

rmohr:
> As I have written in the PR comment, without setting it to permissive I cannot run the logging agent. It is blocked. Moreover, since the logging platform is built to allow selinux denials, it is helpful to run in permissive mode and collect all denials and than create selinux politic.
> 
> The openshift requires selinux to be enabled. Permissive is also enabled. I have tested the cluster and it installs and runs fine.
> 
> However, if it still is a concern to you, we can sort a different solution. I am open to discussion.

You probably have to give it the right setting: https://docs.openshift.com/enterprise/3.1/install_config/persistent_storage/pod_security_context.html#selinuxoptions

petrkotas:
It is solved. Closing

PR #23, opened 2018-07-07
Make ck8s 1.10.3 cluster offline bootable

rmohr:
Download flannel manifests during provision phase.


cynepco3hahue:
The change also relevant for `k8s-1.10.4` and `k8s-1.9.3` images

PR #25, opened 2018-07-16
Added k8s 1.11.0 provider

slintes:
the hash for the new image needs to be added to the readme after being pushed...

Signed-off-by: Marc Sluiter <msluiter@redhat.com>

### DIFF
--- a/k8s/scripts/provision.sh
+++ b/k8s/scripts/provision.sh
@@ -90,31 +90,26 @@ done
 
 kubectl --kubeconfig=/etc/kubernetes/admin.conf get pods -n kube-system
 
-export k8s11=false
+reset_command="kubeadm reset"
+admission_flag="admission-control"
+# k8s 1.11 needs some changes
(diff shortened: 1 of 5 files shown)


k8s/scripts/provision.sh
```diff
@@ -90,14 +90,31 @@ done
 
 kubectl --kubeconfig=/etc/kubernetes/admin.conf get pods -n kube-system
 
-kubeadm reset
+export k8s11=false
+if [[ $version =~ \.([0-9]+) ]] && [[ ${BASH_REMATCH[1]} -ge "11" ]]; then
```
cynepco3hahue: Can you use something more simple, for examlpe:
```
if [[ $version =~ .*-1.11..* ]]; then
  reset_command="kubeadm reset --force"
  admissionflag="enable-admission-plugins"
else
  ...
fi
```

slintes: My check will also work for future versions, the '-ge' means greater or equal. I'd prefer that one (and hope that 1.12 won't need other changes in this area ;) ).
But I can group the 2 ifs together like you did, that's a bit nicer indeed.

PR #26, opened 2018-07-18
Create local volume PV [WIP]

awels:
- Added local volume PV directories.
- Added local volume provisioner.


awels:
WIP, just did this for the OS 3.10 provider, once we work out the exact format, I will do the same thing for the k8s provider.

@rmohr @davidvossel please review.

os-3.10/scripts/provision.sh
```diff
@@ -109,3 +109,156 @@ ansible-playbook -i $inventory_file $openshift_ansible/playbooks/deploy_cluster.
...
+  mkdir /mnt/local-storage/local/disk${i}
+  echo "/var/local/kubevirt-storage/local-volume/disk${i} /mnt/local-storage/local/disk${i} none defaults,bind 0 0" >> /etc/fstab
+done
+# Setup selinux permissions to local volume directories.
+chcon -R unconfined_u:object_r:svirt_sandbox_file_t:s0 /mnt/local-storage/
+# Add privileged to local volume provision service account
+/usr/bin/oc adm policy add-scc-to-user privileged -z local-storage-admin
+
+# Create local volume provisioner yaml
+cat > /tmp/local-volume.yaml << EOF
```
awels: Not sure if I am happy dropping the yaml file in here, it works, but seems messy, maybe a better solution available?

cynepco3hahue: in general, you can create a separate script for it, copy it under gocli and call it from `provision.sh` script, but I do not sure if it worse or better :smile:

rmohr: Yeah, we didn't need manifest so fare because `kubeadm` and `openshift-ansible` were doing everything when we provisioned the cluster initially and stated them later.

We will need them as part of the provision scripts and will also have to ship them as part of the cluster-startup-scripts.

When provisioning the cluster initially, we will have to deploy the storage provisioner too, so that all dependencies are pulled into the vm. Then we need to store the yaml somwhere inside the vm (like I for instance do for the network provisioner here: https://github.com/kubevirt/kubevirtci/pull/23.

I think in the case of the storage provisioner, we can just enable it by default. In case we want to make it opt-in, we would need a switch in `gocli` to then run the script, like petr does it with logging: https://github.com/kubevirt/kubevirtci/pull/9.

In general it might not be too bad to add them to the container (not the vm) and enable them via something like a `feature` flag which also matches the yaml file via `gocli` arguments.

rmohr: Oh, and to make it a little bit more confusing: For the whole lifecycle of a running cluster we use `gocli`. For provisioning we are still using the old bash-written client in `cli`. Did not yet finalize the provision command in `gocli`.

awels: Confused is the right word :). Anyway lets try to figure out what I need to do next. Poking around it looks like I can get something into the container using the Dockerfile under cli. So what if I create a manifest for the storage provisioner, and get it into the container using a Dockerfile copy, then reference that from the provision.sh instead of inlining it there. That way I can re-use the same manifest in the k8s provisioner.sh as well.

I am still trying to wrap my head around all of this, and I need a little bit of hand holding.

awels: So after thinking about this, and playing around with the code on how to get a file into the VM that is running in the container, I have come to the conclusion it is probably going to be much easier if we put the manifests in the kubevirtci github repo, and then in the provision.sh do what @rmohr does in #23 . Grab the manifest with a curl during the execution of the provision.sh, and then use it in node01.sh (since we really only need to execute the manifest once).

I think this will work nicely, it does however present me with a chicken and egg problem. I am using the manifest in the PR that adds the manifest. Which makes it hard to test as the manifest doesn't exist yet.

What do you guys think of this approach, and how would I solve the chicken and egg problem (I could make a separate PR to put the manifest in repo)

rmohr: Also thought about it. You can also save the manifests in the kubevirtci repo and then copy them into the container, like we do here: https://github.com/kubevirt/kubevirtci/blob/master/cli/cli#L183 (just modify the script). Then you can copy it with `scp` into the machine which makes it effectively available for all provision scripts.

* So copying it into the container here: https://github.com/kubevirt/kubevirtci/blob/master/cli/cli#L183
* Then copy it from the container into the vm before this is executed: https://github.com/kubevirt/kubevirtci/blob/master/cli/cli#L185

rmohr: And since you would want to use `scp`, the `ssh.sh` script looks like this: https://github.com/kubevirt/kubevirtci/blob/ce712b232a0985dc4d01ca817f8925222227b9e1/base/scripts/vm.sh#L38.

rmohr: The `scp` command for provisioning should look something like this:

```
scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i vagrant.key ...
```

awels: So that is what I spend most of my time doing yesterday, looking at how to get the manifests into the container and them get them from the container into the VM like you pointed me to. I also came to the conclusion I would have to implement an scp to copy the files to the VM. To me it seemed easier to simply curl them in the provision script instead. But that presents the chicken and egg problem. If there is no issue with implementing an scp.sh script similar to the ssh one. That would work as well.

rmohr: You can also just hardcode the scp command just for your copy-attempt. It will always have the same IP when we provision. `gocli` has an `scp` command included already. Just need to get to it, to also use it for provisioning. You don't have to spend too much time on making it nice in the bash script.

rmohr: And I would prefer the non-curl way because of the circle.

k8s/scripts/provision.sh
```diff
@@ -103,3 +103,20 @@ kubernetesVersion: ${version}
 networking:
   podSubnet: 10.244.0.0/16
 EOF
+
+# Create local-volume directories
+mkdir /var/local/kubevirt-storage
```
j-griffith: mkdir -p and combine the two?

k8s/scripts/provision.sh
```diff
@@ -103,3 +103,20 @@ kubernetesVersion: ${version}
...
+mkdir /var/local/kubevirt-storage/local-volume
+mkdir /mnt/local-storage
+mkdir /mnt/local-storage/local/
```
j-griffith: same as above WRT mkdir -p?

os-3.10/scripts/provision.sh
```diff
@@ -109,3 +109,22 @@ ansible-playbook -i $inventory_file $openshift_ansible/playbooks/deploy_cluster.
...
 /usr/bin/oc adm policy add-cluster-role-to-user cluster-admin admin
+
+# Create local-volume directories
+mkdir /var/local/kubevirt-storage
+mkdir /var/local/kubevirt-storage/local-volume
+mkdir /mnt/local-storage
+mkdir /mnt/local-storage/local/
+for i in {1..10}
+do
+  mkdir /var/local/kubevirt-storage/local-volume/disk${i}
```
j-griffith: so actually you could condense L#s 114-117 down into this for loop using -p I think?

k8s/scripts/provision.sh
awels: Been using linux over a decade, never knew that existed... yeah saves a few lines of code.

k8s/scripts/provision.sh
awels: yes, easier to just do -p in the loop.

os-3.10/scripts/provision.sh
awels: I believe so

os-3.10/scripts/provision.sh
```diff
@@ -109,3 +109,18 @@ ansible-playbook -i $inventory_file $openshift_ansible/playbooks/deploy_cluster.
 /usr/bin/oc create identity allow_all_auth:admin
 /usr/bin/oc create useridentitymapping allow_all_auth:admin admin
 /usr/bin/oc adm policy add-cluster-role-to-user cluster-admin admin
+
+# Create local-volume directories
+for i in {1..10}
```
rmohr: Does this mean that I can create 10 `PVC`s with the storage class `local` per node?

awels: Correct, so in a 3 node cluster it will create 30 PVs, and when you request a PVC, it will bind one of the available ones, and the pod claiming it will be scheduled on the node the PV that was claimed is defined on. Also note I marked the storage class as default, so if you don't specify a storage class it will use the local volume one.

awels: Actually you can add more PVs if you want, by simply mount binding more directories after the fact (like I described in the original documentation PR that spawned this one). I figured 10 PVs per node would a decent default, but you have to ability to expand if needed.

rmohr: Great. Thx for the explanation.

rmohr:
@awels if the k8s and openshift clusters are working for you, we can proceed with building and uploading them. I can do that tomorrow for you. @cynepco3hahue, @slintes  if you have some spare time, just let me know. Otherwise I start tomorrow.

@awels once , we can add the shasums to the readme and then merge your PR. So from my perspective your work is done here. Thx.

cynepco3hahue:
@rmohr I can do it today, so leave it to me

awels:
Excellent, I have tested both the k8s and OS-3.10 providers on my setup, and they worked for me. I tried both single node and 3 node clusters. I was able to claim PVCs, and run a CDI import on them, and start the VM from the import.

cynepco3hahue:
Pushed updated images

PR #27, opened 2018-07-25
K8s provider with multus cni

SchSeba:
Create a new k8s 1.11.0 provider with multus cni.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirtci/27)

### DIFF
--- a/k8s-multus/scripts/node01.sh
+++ b/k8s-multus/scripts/node01.sh
@@ -11,9 +11,5 @@ kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f crd.yml
 kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f flannel-conf.yml
 kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f macvlan-conf.yml
 
-# kubectl create clusterrolebinding multus-node-`hostname` \
-#     --clusterrole=multus-crd-overpowered \
-#     --user=system:node:`hostname`
-
 kubectl --kubeconfig=/etc/kubernetes/admin.conf apply -f https://raw.githubusercontent.com/intel/multus-cni/dev/network-plumbing-working-group-crd-change/examples/multus-with-flannel.yml
 kubectl --kubeconfig=/etc/kubernetes/admin.conf taint nodes node01 node-role.kubernetes.io/master:NoSchedule-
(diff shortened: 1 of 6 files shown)


rmohr:
@SchSeba does it work already?

SchSeba:
@cynepco3hahue 
I only test it with k8s-1.11.0
Added the version variable. ( I was using an export variable locally)
Thanks!

SchSeba:
@rmohr Sure :)

cynepco3hahue:
@rmohr If you do not have other objections I can create the image and push it to docker hub.

k8s-multus/scripts/provision.sh
```diff
@@ -0,0 +1,158 @@
...
+      }
+    }'
+EOF
+
+kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f clusterrole.yml
+kubectl --kubeconfig=/etc/kubernetes/admin.conf create clusterrolebinding multus-node-`hostname` \
+                                                       --clusterrole=multus-crd-overpowered \
+                                                       --user=system:node:`hostname`
+
+kubectl --kubeconfig=/etc/kubernetes/admin.conf apply -f https://raw.githubusercontent.com/intel/multus-cni/dev/network-plumbing-working-group-crd-change/examples/multus-with-flannel.yml
```
rmohr: Can you download the file here and then just provision it from the file here and in the startup script? You don't have to repeate a mistake which we are going to fix: #23.

k8s-multus/scripts/node01.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Wait for the network to really came up
+while [[ `cat /proc/sys/net/ipv4/ip_forward` -eq 0 ]]
+do
+ sleep 2
+done
+
+while [[ ! -f /proc/sys/net/bridge/bridge-nf-call-iptables ]]
+do
+ sleep 2
```
rmohr: @slintes looks like this may be related to the issues you saw?

rmohr: Ok, then it might be better to wait for docker in general?

SchSeba: Done

SchSeba:
@rmohr I make the requested changes.

cynepco3hahue:
Will update the PR with relevant image digest.

PR #28, opened 2018-07-25
Improved network config, needed for kubeadm 1.11 preflight checks

slintes:


slintes:
At least the cluster was started locally and on standard-ci.
Maybe let's wait if kubevirtci is also still happy
(see https://github.com/kubevirt/kubevirt/pull/1353)

PR #29, opened 2018-07-25
Wait for api server to be available before installing local volume provisioner.

awels:
Added a check to see if the API server is available before attempting
to install the local volume provisioner.


awels:
@rmohr This fixes the issue with 1 node OS-3.10 provider where the API server wasn't available before attempting to install the local volume provisioner. Also added the check to k8s provider just in case we try on a slow machine. It also caps the maximum number of retries at 20.

cynepco3hahue:
@rmohr Will do it

PR #32, opened 2018-08-09
update sdn network plugin to use openshift-ovs-networkpolicy

duyanyan:
Add a parameter here to use openshift-ovs-networkpolicy plugin since the networkpolicy test require a dev cluster with openshift-ovs-networkpolicy plugin.

Without this parameter, the dev cluster will use openshift-ovs-subnet plugin by default, pods/vms can be connected across the namespaces. After changing to openshift-ovs-networkpolicy plugin, it is same as subnet if you don't create any networkpolicy rules in the namespace. So I think it is no harm to other testing.


cynepco3hahue:
@duyanyan I pushed the image, can you please update os-3.10 provider(`cluster/os-3.10.0/provider.sh`) under your tests PR

duyanyan:
@cynepco3hahue  I have updated the image in my test PR. Thanks

cynepco3hahue:
@duyanyan I think this one no relevant anymore because multus oc image already has ovs.

duyanyan:
@cynepco3hahue Yes, thank you!

PR #41, opened 2018-09-11
Pass --qemu-args as a single argument to vm.sh

booxter:
A lot of qemu arguments are comprised of multiple words, f.e.:

-device vfio-pci,...

We should pass them into vm.sh as a single argument.

### DIFF
--- a/gocli/cmd/run.go
+++ b/gocli/cmd/run.go
@@ -314,7 +314,7 @@ func run(cmd *cobra.Command, args []string) (err error) {
 		volumes <- vol.Name
 
 		if len(qemu_args) > 0 {
-			qemu_args = "--qemu-args " + qemu_args
+			qemu_args = fmt.Sprintf("--qemu-args '%s'", qemu_args)
 		}
 		node, err := cli.ContainerCreate(ctx, &container.Config{
 			Image: cluster,


booxter:
I would love to test it end-to-end but it's not clear how to build the container locally so that I can replace gocli container with an updated version. I tested the change isolated though.

rmohr:
@booxter sorry for the inconvenience, I got distracted in the middle of converting everything to bazel. You can run `bazel run //:gocli` once. 

You will see output like this:

```
Tagging 349fa64932c766f95e42297aaf6f64388e8be39ddda242bea8a669151aa26d23 as bazel:gocl
```

After you did that, you can set `bazel:gocli` e.g. in you kubevirt/kubevirt repo on the same node. Then it will use this version.

Btw. I think I saw that you want to pass through some devices from the node. Keep in mind that we start all VMs in containers. They don't necessarily see all devices on the node, just in case you get errors like that a device does not exist.

booxter:
@rmohr thanks a lot for the tip! I will try it once I fix another issue in the path of extra args passing to vm.sh that I've found in a different place. (This fix is still legit.)

Yes, I may need to mount some devices from the host into qemu containers, not yet sure though, and if so, how to achieve it.

PR #34, opened 2018-08-10
Openshift 3.10 provider with multus network plugin

SchSeba:
Added openshift 3.10 provider with multus network plugin

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirtci/34)

PR #44, opened 2018-09-17
Added cni plugins deployment files to the manifest folder.

SchSeba:
* scp the files before starting the provision.
* check that all the pods are running before finnish the provison state.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirtci/44)


SchSeba:
@cynepco3hahue @rmohr Can you please take a look ?

This is a follow PR as requested in #34

k8s-multus/scripts/node01.sh
```diff
@@ -11,9 +11,17 @@ done
 
 kubeadm init --config /etc/kubernetes/kubeadm.conf
 
-kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f /etc/kubernetes/multus.yml
-kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f /etc/kubernetes/flannel.yml
-kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f /etc/kubernetes/macvlan-conf.yml
-kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f /etc/kubernetes/ptp-conf.yml
-kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f /etc/kubernetes/bridge-conf.yml
+kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f /tmp/flannel.yaml
```
rmohr: Is there a way that we can simply deploy all files in this directory? E.g. only copying the correct config maps into the tmp directory?

SchSeba: The problem is the copy happens on the cli `docker exec ${VM_CID} /bin/bash -c "scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i vagrant.key -P 22 /manifests/*.yaml vagrant@192.168.66.101:/tmp"`

phoracek:
@SchSeba I think we should install Multus in the way we advertise it in ovs-cni (with shipped manifests). We would have better test coverage for it plus we won't need to maintain yet another way to install multus. The same for ovs-cni.

SchSeba:
@rmohr Please take a look this PR is ready now

rmohr:
> We still need to build the os-multus provider. (the new k8s is already exist)

@SchSeba rebuilding the image.

os-3.10-multus/scripts/provision.sh
```diff
@@ -214,8 +214,30 @@ chcon -R unconfined_u:object_r:svirt_sandbox_file_t:s0 /mnt/local-storage/
...
+
+# Wait for the daemonset to create the ovs file
+while  [ ! -f /usr/bin/ovs-vsctl ]
+do
+    echo "ovs-vsctl not found!"
+    sleep 5
+done
+
+# Create a bridge for the tests
+ovs-vsctl add-br br1
```
rmohr: @SchSeba is this  command persistent? Will the bridge be there after restarts?

SchSeba: Yes I already check it.

I also run the multus functests locally and everything pass.

rmohr:
@SchSeba image is pushed.

SchSeba:
@rmohr Thanks I will try it on https://github.com/kubevirt/kubevirt/pull/1583 and update.

os-3.10-multus/scripts/provision.sh
phoracek: Add a new line please.

SchSeba: Done

manifests/cni-plugins-ds.yaml
```diff
@@ -0,0 +1,42 @@
...
+            memory: "50Mi"
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - name: cnibin
+          mountPath: /opt/cni/bin
+      volumes:
+      - name: cnibin
+        hostPath:
+          path: /opt/cni/bin
```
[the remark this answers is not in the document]
SchSeba: Done

k8s-multus/scripts/provision.sh
```diff
@@ -49,7 +49,13 @@ yum install --nogpgcheck -y \
...
     kubelet-${version} \
     kubectl-${version} \
-    kubernetes-cni
+    kubernetes-cni \
+    openvswitch
+
+systemctl start openvswitch
+systemctl enable openvswitch
+
+ovs-vsctl add-br br1
```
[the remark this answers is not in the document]
SchSeba: I don't think we need this.

The functional tests using affinity to make sure both vm's are in the same node.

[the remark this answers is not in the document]
phoracek: If it is meant only for functional tests, then OK. It's just a bit misleading that we create a network, but when it is used with multiple nodes, it won't work.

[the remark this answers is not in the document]
SchSeba: Yes its only for functional tests.

SchSeba:
@rmohr Adding the image pull policy doesn't fix the problem on pods stuck in containerCreate status on the secondary node so I just remove all the relevant pods (node01.sh script) and the daemonSet recreate them.

k8s-multus/scripts/provision.sh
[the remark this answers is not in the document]
phoracek: Would it be complicated to create this bridge as a part of the functional test? I'd rather not spoil clean deployment, kubevirtci is not only for KubeVirt functional tests.

os-3.10-multus/scripts/node01.sh
```diff
@@ -84,9 +84,12 @@ while [[ $retry_counter -lt 20  && $os_rc -ne 0 ]]; do
 done
 set -e
 
-/usr/bin/oc create -f /tmp/local-volume.yaml
+# Remove the multus pod to recreate it.
```
phoracek: It says node01 in the file name. Does it mean this will be executed on master node?

manifests/ovs-net-vlan100.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: "k8s.cni.cncf.io/v1"
```
phoracek: IMO network definition should not be a part of kubevirtci, it should be created as a part of testing.

os-3.10-multus/scripts/node01.sh
SchSeba: Yes this will execute only on the master node and remove all the pod related, this way nothing is stuck in containerCreation

manifests/ovs-net-vlan100.yaml
SchSeba: Yes we don't need this anymore. We create them on the functest

https://github.com/kubevirt/kubevirt/pull/1583/files#diff-897951be21489aebaa1df0c66894562cR60

SchSeba:
@phoracek 
> Would it be complicated to create this bridge as a part of the functional test? I'd rather not spoil clean deployment, kubevirtci is not only for KubeVirt functional tests.

I am not sure how to do it in test time you have any suggestion?

rmohr:
> > 
> 
> I am not sure how to do it in test time you have any suggestion?

In theory you can run a pod which runs in the host network namespace. Practically we try to not change the host and more try to detect in the tests on what is given to us. Would there be a way to detect that the node is set up in a way which is sufficient for the test and skip it otherwise?

phoracek:
> @phoracek
> 
> > Would it be complicated to create this bridge as a part of the functional test? I'd rather not spoil clean deployment, kubevirtci is not only for KubeVirt functional tests.
> 
> I am not sure how to do it in test time you have any suggestion?

As Roman suggested, running a DS with mounted OVS socket and creating it there or just calling cluster/cli.sh ssh NODES with ovs-vsctl command.

However, if @rmohr is ok with this approach, I won't fight for it.

SchSeba:
I think it is ready I also already build and push the images @phoracek @rmohr please review.

rmohr:
@SchSeba sorry but the same like in the kubevirt repo. Merged genie first and now this one needs a rebase.

SchSeba:
@rmohr Done with the rebase

PR #55, opened 2018-12-26
Enable user namespaces. This configuration is required by buildah.

annastopel:
In order to use buildah tool, it is required to enable max_user_namespaces on cluster node.

### DIFF
--- a/k8s-genie/scripts/provision.sh
+++ b/k8s-genie/scripts/provision.sh
@@ -13,6 +13,9 @@ sed -i '/ swap / s/^/#/' /etc/fstab
 sed -i 's/quiet"/quiet spectre_v2=off nopti hugepagesz=2M hugepages=64"/' /etc/default/grub
 grub2-mkconfig -o /boot/grub2/grub.cfg
 
+# Allow user namespaces
+sysctl -w user.max_user_namespaces=1024
+
 systemctl stop firewalld NetworkManager || :
 systemctl disable firewalld NetworkManager || :
 # Make sure the firewall is never enabled again
--- a/k8s-multus/scripts/provision.sh
+++ b/k8s-multus/scripts/provision.sh
@@ -13,6 +13,9 @@ sed -i '/ swap / s/^/#/' /etc/fstab
 sed -i 's/quiet"/quiet spectre_v2=off nopti hugepagesz=2M hugepages=64"/' /etc/default/grub
 grub2-mkconfig -o /boot/grub2/grub.cfg
 
+# Allow user namespaces
+sysctl -w user.max_user_namespaces=1024
+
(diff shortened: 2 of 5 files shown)


annastopel:
@cynepco3hahue

cynepco3hahue:
To be honest I thought @annastopel wanted to have support for buildah under builder container.
@annastopel Can you please confirm that you need buildah support under the cluster node?

annastopel:
> To be honest I thought @annastopel wanted to have support for buildah under builder container.
> @annastopel Can you please confirm that you need buildah support under the cluster node?

To our purposes it enough to enable the setting on container level, however it is not possible to run systemctl from just any container, you have to allow container to do so as described in https://forums.docker.com/t/systemctl-status-is-not-working-in-my-docker-container/9075/3
Specifically:
```bash
FROM centos:7
MAINTAINER “you” your@email.here
ENV container docker
RUN (cd /lib/systemd/system/sysinit.target.wants/; for i in ; do [ $i == systemd-tmpfiles-setup.service ] || rm -f $i; done); 
rm -f /lib/systemd/system/multi-user.target.wants/;
rm -f /etc/systemd/system/.wants/;
rm -f /lib/systemd/system/local-fs.target.wants/; 
rm -f /lib/systemd/system/sockets.target.wants/udev; 
rm -f /lib/systemd/system/sockets.target.wants/initctl; 
rm -f /lib/systemd/system/basic.target.wants/;
rm -f /lib/systemd/system/anaconda.target.wants/*;
VOLUME [ “/sys/fs/cgroup” ]
CMD ["/usr/sbin/init"]
```
I am not sure it is a good thing to have a container that can execute systemctl unless you really need that. In our case I figure the specific setting can be beneficial to entire system

k8s-genie/scripts/provision.sh
```diff
@@ -13,6 +13,9 @@ sed -i '/ swap / s/^/#/' /etc/fstab
 sed -i 's/quiet"/quiet spectre_v2=off nopti hugepagesz=2M hugepages=64"/' /etc/default/grub
 grub2-mkconfig -o /boot/grub2/grub.cfg
 
+# Allow user namespaces
+systemctl -w user.max_user_namespaces=1024
```
danielBelenky: I couldn't find the `-w` flag in `systemctl` docs. What does it do?
I know that you usually would use `sysctl` for such configurations...

rmohr:
@slintes looks like the dns issue in the CI cluster is not fully resolved yet.

slintes:
> @slintes looks like the dns issue in the CI cluster is not fully resolved yet.

yeah, too bad, I would have liked to get the open PRs tested, pushed and merged today, since I'm on PTO next week...

PR #56, opened 2019-01-08
Added os-3.11.0-multus-sriov provider

booxter:
This provider is based on regular os-3.11-multus provider. It deploys SR-IOV
components (CNI, DP with config file, sriov network attachment) on top.

Note that when deploying from this image, to make SR-IOV device plugin to
allocate VFs from PFs available on the node, one should list PF addresses in
/etc/pcidp/config.json and then restart sriov device plugin pods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@
 * `os-3.11.0` os-3.11.0 cluster based on the centos7 image, provisioned with openshift-ansible
 * `os-3.11-crio` os-3.11 cluster with CRI-O support based on the centos7 image, provisioned
 * `os-3.11-multus` os-3.11 cluster with multus cni support based on the centos7 image, provisioned with openshift-ansible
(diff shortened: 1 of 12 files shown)


booxter:
@rmohr assuming we get this PR approved, we'll need to push a new image to dockerhub. How do I get access to it? I was told you shared permissions with Sebastian in the past.

booxter:
@phoracek do you have a link handy to ConfigMap adoption for DP configuration purposes? I think I'll need to adopt the same here in manifests carried with the CI repo.

phoracek:
@booxter sure https://github.com/kubevirt/kubevirt-ansible/pull/539

booxter:
/hold

This new 'driver' may not be useful for stdci after all since we need bare metal access to the node PFs to make it work, and a driver implies having another layer of virtualization (qemu inside a node image) that we would need to pass PFs through, which is not obvious how to achieve.

Marcin is exploring using 'local' provider instead. In the meantime, I am going to hold off this patch not to introduce images that are not used anywhere.

PR #62, opened 2019-03-04
Add Ceph storage

igoihman:
Ceph setup: 
- Ceph storage demo container.
- Ceph CSI v1.0 deployment on node01.
- Ceph secret.
- Ceph storageclass.

This is currently supported only on k8s-1.13.3


rmohr:
@igoihman @cynepco3hahue can we talk about this offline? I am not sure if we want to squeeze this into the images. It might make more sense, to provide that storage solution from outside (like we do e.g. for the docker registry). We would start more containers, on-demand and the clusters can just pick up the ceph connection.

igoihman:
@rmohr sure, let's schedule a call

igoihman:
@cynepco3hahue @rmohr @slintes Can you please review?

gocli/cmd/run.go
```diff
@@ -13,14 +20,10 @@ import (
...
 	"kubevirt.io/kubevirtci/gocli/docker"
-	"os"
-	"os/signal"
-	"path/filepath"
-	"strconv"
-	"sync"
 )
 
 const NFSGaneshaImage = "docker.io/janeczku/nfs-ganesha@sha256:17fe1813fd20d9fdfa497a26c8a2e39dd49748cd39dbb0559df7627d9bcf4c53"
+const CephImage = "docker.io/ceph/daemon:latest-luminous"
```
rmohr: Pleasue use a non-floating tag. This one looks like it can float. If there is no non-floating tag, you can use the shasum.

k8s/scripts/ceph-csi.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -xe
+
+yum install -y ceph-common
+
+# Deploy RBACs for sidecar containers and node plugins
+kubectl --kubeconfig /etc/kubernetes/admin.conf create -f /tmp/csi-attacher-rbac.yaml
```
rmohr: We should consider adding a ceph subdirectory and do something like `kubectl create -f /tmp/ceph/`. Other components do it right now but it only gets worse if we never start it.

gocli/cmd/run.go
igoihman: done

k8s/scripts/provision.sh
```diff
@@ -132,21 +136,52 @@ fi
 
 $reset_command
 
+# New configuration for kubernetes >= 1.12
```
slintes: >= 1.11 (that's at least what the next line says)

k8s/scripts/provision.sh
```diff
@@ -132,21 +136,52 @@ fi
...
+controllerManagerExtraArgs:
+  feature-gates: "BlockVolume=true,,CSIBlockVolume=true,VolumeSnapshotDataSource=true"
+kind: ClusterConfiguration
+kubernetesVersion: ${version}
+networking:
+  podSubnet: 10.244.0.0/16
+
+EOF
+else
 # TODO new format since 1.11, this old format will be removed with 1.12, see https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/#config-file
```
slintes: this TODO can be removed

igoihman: done

k8s/scripts/provision.sh
igoihman: done

cynepco3hahue:
@igoihman Can you run `chmod +x` on all bash scripts.
Under the job, I can see `/bin/sh: 1: ./provision.sh: Permission denied`

cynepco3hahue:
@igoihman Can you update README
- add new provider line
- add hash line for the provider with the hash `sha256:734690023598ca6e3e91cb5cd7ffd1561f58a42761a9e9c5e755c6d9ca0667f7`

cynepco3hahue:
@igoihman I see that you remove my update for gocli hash, can you restore it
new gocli hash is `sha256:fa7f615a1b07925b27027c57bf09bba0e9874ca92e4f67559556950665598c49`

igoihman:
@cynepco3hahue Sorry, I restored it.

rmohr:
@igoihman great contribution. Thanks!

PR #49, opened 2018-10-23
Enable CustomResourceSubresources feature gate

rmohr:
Needs to be enabled for any cluster which we still use which is prior to
ck8s 1.11 or 3.11.

Allows using the status endpoint for our CRDs to not allow kubevirt
users to override the status section by accident or intentionally.

PR #63, opened 2019-03-04
Fixed enabling of CustomResourceSubresources feature-gate

slintes:
This is a bit strange:

- the code which was committed with https://github.com/kubevirt/kubevirtci/pull/49 doesn't work (CustomResourceSubresources is a feature-gate, and no admission plugin). A cluster built from that code doesn't start, because apiserver fails
- but the k8s 1.10.11 docker image which was pushed with that PR is correct, the feature gate is enabled (I checked the apiserver logs). So no need to push new images with this PR

Signed-off-by: Marc Sluiter <msluiter@redhat.com>


rmohr:
@slintes thx.

Issue #69, opened 2019-04-10
option to pass custom provision file while (make cluster-up)


kunalkushwaha:
/kind enhancement

__Problem:__
 while working behind network proxy, `make cluster-up` fails to setup cluster while downloading CNI config. Also, no `docker pull` or `kubectl apply` is successful as no image gets downloaded due to lack of proxy settings in docker-daemon.

```
.
.
Your Kubernetes master has initialized successfully!

To start using your cluster, you need to run the following as a regular user:

  mkdir -p $HOME/.kube
  sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
  sudo chown $(id -u):$(id -g) $HOME/.kube/config

You should now deploy a pod network to the cluster.
Run "kubectl apply -f [podnetwork].yaml" with one of the options listed at:
  https://kubernetes.io/docs/concepts/cluster-administration/addons/

You can now join any number of machines by running the following on each node
as root:

  kubeadm join 192.168.66.101:6443 --token abcdef.1234567890123456 --discovery-token-ca-cert-hash sha256:23d5e3383df6878d799c4b326f8910d521894dc9a403dcac9dd6a3800236fac3

+ kubectl --kubeconfig=/etc/kubernetes/admin.conf apply -f https://raw.githubusercontent.com/coreos/flannel/v0.9.1/Documentation/kube-flannel.yml
The connection to the server raw.githubusercontent.com was refused - did you specify the right host or port?
provisioning node node01 failed
Makefile:104: recipe for target 'cluster-up' failed
make: *** [cluster-up] Error 1
```
__Expected Behaviour__
- modify provision scripts like `node.sh` or `node<node-no>.sh` etc.
- `gocli run ` should be able to consume custom scripts if indicated explicitly

rmohr:
> kubectl --kubeconfig=/etc/kubernetes/admin.conf apply -f https://raw.githubusercontent.com/coreos/flannel/v0.9.1/Documentation/kube-flannel.yml

@kunalkushwaha that's an issue in the base scripts. They should **never** download on teprovision state anything. They should work offline by never pulling on restart of VMIs.

kunalkushwaha:
#71 fixes issue partially for using `make cluster-up` script without change behind proxy. To create VMI's docker daemon should be able to pull images, which requires proxy settings for docker daemon.

[...]

PR #70, opened 2019-04-10
option added to set docker daemon proxy setting

kunalkushwaha:
--docker-proxy option added to gocli run command.
    This helps to set proxy setting for docker daemon
    without rebuilding of base VM image.
    
Fixes https://github.com/kubevirt/kubevirtci/issues/69 along with https://github.com/kubevirt/kubevirtci/pull/71

Signed-off-by: Kunal Kushwaha <kushwaha_kunal_v7@lab.ntt.co.jp>

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
 ## Versions to use
 
 * `kubevirtci/cli`: `sha256:1dd015dea4f12e6dcb8e31be3eeb677fed96f290ef4a4892a33c43d666053536`
-* `kubevirtci/gocli`: `sha256:fa7f615a1b07925b27027c57bf09bba0e9874ca92e4f67559556950665598c49`
+* `kubevirtci/gocli`: `sha256:837ab7fff4e00d74ad9d6b8952c79dbf3df1db380da1fc13a87df34d70745003`
 * `kubevirtci/base`: `sha256:034de1a154409d87498050ccc281d398ce1a0fed32efdbd66d2041a99a46b322`
 * `kubevirtci/centos:1804_02`: `sha256:70653d952edfb8002ab8efe9581d01960ccf21bb965a9b4de4775c8fbceaab39`
 * **Deprecated**: `kubevirtci/os-3.9.0:`: `sha256:234b3ae5c335c9fa32fa3bc01d5833f8f4d45420d82a8f8b12adc02687eb88b1`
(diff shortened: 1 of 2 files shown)


kunalkushwaha:
This PR enables to overwrite provision scripts while starting k8s cluster, which is helpful in settings like docker daemon proxy settings, downloading networking `[podnetwork].yaml` in `kubectl apply` command etc

```
$ ./cli run --mount-scripts /home/kunal/kubevirtci/k8s/scripts --nodes 1 --memory 5120M --cpu 5 --random-ports  --background --prefix k8s-1.10.11 --registry-volume
kubevirt_registry kubevirtci/k8s-1.10.11@sha256:b97e556795a56b9aa1763ddf3a49322b49f96877dccb7a164bbca779df078536 --ocp-port 8443
```

rmohr:
> Avoids rebuilding of base VM image for every change in environment.

We deliberately want to rebuild the VM image. The trick is that all docker images are already cached and all rpms preinstalled. There seems to be only the issue that it fetches the flannel manifests from the internet again, that is why `kubectl apply` fails. 

I am not against the content of this PR, we can add it, but it is not a fix for #69. We should not fetch manifests from the internet again on `cluster-up`.

kunalkushwaha:
Updated the PR for scope limited to docker daemon proxy settings as discussed in https://github.com/kubevirt/kubevirtci/issues/69#issuecomment-481569171

Along with #71 this will make cluster workable behind proxy. 

//cc @rmohr

rmohr:
oh, we need gocli test runs ...

rmohr:
@kunalkushwaha @slintes, sorry for the confusion, but the jobs were not actually active so far, I manually triggered them. Should work now. Let's see what the results say.

rmohr:
@kunalkushwaha can you add the new gocli digest to the README.md? The digest from the job is `sha256:837ab7fff4e00d74ad9d6b8952c79dbf3df1db380da1fc13a87df34d70745003`. After that the PR can be merged.

rmohr:
/hold

digest needs to be updated.

kunalkushwaha:
> The digest from the job is sha256:837ab7fff4e00d74ad9d6b8952c79dbf3df1db380da1fc13a87df34d70745003. After that the PR can be merged.

@rmohr thanks. I have updated README.md  with suggested digest.

PR #72, opened 2019-04-17
Include e2fsprogs in the tests image to use mkfs

vladikr:


slintes:
ups, forgot that we have the prow workflow here...

PR #75, opened 2019-04-23
enable audit log for os-3.11

mfranczy:
Enable audit logs for os-3.11.

### DIFF
--- a/cluster-provision/os-3.11/scripts/provision.sh
+++ b/cluster-provision/os-3.11/scripts/provision.sh
@@ -139,6 +139,11 @@ all:
         osm_controller_args:
           feature-gates:
           - BlockVolume=true
+        openshift_master_audit_config:
+          enabled: true
+          logFormat: json
+          auditFilePath: "/var/lib/origin/audit-ocp.log"
+          policyFile: "/etc/origin/master/adv-audit.yaml"
         openshift_node_groups:
         - name: node-config-master-infra-kubevirt
           labels:
@@ -177,6 +182,27 @@ all:
           - key: kubeletArguments.pods-per-core
             value:
             - '40'
+EOF
+
+mkdir -p /etc/origin/master
+cat >/etc/origin/master/adv-audit.yaml <<EOF
+apiVersion: audit.k8s.io/v1beta1
+kind: Policy
+rules:
+- level: Request
+  users: ["system:admin"]
+  resources:
+  - group: kubevirt.io
+    resources:
+    - virtualmachines
+    - virtualmachineinstances
+    - virtualmachineinstancereplicasets
+    - virtualmachineinstancepresets
+    - virtualmachineinstancemigrations
+  omitStages:
+  - RequestReceived
+  - ResponseStarted
+  - Panic
 EOF
 
 # Add cri-o variable to inventory file
--- a/cluster-provision/os-3.11/scripts/provision.sh
+++ b/cluster-provision/os-3.11/scripts/provision.sh
@@ -67,22 +67,12 @@ inventory_file="/root/inventory"
 master_ip="192.168.66.101"
 echo "$master_ip node01" >> /etc/hosts
 
-git clone https://github.com/openshift/openshift-ansible.git -b v3.11.0 --depth 1 $openshift_ansible
-
-# Apply fix https://github.com/openshift/openshift-ansible/pull/10459
-# TODO: remove it when the fix will be available under the v3.11.0 tag
-sed -i 's/python-docker/python-docker-py/' $openshift_ansible/playbooks/init/base_packages.yml
-
-# Apply fix https://github.com/openshift/openshift-ansible/issues/11213
-# TODO: remove it when the fix will be available under the v3.11.0 tag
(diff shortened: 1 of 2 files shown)


mfranczy:
@slintes, @rmohr can we enable audit logs for os-3.11? This PR is related to https://github.com/kubevirt/kubevirt/pull/2208, I need them to generate REST API coverage report.
Tests should not be affected in case of coverage failure.

If you're ok with that I will publish a new image later.

slintes:
@mfranczy lgtm, let's see what the test says. The image can be pushed by prow too now, no need to do that manually anymore :) (at least very soonish ;) )

cynepco3hahue:
hm strange error message
```
Error response from daemon: Error processing tar file(exit status 1): unexpected EOF
```
maybe we out of space

cynepco3hahue:
/push-os-3.11.0

cynepco3hahue:
@mfranczy Can you try to build locally, does it work?

mfranczy:
@cynepco3hahue yes I will try locally again, last time it worked (I pushed an image to my own repo mfranczy/kubevirtci:os-3.11.0-audit). Will let you know soon, thanks for looking into that.

mfranczy:
@cynepco3hahue huh, it fails locally now, debugging in progress...

mfranczy:
there is some network problem,
```
oc describe node node01
...
  ----             ------  -----------------                 ------------------                ------                       -------                                                           
  OutOfDisk        False   Thu, 06 Jun 2019 14:07:13 +0000   Thu, 06 Jun 2019 13:56:42 +0000   KubeletHasSufficientDisk     kubelet has sufficient disk space available                       
  MemoryPressure   False   Thu, 06 Jun 2019 14:07:13 +0000   Thu, 06 Jun 2019 13:56:42 +0000   KubeletHasSufficientMemory   kubelet has sufficient memory available                           
  DiskPressure     False   Thu, 06 Jun 2019 14:07:13 +0000   Thu, 06 Jun 2019 13:56:42 +0000   KubeletHasNoDiskPressure     kubelet has no disk pressure                                      
  PIDPressure      False   Thu, 06 Jun 2019 14:07:13 +0000   Thu, 06 Jun 2019 13:56:42 +0000   KubeletHasSufficientPID      kubelet has sufficient PID available
  Ready            False   Thu, 06 Jun 2019 14:07:13 +0000   Thu, 06 Jun 2019 13:56:42 +0000   KubeletNotReady              runtime network not ready: NetworkReady=false reason:NetworkPlugi$
NotReady message:docker: network plugin is not ready: cni config uninitialized
...
```
still checking what's causing the root issue

mfranczy:
the problem is with ansible version, 2.8 is not supported by openshift-ansible 3.11, that was causing the issue.

mfranczy:
small progress, it's failing now on `playbooks/openshift-service-catalog/config.yml`, checking that..

cluster-provision/os-3.11/scripts/provision.sh
```diff
@@ -71,6 +73,10 @@ git clone https://github.com/openshift/openshift-ansible.git -b v3.11.0 --depth
 # TODO: remove it when the fix will be available under the v3.11.0 tag
 sed -i 's/python-docker/python-docker-py/' $openshift_ansible/playbooks/init/base_packages.yml
 
+# Apply fix https://github.com/openshift/openshift-ansible/issues/11213
+# TODO: remove it when the fix will be available under the v3.11.0 tag
```
mfranczy: we execute `git clone https://github.com/openshift/openshift-ansible.git -b v3.11.0 --depth 1 $openshift_ansible`
tag v3.11.0 is quite old and because of that we have to apply bugfixes by ourselves, isn't better to switch to openshift-ansible-3.11.118-1 ([link](https://github.com/openshift/openshift-ansible/releases))?

mfranczy:
so there were two problems, the first one with the latest floating tag for service broker (https://github.com/kubevirt/kubevirtci/pull/75/files#diff-893b8df0effd7156d06a37b26124ef4cR78) and the second one with insufficient memory, I increased from 3G (default) to 7G.

cluster-provision/os-3.11/scripts/provision.sh
slintes: I'd say: go ahead and give it a try :)

slintes:
> so there were two problems, the first one with the latest floating tag for service broker (https://github.com/kubevirt/kubevirtci/pull/75/files#diff-893b8df0effd7156d06a37b26124ef4cR78) and the second one with insufficient memory, I increased from 3G (default) to 7G.

Do we actually need the service broker? Maybe we can just disable it. And maybe other unneeded components too. Might solve the problem and make the image smaller :)

mfranczy:
> Do we actually need the service broker?

I don't know. I was blindly trying to fix provisioners, didn't think about it.

> Maybe we can just disable it. And maybe other unneeded components too. Might solve the problem and make the image smaller :)

Fair point.

cluster-provision/os-3.11/scripts/provision.sh
mfranczy: I thought that somebody may have a reason to use this specific tag, but I will give it a try.

cynepco3hahue:
Great work Marcin!
/lgtm
/approve

PR #84, opened 2019-05-29
Create the ovs br1 on worker nodes for ocp 3.11

SchSeba:
This PR fixes an issue on the ocp provider.

When we request more the one node for the cluster the additional workers
are reinstalled using the openshift-ansible scale-up playbook.

This playbook removes all the ovs configuration done previously,
so this PR just run on every node using the ovs daemonset installed in the
cluster and recreate the br1 bridge if not exist.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirtci/84)

### DIFF
--- a/os-3.11-multus/scripts/node01.sh
+++ b/os-3.11-multus/scripts/node01.sh
@@ -93,3 +93,5 @@ oc -n linux-bridge delete po `oc get po -n linux-bridge | grep bridge | awk '{pr
 oc -n kube-system delete po `oc get po -n kube-system | grep ovs-cni | awk '{print $1}'`
 
 /usr/bin/oc create -f /tmp/local-volume.yaml
+
+oc get po -n openshift-sdn | grep ovs | awk '{print "oc exec -n openshift-sdn",$1,"-- ovs-vsctl --may-exist add-br br1"}' | sh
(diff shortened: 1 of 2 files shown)


SchSeba:
@rmohr can you take a look I moved the change here

os-3.11-multus/scripts/node01.sh
```diff
@@ -93,3 +93,5 @@ oc -n linux-bridge delete po `oc get po -n linux-bridge | grep bridge | awk '{pr
 oc -n kube-system delete po `oc get po -n kube-system | grep ovs-cni | awk '{print $1}'`
 
 /usr/bin/oc create -f /tmp/local-volume.yaml
+
+oc get po -n openshift-sdn | grep ovs | awk '{print "oc exec -n openshift-sdn",$1,"-- ovs-vsctl --may-exist add-br br1"}' | sh
```
cynepco3hahue: hm get `po`?

SchSeba: yeah it's the same as `oc get pod`

cynepco3hahue: hh wasn't aware of such short name

PR #96, opened 2019-06-11
Enable PSP for k8s clusters

rmohr:
Needed to ensure that KubeVirt works with PSPs enabled.


cynepco3hahue:
@rmohr can you please push again and update the README?

PR #99, opened 2019-06-17
setup linux bridge for functional tests

phoracek:
Multus functional tests, starting with [1] will use Linux bridge CNI
as the default CNI plugin for L2. Since we don't yet have a node
network operator available to configure interfaces on nodes, this
patch introduces a workaround. We use a privileged daemonset to
create bridge br10 on every node and configure respective firewall
rules.

[1] https://github.com/kubevirt/kubevirt/pull/2361

### DIFF
--- a/cluster-provision/k8s-multus/scripts/node01.sh
+++ b/cluster-provision/k8s-multus/scripts/node01.sh
@@ -42,3 +42,9 @@ while [[ $retry_counter -lt 20 && $kubectl_rc -ne 0 ]]; do
     retry_counter=$((retry_counter + 1))
 done
 kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f /tmp/local-volume.yaml
+
+# Setup Linux bridge used in functional tests on nodes. This won't be needed once we
+# finish kubernetes-nmstate and use it for the setup instead.
+./cluster-up/kubectl.sh apply -f /tmp/linux-bridge-config-iptables.yaml
+while [ "$(./cluster-up/kubectl.sh get ds linux-bridge-config -o 'jsonpath={.status.numberDesiredNumberScheduled}')" == "0" ]; do sleep 1; done
+while [ "$(./cluster-up/kubectl.sh get ds linux-bridge-config -o 'jsonpath={.status.numberReady}')" != "$(./cluster-up/kubectl.sh get ds linux-bridge-config -o 'jsonpath={.status.desiredNumberScheduled}')" ]; do sleep 1; done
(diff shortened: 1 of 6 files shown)


phoracek:
@cynepco3hahue could you please take a look?

cluster-provision/k8s-multus/scripts/node01.sh
```diff
@@ -42,3 +42,9 @@ while [[ $retry_counter -lt 20 && $kubectl_rc -ne 0 ]]; do
     retry_counter=$((retry_counter + 1))
 done
 kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f /tmp/local-volume.yaml
+
+# Setup Linux bridge used in functional tests on nodes. This won't be needed once we
+# finish kubernetes-nmstate and use it for the setup instead.
+./cluster-up/kubectl.sh apply -f /tmp/linux-bridge-config-iptables.yaml
+while [ "$(./cluster-up/kubectl.sh get ds linux-bridge-config -o 'jsonpath={.status.numberDesiredNumberScheduled}')" == "0" ]; do sleep 1; done
```
cynepco3hahue: I believe the `numberDesiredNumberScheduled` should be equal to the number of workers nodes, so I believe you also will need to check it under the run script

phoracek: Good call.

cluster-provision/k8s-multus/scripts/node01.sh
```diff
@@ -42,3 +42,9 @@ while [[ $retry_counter -lt 20 && $kubectl_rc -ne 0 ]]; do
     retry_counter=$((retry_counter + 1))
 done
 kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f /tmp/local-volume.yaml
+
+# Setup Linux bridge used in functional tests on nodes. This won't be needed once we
+# finish kubernetes-nmstate and use it for the setup instead.
+./cluster-up/kubectl.sh apply -f /tmp/linux-bridge-config-iptables.yaml
```
cynepco3hahue: it should be `kubectl` instead of `./cluster-up/kubectl.sh` this script runs inside of the master node

cluster-provision/manifests/linux-bridge-config-nftables.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: apps/v1
```
SchSeba: No need for this file we can do the same as openshift-sdn use chroot and mount the host into our utility pod.

please take a look on this https://bugzilla.redhat.com/show_bug.cgi?id=1708500

phoracek: Neat, thanks!

PR #100, opened 2019-06-18
Remove deprecated providers

cynepco3hahue:
- k8s-multus-1.10.4
- k8s-multus-1.11.1
- k8s-multus-1.12.2
- k8s-1.9.3
- k8s-1.10.3
- k8s-1.10.4
- os-3.9
- os-3.9-crio
- os-3.10
- os-3.10-crio
- os-3.10-multus


mfranczy:
with that change I think that can be removed, https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/k8s/scripts/provision.sh#L74 (we don't have k8s < 11)

cynepco3hahue:
> with that change I think that can be removed, https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/k8s/scripts/provision.sh#L74 (we don't have k8s < 11)

We still have support for 1.10.11, I think we will remove it once we will have k8s 1.14

PR #108, opened 2019-06-20
Use Exec for non TTY ssh commands gocli

### LINKED ISSUE #107: Calling cli.sh ssh or ssh.sh without TTY does not return the correct error code
When calling from stuff like golang code or jenkins it's not under TTY and it does not return the correct error code (and possibly it does not wait to finish).

To reproduce
```bash
echo "./kubevirtci/cluster-up/cli.sh ssh node01 false; echo \$?" > foo.sh
nohup bash foo.sh  > nohup.log 2>&1 < /dev/null
cat nohup.log # It should be 1 but it's 0
```

Running directly the gocli command with and without -it

```bash
$ docker run --privileged --net=host --rm -v /var/run/docker.sock:/var/run/docker.sock kubevirtci/gocli@sha256:8571161d7956b830646216335453b995ba754e07319dde062241ccc025f5ee00 --prefix k8s-1.13.3 ssh node01 false
$ echo $?
0

$ docker run --privileged -it --net=host --rm -v /var/run/docker.sock:/var/run/docker.sock kubevirtci/gocli@sha256:8571161d7956b830646216335453b995ba754e07319dde062241ccc025f5ee00 --prefix k8s-1.13.3 ssh node01 false
$ echo $?
1
```

Also looks like without "-it" it's not waiting to finish the command

```bash
$ time docker run --privileged --net=host --rm -v /var/run/docker.sock:/var/run/docker.sock kubevirtci/gocli@sha256:8571161d7956b830646216335453b995ba754e07319dde062241ccc025f5ee00 --prefix k8s-1.13.3 ssh node01 sleep 10

real	0m0,391s
user	0m0,064s
sys	0m0,041s
```

qinqon:
In case of non TTY we have to use Exec call from docker library
at gocli since it does the proper stuff, Terminal was just bypassing
the code.

closes #107

### DIFF
--- a/cluster-provision/gocli/cmd/ssh.go
+++ b/cluster-provision/gocli/cmd/ssh.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/docker/docker/client"
 	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh/terminal"
 	"kubevirt.io/kubevirtci/gocli/docker"
 )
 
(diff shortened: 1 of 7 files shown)


qinqon:
/cc @davidvossel @rmohr @slintes 

Can you help me with this ? we have a PR at kubernetes-nmstate that depends on this thing.

cluster-provision/gocli/cmd/ssh.go
```diff
@@ -35,10 +36,25 @@ func ssh(cmd *cobra.Command, args []string) error {
 	}
 
 	// TODO we can do the ssh session with the native golang client
-	exitCode, err := docker.Terminal(cli, prefix+"-"+node, append([]string{"ssh.sh"}, args[1:]...), os.Stdout)
-	if err != nil {
-		return err
+	container := prefix + "-" + node
+	ssh_command := append([]string{"ssh.sh"}, args[1:]...)
+	file := os.Stdout
+	if terminal.IsTerminal(int(file.Fd())) {
```
cynepco3hahue: hm if you the one who define the file to
`file := os.Stdout` how it can be different from `terminal.IsTerminal(int(file.Fd()))`

qinqon: What do you mean ?

cynepco3hahue: ah now I see the `file.Fd()` changes when you connect with and without TTY

PR #105, opened 2019-06-19
Secondary nics

qinqon:
Add --secondary-nics options to gocli
    
The kubernetes-nmstate [1] project is using kubevirtci and it needs
secondary nics to test bonding and linux-bridge functionality.
    
This add necesary qemu-args to create a configurable number of
secondary nics with KUBEVIRT_NUM_SECONDARY_NICS.
    
[1] https://github.com/nmstate/kubernetes-nmstate

PR #110, opened 2019-06-21
Compose correctly the secondary-nics at multinode

qinqon:
From stated here [1] running with multinode and
secondary nics fail since it was composing incorrectly
the quemu args line.

This remove the infinite concatenation and use unique
device names per node.

[1] https://github.com/kubevirt/kubevirtci/pull/105#discussion_r296142403

### DIFF
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -375,6 +375,17 @@ func run(cmd *cobra.Command, args []string) (err error) {
 	// start one vm after each other
 	for x := 0; x < int(nodes); x++ {
 
+		nodeQemuArgs := qemuArgs
+
+		for i := 0; i < int(secondaryNics); i++ {
+			netSuffix := fmt.Sprintf("%d-%d", x, i)
+			nodeQemuArgs = fmt.Sprintf("%s -device virtio-net-pci,netdev=secondarynet%s -netdev tap,id=secondarynet%s,ifname=stap%s,script=no,downscript=no", nodeQemuArgs, netSuffix, netSuffix, netSuffix)
+		}
+
+		if len(nodeQemuArgs) > 0 {
+			nodeQemuArgs = "--qemu-args \"" + nodeQemuArgs + "\""
+		}
+
 		nodeName := nodeNameFromIndex(x + 1)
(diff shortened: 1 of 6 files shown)


cluster-provision/gocli/cmd/run.go
```diff
@@ -373,8 +373,20 @@ func run(cmd *cobra.Command, args []string) (err error) {
 	wg := sync.WaitGroup{}
 	wg.Add(int(nodes))
 	// start one vm after each other
+
```
phoracek: please drop the line, the comment should not be far away from the block it comments

qinqon:
How do I add some multinic/multinode cluster-up cluster-down test here ?

cluster-provision/gocli/cmd/run.go
```diff
@@ -373,8 +373,20 @@ func run(cmd *cobra.Command, args []string) (err error) {
...
 	for x := 0; x < int(nodes); x++ {
 
+		extraArgs := qemuArgs
```
phoracek: this is still only qemuArgs, until you put it into extra args on 387. Mind keeping it as qemuArgs until put into `"--qemu-args \"" + qmuArgs + "\""`

cluster-provision/gocli/cmd/run.go
qinqon: Done

cluster-provision/gocli/cmd/run.go
qinqon: We need a clean variable every iteration so we don't append to next nodes, but you are right this is not extra args, let me rename the var.

mfranczy:
and please update gocli sha

qinqon:
@SchSeba added to readme
@mfranczy already added the sha

mfranczy:
@qinqon but it's old one (before you made changes after @phoracek comments), should be this one `caf1c3f63dc5f3137795084e492a48c802a7ef2e7f7fbe1d67e5cff684d376a3`

qinqon:
@mfranczy ok fixed, would be nice to have that automatize or semi-automatize (changes on gocli spin the push job, and checks for hashes in this project)

mfranczy:
@qinqon thanks, yup that would be nice.

PR #111, opened 2019-06-24
enable auditing for k8s images

mfranczy:
Enable audit logs for k8s images


mfranczy:
@cynepco3hahue do you think it makes sense to add audit log for 1.10? AFAIK we want to get rid of it soon?

cynepco3hahue:
I think 1.11 and 1.13.3 are enough.

PR #119, opened 2019-07-15
Kind sriov cluster

fedepaol:
This PR enables kubevirtci to run a `[kind](https://github.com/kubernetes-sigs/kind)` backed cluster, while moving the host's sriov enabled interface into the namespace of the node of kind.

As per the discussion with @slintes in https://github.com/kubevirt/kubevirt/pull/2419, I split the setup of the cluster and sriov capabilities and merged here into CI.

Adding one note: this setup relies on the fact that kubectl will already be on the host that is running it (CI images have it on board).

PR #122, opened 2019-07-21
Fix sriov kind provider

fedepaol:
This is a followup to https://github.com/kubevirt/kubevirtci/pull/119 . 

After the refactor `KUBEVIRTCI_PATH` env variable was not available anymore to consumers of `common.sh` and for that reason the sriov ci was failing because the sriov configuration script was not able to find the manifests. 

This pr addresses this problem.


slintes:
mh, we are setting KUBEVIRTCI_PATH at way too many places, wondering if there is a better way to handle and export this. But that would be out of scope of this PR.

/lgtm
/approve

PR #130, opened 2019-08-08
Ensure that we use machine type q35

rmohr:


cynepco3hahue:
we need to rebuild all images with a new base image, also the OpenShift 4.1 uses libvirt straight forward so need also to add machine under the xml definition, see - https://github.com/kubevirt/kubevirtci/blob/d749b29f7a2fa6cb7b46c3590040f2a23e5704e0/cluster-provision/okd/scripts/run.sh#L26

PR #140, opened 2019-08-28
Replace the sriov cni image used by the openshift operator with the upstream one.

fedepaol:
In order to fully test the sr-iov feature we need to ship the intel sriov cni plugin instead of the one shipped with the operator (at least, until https://github.com/intel/sriov-cni/pull/88 is merged in the openshift sr-iov operator).

This pr replaces the sriov cni image with the intel one before deploying the operator.


fedepaol:
> assuming it fixes the custom MAC address SR-IOV test case

Tested locally and it did (if I did not anything wrong). Will run in CI tomorrow

PR #141, opened 2019-09-02
Provider for k8s 1.14.6

dhiller:


dhiller:
@slintes don't know, first time here, but I guess I need the other PR merged first so I can start trying to biuld the images, right?

slintes:
yes, just approved it. So you can now run `/test push-k8s-1.14.6` here. Then check logs, and update shasums here. Test the cluster locally. Then we merge here, and you can create a PR on kubevirt, for updating cluster-up (see commit hash in `hack/config-default.sh` and run `make generate`), and another PR on project-infra for adding a Prow job with the new test lane (make it optional first while testing on the kubevirt PR)

dhiller:
/test push-k8s-1.14.6

@slintes thanks for the heads up!

dhiller:
/test push-k8s-1.14.6

@slintes adjusted kubeadm configuration for 1.14. I thought that the configuration would be deployed within the created docker image, so I could just spin up the image and have a look at the configuration that would have been deployed. Unfortunately that does not seem to be the case, right?

slintes:
the config will be in the vm started by the container. From my notes you can get into the VM with:
- `docker exec -it <containerId> /bin/bash`
- `ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no vagrant@192.168.66.101 -i vagrant.key -p 22`

btw, nice improvements in the provision script :+1: 
if it makes life easier: we don't use the k8s 1.11 lane anymore, feel free to remove code branches for it and mark it as deprecated in the readmes :slightly_smiling_face:

cluster-provision/k8s/scripts/provision.sh
```diff
@@ -2,6 +2,11 @@
 
 set -ex
 
+function get_minor_version() {
```
slintes: :+1:

cluster-up/cluster/k8s-1.14.6/provider.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+
+image="k8s-1.14.6@sha256:ec29c07c94fce22f37a448cb85ca1fb9215d1854f52573316752d19a1c88bcb3"
```
slintes: this sha is different from the one in cluster-provision/README.md?
(and yes, having these shasum spread through several files is ugly)

dhiller: Done

PR #146, opened 2019-09-12
hack: add simple pre-flight check to cluster-up

fromanirh:
It still happens nowdays that we run kubevirtci on wrongly provisioned
boxes with incorrect kvm setup.
This patch adds a simple pre-flight check script to catch the most
common issues, hopefully preventing more time wasted in debugging
cluster running on these wrong setups.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 cluster-up:
-	./cluster-up/check.sh && ./cluster-up/up.sh
+	./cluster-up/check.sh
+	./cluster-up/up.sh
 
 cluster-down:
 	./cluster-up/down.sh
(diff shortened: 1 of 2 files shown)


fromanirh:
ping?

cynepco3hahue:
It can be useful for the KubeVirt usage, but it possible scenario when you do not want really want to run VM's under the cluster, can we make it more informative than restrictive, just give the warning message but do not exit.

fromanirh:
> It can be useful for the KubeVirt usage, but it possible scenario when you do not want really want to run VM's under the cluster, can we make it more informative than restrictive, just give the warning message but do not exit.

ack, updating

PR #151, opened 2019-09-17
Allow the kind-sriov provider to have two nodes

fedepaol:
Since now the kind-sriov provider was forced to have only one node, where the sriov devices of the host were moved.
There are some cases where the operator needs to perform a node drain to perform configuration changes, which means that having it working with two nodes is preferrable.

With this PR, we check for the presence of the second node (worker) in the kind cluster and if present, we do the setup on that node instead of the master one.

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
@@ -41,8 +41,6 @@ else
   SRIOV_NODE=$FIRST_WORKER_NODE
 fi
 
-SRIOV_NODE_CMD="docker exec -it -d ${SRIOV_NODE}"
-
 #move the pf to the node
 mkdir -p /var/run/netns/
 export pid="$(docker inspect -f '{{.State.Pid}}' $SRIOV_NODE)"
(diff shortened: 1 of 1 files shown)


booxter:
Don't you need to modify `kind` manifest too to enable two nodes?

cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
```diff
@@ -32,10 +35,18 @@ function deploy_sriov_operator {
   popd
 }
 
+if [[ -z "$(kubectl get nodes | grep $FIRST_WORKER_NODE)" ]]; then
+  SRIOV_NODE=$MASTER_NODE
+else
+  SRIOV_NODE=$FIRST_WORKER_NODE
+fi
+
+SRIOV_NODE_CMD="docker exec -it -d ${SRIOV_NODE}"
```
dhiller: Nit: maybe move this to line 77 to have it more adjacent to the usage.

fedepaol: Done, do you mind re-approving this?

PR #152, opened 2019-09-20
Add gocli qemu option to attach serial to PTY

qinqon:
In case ssh is down or not accesible is not possible
to attach directly to the console with current providers.

This patch add a gocli option 'serial-pty' to attach serial port to pty device
at qemu run so one can do 'socat - /dev/pts/0' and access the virtual machines

### DIFF
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -371,6 +371,10 @@ func run(cmd *cobra.Command, args []string) (err error) {
 		}
 	}
 
+	// Add serial pty so we can do stuff like 'socat - /dev/pts0' to access
+	// the VM console from the container without ssh
+	qemuArgs += " -serial pty"
+
 	wg := sync.WaitGroup{}
 	wg.Add(int(nodes))
 	// start one vm after each other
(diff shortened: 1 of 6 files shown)


cluster-provision/gocli/cmd/run.go
```diff
@@ -71,6 +71,7 @@ func NewRunCommand() *cobra.Command {
 	run.Flags().String("log-to-dir", "", "enables aggregated cluster logging to the folder")
 	run.Flags().Bool("enable-ceph", false, "enables dynamic storage provisioning using Ceph")
 	run.Flags().String("docker-proxy", "", "sets network proxy for docker daemon")
+	run.Flags().Bool("serial-pty", false, "attach serial to /dev/pts/ device")
```
rmohr: I guess we could just always add it.

rmohr: What do you think?

qinqon: I think so, in case we see problems we can go back to optional.

rmohr:
@qinqon no issue with your PR, but obviously with building in the push job. Will have a look when I find time during the day.

qinqon:
@SchSeba thanks good point!

PR #153, opened 2019-09-23
okd-4.1: avoid permission denied running as user

fromanirh:
with this PR, I can now run "make cluster-up" as user e.g. NOT with root account.
There are no intended behaviour changes in the flow, and no changes for the root use case.

### DIFF
--- a/cluster-up/cluster/okd-4.1/provider.sh
+++ b/cluster-up/cluster/okd-4.1/provider.sh
@@ -10,6 +10,17 @@ function _port() {
     ${_cli} ports --prefix $provider_prefix --container-name cluster "$@"
 }
 
+function _install_from_cluster() {
+    local src_cid="$1"
+    local src_file="$2"
+    local dst_perms="$3"
+    local dst_file="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/$4"
+
+    touch $dst_file
+    chmod $dst_perms $dst_file
+    docker exec $src_cid cat $src_file > $dst_file
+}
+
 function up() {
     workers=$(($KUBEVIRT_NUM_NODES-1))
     if [[ ( $workers < 1 ) ]]; then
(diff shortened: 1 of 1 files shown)


cynepco3hahue:
hm I never ran it as user, and as far as I know docker himself need additional permissions(it's the reason to create docker group), do you have some example of the error that you got?

fromanirh:
> hm I never ran it as user, and as far as I know docker himself need additional permissions(it's the reason to create docker group), do you have some example of the error that you got?

I forgot to mention that indeed I do have set docker in order to run it as normal user. Something like (probably not most secure option, but still)
```bash
$ which docker
~/bin/docker
$ cat ~/bin/docker 
#!/bin/sh
exec /usr/bin/sudo /usr/bin/docker $*
```
The error:
```
+ for vm in ${vms}
++ virsh net-dhcp-leases test-1-sfgtk
++ grep test-1-sfgtk-worker-0-7l58x
++ tr / '\t'
++ awk '{print $1}'
++ awk '{print $5}'
+ vm_ip=192.168.126.51
+ ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -q -lcore -i vagrant.key 192.168.126.51
Red Hat Enterprise Linux CoreOS 410.8.20190904.0
WARNING: Direct SSH access to machines is not recommended.

---
chmod: changing permissions of '/home/fromani/kubevirtci/_ci-configs/okd-4.1/.kubectl': Operation not permitted
```
The permissions:
```
$ ls -lah _ci-configs/okd-4.1/
total 73M
drwxrwxr-x. 2 fromani fromani 4.0K Sep 24 09:10 .
drwxrwxr-x. 3 fromani fromani 4.0K Sep 23 15:58 ..
-rw-rw-r--. 1 fromani fromani  335 Sep 24 09:02 config-provider-okd-4.1.sh
-rw-r--r--. 1 fromani fromani 4.2K Sep 24 09:02 .kubeconfig
-rwxr-xr-x. 1 root    root     73M Apr 10 14:18 .kubectl
```

PR #163, opened 2019-10-09
Deploy github.com/intel/network-resources-injector to the cluster.

fedepaol:
The resources injector is a mutating webhook that adds resource requests based on
the needed resources declared on the network attachment definition.
It's going to be deployed together with the sr-iov operator so we want to be sure it's compatible with kubevirt.

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
@@ -9,6 +9,7 @@ MASTER_NODE="${CLUSTER_NAME}-control-plane"
 FIRST_WORKER_NODE="${CLUSTER_NAME}-worker"
 
 OPERATOR_GIT_HASH=b3ab84a316e16df392fbe9e07dbe0667ad075855
+INJECTOR_GIT_HASH=9ffd768cb7886072e81df3ac78ba2997810ceb55
 
 # not using kubectl wait since with the sriov operator the pods get restarted a couple of times and this is
 # more reliable
@@ -36,6 +37,23 @@ function deploy_sriov_operator {
     make deploy-setup-k8s SHELL=/bin/bash OPERATOR_EXEC="${KUBECTL}"
   popd
 }
+
+function deploy_network_resource_injector {
+  WEBHOOK_PATH=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/network-resources-injector-${INJECTOR_GIT_HASH}
+  if [[ ! -d $WEBHOOK_PATH ]]; then
+    curl -L https://github.com/intel/network-resources-injector/archive/${INJECTOR_GIT_HASH}/network-resources-injector.tar.gz | tar xz -C ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/
+  fi
+
+  pushd $WEBHOOK_PATH
+    make image
+    docker tag network-resources-injector localhost:5000/network-resources-injector
+    sed -i 's/network-resources-injector:latest/registry:5000\/network-resources-injector:latest/g' ./deployments/server.yaml
+    docker push localhost:5000/network-resources-injector
+    _kubectl apply -f ./deployments/auth.yaml
+    _kubectl apply -f ./deployments/server.yaml
+  popd
+}
+
 
 if [[ -z "$(_kubectl get nodes | grep $FIRST_WORKER_NODE)" ]]; then
   SRIOV_NODE=$MASTER_NODE
@@ -90,6 +108,10 @@ _kubectl label node $SRIOV_NODE sriov=true
 envsubst < $MANIFESTS_DIR/network_config_policy.yaml | _kubectl create -f -
 
 
+wait_pods_ready
+
+deploy_network_resource_injector
+sleep 5
 wait_pods_ready
 
 ${SRIOV_NODE_CMD} chmod 666 /dev/vfio/vfio
--- a/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
@@ -40,7 +40,7 @@ function deploy_sriov_operator {
 
 function deploy_network_resource_injector {
   webhook_path=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/network-resources-injector-${INJECTOR_GIT_HASH}
-  if [[ ! -d $webhook_path ]]; then
+  if [ ! -d $webhook_path ]; then
     curl -L https://github.com/intel/network-resources-injector/archive/${INJECTOR_GIT_HASH}/network-resources-injector.tar.gz | tar xz -C ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/
   fi
 
--- a/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
@@ -47,7 +47,7 @@ function deploy_network_resource_injector {
   pushd $webhook_path
     make image
     docker tag network-resources-injector localhost:5000/network-resources-injector
-    sed -i 's/network-resources-injector:latest/registry:5000\/network-resources-injector:latest/g' ./deployments/server.yaml
+    sed -i 's#image: network-resources-injector:latest#image: registry:5000/network-resources-injector:latest#' ./deployments/server.yaml
     docker push localhost:5000/network-resources-injector
     _kubectl apply -f ./deployments/auth.yaml
     _kubectl apply -f ./deployments/server.yaml
--- a/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
@@ -39,12 +39,12 @@ function deploy_sriov_operator {
 }
 
 function deploy_network_resource_injector {
-  WEBHOOK_PATH=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/network-resources-injector-${INJECTOR_GIT_HASH}
-  if [[ ! -d $WEBHOOK_PATH ]]; then
+  webhook_path=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/network-resources-injector-${INJECTOR_GIT_HASH}
+  if [[ ! -d $webhook_path ]]; then
     curl -L https://github.com/intel/network-resources-injector/archive/${INJECTOR_GIT_HASH}/network-resources-injector.tar.gz | tar xz -C ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/
   fi
 
-  pushd $WEBHOOK_PATH
+  pushd $webhook_path
     make image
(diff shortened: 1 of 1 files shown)


fedepaol:
/hold
Don't add more noise until the sriov lane is stable again.

cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
```diff
@@ -37,6 +38,23 @@ function deploy_sriov_operator {
...
+function deploy_network_resource_injector {
+  WEBHOOK_PATH=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/network-resources-injector-${INJECTOR_GIT_HASH}
+  if [[ ! -d $WEBHOOK_PATH ]]; then
+    curl -L https://github.com/intel/network-resources-injector/archive/${INJECTOR_GIT_HASH}/network-resources-injector.tar.gz | tar xz -C ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/
+  fi
+
+  pushd $WEBHOOK_PATH
+    make image
+    docker tag network-resources-injector localhost:5000/network-resources-injector
+    sed -i 's/network-resources-injector:latest/registry:5000\/network-resources-injector:latest/g' ./deployments/server.yaml
```
phoracek: You can do `s#ORIGINAL/PATH#NEW/PATH#` (`#` to avoid escaping of `/`). You can also drop `g` since there won't be multiple occurrences of the substring on a line.

cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
```diff
@@ -37,6 +38,23 @@ function deploy_sriov_operator {
   popd
 }
 
+function deploy_network_resource_injector {
+  WEBHOOK_PATH=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/network-resources-injector-${INJECTOR_GIT_HASH}
```
phoracek: `webhook_path`, it is not a constant

cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
```diff
@@ -37,6 +38,23 @@ function deploy_sriov_operator {
...
+  fi
+
+  pushd $WEBHOOK_PATH
+    make image
+    docker tag network-resources-injector localhost:5000/network-resources-injector
+    sed -i 's/network-resources-injector:latest/registry:5000\/network-resources-injector:latest/g' ./deployments/server.yaml
+    docker push localhost:5000/network-resources-injector
+    _kubectl apply -f ./deployments/auth.yaml
+    _kubectl apply -f ./deployments/server.yaml
+  popd
```
phoracek: Why `pushd` and `popd`? this is enclosed in the function scope.

cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
```diff
@@ -92,4 +110,8 @@ envsubst < $MANIFESTS_DIR/network_config_policy.yaml | _kubectl create -f -
 
 wait_pods_ready
 
+deploy_network_resource_injector
+sleep 5
```
phoracek: If you have to use sleep, at least comment why.

cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
fedepaol: Makes sense, thanks for pointing out

cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
fedepaol: Not sure I understood your comment and the relation between pushd and the scope of the function. 
I pushd in the hook checkout so I can run make and apply the files locally, as per the hook installation setup.
Another approach could have been to pass the full path of the makefile and of the manifests. Do you think it's better that way?

fedepaol:
/hold cancel
Since the kind lane seems to be stable again

cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
```diff
@@ -37,6 +38,23 @@ function deploy_sriov_operator {
   popd
 }
 
+function deploy_network_resource_injector {
+  webhook_path=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/network-resources-injector-${INJECTOR_GIT_HASH}
+  if [ ! -d $webhook_path ]; then
+    curl -L https://github.com/intel/network-resources-injector/archive/${INJECTOR_GIT_HASH}/network-resources-injector.tar.gz | tar xz -C ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/
```
slintes: Can we download this beforehand and put needed files into `cluster-up/cluster/kind-k8s-sriov-1.14.2/manifests`?
On the one hand I don't think it's a good idea to modify files in cluster-up during runtime, because projects importing kubevirtci might check if cluster-up is still the same as in this repository (at least kubevirt does, and redownloads cluster-up if it detects modifications).
And on the other hand we seen it too often that CI fails because of failed downloads, either because of temporary networks issues, or because files not being availabel anymore. Could be easily prevented.
WDYT?

cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
```diff
@@ -21,12 +22,12 @@ function wait_pods_ready {
 }
 
 function deploy_sriov_operator {
-  OPERATOR_PATH=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/sriov-network-operator-${OPERATOR_GIT_HASH}
-  if [[ ! -d $OPERATOR_PATH ]]; then
+  operator_path=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/sriov-network-operator-${OPERATOR_GIT_HASH}
+  if [ ! -d $operator_path ]; then
     curl -L https://github.com/openshift/sriov-network-operator/archive/${OPERATOR_GIT_HASH}/sriov-network-operator.tar.gz | tar xz -C ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/
```
slintes: I see the download was here before this PR, but my other comment applied here as well....

cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
fedepaol: Well, the checkout is done in `KUBEVIRTCI_CONFIG_PATH` which should not suffer from the modification issue you mentioned.

However, putting in place some sync mechanism similar to what kubevirt does for CI would not certainly hurt and will make tests run faster. Will work on it.

slintes: Ah sorry, right, this is `_ci-configs` not `cluster-up`.
However, the issue with downloading stuff during CI runtime remains.

fedepaol:
/hold
want to checkout and try to my cluster.
Also, the webhook vendor folder contains a lot of stuff (and we cannot avoid it since we build the hook from there). 
This will reflect in an increase of the size of kubevirtci. Is it still ok @slintes ?

fedepaol:
/hold cancel
Tested

slintes:
> Also, the webhook vendor folder contains a lot of stuff (and we cannot avoid it since we build the hook from there).
> This will reflect in an increase of the size of kubevirtci. Is it still ok @slintes ?

Oh. I missed that we download sources here and build something. I thought we just need a manifest. At least this confirms why I called sick yesterday a bit later, wasn't my best day :D

But... I still think having this in this repo is better than downloading. But maybe just add the tar.gz file, and not the extracted content?

@rmohr wdyt?

rmohr:
> @rmohr wdyt?

Is it possible to add all the extra files into the kind container itself which is used as base? We were really working hard to have an "offline" model for the other lanes.

fedepaol:
> > @rmohr wdyt?
> 
> Is it possible to add all the extra files into the kind container itself which is used as base? We were really working hard to have an "offline" model for the other lanes.

In theory, yes. The thing is, by doing that we loose the possibility to update the k8s version of the kind lane that depends on the node image version. We would need to find an healthy way to maintain and publish that image.

Also, I am not sure I like the fact that we use the node image as a recipient for stuff it does not have anything to do with. 

The problem is only with the webhook, we fetch the sriov-operator repo only to use its manifests. In the generate script I drop the vendor folder which is the one that takes a lot of room. 

If it's a matter of publishing an image, I'd be happier with building and publishing the webhook image somewhere and using the default image for the kind node.

fedepaol:
Adding that we will probably need to build the webhook image anyway, to ship it with our CNAO. I'd pursue that road then, leaving only the operator checked out here (with no vendor it should not hurt, also the generate.sh may be useful for other dependencies).

fedepaol:
@rmohr @slintes I'll revert the last commit.
In the meanwhile, I asked for a new image repo in kubevirt quay. 
Since we are already downloading the operator and so far it's not hurting, can we treat the "slimming down" job in a separate pr that I'll pursue as soon as I have the webhook image pushed on quay?

slintes:
Works for me
Repo is created, network team has access
https://quay.io/repository/kubevirt/cluster-network-addon-resource-injector?tab=info

fedepaol:
/hold cancel

Tested.
@slintes @rmohr now I removed the vendor folder from the two dependencies. They could be stripped even more but I think this is a good compromise between size and hackery.

fedepaol:
So, as discussed on slack, I reverted the pr to downloading the injector each time the same way we were doing with the sr-iov operator. A followup pr will move the manifests and the needed images locally.

rmohr:
/lgtm
/approve

Here a follow-up list of things which need to be addressed, to keep CI functional on older releases:

 * We need a way to provide manifests as part of one of our containers (for instance by using an additional container in the kubevirtci docker namespace as container volume). No curl downloads from external locations.
 * In general the same for the images. Normally we add them to the VM images local docker registry on provision time. External container images have the tendency to disappear.

fedepaol:
> /lgtm
> /approve
> 
> Here a follow-up list of things which need to be addressed, to keep CI functional on older releases:
> 
>     * We need a way to provide manifests as part of one of our containers (for instance by using an additional container in the kubevirtci docker namespace as container volume). No curl downloads from external locations.
> 
>     * In general the same for the images. Normally we add them to the VM images local docker registry on provision time. External container images have the tendency to disappear.

Ok, for the second point if we don't want to pollute the node image is to have an image we control where we put the images we need (and eventually push them to the local registry), right?

Does that apply also if the webhook image is already hosted on kubevirt's quay? Also, the other images are the official openshift ones, which I expect to be less ephemeral, but now I understood and agree with the approach.

rmohr:
> We can just mirror them in our org and push them to the local registry or the nodes when we start the cluster.

Oh and yes, images in kubevirts quay are ok for me, iff they have non-floating tags, because quay deletes untagged images after some time (in contrast to docker hub).

rmohr:
> Also, the other images are the official openshift ones

We would have to clone them. Deprecated images disappear in openshift after some time (just happened with keycloak for instance). That does not necessarily mean that we don't need them anymore for tests on long-supported kubevirt versions.

fedepaol:
Ok, thanks for the clarification!

> > Also, the other images are the official openshift ones
> 
> We would have to clone them. Deprecated images disappear in openshift after some time (just happened with keycloak for instance). That does not necessarily mean that we don't need them anymore for tests on long-supported kubevirt versions.

Ok, thanks for the clarification.

PR #164, opened 2019-10-15
Add a note with another reason to use upstream sriov cni image

booxter:
```release-note
NONE
```


booxter:
@fedepaol we already use `:latest`, so just adding a note about another reason for upstream image.

fedepaol:
/lgtm

Also, your changes u/s were applied back to the sriov-cni plugin shipped with the operator.
https://github.com/openshift/sriov-cni/commits/master
Not sure about the image though, need to ask, so let's stick with the u/s until we sort this out.

PR #173, opened 2019-11-06
Run with net=host okd provisioning

qinqon:


dhiller:
If that works it would be a great thing!

dhiller:
I started a timer when it started downloading the rhcos image. It took two hours to download the image, so I guess the effect is only marginal. IMHO this doesn't help :(

PR #147, opened 2019-09-15
Allow to deploy network operator on providers

SchSeba:
This PR add a variable both for the cli and the gocli commands to control the deployment if the cluster-network-addons-operator.

In addition the PR creates a new k8s-multus-1.15.1 provider and a okd 4.1 with the network operator installed.

With this changes we are going to have two okd 4.1 lanes, one that contains the network operator (for kubevirt functional tests) and one without (for other projects like hco, nmstate, and the cluster-operator project itself)

Use cluster-network-addons-operator version 0.16.0

Signed-off-by: Sebastian Sch <sebassch@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirtci/147)

PR #181, opened 2019-11-13
update CNAO to 0.20.0

phoracek:
The current manifests referenced ancient version 0.8.0.

### DIFF
--- a/cluster-provision/manifests/cna/namespace.yaml
+++ b/cluster-provision/manifests/cna/namespace.yaml
@@ -2,6 +2,6 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: cluster-network-addons-operator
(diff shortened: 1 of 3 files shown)


cynepco3hahue:
@phoracek You will need to provision all providers that install network addons.

phoracek:
@cynepco3hahue that means calling some CI commands (help me here please) and ...?

cynepco3hahue:
it should be `/test push-<provider_name>` after from the job log you can get new image hash and update the README and all relevant places with new hash

SchSeba:
@phoracek there is also this PR https://github.com/kubevirt/kubevirtci/pull/147 that you can use

phoracek:
@SchSeba right, I think we should finish that one. Thanks for sharing!

PR #183, opened 2019-11-14
WIP: providers: okd: add podman support

fromanirh:
We need special treatment to make OKD clusters working with
the `pack8s` helper.
Differently from docker, we can't just call podman, because
we can be running with an user != root. In that case, the
copy step will fail. Hence, we use a special new command
in pack8s to do the copy on our behalf.


fromanirh:
pending implementation side on `pack8s` /cc @ksimon1

fromanirh:
obsoleted by https://github.com/kubevirt/kubevirtci/pull/189

PR #187, opened 2019-11-18
Refactor okd

cynepco3hahue:
Refactor OKD provision and run scripts
    
- move registry yaml configuration to the manifests directory
- move install-config.yaml from base container to manifests directory
- remove pull-token after provision
- ask for pull-token on run time
- remove if condition, what works for 4.3 also works for 4.1
- add the `--installer-pull-secret-file` parameter to the run CLI
- change the gocli package prefix

### DIFF
--- a/cluster-provision/okd/base/Dockerfile
+++ b/cluster-provision/okd/base/Dockerfile
@@ -1,10 +1,6 @@
 FROM fedora@sha256:a66c6fa97957087176fede47846e503aeffc0441050dd7d6d2ed9e2fae50ea8e
 
-# oc binary without SSL dependency
-ENV OC_PKG https://artifacts-openshift-master.svc.ci.openshift.org/repo/openshift-clients-4.0.0-0.alpha.0.1969.af45cda.x86_64.rpm
-
 RUN dnf install -y \
-${OC_PKG} \
 libvirt \
 libvirt-devel \
 libvirt-daemon-kvm \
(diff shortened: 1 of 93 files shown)


cluster-provision/okd/base/Dockerfile
```diff
@@ -1,24 +1,36 @@
...
-httpd-tools && \
+httpd-tools \
+python2-pip && \
 dnf clean all
 
+RUN pip2 install yq
+
+# install golang
+ENV GO_VERSION 1.12.12
+RUN curl https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz -o go.tar.gz && \
```
qinqon: This is going to make okd- providers bigger, but not much though, is that ok ?

oshoval:
Quique and i experienced a problem with /root/.ssh not found 
usually after something else major happend and the script had to collect logs
checking the container it indeed didnt have files at that folder

so maybe worth to add a tuple, either the exiting vagrant one (with renaming it)
or just to add 
ssh-keygen -t rsa -f /root/.ssh/id_rsa -N ''
before cluster create ?

thanks

cluster-provision/okd/base/Dockerfile
cynepco3hahue: the difference is 300MB, IMHO it is not critical when we have images of 14Gb :(

cynepco3hahue:
> Quique and i experienced a problem with /root/.ssh not found
> usually after something else major happend and the script had to collect logs
> checking the container it indeed didnt have files at that folder
> 
> so maybe worth to add a tuple, either the exiting vagrant one (with renaming it)
> or just to add
> ssh-keygen -t rsa -f /root/.ssh/id_rsa -N ''
> before cluster create ?
> 
> thanks

@oshoval Can you please open an issue and I will fix it under the next iteration?

oshoval:
> > Quique and i experienced a problem with /root/.ssh not found
> > usually after something else major happend and the script had to collect logs
> > checking the container it indeed didnt have files at that folder
> > so maybe worth to add a tuple, either the exiting vagrant one (with renaming it)
> > or just to add
> > ssh-keygen -t rsa -f /root/.ssh/id_rsa -N ''
> > before cluster create ?
> > thanks
> 
> @oshoval Can you please open an issue and I will fix it under the next iteration?

sure thanks
https://github.com/kubevirt/kubevirtci/issues/188

qinqon:
@cynepco3hahue looks like openshift installer libvirt is broken https://github.com/openshift/installer/pull/2657

PR #189, opened 2019-11-21
providers: okd: add podman support

ksimon1:
We need special treatment to make OKD clusters working with
the `pack8s` helper.
Differently from docker, we can't just call podman, because
we can be running with an user != root. In that case, the
copy step will fail. Hence, we use a exec command co copy files to mounted folder (between conteiner and host)
//cc @fromanirh 
This PR is upgrade of https://github.com/kubevirt/kubevirtci/pull/183
Signed-off-by: Karel Šimon <ksimon@redhat.com>

### DIFF
--- a/cluster-up/cluster/okd-4.1/provider.sh
+++ b/cluster-up/cluster/okd-4.1/provider.sh
@@ -12,11 +12,18 @@ function _install_from_cluster() {
     local src_cid="$1"
     local src_file="$2"
     local dst_perms="$3"
+    local dst_name="$4"
     local dst_file="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/$4"
+    local pack8s_path="$5"
 
-    touch $dst_file
+    if [ "$KUBEVIRTCI_RUNTIME" = "podman" ]; then
+        pack8s exec $src_cid cp $src_file /tmp/$dst_name
+        cp $pack8s_path/$dst_name $dst_file
+    else
+        touch $dst_file
+        docker exec $src_cid cat $src_file > $dst_file
+    fi
     chmod $dst_perms $dst_file
-    docker exec $src_cid cat $src_file > $dst_file
 }
 
 function up() {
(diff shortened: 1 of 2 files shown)


cluster-up/cluster/okd-4.1/provider.sh
```diff
@@ -36,13 +43,23 @@ function up() {
         params=" --ocp-console-port $OKD_CONSOLE_PORT ${params}"
     fi
 
+    if [ "$KUBEVIRTCI_RUNTIME" = "podman" ]; then
+        mkdir -p /tmp/pack8s
```
fromanirh: I'd use `mktemp` (with a suitable prefix, see man page), mostly to reduce the risk of the setup failing for existing directory. The other reason is for added security, but I don't think we need much security in this case.

cluster-up/cluster/okd-4.1/provider.sh
```diff
@@ -12,13 +12,20 @@ function _port() {
 
 function _install_from_cluster() {
     local src_cid="$1"
-    local src_file="$2"
-    local dst_perms="$3"
-    local dst_file="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/$4"
+    local src_file_path="$2"
+    local src_file_name="$3"
```
fromanirh: no need to change the function signature, you can use `basename`, see man page

cluster-up/cluster/okd-4.1/provider.sh
```diff
@@ -36,13 +43,23 @@ function up() {
...
+    else
+        cluster_container_id=$(docker ps -f "name=$provider_prefix-cluster" --format "{{.ID}}")
+    fi
     # Copy k8s config and kubectl
-    cluster_container_id=$(docker ps -f "name=$provider_prefix-cluster" --format "{{.ID}}")
 
-    _install_from_cluster $cluster_container_id /bin/oc 0755 .kubectl
-    _install_from_cluster $cluster_container_id /root/install/auth/kubeconfig 0644 .kubeconfig
+    _install_from_cluster $cluster_container_id /bin oc 0755 .kubectl
+    _install_from_cluster $cluster_container_id /root/install/auth kubeconfig 0644 .kubeconfig
```
fromanirh: I prefer to leave this unchanged, see above

cluster-up/cluster/okd-4.3/provider.sh
```diff
@@ -12,13 +12,20 @@ function _port() {
 
 function _install_from_cluster() {
```
fromanirh: same comments as per okd-4.1 provider

cluster-up/cluster/okd-4.3/provider.sh
```diff
@@ -12,13 +12,20 @@ function _port() {
...
-    local dst_file="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/$4"
+    local src_file_path="$2"
+    local src_file_name="$3"
+    local dst_perms="$4"
+    local dst_file="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/$5"
+    local src_file="$src_file_path/$src_file_name"
 
-    touch $dst_file
+    if [ "$KUBEVIRTCI_RUNTIME" = "podman" ]; then
+        pack8s exec $src_cid --commands "cp $src_file /tmp/$src_file_name"
```
fromanirh: outdated now :)

fromanirh:
I suggest to rename this PR (title) in " providers: okd: add podman support"

ksimon1:
@dhiller can you please review this PR?

ksimon1:
https://github.com/kubevirt/kubevirtci/issues/253

PR #190, opened 2019-11-26
Add Connect to container script

oshoval:
use cluster-up/container.sh in order to connect to the container
oc / kubectl commands will be available there since the KUBECONFIG is exported,
(also before it is populated outside of the container)

### DIFF
--- a/cluster-up/container.sh
+++ b/cluster-up/container.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [[ $KUBEVIRT_PROVIDER =~ okd.* ]]; then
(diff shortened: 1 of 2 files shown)


cynepco3hahue:
Please signoff the PR and also can you please provide the make target for the script.

oshoval:
/cc @qinqon  can you please review ?
thanks

cluster-up/container.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [[ $KUBEVIRT_PROVIDER =~ (ocp|okd).* ]]; then
```
cynepco3hahue: IIRC it supported only for OKD

PR #193, opened 2019-12-02
Update readme about make connect

oshoval:


cynepco3hahue:
@oshoval  Can you add the note that it will work only for the OKD provider?

cynepco3hahue:
Missed the fact that a new section was added only under the OKD section.
/lgtm
/approve

PR #195, opened 2019-12-04
Parameterize cluster dir

qinqon:


cluster-up/hack/common.sh
```diff
@@ -15,6 +15,8 @@ if [ -z "$KUBEVIRTCI_CONFIG_PATH" ]; then
     )"
 fi
 
+
+KUBEVIRTCI_CLUSTER_PATH=${KUBEVIRTCI_CLUSTER_PATH:-${KUBEVIRTCI_PATH}/cluster-up/cluter}
```
oshoval: cluster or cluter ?

cynepco3hahue: should be cluster

PR #198, opened 2019-12-05
Add ocp-4.3 provider

qinqon:

### DIFF
--- a/cluster-provision/ocp/4.3/publish.sh
+++ b/cluster-provision/ocp/4.3/publish.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+docker tag kubevirtci/ocp-4.3-provision:latest quay.io/openshift-cnv/kubevirtci-ocp-4.3:latest
+docker push quay.io/openshift-cnv/kubevirtci-ocp-4.3:latest
--- a/cluster-up/cluster/ocp-4.3/README.md
+++ b/cluster-up/cluster/ocp-4.3/README.md
@@ -1,0 +1,45 @@
+# OCP 4.3 in ephemeral containers
+
+Provides a pre-deployed OCP with version 4.3 purely in docker
+containers with libvirt. The provided VMs are completely ephemeral and are
+recreated on every cluster restart. The KubeVirt containers are built on the
+local machine and are then pushed to a registry which is exposed at
+`localhost:5000`.
+
+## Bringing the cluster up
+
+```bash
+export KUBEVIRT_PROVIDER=ocp-4.3
+export KUBEVIRT_NUM_NODES=3 # master + two workers
+make cluster-up
+```
+
+The cluster can be accessed as usual:
+
+```bash
+$ cluster/kubectl.sh get nodes
+NAME                          STATUS   ROLES    AGE   VERSION
+test-1-82xp6-master-0         Ready    master   62m   v1.12.4+509916ce1
+test-1-82xp6-worker-0-wxf27   Ready    worker   57m   v1.12.4+509916ce1
+```
+
+## Bringing the cluster down
+
+```bash
+export KUBEVIRT_PROVIDER=ocp-4.3
+make cluster-down
+```
+
+This destroys the whole cluster. Recreating the cluster is fast, since OCP is
+already pre-deployed. The only state which is kept is the state of the local
+docker registry.
+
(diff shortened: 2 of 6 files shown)


phoracek:
I think it would be nicer towards users if we fail with "Please provide URL of image XY" than "Not found. You don't have access to private images, tough luck".

cluster-up/cluster/ocp-4.3/README.md
```diff
@@ -0,0 +1,45 @@
...
+containers with libvirt. The provided VMs are completely ephemeral and are
+recreated on every cluster restart. The KubeVirt containers are built on the
+local machine and are then pushed to a registry which is exposed at
+`localhost:5000`.
+
+## Bringing the cluster up
+
+```bash
+export KUBEVIRT_PROVIDER=okd-4.3
+export KUBEVIRT_NUM_NODES=3 # master + two workers
```
dhiller: I thought we would only need one worker?

cluster-provision/ocp/4.3/publish.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+docker tag kubevirtci/ocp-4.3-provision:latest quay.io/openshift-cnv/kubevirtci-ocp-4.3:latest
+docker push quay.io/openshift-cnv/kubevirtci-ocp-4.3:latest
```
dhiller: I understand this repo is private? If so, how would we login here?

cluster-up/cluster/ocp-4.3/provider.sh
```diff
@@ -0,0 +1,54 @@
...
+    params="--random-ports --background --prefix $provider_prefix --master-cpu 6 --workers-cpu 6 --secondary-nics ${KUBEVIRT_NUM_SECONDARY_NICS} --registry-volume $(_registry_volume) --workers $workers --container-registry quay.io ${image}"
+    if [[ ! -z "${RHEL_NFS_DIR}" ]]; then
+        params=" --nfs-data $RHEL_NFS_DIR ${params}"
+    fi
+
+    if [[ ! -z "${OKD_CONSOLE_PORT}" ]]; then
+        params=" --ocp-console-port $OKD_CONSOLE_PORT ${params}"
+    fi
+
+    ${_cli} run okd ${params}
```
dhiller: Maybe we should add `ocp` at a later time instead of reusing `okd` here?

cluster-up/cluster/ocp-4.3/README.md
qinqon: This is just an example of how you can start it up with more workers if needed.

cluster-provision/ocp/4.3/publish.sh
qinqon: docker login, same we do with okd.

cluster-up/cluster/ocp-4.3/provider.sh
qinqon: Well is lready using all the OKD scripts, let's just add it in case it gets very different

qinqon:
@phoracek what about adding another env var so you can override the URL ?

phoracek:
@qinqon giving a default that is not useful for anyone outside the firewall feels really rude. I'd like to here someone else's take on this, maybe I'm too stubborn.

rmohr:
/approve
/cc @dhiller 

overall looks good. @dhiller can you look for obvious issues and lgtm?

oshoval:
great
once its merged, we should fix the helper script to support ocp as well
ssh.sh
container.sh 
and so on

cluster-up/cluster/ocp-4.3/provider.sh
```diff
@@ -0,0 +1,54 @@
...
+    local dst_file="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/$4"
+
+    touch $dst_file
+    chmod $dst_perms $dst_file
+    docker exec $src_cid cat $src_file > $dst_file
+}
+
+function up() {
+    workers=$(($KUBEVIRT_NUM_NODES-1))
+    if [[ ( $workers < 1 ) ]]; then
```
dhiller: Nit: as a matter of taste, I'd rather exit with an errorcode here, to make the usage more clear. But thats just IMHO.

PR #199, opened 2019-12-09
Fix external provider for openshift-ci usage

dhiller:
Fix needed to use the external provider, otherwise it won't find the configuration.

Signed-off-by: Daniel Hiller <daniel.hiller.1972@gmail.com>

### DIFF
--- a/cluster-up/cluster/external/provider.sh
+++ b/cluster-up/cluster/external/provider.sh
@@ -12,13 +12,20 @@ function prepare_config() {
         exit 1
     fi
 
-    cat > "${BASE_PATH}/$KUBEVIRT_PROVIDER/config-provider-$KUBEVIRT_PROVIDER.sh" <<EOF
+    PROVIDER_CONFIG_FILE_PATH="${BASE_PATH}/$KUBEVIRT_PROVIDER/config-provider-$KUBEVIRT_PROVIDER.sh"
+
+    cat > "$PROVIDER_CONFIG_FILE_PATH" <<EOF
 kubeconfig=\${KUBECONFIG}
 docker_tag=\${DOCKER_TAG}
 docker_prefix=\${DOCKER_PREFIX}
 manifest_docker_prefix=\${DOCKER_PREFIX}
 image_pull_policy=\${IMAGE_PULL_POLICY:-Always}
 EOF
+
+    if which oc; then
+        echo "oc=$(which oc)" >> "$PROVIDER_CONFIG_FILE_PATH"
+    fi
+
 }
 
 # The external cluster is assumed to be up.  Do a simple check
--- a/cluster-up/cluster/external/provider.sh
+++ b/cluster-up/cluster/external/provider.sh
@@ -8,9 +8,9 @@ function prepare_config() {
     BASE_PATH=${KUBEVIRTCI_CONFIG_PATH:-$PWD}
 
     # required for running tests within openshift ci, otherwise tests don't find the configuration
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/external/provider.sh
```diff
@@ -5,11 +5,17 @@ function _kubectl() {
...
 function prepare_config() {
-    cat >hack/config-provider-external.sh <<EOF
-docker_tag=devel
-docker_prefix=${DOCKER_PREFIX}
-manifest_docker_prefix=${DOCKER_PREFIX}
-image_pull_policy=${IMAGE_PULL_POLICY:-Always}
+    BASE_PATH=${KUBEVIRTCI_CONFIG_PATH:-$PWD}
+
+    # required for running tests within openshift ci, otherwise tests don't find the configuration
+    ln -f -s ${KUBECONFIG:-$HOME/.kube/config} ${BASE_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
```
danielBelenky: Could you wrap the expressions here with quotes? Though unlikely, but if KUBECONFIG will have a space in its path this command will fail.

I'd suggest quoting always unless you explicitly want bash to split your expression

dhiller: Yeah, thought so too. I'd rather fix all `shellcheck` warnings but didn't want to go too far here ;-)

dhiller: Done.

cluster-up/cluster/external/provider.sh
```diff
@@ -5,19 +5,24 @@ function _kubectl() {
...
 function prepare_config() {
-    cat >hack/config-provider-external.sh <<EOF
-docker_tag=devel
-docker_prefix=${DOCKER_PREFIX}
-manifest_docker_prefix=${DOCKER_PREFIX}
-image_pull_policy=${IMAGE_PULL_POLICY:-Always}
+    BASE_PATH=${KUBEVIRTCI_CONFIG_PATH:-$PWD}
+
+    # required for running tests within openshift ci, otherwise tests don't find the configuration
+    ln -f -s "${KUBECONFIG:-$HOME/.kube/config}" "${BASE_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig"
```
rmohr: I am not sure what you mean by that. You can set the config location on the tests binary. Is that what you try to solve here?

It probably does not work right now, but the provider is supposed to just find the config at a default location or in KUBECONFIG.

dhiller: Yes, that is exactly what I'm trying to solve here. Within openshift-ci `KUBECONFIG` is set to `/tmp/admin.kubeconfig`. Unfortunately the tests won't find the configuration unless I symlink to the target configuration.

I agree that this is a symptom, not the cause. Opened #202 for this.

dhiller:
/cc @rmohr @danielBelenky 

can you please re-review?

rmohr:
/lgtm
/approve
/hold

let's reverify that it works and take it in.

cluster-up/cluster/external/provider.sh
```diff
@@ -5,19 +5,33 @@ function _kubectl() {
...
+    if [ -z "${KUBECONFIG}" ]; then
+        echo "KUBECONFIG is not set!"
+        exit 1
+    fi
+
+    PROVIDER_CONFIG_FILE_PATH="${BASE_PATH}/$KUBEVIRT_PROVIDER/config-provider-$KUBEVIRT_PROVIDER.sh"
+
+    cat > "$PROVIDER_CONFIG_FILE_PATH" <<EOF
+kubeconfig=\${KUBECONFIG}
+docker_tag=\${DOCKER_TAG}
```
rmohr: should we fall back to `devel` here?

PR #201, opened 2019-12-10
Re-push okd-4.2 provider

qinqon:
It has being pushed with incompatible docker version and
it cannot be pulled with default docker version at some
redhat distros like fedora.


cynepco3hahue:
@qinqon I think it also relevant for 4.1

qinqon:
@cynepco3hahue I have use wrong base commit, I have to re-do

PR #208, opened 2019-12-12
WIP: okd-provider: rotate certs on startup

danielBelenky:
Since we use static images for the providers, we get a cert during
the provisioning of the provider that expires after 30 days.

With this change, every time we run the cluster, we force openshift
 to rotate the certs our tests.

*Relevant reading:*
https://docs.openshift.com/container-platform/4.1/authentication/certificates/service-serving-certificate.html#manually-rotate-service-ca_service-serving-certificate
https://docs.openshift.com/container-platform/4.1/backup_and_restore/disaster_recovery/scenario-3-expired-certs.html

Signed-off-by: Daniel Belenky <dbelenky@redhat.com>

### DIFF
--- a/cluster-provision/okd/4.1/constants
+++ b/cluster-provision/okd/4.1/constants
@@ -1,0 +1,4 @@
+# Environment file with constants used by the provider.
+# To use it, source it from your script.
+
+OCP_RELEASE_IMAGE=docker.io/kubevirtci/ocp-release:4.1.24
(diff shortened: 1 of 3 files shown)


oshoval:
@danielBelenky 
maybe we can take providers with dead cert, run it once on them, and then shutdown the vms 
run docker commit and publish ? (in prow)

in case we will want to revive a provider without reprovision 
for example if reprovision would failed due to a bug on openshift-installer
which happens of course

PR #212, opened 2019-12-16
Add scale-up script

oshoval:


cluster-up/scale.sh
```diff
@@ -0,0 +1,144 @@
...
+
+  while true
+  do
+     get_state
+     [ $STATE != "shut" ] || break
+     sleep 5
+  done
+
+  mv /tmp/el77-worker-1.img /tmp/el77-worker-ocp.img
+fi
```
alonSadan: do we need a seperate `if` for this , or this can be onserted into this `if` : https://github.com/kubevirt/kubevirtci/pull/212/files#diff-18e81d0671724ea26dcf46c3a229f272R91-R97

-

rmohr:
/hold

why is this needed and what does it try  to solve? I don't think that the ocp providers right now support KUBEVIRT_NUM_NODES. Why can't it just be integrated in that flow?

cluster-up/scale.sh
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/bash -x
+
+# Add Rhel 7.7 Worker
+
+CUSTOM_IMAGE=${CUSTOM_IMAGE:-"0"}
+PROVISION_MODE=${PROVISION_MODE:-"0"}
+REPO_FILE=${REPO_FILE:-"0"}
+
+BASE_IMAGE="rhel-server-7.7-x86_64-kvm.qcow2"
```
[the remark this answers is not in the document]
qinqon: +1 on parameterize it.

cluster-up/scale.sh
```diff
@@ -0,0 +1,144 @@
...
+fi
+
+dnf install -y https://repo.fedora.md/centos/7/virt/x86_64/ovirt-4.2/common/ansible-2.8.2-1.el7.noarch.rpm
+dnf install -y openshift-ansible openshift-clients jq
+
+NETLIST=$(virsh net-list | grep -v def | grep active | awk '{print $1}')
+
+cd /tmp
+if [ $CUSTOM_IMAGE = "0" ] || [ $PROVISION_MODE = "1" ]; then
+   qemu-img create -f qcow2 el77-worker-1.img 20G
```
qinqon: can we rename el77-worker-1.img to just worker-1.img so we remove more references to rhel-7 ?

cluster-up/scale.sh
```diff
@@ -0,0 +1,144 @@
...
+   virt-resize --expand /dev/sda1 $BASE_IMAGE el77-worker-1.img
+else
+   cp $CUSTOM_IMAGE el77-worker-1.img
+fi
+
+virt-customize -a el77-worker-1.img --root-password password:123456 --ssh-inject root:file:/root/.ssh/id_rsa.pub --selinux-relabel --timezone Europe/Berlin --hostname $NETLIST-worker-1 --uninstall cloud-init,kexec-tools,postfix
+virt-install --name $NETLIST-worker-1 --description "$NETLIST-worker-1" --os-type=linux --os-variant=rhel7.6 --vcpus=4 --ram 7168 --rng /dev/urandom --import --disk el77-worker-1.img --network network=$NETLIST --check all=off --connect qemu:///system --graphics none -m 52:54:00:72:5c:30 --noautoconsole --cpu host-passthrough
```
qinqon: add a os-variant calculation code before this:
something like
```bash
if [[ $BASE_IMAGE =~ rhel-server-7 ]]; then
VIRT_INSTALL_OS_VARIANT=rhel-7.6
else
VIRT_INSTALL_OS_VARIANT=fedora-28 // or whatever is good
fi

```
And then `then --os-variant=$VIRT_INSTALL_OS_VARIANT`

cluster-up/scale.sh
```diff
@@ -0,0 +1,144 @@
...
+  done
+fi
+
+# Enable nested virtualization
+run_ssh rmmod kvm-intel
+run_ssh "echo 'options kvm-intel nested=1' >> /etc/modprobe.d/dist.conf"
+run_ssh modprobe kvm-intel
+
+# Add vagrant public key
+run_ssh "echo 'ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEA6NF8iallvQVp22WDkTkyrtvp9eWW6A8YVr+kz4TjGYe7gHzIw+niNltGEFHzD8+v1I2YJ6oXevct1YeS0o9HZyN1Q9qgCgzUFtdOKLv6IedplqoPkcmF0aYet2PkEDo3MlTBckFXPITAMzF8dJSIFo9D8HfdOV0IAdx4O7PtixWKn5y2hMNG0zQPyUecp4pzC6kivAIhyfHilFR61RGL+GPXQ2MWZWFYbAGjyiYJnAmCP3NOTd0jMZEnDkbUvxhMmBYSdETk1rRgm+R4LOzFUGaHqHDLKLX+FIPKcF96hrucXzcWyLbIbEgE98OHlnVYCzRdK8jlqm8tehUc9c9WhQ== vagrant insecure public key' >> /root/.ssh/authorized_keys"
```
qinqon: You can generate the pub key from private so 
``` bash
vagrant_pub_key=$(ssh-keygen -y -f /vagrant.key)
run_ssh "echo $vagrant_pub_key >> /root/.ssh/authorized_keys"
```

cluster-up/scale.sh
```diff
@@ -0,0 +1,144 @@
...
+fi
+
+virt-customize -a el77-worker-1.img --root-password password:123456 --ssh-inject root:file:/root/.ssh/id_rsa.pub --selinux-relabel --timezone Europe/Berlin --hostname $NETLIST-worker-1 --uninstall cloud-init,kexec-tools,postfix
+virt-install --name $NETLIST-worker-1 --description "$NETLIST-worker-1" --os-type=linux --os-variant=rhel7.6 --vcpus=4 --ram 7168 --rng /dev/urandom --import --disk el77-worker-1.img --network network=$NETLIST --check all=off --connect qemu:///system --graphics none -m 52:54:00:72:5c:30 --noautoconsole --cpu host-passthrough
+
+function getip {
+  IP=$(virsh domifaddr $NETLIST-worker-1  | awk '{ print $4 }' | tail -2 | head -1 | awk -F/ '{print $1}')
+}
+
+function run_ssh {
```
qinqon: You can use HEREDOCS with ssh to run at embeeded script 
https://linuxize.com/post/bash-heredoc/#using-heredoc-with-ssh

cluster-up/scale.sh
```diff
@@ -0,0 +1,144 @@
...
+
+# Add vagrant public key
+run_ssh "echo 'ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEA6NF8iallvQVp22WDkTkyrtvp9eWW6A8YVr+kz4TjGYe7gHzIw+niNltGEFHzD8+v1I2YJ6oXevct1YeS0o9HZyN1Q9qgCgzUFtdOKLv6IedplqoPkcmF0aYet2PkEDo3MlTBckFXPITAMzF8dJSIFo9D8HfdOV0IAdx4O7PtixWKn5y2hMNG0zQPyUecp4pzC6kivAIhyfHilFR61RGL+GPXQ2MWZWFYbAGjyiYJnAmCP3NOTd0jMZEnDkbUvxhMmBYSdETk1rRgm+R4LOzFUGaHqHDLKLX+FIPKcF96hrucXzcWyLbIbEgE98OHlnVYCzRdK8jlqm8tehUc9c9WhQ== vagrant insecure public key' >> /root/.ssh/authorized_keys"
+
+# Add core user
+run_ssh useradd core
+run_ssh "echo core | passwd --stdin core"
+run_ssh mkdir -p /home/core/.ssh
+run_ssh cp .ssh/authorized_keys /home/core/.ssh
+run_ssh chown -R core /home/core/.ssh
```
qinqon: Use bash HEREDOC

oshoval:
@rmohr 
this one adds a custom os worker, for example rhel 7.7
KUBEVIRT_NUM_NODES is to scale with RHCOS workers once it will work, RHCOS scaling would use inheritance flow which is different from the scale-up
we followed openshift instructions on how to scale with rhel 7.7 for example

cluster-up/scale.sh
```diff
@@ -0,0 +1,144 @@
...
+
+cp $KUBECONFIG /tmp
+sed -i 's/127.0.0.1/192.168.126.11/' /tmp/kubeconfig
```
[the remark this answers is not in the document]
oshoval: since its in ' ' , it wont expend $PARAMETER
this ip is used as constant on more places, if crucial i will think how to fix it

cluster-up/scale.sh
oshoval: its possible indeed, just matter of taste if we want to separate the ansible run from the 
epilogue

cluster-up/scale.sh
oshoval: okd 4.3 still doesnt support fedora as far as i know but just rhel 7.6+ not even include 8.0 (rhcos / fcos as base images),
it would need to add a redundent var to scale.sh that i removed (soon will update the PR with the change), do we want it ?

rmohr:
> @rmohr
> this one adds a custom os worker, for example rhel 7.7
> KUBEVIRT_NUM_NODES is to scale with RHCOS workers once it will work, RHCOS scaling would use inheritance flow which is different from the scale-up
> we followed openshift instructions on how to scale with rhel 7.7 for example

Please create an extra provider which has the full setup (in this case mixed operating system versions for workers) done during the creation phase. There is no need for `scale-up`. kubevirtci ships fully contained, and offline bootable clusters, so at least all needed containers need to be prepulled during the provision phase already.

oshoval:
@rmohr 
ok thanks,
just a site note, also the make scale-up has a preprovision part, that we use,
does it change the picture ?
it takes a pre-provisioned qcow that include the ocp base image , start a vm with it, and uses ansbile official scaling procedure to add it to an existing cluster

rmohr:
Not really. Ansible will do all kinds of unpredictable things which we just want to not deal with when we just want to test our code changes.

oshoval:
i patch the ansible playbook to skip them , it was mandatory because else the base images are deleted from the CI,
so once we provision the ocp itself we also provision a qcow with the same ocp base, and then we mark the podman pull tasks of ansible to be skipped

rmohr:
I don't see why we need an extra command. Just create an extra lane, provision it an dship it so that it works with `make cluster-up`. I would want to avoid any extra complexities on the consumer side if not absolutely necessary.

oshoval:
ok thanks, will do

PR #223, opened 2019-12-24
Add provision test support for ocp

oshoval:


oshoval:
use 
in order to test provisioned local image before publishing it

cluster-up/cluster/ocp-4.3/provider.sh
```diff
@@ -44,7 +44,11 @@ function up() {
     user=$(echo $auth |awk -F: '{print $1}')
     password=$(echo $auth |awk -F: '{print $2}')
 
-    ${_cli} run okd ${params} --container-registry-user $user --container-registry-password $password
+    if [ -z $KUBEVIRTCI_PROVISION_CHECK ]; then
+      ${_cli} run okd ${params} --container-registry-user $user --container-registry-password $password --container-registry ${container_registry}
```
dankenigsberg: I don't understand the code here, but it is better to set
```
   params="$params --container-reg $container-registry"
fi
```
over writing a long line twice. Code duplication is evil.

oshoval: tried that at first
it cant work that way because you need --container-reg "" explicit
unless there is a way to do that

dankenigsberg:
```
else
    params="$params --container-reg ''"
```

oshoval: tried that before and also now
it gives this instead of just kubevirtci/ocp-4.3-provision:latest
2019/12/29 09:14:56 failed to download ''/kubevirtci/ocp-4.3-provision:latest: invalid reference format

oshoval: params=" --container-registry= $params"
this might work
tested it dry and it seems fine, but better to test it fully on next provision
meanwhile if its ok i will current code ?

cluster-up/cluster/images.sh
```diff
@@ -14,7 +14,13 @@ IMAGES[k8s-multus-1.13.3]="k8s-multus-1.13.3@sha256:c0bcf0d2e992e5b4d96a7bcbf988
 IMAGES[okd-4.1]="okd-4.1@sha256:e7e3a03bb144eb8c0be4dcd700592934856fb623d51a2b53871d69267ca51c86"
 IMAGES[okd-4.2]="okd-4.2@sha256:a830064ca7bf5c5c2f15df180f816534e669a9a038fef4919116d61eb33e84c5"
 IMAGES[okd-4.3]="okd-4.3@sha256:63abc3884002a615712dfac5f42785be864ea62006892bf8a086ccdbca8b3d38"
-IMAGES[ocp-4.3]="ocp-4.3@sha256:86a2720562809462803aa2de066a50cd725af141d4971daf04895d0b3d75d4cc"
+
+if [ -z $KUBEVIRTCI_PROVISION_CHECK ]; then
```
phoracek: I don't understand how is this related to "Add provision test support for ocp". Could you elaborate in the commit message and, if needed, in an inline comment?

Is the parameter newly introduced? If it is, could you document it somewhere?

cluster-up/cluster/images.sh
```diff
@@ -14,7 +14,13 @@ IMAGES[k8s-multus-1.13.3]="k8s-multus-1.13.3@sha256:c0bcf0d2e992e5b4d96a7bcbf988
...
+  IMAGES[ocp-4.3]="ocp-4.3@sha256:86a2720562809462803aa2de066a50cd725af141d4971daf04895d0b3d75d4cc"
+else
+  IMAGES[ocp-4.3]="ocp-4.3-provision:latest"
```
dhiller: I don't think it's a good idea to do this change here, as it's a ocp specific thing. I'd rather put this into the ocp 4.3 `provider.sh` and work towards cleaning up that prefix oddity in a later move.

cluster-up/cluster/ocp-4.3/provider.sh
dhiller: Additionally the indentation looks astray from the default here.

cluster-up/cluster/images.sh
oshoval: we need it also for okd i think, so maybe worth putting logic in a common place for them ?

oshoval: done, please review

cluster-up/cluster/images.sh
oshoval: wasnt new, but added to commit message, readme, and comment to .sh file
done, please review

cluster-up/cluster/ocp-4.3/provider.sh
oshoval: fixed both issues

README.md
```diff
@@ -100,3 +100,9 @@ Accessing OKD UI
 ```
 TODO - in the process of working out the details here.
 ```
+
+In order to check newly created provider run,
+this will point to the local created provider upon cluster-up
```
phoracek: Sorry, in afraid I still don't understand. What "created provider run" and "local created provider"? What does the check do? When would I use it?

cluster-up/cluster/images.sh
```diff
@@ -3,21 +3,27 @@
...
+    IMAGES[okd-4.2]="okd-4.2@sha256:a830064ca7bf5c5c2f15df180f816534e669a9a038fef4919116d61eb33e84c5"
+    IMAGES[okd-4.3]="okd-4.3@sha256:63abc3884002a615712dfac5f42785be864ea62006892bf8a086ccdbca8b3d38"
+    IMAGES[ocp-4.3]="ocp-4.3@sha256:86a2720562809462803aa2de066a50cd725af141d4971daf04895d0b3d75d4cc"
+fi
 export IMAGES
 
-image="${IMAGES[$KUBEVIRT_PROVIDER]:-${KUBEVIRT_PROVIDER}:latest}"
-export image
+IMAGE_SUFFIX=""
+if [[ $KUBEVIRT_PROVIDER =~ (ocp|okd).* ]]; then
```
phoracek: Why only for OpenShift and not for Kubernetes?

cluster-up/cluster/ocp-4.3/provider.sh
```diff
@@ -44,6 +44,13 @@ function up() {
     user=$(echo $auth |awk -F: '{print $1}')
     password=$(echo $auth |awk -F: '{print $2}')
 
+    # If provision test mode is on, use local image
+    if [ -z $KUBEVIRTCI_PROVISION_CHECK ]; then
+        params=" --container-registry ${container_registry} $params"
+    else
+        params=" --container-registry= $params"
```
phoracek: What does '--container-registry=' alone do?

README.md
oshoval: after some create a new provider, using ./cluster-provision/ocp/4.3/provision.sh
for example, it creates a container, runs openshift-installer in it, shutdown machines, and then commit the image (locally), before publishing it to quay, we should check cluster-up with that local image instead of taking the official one from quay.

another case is for example if we want to alter provision.sh to fix bugs / extend it with another rhel 7.7 machine, we should check it

this export allow us to quick point on the local provisioned container
it existed for k8s (with little problems that are addressed as well), i added it in a single interface for okd/ocp as well

cluster-up/cluster/images.sh
oshoval: just ocp okd have a "...-provision" suffix needed, (afterward we tag them in publish.sh),
k8s provided image doesnt have it

cluster-up/cluster/ocp-4.3/provider.sh
oshoval: it is mandatory that when you dont want gocli add a prefix like docker (default) / quay
to the image its going to look for,
this is the method for telling it to add a null string as prefix, needed when the image is local (after provision its local)

dhiller: IMHO it's a sign of a smell here that we're confusing the user with an empty string. I think we should rather use an empty string as default value for gocli and then set an external location explicitly instead of what we're doing now: setting the parameter to an empty string. But that's a follow up issue.

PR #228, opened 2020-01-13
Connect to the right cluster container

ormergi:
When connecting to the cluster container with "make connect",
the grep command returns more then one container id and raises an error.

Adding extra post-fix to the search pattern making the result more
accurate and returning the right container id.

### DIFF
--- a/cluster-up/container.sh
+++ b/cluster-up/container.sh
@@ -3,7 +3,8 @@
 set -e
 
 if [[ $KUBEVIRT_PROVIDER =~ (ocp|okd).* ]]; then
-    CONTAINER=$(docker ps | grep kubevirt | grep $KUBEVIRT_PROVIDER | awk '{print $1}')
+    POSTFIX="cluster"
+    CONTAINER=$(docker ps | grep kubevirt | grep "${KUBEVIRT_PROVIDER}-${POSTFIX}" | awk '{print $1}')
     if [ -z $CONTAINER ]; then
         echo "container was not found"
         exit 0


cluster-up/container.sh
```diff
@@ -3,7 +3,9 @@
 set -e
 
 if [[ $KUBEVIRT_PROVIDER =~ (ocp|okd).* ]]; then
-    CONTAINER=$(docker ps | grep kubevirt | grep $KUBEVIRT_PROVIDER | awk '{print $1}')
+    POSTFIX="cluster"
+    PATTERN="${KUBEVIRT_PROVIDER}-${POSTFIX}"
+    CONTAINER=$(docker ps | grep kubevirt | grep $PATTERN | awk '{print $1}')
```
oshoval: would suggest to add a grep cluster and leave it one liner, but up to you,
one liners are good because they can be adapted on other cases as well for example

dhiller: I'd second that, maybe just even compress into one `grep` (with added `cluster`) along the way.

ormergi: I use $PATTERN in order to separate the "what to search" logic form the "how to search" logic, in a way it will be more easy to test and debug in case the container name would change for any reason.
I left "grep kubevirt"  in case that there is for some unexpected reason another container with the same name but different image registry.
I understand if it may sound redundant or over-kill for this small example.

dhiller: I understand you want to make clear that in case of ocp and okd there's a POSTFIX required. I'm fine with that. However, `PATTERN` I find redundant, because grep always searches for a pattern.

Maybe we can settle on sth like this:
```
POSTFIX="cluster"
CONTAINER=$(docker ps | grep "kubevirtci/${KUBEVIRT_PROVIDER}-${POSTFIX}" | awk '{print $1}')
```

ormergi: We still need the "grep kubevirt" because the "cluster" postfix is on the name column and the "kubevirt" string is on the image name column.

Thanks for understanding, I will change the commit.

oshoval: grep -E '1.*2' this can be used if desired one grep, but no biggie,
imo a var on its own line, which always have fixed value is redundant,
even for debug of futuristic cases.

but up to you
LGTM

PR #229, opened 2020-01-14
Fix ssh.sh for ocp-4.3 provider

oshoval:
Since the worker ip might end in 51 or 52, haproxy need to be updated
accordingly in order to allow using ssh.sh to connect to the worker

### DIFF
--- a/cluster-up/cluster/images.sh
+++ b/cluster-up/cluster/images.sh
@@ -16,7 +16,7 @@ if [ -z $KUBEVIRTCI_PROVISION_CHECK ]; then
     IMAGES[okd-4.1]="okd-4.1@sha256:e7e3a03bb144eb8c0be4dcd700592934856fb623d51a2b53871d69267ca51c86"
     IMAGES[okd-4.2]="okd-4.2@sha256:a830064ca7bf5c5c2f15df180f816534e669a9a038fef4919116d61eb33e84c5"
     IMAGES[okd-4.3]="okd-4.3@sha256:63abc3884002a615712dfac5f42785be864ea62006892bf8a086ccdbca8b3d38"
-    IMAGES[ocp-4.3]="ocp-4.3@sha256:4f537cbf6cf068eec9eadf2e340a642f2fda97d1b59b0aeebd615c19ce8e4ca6"
+    IMAGES[ocp-4.3]="ocp-4.3@sha256:8f59d625852ef285d6ce3ddd6ebd3662707d2c0fab19772b61dd0aa0f6b41e5f"
 fi
 export IMAGES
 
--- a/cluster-provision/okd/scripts/run.sh
+++ b/cluster-provision/okd/scripts/run.sh
@@ -89,6 +89,10 @@ if [[ ${worker_node_ip} != "192.168.126.51" ]]; then
   <hostname>console-openshift-console.apps.test-1.tt.testing</hostname>
   <hostname>oauth-openshift.apps.test-1.tt.testing</hostname>
 </host>" --live --config
+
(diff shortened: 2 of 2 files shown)


cluster-up/cluster/images.sh
```diff
@@ -16,7 +16,7 @@ if [ -z $KUBEVIRTCI_PROVISION_CHECK ]; then
     IMAGES[okd-4.1]="okd-4.1@sha256:e7e3a03bb144eb8c0be4dcd700592934856fb623d51a2b53871d69267ca51c86"
     IMAGES[okd-4.2]="okd-4.2@sha256:a830064ca7bf5c5c2f15df180f816534e669a9a038fef4919116d61eb33e84c5"
     IMAGES[okd-4.3]="okd-4.3@sha256:63abc3884002a615712dfac5f42785be864ea62006892bf8a086ccdbca8b3d38"
-    IMAGES[ocp-4.3]="ocp-4.3@sha256:4f537cbf6cf068eec9eadf2e340a642f2fda97d1b59b0aeebd615c19ce8e4ca6"
+    IMAGES[ocp-4.3]="ocp-4.3@sha256:8f59d625852ef285d6ce3ddd6ebd3662707d2c0fab19772b61dd0aa0f6b41e5f"
```
rmohr: Where is this new sha coming from?

oshoval: i took the previous container image, 
docker create, cped the file there, commit, and published it to quay

rmohr: Is this the current way of updating this private image or are there jobs defined already?

oshoval: there is the way to create a whole new provider using provision.sh, 
there isnt automation, we run it manually, it takes couple hours to create, test and publish,
this quick method saves time and totally works, (tested)

running it now with the ocp-4.3 lane you created on #2950 of kubevirt
(it fixes a problem it detected there)

qinqon: we have jobs to publish images for ocp provider already, let's do this way for now, but we should run provider jobs to check that hash is like the one it creates.

oshoval: i did check both via catting the file in the container + checking the functional
and also cluster-up shows it is on prow

</host>' --live --config
Updated network test-1-27qv4 persistent config and live state
+ sed -i s/192.168.126.51/192.168.126.52/ /etc/haproxy/haproxy.cfg
+ pkill haproxy
+ haproxy -f /etc/haproxy/haproxy.cfg

rmohr: > there is the way to create a whole new provider using provision.sh,
> there isnt automation, we run it manually, it takes couple hours to create, test and publish,
> this quick method saves time and totally works, (tested)

Well the issue is that this is intransparent (it is not clear if the changes here are really the ones in the image) and while it looks like it may save time for you, it can very well cost the next one working on this extra time.

oshoval: i c, but i did triple checks, in the small chance it wont work, we would learn and do better next time, i didnt want to waste hours creating new provider, testing it (and creating a new qcow preprovisioned for scale-up that we need still)

run.sh is copied to the container as is as part of provision.sh 
i needed just that snippet so performed it manually

rmohr: > we have jobs to publish images for ocp provider already, let's do this way for now, but we should run provider jobs to check that hash is like the one it creates.

@qinqon the digest is not at all related to the content of the image. You will just get another random digest. Just pushing basically means that anything can be in there. Not that we did not do that in the past, but that is one of the reasons why we added the jobs.

PR #231, opened 2020-01-15
Use jq container at ocp-4.3 cluster-up

qinqon:
Using jq as a container we don't care a bout underlaying OS to
install jq.

### DIFF
--- a/cluster-up/cluster/ocp-4.3/provider.sh
+++ b/cluster-up/cluster/ocp-4.3/provider.sh
@@ -40,7 +40,7 @@ function up() {
     fi
 
     # The auth has the format base64(user:password)
-    auth=$(cat ~/.docker/config.json  | jq -r '.auths["'$container_registry'"]["auth"]' |base64 -d)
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/ocp-4.3/provider.sh
```diff
@@ -40,7 +40,7 @@ function up() {
     fi
 
     # The auth has the format base64(user:password)
-    auth=$(cat ~/.docker/config.json  | jq -r '.auths["'$container_registry'"]["auth"]' |base64 -d)
+    auth=$(cat ~/.docker/config.json  | docker run --rm -i imega/jq -r '.auths["'$container_registry'"]["auth"]' |base64 -d)
```
rmohr: Can you use a specific tag or a shasum?

qinqon: Add :1.6, I am runing it at check-up-ocp-4.3 job.

PR #237, opened 2020-01-16
Update ocp-4.3 provider with new version

qinqon:


oshoval:
it isnt good to run whole at once, they are depended, 
check-up runs atm the current hash, it is known to work with current
maybe we want it to run at most the new created one , so it should be the last of the 3

and the push must use the output of provision i guess,

what you think ?

qinqon:
@oshoval they are not dependant, they run standalone, let's wait to see if they passes if this is the key we have a sane ocp job set.

qinqon:
/test check-provision-ocp-4.3

Changed the secret, let's retry

qinqon:
/test check-up-ocp-4.3

testing with new hash generated at push-ocp-4.3

oshoval:
nice
but please pay attention that this is problematic, because after the push, the current used image cant be pulled, so ocp lane for the users is broken, we need to address the tagging such both images will be available (there is an issue open for that as well)

cluster-up/cluster/images.sh
```diff
@@ -15,8 +15,7 @@ if [ -z $KUBEVIRTCI_PROVISION_CHECK ]; then
     IMAGES[k8s-multus-1.13.3]="k8s-multus-1.13.3@sha256:c0bcf0d2e992e5b4d96a7bcbf988b98b64c4f5aef2f2c4d1c291e90b85529738"
     IMAGES[okd-4.1]="okd-4.1@sha256:e7e3a03bb144eb8c0be4dcd700592934856fb623d51a2b53871d69267ca51c86"
     IMAGES[okd-4.2]="okd-4.2@sha256:a830064ca7bf5c5c2f15df180f816534e669a9a038fef4919116d61eb33e84c5"
-    IMAGES[okd-4.3]="okd-4.3@sha256:63abc3884002a615712dfac5f42785be864ea62006892bf8a086ccdbca8b3d38"
```
oshoval: why removed okd ?

cluster-up/cluster/images.sh
```diff
@@ -15,8 +15,7 @@ if [ -z $KUBEVIRTCI_PROVISION_CHECK ]; then
...
-    IMAGES[ocp-4.3]="ocp-4.3@sha256:8f59d625852ef285d6ce3ddd6ebd3662707d2c0fab19772b61dd0aa0f6b41e5f"
+    IMAGES[ocp-4.3]="ocp-4.3@sha256:d3878b275c607b6808e82ba2cedff9410af8ec3bd45b1d5ea461e04bf0244768"
```
oshoval: if this is safe to merge , please do , because currently kubevirtci ocp-4.3 is broken since that push made the previous hash undownloadable

qinqon: Restored, thanks mate

cluster-up/cluster/images.sh
qinqon: Wrong key I supose :-), has testore it again.

oshoval:
LGTM

PR #242, opened 2020-01-20
Configure br_netfilter to fix DNS

booxter:
This is needed by SR-IOV tests that may need to resolve DNS (specifically, to fetch agent binary into Fedora VMIs).


booxter:
This is not required for the lane but would be of help for those trying to run the suite on older kernels like from CentOS 7.x that have these knobs disabled by default.

PR #243, opened 2020-01-21
Fix kubelet config for nodes

dhiller:
On nodes the kubelet config is meant to be extended so that
it contains specific configuration. However the approach in
extending it lead to a corrupt configuration file that overwrote
required parameters.

Side note: Add -c parameter for skipping cluster-down to
check-cluster-up.sh .


dhiller:
Tests on kubevirt still failing, looks like this is not the only issue...

/hold

PR #244, opened 2020-01-22
Upgrade k8s nodes to centos7-2001_01

qinqon:
We have found issues with the DNS libraries, upgrading them fixes the issue
Let's upgrade the centos7 version all together.

Fixing libs:
dhclient              x86_64      12:4.2.5-77.el7.centos       base      285 k
dhcp-common           x86_64      12:4.2.5-77.el7.centos       base      176 k
dhcp-libs             x86_64      12:4.2.5-77.el7.centos       base      133 k


dhiller:
Oh, interesting. I found another misconfiguration in my search for why the provider does not comply with the e2e tests. kubevirt/kubevirtci#243 just FYI @qinqon

qinqon:
@dhiller also something is weird at master, we (knmstate) end up with empty resolv.conf but we install newer networkmanager, weird thing is that only happend on master when I spin it with KUBEVIRT_NUM_NODES=2

rmohr:
> /test push-centos7

Ah right that was the right one.

Now you have to update the shasum with `sha256:ce64b07d7daf982da6659af35d7be6a1febe540421e93799a3154a04c4940318` from the build and after this calling the cluster build jobs.

cluster-provision/README.md
```diff
@@ -17,7 +17,7 @@
 * `kubevirtci/cli`: `sha256:1dd015dea4f12e6dcb8e31be3eeb677fed96f290ef4a4892a33c43d666053536`
 * `kubevirtci/gocli`: `sha256:8dc7a694e67fadfbb337d59dfc269253079e31dca62e5298361dd464a82adc4b`
 * `kubevirtci/base`: `sha256:850ac2e2828610b5f35f004f2a8a1ab23609a4c7891c8a1b68cbb7eef5f5dda0`
-* `kubevirtci/centos:1905_01`: `sha256:4b292b646f382d986c75a2be8ec49119a03467fe26dccc3a0886eb9e6e38c911`
+* `kubevirtci/centos:1905_01`: `sha256:6f2548dcc23489d0c945aef516781ae2ea678424c3760d1dafa0a83d29411713`
```
dhiller: Is there a change missing here? I mean you changed the centos version tag, right?

cluster-provision/centos7/build.sh
```diff
@@ -1,3 +1,10 @@
-#!/bin/bash
+#!/bin/bash -e
 
-docker build . -t kubevirtci/centos:1905_01
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+source ../images.sh
+
+echo $IMAGES
```
dhiller: This is a leftover of debugging I suppose?

dhiller:
Just holding in case you want to work on this still.

qinqon:
@dhiller It also contains the hash update for k8s 1.17.0 is the one that we are going to use a knmstate, let me know if we want to update the others.

dhiller:
IDK, I think _if_ we want to update we would also update the k8s versions to latest in the same step. AFAIR others are working for now, or did we have issues recently?

dhiller:
^^ @qinqon

qinqon:
@dhiller we have just foudn the issue at 1.16 and 1.17 but we are going to use 1.17 so just updating this is good enough I suppose.

PR #245, opened 2020-01-22
replace PF ID selector with vendor and device ID

phoracek:
Since the pfNames selector seems to be broken, we replace it with vendor
and deviceID. Due to this change, we may be moving more PFs to the pod than
needed. However, it should not be an issue.


danielBelenky:
let's run the corresponding test... when it passes, we can merge
/test kind-k8s-sriov-1.14.2

phoracek:
Resolved in https://github.com/kubevirt/kubevirtci/pull/248

Issue #238, opened 2020-01-16
ClusterIP services may not accessible in multi node 1.17.0 environment


mhenriks:
It appears that clusterip services are only accessible on nodes that have an endpoint for that service.  

For example, take the `cdi-api` service.  It has one endpoint on node02.

```
➜  containerized-data-importer git:(master) ✗ k get services -n cdi cdi-api
NAME      TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
cdi-api   ClusterIP   10.96.239.43   <none>        443/TCP   20m

➜  containerized-data-importer git:(master) ✗ k get endpoints -n cdi cdi-api
NAME      ENDPOINTS         AGE
cdi-api   10.244.1.5:8443   17m

➜  containerized-data-importer git:(master) ✗ k get endpoints -n cdi cdi-api -o yaml
apiVersion: v1
kind: Endpoints
metadata:
  annotations:
    endpoints.kubernetes.io/last-change-trigger-time: "2020-01-16T15:58:17Z"
  creationTimestamp: "2020-01-16T15:58:11Z"
  labels:
    cdi.kubevirt.io: cdi-apiserver
    operator.cdi.kubevirt.io/createVersion: latest
  name: cdi-api
  namespace: cdi
  resourceVersion: "999"
  selfLink: /api/v1/namespaces/cdi/endpoints/cdi-api
  uid: 480ddcea-5bde-436a-b8e3-d93af2db8204
subsets:
- addresses:
  - ip: 10.244.1.5
    nodeName: node02
    targetRef:
      kind: Pod
      name: cdi-apiserver-7859f9fd8c-xjvrj
      namespace: cdi
      resourceVersion: "997"
      uid: 31cbb07b-39af-4ecc-83c3-323bad15a836
  ports:
  - port: 8443
    protocol: TCP
```

The service is accessible from node02

```
[vagrant@node02 ~]$ curl --insecure https://10.96.239.43/apis
{
 "kind": "APIGroupList",
 "apiVersion": "v1",
 "groups": [
  {
   "kind": "APIGroup",
   "apiVersion": "v1",
   "name": "upload.cdi.kubevirt.io",
   "versions": [
    {
     "groupVersion": "upload.cdi.kubevirt.io/v1alpha1",
     "version": "v1alpha1"
    }
   ],
   "preferredVersion": {
    "groupVersion": "upload.cdi.kubevirt.io/v1alpha1",
    "version": "v1alpha1"
[...]

PR #246, opened 2020-01-23
Create k8s provider with Fedora31 nodes.

ormergi:
Creating new provider with k8s 1.17.0 version which deploys fedora31 nodes (instead of centos7).

This allow you to test the latest packages update on a full functioning cluster (e.g: copr repos, nmstate).

*Currently all the images are on a different registry, when this will approved I will upload it to kubevirtci registry.


cluster-provision/cli/cli
```diff
@@ -190,6 +190,14 @@ if [ "$COMMAND" == "provision" ] ; then
   docker exec ${VM_CID} /bin/bash -c "scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i vagrant.key -P 22 /manifests/ceph/*.yaml vagrant@192.168.66.101:/tmp/ceph"
   docker exec ${VM_CID} /bin/bash -c "ssh.sh mkdir -p /tmp/cna"
   docker exec ${VM_CID} /bin/bash -c "scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i vagrant.key -P 22 /manifests/cna/*.yaml vagrant@192.168.66.101:/tmp/cna"
+  if [[ ${PREFIX} =~ "fedora" ]]; then
```
qinqon: No need for quotes

cluster-provision/fedora/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM kubevirtci/base
+
+ARG fedora_version
+
+RUN echo "Fedroa31 version $fedora_version"
```
qinqon: Typo: Fedora

cluster-provision/fedora/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM kubevirtci/base
```
qinqon: We need a hash here.

cluster-provision/cli/cli
```diff
@@ -190,6 +190,14 @@ if [ "$COMMAND" == "provision" ] ; then
...
+    # Workaround for for vms that runs fedora31+ and docker,
+    # In order to work with docker on Fedora31 it is necessary to revert back to cgroups v1.
+    echo "Configuring cgroup version"
+    docker exec ${VM_CID} /bin/bash -c "ssh.sh sudo /bin/bash -s < /scripts/cgroups.sh &"
```
qinqon: is more elegant to use wait [1] you just get pid from proces and pass it to wait
https://www.gnu.org/software/bash/manual/html_node/Job-Control-Builtins.html#index-wait

also implement all that in the cgroups.sh script so it feels like a blocking script outside, doing conditionals with docker exec lines is super weird.

cluster-provision/images.sh
```diff
@@ -3,6 +3,8 @@
 set -e
 
 declare -A IMAGES
+IMAGES[fedora31]=""
```
qinqon: We don't need it ?

cluster-provision/k8s-fedora/1.17.0/provision.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
qinqon: cluster-provision/k8s/1.17.0/fedora/...

cluster-up/cluster/images.sh
```diff
@@ -5,6 +5,7 @@ set -e
 declare -A IMAGES
 IMAGES[gocli]="gocli@sha256:e48c7285ac9e4e61fe0f89f35ac5f9090497ea7c8165deeadb61e464c88d8afd"
 if [ -z $KUBEVIRTCI_PROVISION_CHECK ]; then
+    IMAGES[k8s-fedora-1.17.0]="k8s-fedora-1.17.0"
```
qinqon: Create a push-k8s-fedora-1.17.0 prow job and update this has with the output from that.

cluster-provision/fedora/Dockerfile
ormergi: There is no hash because I built a new base image and uploaded it to my own registry because I dont have permissions to kubevirtci registry

dhiller:
Just a general question: you are aware that we have some misconfiguration in the 1.17 provider, right? Did you find anything about it that fixed the behaviour we observed? Please see kubevirtci#238 for more details.

Maybe you can check by deploying kubevirt on the locally provisioned cluster and running the tests, should work by using `KUBEVIRTCI_PROVISION_CHECK=1`.

cluster-provision/cli/cli
ormergi: done

cluster-provision/fedora/Dockerfile
ormergi: done

cluster-provision/fedora/Dockerfile
dhiller: I think Quique already wrote it somewhere else, there should be a prow job to push the images (or you need to create a new one). Prow has the credentials, we don't push the images from local.

cluster-provision/images.sh
dhiller: I suspect this is a placeholder where you plan to insert the image sha.

cluster-provision/k8s-fedora/check-cluster-up.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
dhiller: IDK, maybe we can just use a symlink here instead of copying?

cluster-provision/k8s-fedora/scripts/cgroups.sh
```diff
@@ -0,0 +1,41 @@
...
+
+os=$(get_os_id)
+os_version=$(get_os_major_version)
+cgroup_hierarchy=$(get_cgroup_hierarchy)
+echo "OS: $os $os_version unified_cgroup_hierarchy: $cgroup_hierarchy"
+if [[ $os =~ $TARGET_ID ]]; then    
+    if [ $os_version == $TARGET_ID_VERSION ]; then
+        # Revert cgroup to v1 in order to work with docker.
+        if [ $cgroup_hierarchy != 0 ]; then
+            grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
```
oshoval: instead of using grubby, wont it be better to manually update grub and build it ?

oshoval:
nice
one thing , even not on this PR scope, 
maybe worth to think on about a way to not duplicate the files, but just souce them and override the needed stuff ?

dhiller:
> nice
> one thing , even not on this PR scope,
> maybe worth to think on about a way to not duplicate the files, but just souce them and override the needed stuff ?

Thinking the same, please keep in mind that you can check in symlinks with git.

cluster-provision/images.sh
ormergi: Exactly :)
This  is a placeholder fot the VM container image sha.

cluster-provision/k8s-fedora/scripts/cgroups.sh
ormergi: IMO with grubby it is more straightforward.

cluster-provision/k8s-fedora/check-cluster-up.sh
ormergi: I copied it is case it will be necessary to add more tests in case nodes are fedora, your call.

cluster-provision/k8s-fedora/1.17.0/provision.sh
ormergi: I structured the directories that way in order to keep that same structure as k8s directory,
then we could add more directories for other k8s versions.

cluster-provision/k8s-fedora/scripts/cgroups.sh
oshoval: instead of modifying /etc/default/grub file and using grub2-mkconfig ?
np

cluster-provision/cli/cli
ormergi: Just to check if i got it,
do you mean remove the If clause, remove the while ssh_ready loop
and put wait before invoking cgroups.sh script?

cluster-provision/fedora/Dockerfile
ormergi: Will do

ormergi:
> Just a general question: you are aware that we have some misconfiguration in the 1.17 provider, right? Did you find anything about it that fixed the behaviour we observed? Please see #238 for more details.
> 
> Maybe you can check by deploying kubevirt on the locally provisioned cluster and running the tests, should work by using `KUBEVIRTCI_PROVISION_CHECK=1`.

Yes I checked and the same problem occurs, there is an open issue in flannel repo for this problem
https://github.com/coreos/flannel/issues/1245

ormergi:
> > Just a general question: you are aware that we have some misconfiguration in the 1.17 provider, right? Did you find anything about it that fixed the behaviour we observed? Please see #238 for more details.
> > Maybe you can check by deploying kubevirt on the locally provisioned cluster and running the tests, should work by using `KUBEVIRTCI_PROVISION_CHECK=1`.
> 
> Yes I checked and the same problem occurs, there is an open issue in flannel repo for this problem
> [coreos/flannel#1245](https://github.com/coreos/flannel/issues/1245)

A workaround for this issue #256

ormergi:
> /test check-provision-k8s-fedora-1.17.0

I think we need to build "cluster-provision/base" and "cluster-provision/fedora" images first.

qinqon:
/hold

We have break the k8s provision with this.

qinqon:
error:
```

Your Kubernetes control-plane has initialized successfully!

To start using your cluster, you need to run the following as a regular user:

  mkdir -p $HOME/.kube
  sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
  sudo chown $(id -u):$(id -g) $HOME/.kube/config

You should now deploy a pod network to the cluster.
Run "kubectl apply -f [podnetwork].yaml" with one of the options listed at:
  https://kubernetes.io/docs/concepts/cluster-administration/addons/

Then you can join any number of worker nodes by running the following on each as root:

kubeadm join 192.168.66.101:6443 --token abcdef.1234567890123456 \
    --discovery-token-ca-cert-hash sha256:f4e023786e5c238de1d22f070685c63266f25fe55bed797f27e8e21ed9bfef80 
/bin/bash: line 119: syntax error near unexpected token `fi'
```

cluster-provision/fedora/Dockerfile
qinqon: Have run push-base job so we will have a hash soon.

qinqon: sha256:cb9a1cb7740f9b717d4988ffe3d5a7e41f38ed9b2fe9013458b225de19b7c3b4

ormergi: done

cluster-provision/images.sh
qinqon: job running, let's wait for it to finish so we have the final hash.

qinqon:
/hold cancel

We have calico now.

dhiller:
@ormergi you need to rebase.

ormergi:
> @ormergi you need to rebase.

done

PR #248, opened 2020-01-27
Fix kind lane issues: pin to sriov-config-daemon version.

fedepaol:
Also, move all the pfs into the node's namespace so the SR-IOV operator
does not gets confused from what it sees in sysfs and what devices are in the
current namespace.

This very same changes where tested successfully in https://github.com/kubevirt/kubevirt/pull/3023
Sample runs:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3023/pull-kubevirt-e2e-kind-k8s-sriov-1.14.2/1221708467652268032
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3023/pull-kubevirt-e2e-kind-k8s-sriov-1.14.2/1220762234578276352

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
@@ -33,7 +33,7 @@ function deploy_sriov_operator {
     # be merged in openshift sriov operator. We need latest since that feature was not tagged yet
     sed -i '/SRIOV_CNI_IMAGE/!b;n;c\              value: nfvpe\/sriov-cni' ./deploy/operator.yaml
     sed -i 's#image: quay.io/openshift/origin-sriov-network-operator$#image: quay.io/openshift/origin-sriov-network-operator:4.2#' ./deploy/operator.yaml
-
+    sed -i 's#value: quay.io/openshift/origin-sriov-network-config-daemon$#value: quay.io/openshift/origin-sriov-network-config-daemon:4.2#' ./deploy/operator.yaml
     # on prow nodes the default shell is dash and some commands are not working
     make deploy-setup-k8s SHELL=/bin/bash OPERATOR_EXEC="${KUBECTL}"
   popd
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
```diff
@@ -117,4 +119,4 @@ deploy_network_resource_injector
 sleep 5
 wait_pods_ready
 
-${SRIOV_NODE_CMD} chmod 666 /dev/vfio/vfio
+${SRIOV_NODE_CMD} chmod 666 /dev/vfio/vfio
```
phoracek: Please add a newline

cluster-up/cluster/kind-k8s-sriov-1.14.2/config_sriov.sh
```diff
@@ -78,16 +78,18 @@ for ifs in "${sriov_pfs[@]}"; do
...
 
-  # These values are used to populate the network definition policy yaml.
-  # We just use the first suitable pf
-  # We need the num of vfs because if we don't set this value equals to the total, in case of mellanox
-  # the sriov operator will trigger a node reboot to update the firmware
-  export NODE_PF="$ifs_name"
-  export NODE_PF_NUM_VFS=$(cat /sys/class/net/"$NODE_PF"/device/sriov_totalvfs)
-  break
+
+  if [ -z "$NODE_PF" ]; then
```
phoracek: Why do we need this change? Is it so the value is set only during the first iteration? If so, could you either refactor the loop or at least add a comment before this line explaining it?

fedepaol: > Why do we need this change? Is it so the value is set only during the first iteration? If so, could you either refactor the loop or at least add a comment before this line explaining it?

Even though it's not probably related to the CI failures, there were a lot of errors in the node configuration daemon because it was finding devices in sysfs but it wasn't find the related link in the namespace (this is because we were moving only the first device in the node's name space).
With this change, the configuration daemon will see the same things it would in a non-dockerized environment, preventing additional errors.

We pick only one PF to inject it in the network configuration as we were doing before.

phoracek: I still don't get the logic of `-z` here.

fedepaol: It's to choose the first (one) PF as I wrote before:

We pick only one PF to inject it in the network configuration as we were doing before.

booxter: Probably a deficiency in the operator if that breaks it, but I guess since we have blacklists, it's good enough for environments that rely on a PF for host connectivity. I am not thrilled but I'm ok with it.

PR #256, opened 2020-01-29
Deploy Calico cni with k8s-1.17

ormergi:
This PR follows #254 draft,
flannel cni plugin has comparability issues with k8s-1.17 https://github.com/coreos/flannel/issues/1245,
causing slow connectivity between services over cluster nodes.

This PR deploys calico pod network plugin over k8s-1.17 instead also for better performance.


cluster-provision/k8s/scripts/provision.sh
```diff
@@ -104,14 +104,20 @@ sysctl --system
 echo bridge >> /etc/modules
 echo br_netfilter >> /etc/modules
 
-kubeadm init --pod-network-cidr=10.244.0.0/16 --kubernetes-version v${version} --token abcdef.1234567890123456
-flannel_manifest="/tmp/flannel.yaml"
-if [[ $minor_version -ge "16" ]]; then
-    flannel_manifest="/tmp/flannel-ge-16.yaml"
+pod_cird="10.244.0.0/16"
```
dhiller: s/pod_cird/pod_cidr/

cluster-provision/k8s/scripts/node01.sh
```diff
@@ -10,14 +10,18 @@ done
 
 kubeadm init --config /etc/kubernetes/kubeadm.conf
 
+pod_cird="10.244.0.0/16"
```
dhiller: s/pod_cird/pod_cidr/

cluster-provision/k8s/scripts/node01.sh
```diff
@@ -10,14 +10,18 @@ done
...
 version=`kubectl version --short --client | cut -d":" -f2 |sed  's/ //g' | cut -c2- | cut -d"." -f2`
-flannel_manifest="/tmp/flannel.yaml"
-if [[ $version -ge "16" ]]; then
-    flannel_manifest="/tmp/flannel-ge-16.yaml"
+network_plugin_manifest="/tmp/flannel.yaml"
+if [[ $version -ge "17" ]]; then
+    network_plugin_manifest="/tmp/calico.yaml"
+    sed -i -e "s?192.168.0.0/16?$pod_cird?g" "$network_plugin_manifest"
```
dhiller: s/pod_cird/pod_cidr/

cluster-provision/k8s/scripts/provision.sh
```diff
@@ -104,14 +104,20 @@ sysctl --system
...
+kubeadm init --pod-network-cidr=$pod_cird --kubernetes-version v${version} --token abcdef.1234567890123456
+
+network_plugin_manifest="/tmp/flannel.yaml"
+if [[ $minor_version -ge "17" ]]; then 
+    network_plugin_manifest="/tmp/calico.yaml"
+    sed -i -e "s?192.168.0.0/16?$pod_cird?g" "$network_plugin_manifest"
```
dhiller: s/pod_cird/pod_cidr/

cluster-provision/k8s/scripts/node01.sh
```diff
@@ -10,14 +10,20 @@ done
...
+pod_cidr="10.244.0.0/16"
 version=`kubectl version --short --client | cut -d":" -f2 |sed  's/ //g' | cut -c2- | cut -d"." -f2`
-flannel_manifest="/tmp/flannel.yaml"
-if [[ $version -ge "16" ]]; then
-    flannel_manifest="/tmp/flannel-ge-16.yaml"
+
+network_plugin_manifest="/tmp/flannel.yaml"
+if [[ $version -ge "17" ]]; then
+    network_plugin_manifest="/tmp/calico.yaml"
+    sed -i -e "s?$default_cidr?$pod_cird?g" "$network_plugin_manifest"
```
dhiller: s/pod_cird/pod_cidr/

cluster-provision/k8s/scripts/provision.sh
```diff
@@ -104,14 +104,21 @@ sysctl --system
...
-if [[ $minor_version -ge "16" ]]; then
-    flannel_manifest="/tmp/flannel-ge-16.yaml"
+default_cidr="192.168.0.0/16"
+pod_cidr="10.244.0.0/16"
+kubeadm init --pod-network-cidr=$pod_cird --kubernetes-version v${version} --token abcdef.1234567890123456
+
+network_plugin_manifest="/tmp/flannel.yaml"
+if [[ $minor_version -ge "17" ]]; then 
+    network_plugin_manifest="/tmp/calico.yaml"
+    sed -i -e "s?$default_cidr?$pod_cird?g" "$network_plugin_manifest"
```
dhiller: s/pod_cird/pod_cidr/

cluster-provision/manifests/calico.yaml
```diff
@@ -0,0 +1,793 @@
...
+  name: calico-kube-controllers
+  namespace: kube-system
+---
+# Source: calico/templates/calico-etcd-secrets.yaml
+
+---
+# Source: calico/templates/calico-typha.yaml
+
+---
+# Source: calico/templates/configure-canal.yaml
```
phoracek: please add a newline

cluster-provision/k8s/scripts/node01.sh
```diff
@@ -10,14 +10,20 @@ done
 
 kubeadm init --config /etc/kubernetes/kubeadm.conf
```
phoracek: maybe we should add `--pod-network-cidr=$pod_cidr` here as well?

cluster-provision/manifests/calico.yaml
```diff
@@ -0,0 +1,793 @@
+---
```
phoracek: Please add a note that this file is copied from Calico documentation and that no changes should be done on it (we have the `sed` replacing the default route and that is good, when we want a new version of calico, we just copy a new file from docs, no need to adjust it here.

cluster-provision/k8s/scripts/node01.sh
ormergi: When invoking 'kubeadm init' you cannot combine --config with any other parameter.
I verified and it works like that.

cluster-provision/manifests/calico.yaml
ormergi: done

cluster-provision/k8s/scripts/node01.sh
ormergi: done

cluster-provision/k8s/scripts/provision.sh
ormergi: done

cluster-provision/k8s/scripts/provision.sh
ormergi: done

cluster-provision/k8s/scripts/node01.sh
ormergi: done

cluster-provision/k8s/scripts/node01.sh
ormergi: done

cluster-provision/k8s/scripts/provision.sh
ormergi: done

cluster-provision/manifests/calico.yaml
ormergi: done

cluster-provision/k8s/scripts/node01.sh
```diff
@@ -10,14 +10,20 @@ done
...
 
+default_cidr="192.168.0.0/16"
+pod_cidr="10.244.0.0/16"
 version=`kubectl version --short --client | cut -d":" -f2 |sed  's/ //g' | cut -c2- | cut -d"." -f2`
-flannel_manifest="/tmp/flannel.yaml"
-if [[ $version -ge "16" ]]; then
-    flannel_manifest="/tmp/flannel-ge-16.yaml"
+
+network_plugin_manifest="/tmp/flannel.yaml"
```
qinqon: I have do a search and no one is using node01.sh, I think it's safe to delete so we don't have the flannel/calico installation logic duplicated.

cluster-provision/manifests/calico.yaml
qinqon: Can we "curl" it ?

phoracek:
I don't think we should curl. We want to have this as stable and self-contained as possible. With the node01, you may be up to something.

cluster-provision/k8s/scripts/node01.sh
ormergi: I think this script is being used [here](https://github.com/kubevirt/kubevirtci/blob/db3913a3a919bbc20c3b7044dffc70c009fdc3bd/cluster-provision/cli/cli#L242)

qinqon: Then let's put the plugin selection at common place, the logic is big enough now for that since we have the default cidr replacement and all.

cluster-provision/manifests/calico.yaml
qinqon: ok let's n ot curl it.

cluster-provision/k8s/scripts/node01.sh
ormergi: Done

qinqon:
1.17 provider is failing, this feels wrong `kubectl apply -f [podnetwork].yaml`:
```
You should now deploy a pod network to the cluster.
Run "kubectl apply -f [podnetwork].yaml" with one of the options listed at:
  https://kubernetes.io/docs/concepts/cluster-administration/addons/

Then you can join any number of worker nodes by running the following on each as root:

kubeadm join 192.168.66.101:6443 --token abcdef.1234567890123456 \
    --discovery-token-ca-cert-hash sha256:dfcf0bd9b5f3c549d7d258b4bde5704642df26489eef418c11f2ebedaa092188 
+ default_cidr=192.168.0.0/16
+ pod_cidr=10.244.0.0/16
++ kubectl version --short --client
++ cut -d: -f2
++ sed 's/ //g'
++ cut -c2-
++ cut -d. -f2
+ version=17
+ kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f ''
Error: must specify one of -f and -k

Create a resource from a file or from stdin.
```

qinqon:
So this is the command failing:
```
docker run --privileged --net=host --rm -v /var/run/docker.sock:/var/run/docker.sock kubevirtci/gocli@sha256:e48c7285ac9e4e61fe0f89f35ac5f9090497ea7c8165deeadb61e464c88d8afd run --container-registry= --nodes 2 --memory 5120M --cpu 5 --secondary-nics 0 --random-ports --background --prefix check-provision-k8s-1.17.0 --registry-volume check-provision-k8s-1.17.0_registry kubevirtci/k8s-1.17.0:latest
```
Do we have to re-create gocli container to include cli/cli changes ?

cluster-provision/k8s/scripts/node01.sh
```diff
@@ -2,6 +2,10 @@
 
 set -ex
 
+source cnis-map.sh
```
qinqon: Use KUBEVIRTCI_PATH

PR #266, opened 2020-02-04
Add ocp-cnao-4.4

qinqon:
It uses CNAO 0.25.0

### DIFF
--- a/cluster-provision/ocp-cnao/4.4/provision.sh
+++ b/cluster-provision/ocp-cnao/4.4/provision.sh
@@ -1,0 +1,39 @@
+#!/bin/bash -x
+
+PARENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/../.. && pwd )"
+KUBEVIRTCI_DIR="$( cd ${PARENT_DIR}/../kubevirtci && pwd)"
+
+okd_base_hash="sha256:73ede51ce464546a82b81956b7f58cf98662a4c5fded9c659b57746bc131e047"
+gocli_image_hash="sha256:220f55f6b1bcb3975d535948d335bd0e6b6297149a3eba1a4c14cad9ac80f80d"
+prefix=ocp-cnao-4.4-provision
+
+gocli="docker run \
+--privileged \
+--net=host \
+--rm -t \
+-v /var/run/docker.sock:/var/run/docker.sock \
+-v ${PARENT_DIR}:${PARENT_DIR} \
+docker.io/kubevirtci/gocli@${gocli_image_hash}"
+
+provisioner_container_id=$(docker ps --filter name=$prefix-cluster --format {{.ID}})
+docker kill $provisioner_container_id
+docker container rm $provisioner_container_id
+
+${gocli} provision okd \
+--prefix $prefix \
+--dir-scripts ${PARENT_DIR}/okd/scripts \
+--dir-manifests ${PARENT_DIR}/manifests \
+--dir-hacks ${PARENT_DIR}/okd/hacks \
+--master-memory 10240 \
+--workers-memory 8192 \
+--workers-cpu 4 \
+--networking-type OVNKubernetes \
+--installer-pull-secret-file ${INSTALLER_PULL_SECRET} \
+--installer-repo-tag release-4.4 \
+--installer-release-image registry.svc.ci.openshift.org/ocp/release:4.4 \
+"kubevirtci/okd-base@${okd_base_hash}"
+rc=$?
+
+
+
+exit $rc
--- a/cluster-provision/okd/scripts/provision.sh
+++ b/cluster-provision/okd/scripts/provision.sh
@@ -208,15 +208,24 @@ oc adm policy add-cluster-role-to-user cluster-admin admin
 
 if [ "${CNAO}" == "true" ]; then
     # Apply network addons
-    oc create -f /manifests/cna/namespace.yaml
-    oc create -f /manifests/cna/network-addons-config.crd.yaml
-    oc create -f /manifests/cna/operator.yaml
-    oc create -f /manifests/cna/network-addons-config-example.cr.yaml
 
-     # Wait until all the network components are ready
-    until oc wait networkaddonsconfig cluster --for condition=Available --timeout=100s; do
-        sleep 10
+    # TODO: pass CNAO version at gocli
+    cnao_version=0.25.0
+    cnao_path=/manifests/cnao/$cnao_version/
+    oc create -f $cnao_path/namespace.yaml
+    oc create -f $cnao_path/network-addons-config.crd.yaml
+    oc create -f $cnao_path/operator.yaml
+
+    # Wait until cnao is ready
+    oc wait deployment -n cluster-network-addons cluster-network-addons-operator --for condition=Available --timeout=5m
+
+    # Import CNAO network components images to store them at provider
+    cnao_subimages=$(oc get deployment -n cluster-network-addons cluster-network-addons-operator -o json \
+        |jq -r '.spec.template.spec.containers[0].env | [ .[] | select( .name | contains("IMAGE")) ][].value')
+    for subimage in $cnao_subimages; do
+        oc import-image $subimage --confirm
     done
+
 fi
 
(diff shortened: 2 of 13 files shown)


.gitignore
```diff
@@ -9,3 +9,6 @@ _ci-configs/*
 
 # Bazel directories
 cluster-provision/gocli/bazel-*
+
+# vim swap files
+*.sw*
```
phoracek: This should be in your own globalignore, not in the one for this project.

cluster-provision/okd/scripts/provision.sh
```diff
@@ -208,10 +208,14 @@ oc adm policy add-cluster-role-to-user cluster-admin admin
...
-    oc create -f /manifests/cna/operator.yaml
-    oc create -f /manifests/cna/network-addons-config-example.cr.yaml
+
+    # TODO: pass CNAO version at gocli
+    cnao_version=0.25.0
+    cnao_path=/manifests/cnao/$cnao_version/
+    oc create -f $cnao_path/namespace.yaml
+    oc create -f $cnao_path/network-addons-config.crd.yaml
+    oc create -f $cnao_path/operator.yaml
+    oc create -f $cnao_path/network-addons-config-example.cr.yaml
```
phoracek: This should be created by the user, not during provision. That way users can decide what to deploy. That may be useful e.g. in case of kubemacpool where we want to deploy everything but kubemacpool while testing master.

qinqon: You mean the CR right ?

.gitignore
qinqon: Upsy

cluster-provision/okd/scripts/provision.sh
phoracek: Yes.

Actually, we need to create the example here, wait for all images to appear and then remove the CR and wait for the cleanup to be done. That way we cache all images but don't leave anything running.

qinqon: But CNAO does not support delete, I think.

phoracek: You delete the whole NetworkAddonsConfig CR, thanks to owner references, that removes all the deployed components. We don't support partial removal.

qinqon: ack, let's add this.

cluster-provision/okd/scripts/provision.sh
```diff
@@ -208,13 +208,16 @@ oc adm policy add-cluster-role-to-user cluster-admin admin
...
+    # TODO: pass CNAO version at gocli
+    cnao_version=0.25.0
+    cnao_path=/manifests/cnao/$cnao_version/
+    oc create -f $cnao_path/namespace.yaml
+    oc create -f $cnao_path/network-addons-config.crd.yaml
+    oc create -f $cnao_path/operator.yaml
+
+     # Wait until cnao is ready
+    until oc wait deployment -n cluster-network-addons cluster-network-addons-operator --for condition=Available --timeout=100s; do
         sleep 10
```
phoracek: The timeout is not long enough, I would set it to 5 minutes, lot of images are pulled in the process.

Also, there is no need for `until`, just use:

```
oc wait deployment -n cluster-network-addons cluster-network-addons-operator --for condition=Available --timeout=5m
```

cluster-provision/okd/scripts/provision.sh
```diff
@@ -208,13 +208,16 @@ oc adm policy add-cluster-role-to-user cluster-admin admin
...
     done
 fi
```
phoracek: What about the removal of the `NetworkAddonsConfig` once it is deployed?

cluster-provision/ocp-cnao/4.4/provision.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash -x
```
phoracek: Do we need to copy paste the whole thing? Cannot we share some code?

cluster-up/cluster/ocp-cnao-4.4/README.md
```diff
@@ -0,0 +1,59 @@
+# OCP 4.4 in ephemeral containers
```
phoracek: This should mention CNAO in the title

cluster-provision/okd/scripts/provision.sh
qinqon: This is not netowrkaddonsconfig this is just the operator.

cluster-provision/ocp-cnao/4.4/provision.sh
qinqon: Yep, just like the other providers, at the end most of it are just parameters to gocli, also it does not change that much.

qinqon:
we have auth issues:
```
Run provision script
+ '[' '!' -f /etc/installer/token ']'

+ '[' '!' -z registry.svc.ci.openshift.org/ocp/release:4.4 ']'

++ oc adm release info -a /etc/installer/token registry.svc.ci.openshift.org/ocp/release:4.4 --commits

++ grep installer

++ awk '{print $3}'

++ head -n 1

error: unable to read image registry.svc.ci.openshift.org/ocp/release:4.4: Head https://registry.svc.ci.openshift.org/v2/ocp/release/manifests/4.4: unauthorized: authentication required

+ export INSTALLER_COMMIT=

+ INSTALLER_COMMIT=

+ compile_installer

+ local 'build_pkgs=git gcc-c++'

+ dnf install -y git gcc-c++

Fedora 28 - x86_64 -  0% [                    ] ---  B/s |   0  B     --:-- ETA
Fe
```

qinqon:
/test push-ocp-cnao-4.4 

secret updated

cluster-provision/ocp-cnao/4.4/provision.sh
phoracek: To record it: We can install CNAO into basic OCP and prepull all the images. Particular features then can be enabled by the user of the provider if needed.

qinqon:
/test push-ocp-cnao-4.4

It has provision allright but there where autorization issues at quay.io/kubevirtci are fixed now.

cluster-provision/ocp-cnao/4.4/provision.sh
qinqon: I am keeping two providers, since they will have different size and you don't want the extra time cost if you don't need CNAO.

phoracek: I especially don't want extra time spending on maintenance of another provider. Could you show what is the actual difference in size? It gets cached on one's host, so I don't see it as that big of an issue. Please prove me wrong.

phoracek: Also, if you happen to be running both providers on your laptop, you end up with both saved on your disk.

qinqon: Let's wait for push size is at quay.io, if it does not worth new provider we just retag it.

phoracek:
@qinqon any update on sizes of those images?

qinqon:
@phoracek so comparing size between ocp-4.4 and ocp-cnao-4.4 is like 9gb vs 11gb, I think it's worth to have the two.
https://quay.io/repository/kubevirtci/ocp-4.4?tab=tags
https://quay.io/repository/kubevirtci/ocp-cnap-4.4?tab=tags

phoracek:
@qinqon I don't think that relative difference is that big to be worth maintaining two images.

How big is the shared layer between the two? I'm afraid we end-up with having two images on your host with total 20 GB instead of one with 11.

qinqon:
@phoracek my main concern is not the memory needed at host if you run the two, but the time it gets at prow jobs to download those (it's a lot since we don't have a proxy cache yet for quay.io), but you maybe right we can keep it simple and optimiza if we see a problem.

qinqon:
so looks like something is broken at network level at cluster-up
```
Container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: Missing CNI default network
```

qinqon:
Looks like vanilla ocp-4.4 is not able to provision

PR #272, opened 2020-02-05
Enable local registry in containerd when using kind-ipv6

eranco74:
This change is required for running ```make sync``` from KubeVirt on a cluster with IPv6.
Currently the make fail to deploy kubevirt containers since the cluster doesn't have access to the local registry.

Reference for the configuration: https://kind.sigs.k8s.io/docs/user/local-registry/

### DIFF
--- a/cluster-up/cluster/kind/manifests/kind-ipv6.yaml
+++ b/cluster-up/cluster/kind/manifests/kind-ipv6.yaml
@@ -1,8 +1,11 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
+containerdConfigPatches:
+- |-
+  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."registry:5000"]
+    endpoint = ["http://registry:5000"]
 networking:
   ipFamily: ipv6
   apiServerAddress: "::1"
 nodes:
 - role: control-plane
-


dankenigsberg:
As a casual reviewer, I'd appreciate if you (always) state the motivation for a PR. In this case I believe that it is being able to `make sync` from KubeVirt.

qinqon:
@eranco74 why we don't have same issue at the other kind provider ?

rmohr:
> @eranco74 why we don't have same issue at the other kind provider ?

That is an interesting question, since we also explicitly allow this registry on all other providers to make it work ...

eranco74:
> > @eranco74 why we don't have same issue at the other kind provider ?
> 
> That is an interesting question, since we also explicitly allow this registry on all other providers to make it work ...

I don't know.
I now see that we do it for kind as well [here](https://github.com/kubevirt/kubevirtci/blob/master/cluster-up/cluster/kind/common.sh#L49)
But it didn't work for my IPv6 cluster.

eranco74:
@rmohr seems that we move from kind v0.3.0  to kind v0.7.0  and in 
kind v0.6.0 there is a new config feature `containerdConfigPatches` that is used for the local registry see [here](https://github.com/kubernetes-sigs/kind/blob/8643089fc3dca2db64c5b3b36e69054f61d22d88/site/content/docs/user/local-registry.md)

PR #281, opened 2020-02-13
Configure calico additional settings to work with NetworkManager

ormergi:
This PR configuring NetworkManager to not manage calicos interfaces.

According to [Calico documentation](https://docs.projectcalico.org/v3.9/maintenance/troubleshooting#configure-networkmanager) it is necessary to configure NetworkManager to not manage calico's 
interfaces to prevent denial of service and disconnections, and extend conntrack capacity for better iptabels performance

If it will be necessary to configure additional settings related to the cni plugin ,whether it is calico or another plugin, this should be added to the config-cni.sh script.
Notice that the config-cni.sh script,  located at cluster-provision/common-scripts, copied to the VM and invoked during provision. 


qinqon:
/lgtm
/approve

Super nice

PR #284, opened 2020-02-14
Add kubevirt local testing, reorg sections in separate documents

dhiller:
/cc @qinqon @phoracek

### DIFF
--- a/K8S.md
+++ b/K8S.md
@@ -8,7 +8,7 @@ cd kubevirtci
                                                                                       
 Start multi node k8s cluster with 2 nics                                              
 ```                                                                                   
-export KUBEVIRT_PROVIDER=k8s-1.13.3 KUBEVIRT_NUM_NODES=2 KUBEVIRT_NUM_SECONDARY_NICS=1
+export KUBEVIRT_PROVIDER=k8s-1.16.2 KUBEVIRT_NUM_NODES=2 KUBEVIRT_NUM_SECONDARY_NICS=1
 make cluster-up                                                                       
 ```                                                                                   
                                                                                       
--- a/K8S.md
+++ b/K8S.md
@@ -1,0 +1,48 @@
+# [kubevirtci](README.md): Getting Started with a multi-node Kubernetes Provider                              
+                                                                                      
+Download this repo                                                                    
+```                                                                                   
+git clone https://github.com/kubevirt/kubevirtci.git                                  
+cd kubevirtci                                                                         
+```                                                                                   
+                                                                                      
+Start multi node k8s cluster with 2 nics                                              
+```                                                                                   
+export KUBEVIRT_PROVIDER=k8s-1.13.3 KUBEVIRT_NUM_NODES=2 KUBEVIRT_NUM_SECONDARY_NICS=1
+make cluster-up                                                                       
+```                                                                                   
+                                                                                      
+Stop k8s cluster                                                                      
+```                                                                                   
+make cluster-down                                                                     
+
+Use provider's kubectl client with kubectl.sh wrapper script               
+```                                                                        
+cluster-up/kubectl.sh get nodes                                            
+cluster-up/kubectl.sh get pods --all-namespaces                            
+```                                                                        
+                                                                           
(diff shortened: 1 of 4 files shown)


OKD.md
```diff
@@ -0,0 +1,85 @@
...
+
+Download this repo
+```
+git clone https://github.com/kubevirt/kubevirtci.git
+cd kubevirtci
+```
+
+Start okd cluster (pre-configured with a master and worker node)
+```
+export KUBEVIRT_PROVIDER=okd-4.1
```
oshoval: okd 4.1 is broken, current we use mainly ocp-4.3 , so maybe better to refer it ?

oshoval: on the other hand, since ocp need quay access, maybe its good to be on okd

K8S.md
```diff
@@ -0,0 +1,48 @@
...
+                                                                                      
+Download this repo                                                                    
+```                                                                                   
+git clone https://github.com/kubevirt/kubevirtci.git                                  
+cd kubevirtci                                                                         
+```                                                                                   
+                                                                                      
+Start multi node k8s cluster with 2 nics                                              
+```                                                                                   
+export KUBEVIRT_PROVIDER=k8s-1.13.3 KUBEVIRT_NUM_NODES=2 KUBEVIRT_NUM_SECONDARY_NICS=1
```
ormergi: Lets set the default provider to be 1.17 or 1.16

K8S.md
```diff
@@ -0,0 +1,48 @@
...
+cluster-up/ssh.sh node01                                                   
+```                                                                        
+                                                                           
+Attach to node console with socat and pty (non okd4 providers)             
+```                                                  
+# Get node01 container id                            
+node01_id=$(docker ps |grep node01 |awk '{print $1}')
+                                                     
+# Install socat                                      
+docker exec $node01_id yum install -y socat
```
qinqon: I thing this is no longer needed since socat is allready installed.

KUBEVIRT_LOCAL_TESTING.md
```diff
@@ -0,0 +1,40 @@
...
+With the changes in place you also can directly execute [`make functest`](https://github.com/kubevirt/kubevirt/blob/master/docs/getting-started.md#testing) against a cluster locally with kubevirt that was provisioned using `kubevirtci`.
+
+Steps:
+
+### kubevirtci: provision cluster locally
+
+```bash
+cd .../kubevirtci
+cd cluster-provision/k8s/1.17.0
+./provision.sh
```
qinqon: Add note that provision.sh also runs up.sh so it also checks that cluster-up is working

Also that cluster-up/... script could need tunning for new k8s versions.

OKD.md
```diff
@@ -0,0 +1,85 @@
+# [kubevirtci](README.md): Getting Started with multi-node OKD Provider
```
qinqon: This is all the same for OCP maybe we can say so.

K8S.md
dhiller: Will set it to 1.16.2.

OKD.md
dhiller: IDK, anyone using OCP needs access to the quay registry? So should we?

K8S.md
dhiller: On k8s-1.16.2
```
> docker exec -it $node01_id socat - /dev/pts/0
OCI runtime exec failed: exec failed: container_linux.go:349: starting container process caused "exec: \"socat\": executable file not found in $PATH": unknown
```

dhiller: So it is needed (at least for <= k8s-1.16.2)

dhiller: I'd leave it in the as it doesn't hurt.

KUBEVIRT_LOCAL_TESTING.md
dhiller: Done.

dhiller: I didn't address the second comment, because that should be an issue IMHO.

PR #289, opened 2020-02-19
Untangle k8s providers

rmohr:
Untangle the k8s providers so that they can be simply copied and minor versions can just be set at one place.

This helps automation and users by easily detecting differences and cloning providers without having to care about ifs/elses in the code while at the same time ensuring that we don't drag around old if/else exceptions forever.

What this PR does:

 * The following providers are affected: `k8s/1.14`, `k8s/1.15`, `k8s/1.16`, `k8s/1.17`
 * Extracting the vanilla k8s/* common code from `common-scripts`
 * Extract the vanilla k8s/* common manifests from `manifests`
 * Making each provider completely independent of each other
 * The providers now live in directories with the form `<major>.<minor>`
 * Each provider contains a `version` file with `<major>.<minor>.<patch>`
 * It will still be ok, to pin specific versions by creating providers in directories like `k8s/1.15.2`, but the goal is to implement a sliding window of latest supported k8s versions, so the patch level of the `<major>.<minor>` providers will be periodically updated in the future.


rmohr:
Should be ready for review. The jobs which fail here were run in their new form from here: https://github.com/kubevirt/project-infra/pull/358. The results look good and the shasum is coming from them.

Here the results:

https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/289/check-provision-k8s-1.14/1230178340275490816
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/289/check-provision-k8s-1.15/1230181420886921216
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/289/check-provision-k8s-1.16/1230181420891115520
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/289/check-provision-k8s-1.17/1230181420903698432

https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/289/push-k8s-1.14/1230194020941041664
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/289/push-k8s-1.15/1230194020953624577
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/289/push-k8s-1.16/1230194020953624578
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/289/push-k8s-1.17/1230194020953624576

rmohr:
I would suggest to first merge https://github.com/kubevirt/project-infra/pull/358 immediately followed by this one if people are ok with the content.

rmohr:
/hold

Holding so that merging can be properly orchestrated if people agree with the content.

PR #292, opened 2020-02-21
Use centos8 for k8s 1.17 provider nodes

qinqon:
Some projects like kubernetes-nmstate has big dependency on the k8s nodes os version,  using centos8 allow to install NetworkManager from copr without using latest git also there are differences from systemd and it affects how it works after a reboot, some issues are explained here [1]

[1] https://github.com/nmstate/nmstate/issues/733


qinqon:
base centos8 hash taken from https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/292/push-centos8/1230815495175802880

qinqon:
We need to change nm to use dhcpient so we have expected behaviour 

https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html

dankenigsberg:
@qinqon in your top comment of the conversation, always include the motivation for the change, please. I think can guess your reasons (nftables?) but I should not be guessing.

qinqon:
@dankenigsberg done

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -2,6 +2,14 @@
 
 set -ex
 
+# Resize root partition
+dnf install -y cloud-utils-growpart
+growpart /dev/vda 1
+xfs_growfs -d /
+
+# Set hostname right
+dhclient
```
ormergi: We might need this in every reboot this vm will do

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -24,6 +32,37 @@ yum -y remove firewalld
...
+
+# Create /etc/docker directory.
+mkdir /etc/docker
+
+# Setup docker daemon
+cat << EOF > /etc/docker/daemon.json
+{
+  "insecure-registries" : ["registry:5000"],
+  "log-driver": "json-file",
+  "exec-opts": ["native.cgroupdriver=systemd"]
```
ormergi: If you would like to add ipv6 support:
`
  "ipv6": true,
  "fixed-cidr-v6": "2001:db8:1::/64"
`

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -34,29 +73,25 @@ repo_gpgcheck=1
...
+# k8s repo is no longer needed and it introduce some
+# gpg key non interactive dnf install -y related to
+# gpg keys (using rpm --import does not fix it)
+rm /etc/yum.repos.d/kubernetes.repo
+
+# Ensure iptables tooling does not use the nftables backend
+# TODO: Not working at centos8
+# update-alternatives --set iptables /usr/sbin/iptables-legacy
+
+# Enable and state kubelet service
```
dhiller: s/state/start/ ? Why now? It's done in line 69 also?

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -34,29 +73,25 @@ repo_gpgcheck=1
...
     kubeadm-${version} \
     kubelet-${version} \
     kubectl-${version} \
-    kubernetes-cni-0.6.0
+    kubernetes-cni
+
+# k8s repo is no longer needed and it introduce some
+# gpg key non interactive dnf install -y related to
+# gpg keys (using rpm --import does not fix it)
+rm /etc/yum.repos.d/kubernetes.repo
```
ormergi: Set [this](https://github.com/kubevirt/kubevirtci/pull/292/files#diff-4df819f2ca7dae6d02fae2a568d619fdR72) to '0' to prevent those issues

cluster-provision/k8s/1.17/provision.sh
qinqon: I forgot to remove this, we are using dhclient as dhcp option in NM.

cluster-provision/k8s/1.17/provision.sh
qinqon: This is compatible with ipv4 right ?

cluster-provision/k8s/1.17/provision.sh
qinqon: ack let's try it out.

cluster-provision/k8s/1.17/provision.sh
qinqon: I think it's not needed let's remove it and test

cluster-provision/k8s/1.17/provision.sh
ormergi: yes

PR #293, opened 2020-02-23
Revert CNI to kindnet, due to kind not fully support ipv6.

ormergi:
This PR reverting calico cni plugin installation.
According to kind documentation, due to kind not fully support ipv6 and ipv6-dual stack with different CNI plugins then kindnet.
https://kind.sigs.k8s.io/docs/user/configuration/#ip-family

### DIFF
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -111,8 +111,10 @@ function kind_up() {
         echo "no ipv6, safe to install flannel"
         _kubectl create -f $KIND_MANIFESTS_DIR/kube-flannel.yaml
     else
-        echo "ipv6 enabled, using calico"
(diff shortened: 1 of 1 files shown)


phoracek:
@ormergi please link the documentation you are mentioning.

qinqon:
/lgtm

Let's add the links and we approve

ormergi:
Updated PR comment, added reference for ipv6 issues

phoracek:
The linked documentation does not say anything about CNI or IPv6 being limited to kindnet. @ormergi is there a reference to that or did you just find it out by trying?

ormergi:
When working with kind there is a limited ipv6 support in general no matter what CNI we deploy.
I think we should revert back to kindnet until there will be full support for ipv6,
even wait for ipv6-dualStack which is more popular in ipv6 environments.
At this state this provider dosent work properly.

phoracek:
/approve

Thanks

PR #294, opened 2020-02-26
k8s-1.17: Fix non interactive dnf install -y

qinqon:


cluster-provision/k8s/1.17/provision.sh
```diff
@@ -68,7 +68,7 @@ name=Kubernetes
 baseurl=http://yum.kubernetes.io/repos/kubernetes-el7-x86_64
```
ormergi: try this url https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64
before you change gpg check

qinqon:
@dhiller I have tried it at kubernetes-nmstate jobs it does not work, also by default repogpgcheck is 0 if you don't specify so it's not that bad also we review the content of .repo file.

PR #295, opened 2020-02-27
Increase k8s provider number of cpus to 6

qinqon:
The k8s 1.17 provider is using calico and it takes from CPUs so some
kubevirt test are failing at lacking cpus at nodes, increasing this
fix the test and does not affect the rest of them.


slintes:
ah, that's why my laptop is unresponsive for 30 minutes after a `make cluster-up` :wink:

PR #303, opened 2020-03-05
k8s-1.17: Wait for transient hostname

qinqon:


cluster-provision/k8s/1.17/node01.sh
```diff
@@ -2,6 +2,11 @@
 
 set -ex
 
+while ! hostnamectl  |grep Transient ; do
+    echo "Waitting for dhclient to set the hostname from dnsmasq"
```
phoracek: Waiting

cluster-provision/k8s/1.17/node01.sh
```diff
@@ -2,6 +2,11 @@
 
 set -ex
 
+while ! hostnamectl  |grep Transient ; do
```
phoracek: It would be safer to do this with a timeout.

PR #304, opened 2020-03-05
Fix ocp 4.3 provider by updating patch file, reprovison

oshoval:


oshoval:
@rmohr 
which lanes should we check here beside up ?
image was provisioned on prow itself

oshoval:
@danielBelenky

oshoval:
check-up-ocp-4.3 fails on docker login
we had to recreate the repo at quay, does the present need to be updated ?

PR #305, opened 2020-03-06
Replace KUBEVIRT_WITH_CNAO to k8s-cnao-1.17 fake provider

qinqon:
The kubevirt automation expect just a target and parse this target to skip or focus tests, let's continue with this approach creating this fake provider, behind the scines is just the k8s-1.17 provider but we pass some flag to the infra scripts.

### DIFF
--- a/cluster-up/cluster/k8s-provider-common.sh
+++ b/cluster-up/cluster/k8s-provider-common.sh
@@ -21,7 +21,7 @@ function up() {
     prepare_config
 
     # Activate cluster-network-addons-operator if flag is passed
-    if [ "$KUBEVIRT_WITH_CNAO" == "true" ]; then
+    if [ "$WITH_CNAO" == "true" ]; then
         kubectl="${_cli} --prefix $provider_prefix ssh node01 -- sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf"
 
         $kubectl create -f /opt/cnao/namespace.yaml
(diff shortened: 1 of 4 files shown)


cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -21,7 +21,7 @@ function up() {
     prepare_config
 
     # Activate cluster-network-addons-operator if flag is passed
-    if [ "$KUBEVIRT_WITH_CNAO" == "true" ]; then
+    if [ "$WITH_CNAO" == "true" ]; then
```
phoracek: I recall Roman commenting on the benefit of keeping it namespaced.

cluster-up/cluster/k8s-1.17/README.md
```diff
@@ -23,15 +23,6 @@ node01    NotReady   master    31s       v1.17.1
 node02    NotReady   <none>    5s        v1.17.1
 ```
 
-## Bringing the cluster up with cluster-network-addons-operator provisioned
```
phoracek: Why do you drop the documentation? I find it useful.

qinqon: I have move it.

cluster-up/cluster/k8s-provider-common.sh
qinqon: This is kind of internal var, that I didn't namespaced it.

rmohr: It should probably default somewhere to false then. Strictly speaking. Not that we enforce that in general anywhere ;)

PR #308, opened 2020-03-10
re-enable machineConfigDaemon

alonSadan:
In this PR we un-delete the machineConfigDaemon
which allowes updates on nodes using machineConfig.
right now it is deleted because it is not possiable
to update configurations of nodes on a cluster
with just one master.
We should consider provision a cluster with more
then one master node, so that we can re-enable
updates on the machines configuration.

Signed-off-by: alonSadan <asadan@redhat.com>

### DIFF
--- a/cluster-provision/okd/scripts/provision.sh
+++ b/cluster-provision/okd/scripts/provision.sh
@@ -256,27 +256,28 @@ EOF
 timeout 30m bash -c "until oc -n openshift-machine-config-operator wait machineconfigpools worker --for condition=Updating --timeout=30m; do sleep 3; done"
 timeout 30m bash -c "until oc -n openshift-machine-config-operator wait machineconfigpools worker --for condition=Updated --timeout=30m; do sleep 3; done"
 
-# Disable updates of machines configurations, because on the update the machine-config
-# controller will try to drain the master node, but it not possible with only one master
-# so the node will stay in cordon state forewer.
-# It will prevent any updates of the kubelet or registries configuration on the machine
-# so if you need some, please add it before these lines
+if [ $MASTERS -eq 1 ]; then
+    # Disable updates of machines configurations, because on the update the machine-config
+    # controller will try to drain the master node, but it not possible with only one master
+    # so the node will stay in cordon state forewer.
+    # It will prevent any updates of the kubelet or registries configuration on the machine
+    # so if you need some, please add it before these lines
 
-# Scale down cluster version operator to prevent updates of other operators
-until oc -n openshift-cluster-version scale --replicas=0 deploy cluster-version-operator; do
-    sleep 5
-done
+    # Scale down cluster version operator to prevent updates of other operators
+    until oc -n openshift-cluster-version scale --replicas=0 deploy cluster-version-operator; do
+        sleep 5
+    done
 
-# Scale down machine-config-operator to prevent re-creation of master machineconfigpools
-until oc -n openshift-machine-config-operator scale --replicas=0 deploy machine-config-operator; do
-    sleep 5
-done
+    # Scale down machine-config-operator to prevent re-creation of master machineconfigpools
+    until oc -n openshift-machine-config-operator scale --replicas=0 deploy machine-config-operator; do
+        sleep 5
+    done
 
-# Delete machine-config-daemon to prevent configuration updates
-until oc -n openshift-machine-config-operator delete ds machine-config-daemon; do
-    sleep 5
-done
-
+    # Delete machine-config-daemon to prevent configuration updates
+    until oc -n openshift-machine-config-operator delete ds machine-config-daemon; do
+        sleep 5
+    done
+fi
 # get_value fetches command output, with retry and timeout
 # syntax nodes=$(get_value 10 oc get nodes)
 # first parameter is the number of iterations to try, each has 6 seconds delay


alonSadan:
@oshoval @qinqon

cluster-provision/okd/scripts/provision.sh
```diff
@@ -249,15 +250,18 @@ until oc -n openshift-cluster-version scale --replicas=0 deploy cluster-version-
     sleep 5
 done
 
+
 # Scale down machine-config-operator to prevent re-creation of master machineconfigpools
 until oc -n openshift-machine-config-operator scale --replicas=0 deploy machine-config-operator; do
```
alonSadan: maybe there are other lines that needs to be deleted like this one. wasn't sure.

slintes:
just fyi: the CodeReadyContainers also runs 1 master only IIRC, maybe sync with them on what workarounds you need?
https://github.com/code-ready/crc/

alonSadan:
> just fyi: the CodeReadyContainers also runs 1 master only IIRC, maybe sync with them on what workarounds you need?
> https://github.com/code-ready/crc/

Thanks.
also @cynepco3hahue  suggseted to pause the master-MCP when performing updates.

oshoval:
Hi
imo please use an if condition to check how many masters you have and act accordingly ?
also you would need to reprovision with prow so the provider will have this change on 4.4 3 masters i guess

keep in mind that in the future we might remove 3 masters and back to 1 master
in this case we would need to think about this feature again and see 
(but lets reach the bridge)

alonSadan:
> Hi
> imo please use an if condition to check how many masters you have and act accordingly ?
> also you would need to reprovision with prow so the provider will have this change on 4.4 3 masters i guess
> 
> keep in mind that in the future we might remove 3 masters and back to 1 master
> in this case we would need to think about this feature again and see
> (but lets reach the bridge)

added the if condition:)

draft.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
oshoval, line 1: O_o whats this file ?

cluster-provision/okd/scripts/provision.sh
```diff
@@ -302,13 +281,47 @@ function get_value()
...
+
+
 masters=$(get_value 50 oc get nodes -l node-role.kubernetes.io/master -o'custom-columns=name:metadata.name' --no-headers)
+
 if [[ ${masters} == "" ]]; then
     echo "Failed to get masters list, exiting"
     exit 1
 fi
 
+if [[ "$(echo ${masters[@]} | wc -w)" -le "1" ]]; then
```
oshoval, line 301: Awesome,
imo i would use this var
https://github.com/kubevirt/kubevirtci/blob/0c328ba07125320f429d136f3ce1afe20cae32b0/cluster-provision/okd/scripts/provision.sh#L16

cluster-provision/okd/scripts/provision.sh
```diff
@@ -302,13 +281,47 @@ function get_value()
...
 
-# Remove master schedulable taint from masters
+# in case of one master-node: 
+# Disable updates of machines configurations, because on the update the machine-config
+# controller will try to drain the master node, but it not possible with only one master
+# so the node will stay in cordon state forewer.
+# It will prevent any updates of the kubelet or registries configuration on the machine
+# so if you need some, please add it before these lines
+
+#uncomment the next lines in case of one master
```
oshoval, line 291: i would remove this

draft.sh
alonSadan: oops

cluster-provision/okd/scripts/provision.sh
```diff
@@ -249,15 +250,18 @@ until oc -n openshift-cluster-version scale --replicas=0 deploy cluster-version-
...
 done
 
+
 # Scale down machine-config-operator to prevent re-creation of master machineconfigpools
 until oc -n openshift-machine-config-operator scale --replicas=0 deploy machine-config-operator; do
     sleep 5
 done
 
-# Delete machine-config-daemon to prevent configuration updates
-until oc -n openshift-machine-config-operator delete ds machine-config-daemon; do
```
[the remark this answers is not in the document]
alonSadan: OK. where do you think It's the right place to groom it?

cluster-provision/okd/scripts/provision.sh
alonSadan: DONE

cluster-provision/okd/scripts/provision.sh
alonSadan: DONE

alonSadan:
The weird-a-like code shifting wa made because I wanted to use `masters` var and the function get_value(), which in the previous revision were defined after machineConfig was disabled

cluster-provision/okd/scripts/provision.sh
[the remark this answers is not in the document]
oshoval: you already added masters check i think ?

cluster-provision/okd/scripts/provision.sh
```diff
@@ -256,27 +256,28 @@ EOF
...
-# Scale down cluster version operator to prevent updates of other operators
-until oc -n openshift-cluster-version scale --replicas=0 deploy cluster-version-operator; do
-    sleep 5
-done
-
-# Scale down machine-config-operator to prevent re-creation of master machineconfigpools
-until oc -n openshift-machine-config-operator scale --replicas=0 deploy machine-config-operator; do
-    sleep 5
-done
+if [ MASTERS -eq 1 ]; then
```
oshoval, line 259: should be $MASTERS,
please re-provision the two lanes and check each does what it should

alonSadan: oops. stupid mistake. fixed

alonSadan:
@qinqon  can you please take a look as well? 
when we have more than one master we can leave both the `machine-config-operator` and `cluster-version-operator` ,  and not just the `machine-config-daemon`?

rmohr:
/close

This provider does not exist anymore.

PR #313, opened 2020-03-18
Make base image ipv6 ready and use it for centos8

rmohr:
Let dnsmasq advertise ipv6 addresses which can reach each other over the
internal br0 bridge in addition to the ipv4 addresses.

The k8s 1.17 cluster which is based on centos8, is configured to automatically ask for an ipv6 address. The IPv6 addresses will be assigned as follows: `node01: fd00::101`, `node02: fd00::101`, and so forth.

### DIFF
--- a/cluster-provision/k8s/1.17/provision.sh
+++ b/cluster-provision/k8s/1.17/provision.sh
@@ -2,6 +2,13 @@
 
 set -ex
 
+# Configure IPv6 DHCP
+nmcli connection modify "System eth0" ipv6.method dhcp
+nmcli connection modify "System eth0" ipv6.may-fail no
+nmcli connection modify "System eth0" connection.autoconnect yes
+nmcli connection modify "System eth0" ipv6.addr-gen-mode eui64
+nmcli connection down "System eth0" && sudo nmcli connection up "System eth0"
+
 # Resize root partition
 dnf install -y cloud-utils-growpart
 if growpart /dev/vda 1; then
--- a/cluster-provision/centos8/scripts/dnsmasq.sh
+++ b/cluster-provision/centos8/scripts/dnsmasq.sh
@@ -6,8 +6,10 @@ NUM_NODES=${NUM_NODES-1}
 NUM_SECONDARY_NICS=${NUM_SECONDARY_NICS-0}
 
 ip link add br0 type bridge
+echo 0 > /proc/sys/net/ipv6/conf/br0/disable_ipv6
 ip link set dev br0 up
 ip addr add dev br0 192.168.66.02/24
+ip -6 addr add fd00::1 dev br0
 
 # Create secondary networks
 for snet in $(seq 1 ${NUM_SECONDARY_NICS}); do
@@ -20,7 +22,7 @@ for i in $(seq 1 ${NUM_NODES}); do
   ip tuntap add dev tap${n} mode tap user $(whoami)
   ip link set tap${n} master br0
   ip link set dev tap${n} up
-  DHCP_HOSTS="${DHCP_HOSTS} --dhcp-host=52:55:00:d1:55:${n},192.168.66.1${n},node${n},infinite"
+  DHCP_HOSTS="${DHCP_HOSTS} --dhcp-host=52:55:00:d1:55:${n},192.168.66.1${n},[fd00::1${n}],node${n},infinite"
   for s in $(seq 1 ${NUM_SECONDARY_NICS}); do
     tap_name=stap$(($i - 1))-$(($s - 1))
     ip tuntap add dev $tap_name mode tap user $(whoami)
@@ -34,4 +36,4 @@ iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
 iptables -A FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
 iptables -A FORWARD -i br0 -o eth0 -j ACCEPT
 
-exec dnsmasq -d ${DHCP_HOSTS} --dhcp-range=192.168.66.10,192.168.66.200,infinite
+exec dnsmasq -d ${DHCP_HOSTS} --dhcp-range=192.168.66.10,192.168.66.200,infinite --dhcp-range=::,static
--- a/cluster-provision/gocli/cmd/ports.go
+++ b/cluster-provision/gocli/cmd/ports.go
@@ -93,7 +93,7 @@ func ports(cmd *cobra.Command, args []string) error {
 		case utils.PortNameOCPConsole:
 			err = utils.PrintPublicPort(utils.PortOCPConsole, containers[0].Ports)
 		case utils.PortNameVNC:
-			err = utils.PrintPublicPort(utils.PortVNC, containers[0].Ports)
+			err = utils.PrintPublicPort(utils.VNCPortStartRange, containers[0].Ports)
 		}
 
 		if err != nil {
--- a/cluster-provision/gocli/cmd/provision.go
+++ b/cluster-provision/gocli/cmd/provision.go
@@ -69,7 +69,7 @@ func provision(cmd *cobra.Command, args []string) error {
 	portMap := nat.PortMap{}
 
 	utils.AppendIfExplicit(portMap, utils.PortSSH, cmd.Flags(), "ssh-port")
-	utils.AppendIfExplicit(portMap, utils.PortVNC, cmd.Flags(), "vnc-port")
+	utils.AppendIfExplicit(portMap, utils.VNCPortStartRange+1, cmd.Flags(), "vnc-port")
 
 	qemuArgs, err := cmd.Flags().GetString("qemu-args")
 	if err != nil {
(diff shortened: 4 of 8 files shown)


qinqon:
@phoracek @AlonaKaplan with this we will be able to test ipv6 at k8s qemu providers instead of kind.

phoracek:
@qinqon with this and another PR configuring Kubernetes to run on IPv6. But it is a good start. Kudos!

cluster-provision/base/scripts/dnsmasq.sh
```diff
@@ -6,15 +6,17 @@ NUM_NODES=${NUM_NODES-1}
...
 ip addr add dev br0 192.168.66.02/24
+ip -6 addr add 2001:0db8:0:f101::1 dev br0
 
 for i in $(seq 1 ${NUM_NODES}); do
   n="$(printf "%02d" ${i})"
   ip tuntap add dev tap${n} mode tap user $(whoami)
   ip link set tap${n} master br0
   ip link set dev tap${n} up
-  DHCP_HOSTS="${DHCP_HOSTS} --dhcp-host=52:55:00:d1:55:${n},192.168.66.1${n},node${n},infinite"
+  DHCP_HOSTS="${DHCP_HOSTS} --dhcp-host=52:55:00:d1:55:${n},192.168.66.1${n},[2001:0db8:0:f101::1${n}],node${n},infinite"
```
phoracek: 2001:0db8: is used for doc examples https://en.wikipedia.org/wiki/Reserved_IP_addresses.

I think we should use fd00::/8 for local addressing.

rmohr: done.

cluster-provision/base/scripts/dnsmasq.sh
```diff
@@ -28,4 +30,4 @@ iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
 iptables -A FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
 iptables -A FORWARD -i br0 -o eth0 -j ACCEPT
 
-exec dnsmasq -d ${DHCP_HOSTS} --dhcp-range=192.168.66.10,192.168.66.200,infinite
+exec dnsmasq -d ${DHCP_HOSTS} --dhcp-range=192.168.66.10,192.168.66.200,infinite --dhcp-range=::,static
```
phoracek: Sorry I missed this originally, but the range should be from fd00/8 too, shouldn't it? With the exception of ::1.

rmohr: This configures a "static" dhcp configuration. That means that it will only offer IPs to requests where the mac is  specified in ${DHCP_HOSTS}.

phoracek: Oh, my bad, I missed the for cycle. Thanks for the explanation.

phoracek:
@maiqueb this PR introduces a DHCP server offering IPv6 addresses to kubevirtci nodes. It seems to have issues receiving the IP on the node side. Could it be due to the prefix bug you were describing?

maiqueb:
> @maiqueb this PR introduces a DHCP server offering IPv6 addresses to kubevirtci nodes. It seems to have issues receiving the IP on the node side. Could it be due to the prefix bug you were describing?

The issue I described is characterized by the nodes receiving the IP addresses correctly, but, having a misconfigured prefix length - on newer versions of dhclient / NM, the prefix length will default to 128 bits.

If you notice the IPs have a prefix length of 128 bits on the ipv6 addresses, that's definitely it. AFAIK, dynamic ipv6 relies on the RA messages to configure the prefix length... 

Not sure you can hack something simpler, but the usual solution is to add the router advertisement daemon (radvd) into the mix.

rmohr:
/hold

node01 comes up properly, but node02 does not. Probably related to the mac address change. Will look into it.

ormergi:
> /hold
> 
> node01 comes up properly, but node02 does not. Probably related to the mac address change. Will look into it.

Hey this is might be related to the IPv4 address that passes to kubeadm join .. in [nodes.sh](https://github.com/kubevirt/kubevirtci/blob/256276b9c47df8d6db26b800c6fca35e8378b83c/cluster-provision/k8s/1.17/nodes.sh#L42) script

rmohr:
> Hey this is might be related to the IPv4 address that being passed to kubeadm join .. in [nodes.sh](https://github.com/kubevirt/kubevirtci/blob/256276b9c47df8d6db26b800c6fca35e8378b83c/cluster-provision/k8s/1.17/nodes.sh#L42) script

Good idea, but this PR only makes the VMs ready for ipv6, it does not yet create an IPv6 cluster. It turned out that the issue was that the local-link certificates were not created based on the mac address. As a result `node02` ended up with the same local-link addresses like `node01`. NetworkManager detected that and refused to continue the boot.

The solution is to tell NetworkManager to generate them based on the mac address:

```
nmcli connection modify "System eth0" ipv6.addr-gen-mode eui64
```

cluster-provision/base/scripts/dnsmasq.sh
qinqon: But that's for the primary nics, aren't we going to set ipv6 addresses at secondary nics ?

oshoval:
Hi, whats is the current state of this PR please?
i would need to create with it an ipv6 provider which isnt kind
thanks

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -2,6 +2,12 @@
 
 set -ex
 
+# Configure IPv6 DHCP
+nmcli connection modify "System eth0" ipv6.method dhcp
+nmcli connection modify "System eth0" ipv6.may-fail no
+nmcli connection modify "System eth0" connection.autoconnect yes
+nmcli connection down "System eth0" && sudo nmcli connection up "System eth0"
+
```
oshoval, line 10: is 
nmcli connection modify "System eth0" ipv6.addr-gen-mode eui64
should be added here  as Roman said ?

rmohr:
> Hi, whats is the current state of this PR please?
> i would need to create with it an ipv6 provider which isnt kind
> thanks

The IP assignment does not always work
still have to finalize it.

rmohr:
and

> The IP assignment does not always work
> still have to finalize it.

And I am facing router advertisment issues with dnsmasq.

oshoval:
Hi, FYI please i case needed,
https://github.com/projectcalico/calico/issues/2191
calico has problem with kernel 4.x unless ipv6 is disabled
k8s providers use calico atm

thanks

EdDev:
> As a result node02 ended up with the same local-link addresses like node01. NetworkManager detected that and refused to continue the boot.
> 
> The solution is to tell NetworkManager to generate them based on the mac address

Ending up with the same address is not something that should happen, no matter in which mode it is. The only reason to force generating the address based on the mac address is for consistency over reboots (i.e. the address may change between reboots). Do you really need to reboot the VM in the middle?

cluster-provision/base/scripts/dnsmasq.sh
EdDev: Regarding the configuration here, I am afraid it is a bit more complex than this one liner.
See dnsmasq configuration from the nmstate project: https://github.com/nmstate/nmstate/blob/2cf0bb8812415a96b0f35d1f0d4e4ef12946ea6d/tests/integration/dynamic_ip_test.py#L85

cluster-provision/k8s/1.17/provision.sh
rmohr: Is now there.

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -2,6 +2,12 @@
 
 set -ex
 
+# Configure IPv6 DHCP
+nmcli connection modify "System eth0" ipv6.method dhcp
+nmcli connection modify "System eth0" ipv6.may-fail no
+nmcli connection modify "System eth0" connection.autoconnect yes
+nmcli connection down "System eth0" && sudo nmcli connection up "System eth0"
```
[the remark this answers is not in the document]
rmohr: might be a good change.

rmohr:
Here the latest update on this PR:

 * The nodes don't pick up the the offered DHCP lease. It seems like after the initial provision, networkmanager is  still caching something somewhere and wants to stick with the old settings.
 * dnsmasq sound send out router advertisments with the right network segment. It seems like they are ignored. Therefore sometimse you get a /64 and sometimes a /128 assigned.

cluster-provision/base/scripts/dnsmasq.sh
rmohr: Can very well be. I can only confirm that if the IPs and the segments work, the VMs can communicate with each other.

rmohr:
> Ending up with the same address is not something that should happen, no matter in which mode it is. The only reason to force generating the address based on the mac address is for consistency over reboots (i.e. the address may change between reboots). Do you really need to reboot the VM in the middle?

We provision one golden image in the provision step. We then start the nodes with different macs than were used back then. What I try to work around here is that NetworkManager generates a network id and uses this network id to detect configurations and caches. NetworkManager does then not ask for a new lease if the mac address for this configuration changed.

oshoval:
Thanks Roman and all
continuing at
https://github.com/kubevirt/kubevirtci/pull/345

rmohr:
/close

Closing in favour of #345

PR #314, opened 2020-03-18
Fix call to provision.sh in local testing docs

jean-edouard:
./provision.sh happens to be a valid command, but it's the script from
the parent directory that should actually be called to build the image

### DIFF
--- a/KUBEVIRT_LOCAL_TESTING.md
+++ b/KUBEVIRT_LOCAL_TESTING.md
@@ -9,7 +9,7 @@ Steps:
 ```bash
 cd .../kubevirtci
 cd cluster-provision/k8s/1.17.0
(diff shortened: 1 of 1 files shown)


KUBEVIRT_LOCAL_TESTING.md
```diff
@@ -9,15 +9,15 @@ Steps:
...
+../provision.sh      # this also calls cluster-up to check whether the cluster will really start and have the pods ready
 ```
 
 ### kubevirt: prepare for tests
 
 ```bash
 # set local provision test flag
-# sync changes to kubevirt cluster-up
+# sync changes to kubevirt cluster-up, notably cluster/images.sh
```
rmohr: I guess that we still have to document that cluster/images.sh needs to be manually updated, but a aprt from that looks good.

PR #315, opened 2020-03-19
Validate cluster nodes labels and taints in kind providers

ormergi:
This PR validates that worker  nodes has the correspond role label, and remove NoShchedule taint from masters nodes.

I encounter this while working on kind-k8s-1.17.0-ipv6 lane causing VM migration test to fail with the following error on the destination vm virt-launcher pod:
0/2 nodes are available: 1 node(s) didn't match pod affinity/anti-affinity, 1 node(s) had taints that the pod didn't tolerate.

Due to inconsistent with labels and master nodes taints on multi-node clusters,
it is necessary to remove taint NoSchedule and set role labels properly.

Signed-Off-by: Or Mergi <ormergi@redhat.com>

### DIFF
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -12,6 +12,9 @@ KUBECTL="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl --kubeconfig=${KU
 
 REGISTRY_NAME=${CLUSTER_NAME}-registry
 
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/kind/common.sh
```diff
@@ -99,6 +102,25 @@ function prepare_workers() {
     done
 }
 
+function _validate_nodes_roles() {
```
qinqon: nit: I would call this._fix_node_labels

ormergi: done

PR #317, opened 2020-03-19
kind: Don't relabel nodes with with schedule=true if not needed

qinqon:
When we startup a kind provider with just one node, it fails since master node is schedulable, and we cannot taint it with `kubectl taint nodes $node node-role.kubernetes.io/master:NoSchedule-`.

This PR label master nodes with `kubevirt.io/schedulable=true` only if we can taint the master nodes.

Also extra mile we have create a pair of helper functions to retrieve nodes and pods.

### DIFF
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -102,19 +102,28 @@ function prepare_workers() {
     done
 }
 
+function _get_nodes() {
+    _kubectl get nodes --no-headers
+}
+
+function _get_pods() {
+    _kubectl get pods --all-namespaces --no-headers
+}
+
 function _fix_node_labels() {
     # Due to inconsistent labels and taints state in multi-nodes clusters,
     # it is nessecery to remove taint NoSchedule and set role labels manualy:
     #   Master nodes might lack 'scheduable=true' label and have NoScheduable taint.
     #   Worker nodes might lack worker role label.
-    master_nodes=$(_kubectl get nodes --no-headers | grep -i $MASTER_NODES_PATTERN | awk '{print $1}')
+    master_nodes=$(_get_nodes | grep -i $MASTER_NODES_PATTERN | awk '{print $1}')
     for node in ${master_nodes[@]}; do
-        # removing NoSchedule taint
-        _kubectl taint nodes $node node-role.kubernetes.io/master:NoSchedule-
-        _kubectl label node $node kubevirt.io/schedulable=true
+        # removing NoSchedule taint if is there
+        if _kubectl taint nodes $node node-role.kubernetes.io/master:NoSchedule-; then
+            _kubectl label node $node kubevirt.io/schedulable=true
+        fi
     done
 
-    worker_nodes=$(_kubectl get nodes --no-headers | grep -i $WORKER_NODES_PATTERN | awk '{print $1}')
+    worker_nodes=$(_get_nodes | grep -i $WORKER_NODES_PATTERN | awk '{print $1}')
     for node in ${worker_nodes[@]}; do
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/kind/common.sh
```diff
@@ -145,26 +153,27 @@ function setup_kind() {
 
     _wait_kind_up
     _kubectl cluster-info
+    if [ $(nodes |wc -l) -ge 2 ]; then
```
ormergi: Do we have access to KUBEVIRT_NUM_NODES parameter in this context?

qinqon: I think so, but let's do this thing depending on the system.

qinqon:
@ormergi agree let's add a verb there also add the internal function prefix _

cluster-up/cluster/kind/common.sh
```diff
@@ -145,26 +153,27 @@ function setup_kind() {
 
     _wait_kind_up
     _kubectl cluster-info
+    if [ $(_get_nodes |wc -l) -ge 2 ]; then
+        _fix_node_labels
```
phoracek: I would expect it to fix the labels all the time! Why only if >= 2. (asking as a casual user reading this for the first time) Could you move the if into the function? There if will be clearer why we need it. Also, you have to do some labeling even with a single node - mark the master with worker label

cluster-up/cluster/kind/common.sh
```diff
@@ -102,19 +102,27 @@ function prepare_workers() {
...
+    _kubectl get pods --all-namespaces --no-headers
+}
+
 function _fix_node_labels() {
     # Due to inconsistent labels and taints state in multi-nodes clusters,
     # it is nessecery to remove taint NoSchedule and set role labels manualy:
     #   Master nodes might lack 'scheduable=true' label and have NoScheduable taint.
     #   Worker nodes might lack worker role label.
-    master_nodes=$(_kubectl get nodes --no-headers | grep -i $MASTER_NODES_PATTERN | awk '{print $1}')
+    master_nodes=$(_get_nodes | grep -i $MASTER_NODES_PATTERN | awk '{print $1}')
```
danielBelenky: Though not directly related to you change, since you already move stuff into functions, I think that we have an opportunity to improve here:

I think that the `node-role.kubernetes.io/master` label can tell you exactly who the masters are. Also, you don't need to use `awk`, you can use `kubectl's output formatter instead... maybe something like:

```
function _get_master_names() {
	_kubectl get nodes \
		--no-headers \
		-l node-role.kubernetes.io/master="" \
		-o custom-columns=:.metadata.name
}
```

Since this is not directly related, feel free not to do this change

ormergi: Hi Daniel, we search for worker nodes by name because there seem to be an issue with the worker labeling in KIND with multi-nodes clusters, this is why I created PR this in the first place.

danielBelenky: Ah I see... this is why an informative commit message is important :smile:

cluster-up/cluster/kind/common.sh
qinqon: Ok let's try to mark as workter too, the erro was with the taint since node is not a NonSchedule.

cluster-up/cluster/kind/common.sh
qinqon: @danielBelenky yep, totally my fault, I am reworking a little this so change is smaller, just don't relabel nodes if not needed.

danielBelenky:
@qinqon hey, could you please add the description to the commit message to? The github description is not being committed to git log...

PR #318, opened 2020-03-19
Update the sha256 in images.sh in the k8s provision script

jean-edouard:


jean-edouard:
This is actually not technically needed, since images.sh defaults to `:latest` if `KUBEVIRTCI_PROVISION_CHECK` is set.

qinqon:
/hold
You cannot update images.sh if image is not pushed

jean-edouard:
This was for local testing use-cases, but even there it doesn't make much sense, closing.

PR #319, opened 2020-03-20
kind: fix SR-IOV 1.17 provider

phoracek:
We can't pass the SHELL settings to a makefile as a variable. Instead, we need to set it as an argument of make.

In the current code, we try to create SriovNodeNetworkPolicy before we fully configure SR-IOV operator's webhook. The code was reordered to fix that.

With this PR we also enforce that the cluster is started with 2 nodes, which is the only supported configuration now.

This also drops SR-IOV 1.14 provider which is outdated and does not work. We don't intend to fix it.

With these changes, the SR-IOV lane should work again.

This change is being tested in https://github.com/kubevirt/kubevirt/pull/3185.


phoracek:
/hold

Still testing on kubevirt.

phoracek:
Verified in https://github.com/kubevirt/kubevirt/pull/3185

PR #322, opened 2020-03-21
add local registry mirror to kind base

phoracek:
In order for kind provider to work with make cluster-sync, we need to add
registry mirrors.


phoracek:
Verified in https://github.com/kubevirt/kubevirt/pull/3185

PR #323, opened 2020-03-24
kind: deploy local volume provisioner

maiqueb:
Kind default StorageClass is named 'standard', while kubevirt
func tests assume it to be called 'local' - as is for all the
other providers.
    
To enable the kind-k8s-1.17-ipv6 provider to work out of the box,
the `local-volume-provisioner` daemonset must be deployed. It will
take care of:
  - create the 'local' storage class
  - discover all mount points in its configuration path (via config
    map) and expose those mount points as PersistentVolumes.
    
For more information, check the local static provisioner repo,
located at [0].
    
This will enable the functional_tests labeled as 'PVC' to be
executed.
    
[0] - https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner

Signed-off-by: Miguel Duarte Barroso <mdbarroso@redhat.com>

### DIFF
--- a/cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
+++ b/cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
@@ -2,6 +2,8 @@
 
 set -e
 
+DOCKER="${CONTAINER_RUNTIME:-docker}"
+
 export IPV6_CNI="yes"
 export CLUSTER_NAME="kind-1.17.0"
 export KIND_NODE_IMAGE="kindest/node:v1.17.0"
@@ -11,4 +13,27 @@ source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
 function up() {
     cp $KIND_MANIFESTS_DIR/kind-ipv6.yaml ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
     kind_up
+    # remove the rancher.io kind default storageClass
+    _kubectl delete sc standard
+
+    nodes=$(_kubectl get nodes -o=custom-columns=:.metadata.name | awk NF)
+    for node in $nodes; do
+        # Create local-volume directories, which, on other providers, are pre-provisioned.
+        # For more info, check https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/STORAGE.md
+        for i in {1..10}; do
+            mount_disk $node $i
+        done
+        $DOCKER exec -it $node bash -c "chmod -R 777 /var/local/kubevirt-storage/local-volume"
+    done
+
+    # create the `local` storage class - which functional tests assume to exist
+    _kubectl apply -f $KIND_MANIFESTS_DIR/local-volume.yaml
+}
+
+function mount_disk() {
+    local node=$1
+    local idx=$2
+    $DOCKER exec -it $node bash -c "mkdir -p /var/local/kubevirt-storage/local-volume/disk${idx}"
+    $DOCKER exec -it $node bash -c "mkdir -p /mnt/local-storage/local/disk${idx}"
+    $DOCKER exec -it $node bash -c "mount -o bind /var/local/kubevirt-storage/local-volume/disk${idx} /mnt/local-storage/local/disk${idx}"
 }
--- a/cluster-up/cluster/kind/manifests/local-volume.yaml
+++ b/cluster-up/cluster/kind/manifests/local-volume.yaml
@@ -1,0 +1,130 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+ name: local
+ annotations:
+  storageclass.kubernetes.io/is-default-class: "true"
+provisioner: kubernetes.io/no-provisioner
+volumeBindingMode: WaitForFirstConsumer
+reclaimPolicy: Delete
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: local-storage-config
+data:
+  storageClassMap: |
+    local:
+       hostDir: /mnt/local-storage/local
+       mountDir: /mnt/local-storage/local
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: local-storage-provisioner-pv-binding
+subjects:
+- kind: ServiceAccount
+  name: local-storage-admin
+  namespace: default
+roleRef:
+  kind: ClusterRole
+  name: system:persistent-volume-provisioner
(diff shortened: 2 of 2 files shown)


cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+DOCKER="${CONTAINER_RUNTIME:-docker}"
```
qinqon: This is unrelated

cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
```diff
@@ -11,4 +13,24 @@ source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
 function up() {
     cp $KIND_MANIFESTS_DIR/kind-ipv6.yaml ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
     kind_up
+    # remove the rancher.io kind default storageClass
+    _kubectl delete sc standard
+
+    nodes=$(_kubectl get nodes -o=custom-columns=:.metadata.name | awk NF)
+    for node in $nodes; do
+        # Create local-volume directories
+        for i in {1..10}; do
```
qinqon: why 10 ?

cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
```diff
@@ -11,4 +13,24 @@ source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
...
+        done
+        $DOCKER exec -it $node bash -c "chmod -R 777 /var/local/kubevirt-storage/local-volume"
+    done
+
+    # create the `local` storage class - which functional tests assume to exist
+    _kubectl apply -f $KIND_MANIFESTS_DIR/local-volume.yaml
+}
+
+function mount_disk() {
+    $DOCKER exec -it $node bash -c "mkdir -p /var/local/kubevirt-storage/local-volume/disk${i}"
```
qinqon: pass index as argument
```bash
local idx=$1
```

cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
awels: 10 is basically a number that seemed large enough to give enough PVs for most tests at the time. With the ability to add more later, it seemed good enough. I still haven't found any cases where 10 was not more than enough.

cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
```diff
@@ -11,4 +13,25 @@ source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
...
+    for node in $nodes; do
+        # Create local-volume directories
+        for i in {1..10}; do
+            node=$node mount_disk $i
+        done
+        $DOCKER exec -it $node bash -c "chmod -R 777 /var/local/kubevirt-storage/local-volume"
+    done
+
+    # create the `local` storage class - which functional tests assume to exist
+    _kubectl apply -f $KIND_MANIFESTS_DIR/local-volume.yaml
```
awels: Any reason not to use the existing one in cluster-provision/manifests/local-volume.yaml

awels: So from what I understand, there is a provision step that builds the images, and in that step that file is copied onto the nodes. And during the cluster startup the yaml is applied from the node. But this is in the VM case, I guess kind is different?

maiqueb: Yes, seems we just fetch the `kind` that k8s-sigs provides: https://github.com/kubevirt/kubevirtci/blob/master/cluster-up/cluster/kind/common.sh#L35

awels: Interesting. Well I simply want to avoid having the same file in 2 places so we don't over time get drift between the two. I don't care that much where it lives.

cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
qinqon: Can you add q comment about it ?

qinqon: Can you add q comment about it ?

qinqon: Can you add q comment about it ?

cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
qinqon: Fair enough

cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
qinqon: At kubevirtci we are doing transition on having providers isoletad, for easy debugging and releasing, even if it means duplication, in case of running needed is much better, and we don't have that much.

cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
maiqueb: This is actually public facing information, available in the [kubevirtci storage README](https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/STORAGE.md#storage-providers-inside-the-generated-images) - "By default, we have pre-provisioned 10 local volumes per node, which are available to make requests against. You can manually bind mount more directories to increase this number.".

Added that link to the already existing comment, hope that's enough.

qinqon: good

cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
```diff
@@ -11,4 +13,26 @@ source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
...
+}
+
+function mount_disk() {
+    local idx=$1
+    $DOCKER exec -it $node bash -c "mkdir -p /var/local/kubevirt-storage/local-volume/disk${idx}"
```
qinqon: Same for node
local node=$2 so it's easier to follow.

maiqueb: Done.

PR #327, opened 2020-03-26
Label k8s non master nodes as workers

qinqon:
They were missing the node-role label at all and this prevent some
testing to work propertly.

### DIFF
--- a/cluster-up/cluster/k8s-provider-common.sh
+++ b/cluster-up/cluster/k8s-provider-common.sh
@@ -20,9 +20,14 @@ function up() {
     # Make sure that local config is correct
     prepare_config
 
+    kubectl=${KUBEVIRTCI_PATH}/kubectl.sh
+
+    # Label all the non master nodes as workers, we have to do here since, after k8s 1.16 is not possible to do
+    # at kubelet [1]
+    # [1] https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/user/troubleshooting.md
+    $kubectl label node -l '!node-role.kubernetes.io/master' node-role.kubernetes.io/worker=''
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -20,9 +20,15 @@ function up() {
     # Make sure that local config is correct
     prepare_config
 
+    kubectl=${KUBEVIRTCI_PATH}/kubectl.sh
+
+    # Label all the non master nodes as workers, we have to do here since, after k8s 1.16 is not possible to do
+    # at kubelet [1]
+    # [1] https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/user/troubleshooting.md
+    $kubectl get nodes --no-headers -l '!node-role.kubernetes.io/master' -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}' | xargs -I{} $kubectl label node {} node-role.kubernetes.io/worker=''
```
ormergi: This will give us the node name in more simple way:
kubectl get nodes --no-headers -l '!node-role.kubernetes.io/master' -o custom-columns=:.metadata.name

danielBelenky: You don't need `xargs` here...

```
kubectl label node -l '!node-role.kubernetes.io/master' role.kubernetes.io/worker=''
```

Could you elaborate why you need this label? the single node setup doesn't seem to have it and everything works...

qinqon: For kubernetes-nmstate we want to test some destructive node network changes only at workers we use kubevirtci with multinode for that, having the worker node-role is very convenient and also is what ocp has.

ormergi:
lgtm

PR #328, opened 2020-03-31
Label 1 node cluster with 'master,worker' roles

qinqon:
In case KUBEVIRT_NUM_NODES=1 is used only a master node is stareted
but it's missing worker node-role so it may be an issue with deploys or test
that depends on worker node-role.

### DIFF
--- a/cluster-up/cluster/k8s-provider-common.sh
+++ b/cluster-up/cluster/k8s-provider-common.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -eo pipefail
 
 source ${KUBEVIRTCI_PATH}/cluster/ephemeral-provider-common.sh
 
--- a/cluster-up/cluster/k8s-provider-common.sh
+++ b/cluster-up/cluster/k8s-provider-common.sh
@@ -4,6 +4,19 @@ set -e
 
 source ${KUBEVIRTCI_PATH}/cluster/ephemeral-provider-common.sh
 
+function _label_workers() {
+    # Label all the non master nodes as workers, we have to do here since, after k8s 1.16 is not possible to do
+    # at kubelet [1]
+    # [1] https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/user/troubleshooting.md
+    number_of_nodes=$($kubectl get node  --no-headers |wc -l)
+    if [ $number_of_nodes -eq 1 ]; then
+        label="node-role.kubernetes.io/master"
+    else
+        label="!node-role.kubernetes.io/master"
+    fi
+    $kubectl label node -l $label node-role.kubernetes.io/worker=''
+}
+
 function up() {
     ${_cli} run $(_add_common_params)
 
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -4,6 +4,19 @@ set -e
 
 source ${KUBEVIRTCI_PATH}/cluster/ephemeral-provider-common.sh
 
+function _label_workers() {
+    # Label all the non master nodes as workers, we have to do here since, after k8s 1.16 is not possible to do
+    # at kubelet [1]
+    # [1] https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/user/troubleshooting.md
+    number_of_nodes=$($kubectl get node  --no-headers |wc -l)
```
oshoval: in rare case that kubectl temporally fails this would give 0 because the wc at the end
so maybe worth to first make sure the left most command success, 
theoretically set -o pipefail && $kubectl might work , but one time tried it and it didnt , guess i didnt do it right,
just until can be used if command splitted

qinqon: Ahh thanks, let me fix it.

ormergi: In case something bad happaned during deployment and you reach to this point,
`kubectl get nodes` will fail with server not reachable error, then wc -l will return 0.
I think we should replace else with` elif [ $number_of_nodes -gt 1 ]`

We can also verify that all nodes that the user requested are up with comparing $number_of_nodes with KUBEVIRT_NUM_NODE

oshoval: the pipe should fail it as a whole
the only other alternative i see is to do a redo on the command
else the value that will be used isnt right

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -1,9 +1,22 @@
 #!/usr/bin/env bash
 
-set -e
+set -eo pipefail
```
dhiller: This file is sourced, so I don't think it is a good idea to set this here. I'd rather prefer this set in the calling scripts.

dhiller: Although this is done in other places several times already? WDYT?

qinqon: Let's run provision tests (they run cluster-up too) so we see if we good ?

oshoval: you can also if you want use just to that command
set -o pipefail && command

btw this phenomena will potentially hit us on more places, silently
i saw it happens when i created 3 masters PR  , where it happens a lot

qinqon:
/hold
wait until we run check-provision to check that pipefail does not break anything.

qinqon:
We found an issue labeling workers so we have revert it, now I have put together the non master labeling and the mono node cluster labeling here https://github.com/kubevirt/kubevirtci/pull/332

PR #336, opened 2020-04-05
Enable PodPreset on Kind IPv6 in order to enable migration

oshoval:
Migration require unique UUID for each virt-launcher pod,
we are using PodPreset for the launchers, so this PR
enables PodPreset and generate a unique UUID per node, that will
be mounted using the PodPreset on the virt-launcher


danielBelenky:
/lgtm
/approve

where you want to create the `PodPreset` object itself?

oshoval:
> /lgtm
> /approve
> 
> where do you want to create the `PodPreset` object itself?

Thanks 
still WIP, in utils in case its kind-ipv6 (later will extended to other kinds according need)
Thanks Fabian for the lead of this method

https://github.com/kubevirt/kubevirt/pull/3221/commits/b2a8cbba00b2134b4a85c109ecd758e177281b6a#diff-a5a178b4c66835110ceff63c61d28c96R1183

PR #337, opened 2020-04-06
dnsmasq: Create secondary networks

EdDev:
Create a network for each secondary NIC.
On each node, the secondary NIC X will be connected to network X
where the network is represented by a bridge (brX).

With this change, the secondary nics are no longer connected to the pod
network and therefore also no longer have a dynamic IP.

One of the known issues this change solves is the console/vnc access to
the KubeVirt VM/s when secondary nics are defined.


cluster-provision/base/scripts/dnsmasq.sh
```diff
@@ -9,6 +9,12 @@ ip link add br0 type bridge
 ip link set dev br0 up
 ip addr add dev br0 192.168.66.02/24
 
+# Create secondary networks
```
phoracek, line 12: Could you run DHCP servers on each network? It would help us with testing on kubernetes-nmstate.

qinqon: It will not, in fact we are flushing them at cluster-up https://github.com/nmstate/kubernetes-nmstate/blob/master/hack/flush-secondary-nics.sh

qinqon:
I think we are safe here related to cluster wide unique mac address at secondary nics since we are setting them syntetically here https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/gocli/cmd/run.go#L397-L400

qinqon:
@EdDev after push-k8s-1.17 a sha256 will be generated with the pushed image you will have to update cluster-up/cluster/images.sh with that, I think we have to update only k8s-1.17 provider since the others use flannel so there is no issue (I think).

cluster-provision/base/scripts/dnsmasq.sh
phoracek: What do you mean? We flush the assigned IPs initially to get a clean start. Then however, during test run, we create policies obtaining IP on those ifaces via DHCP. So we need a DHCP server.

EdDev: It may fit a separate PR, but I doubt we will find a good reason to have it as a service of the infra.
We could probably stick with static addresses and if a need to test with dynamic ones appears, a dedicated pod can be ran with the DHCP server.

qinqon: Ahh right, let me investigate if we can make dnsmasq run on different bridges...

maiqueb: It *might* be needed to have each dnsmasq on a different netns - not really sure, please try without first.

EdDev: IMO we should not have this in the infra. A dhcp server should be loaded by the tests in a pod connected to that network.

phoracek: The provider gives its user a prepared infrastructure with hosts connected through networks. When I'm given an infrastructure, I would expect that DHCP is part of the package. We can of course do via a pod on the client-side, but I see it as an unnecessary effort for knmstate and other users.

qinqon: @phoracek is ok we do this at follow up PR ? it's only needed for kubernetes-nmstate.

EdDev: @phoracek IMO using dynamic IP is a bad idea in an enterprise env, and these secondary networks are supposed to represent such enterprise env if I understand correctly.
I think it will be better to control if you want DHCP or not.

But if you still think it is better to have it by default and without the options to control it from the tests, then a following PR can propose it indeed.

phoracek: Make sense. I don't mind dnsmasq being called explicitly by kubevirtci user (e.g. knmstate).

EdDev:
I will prefer to squash this myself over the bot doing it.
If it is ready, please let me know and I will do it.

qinqon:
We can squash the PR to have a proper commit message with just one Signed-off-by

PR #338, opened 2020-04-07
Delete kubevirtci base

qinqon:
It's included as part of centos7,centos8 and fedora base images, this
reduce the provision steps needed when when a change at base image is done,
helping with automation effort and dev experience.


rmohr:
Ugh, the checks are not  required ... :/

PR #341, opened 2020-04-09
Go back to using a fixed 35G disk size for Centos 8

jean-edouard:
PR #338 removed the hardcoded 35G qcow2 disk size to instead match the original size.
The original size for Centos is currently 10G, which is not enough for test VMs.


rmohr:
I believe that removing the hard-coded value was done intentionally for centos7. Seems like it is not good for centos8. Maybe only do it for centos8 right now and only update the 1.17 cluster.

jean-edouard:
@rmohr, just narrowed down to Centos 8 only

rmohr:
Looks good to me. You can proceed with the upload if you like. Thanks.

cluster-provision/centos8/scripts/vm.sh
```diff
@@ -68,10 +68,8 @@ fi
 
 calc_next_disk
 
-disk_size=$(qemu-img info --output json ${last} | jq '.["virtual-size"]')
-
-echo "Creating disk \"${next} backed by ${last} with size ${disk_size}\"."
-qemu-img create -f qcow2 -o backing_file=${last} ${next} ${disk_size}
+echo "Creating disk \"${next} backed by ${last} with size 35G\"."
+qemu-img create -f qcow2 -o backing_file=${last} ${next} 35G
```
qinqon, line 72: Can use 35G as default in case the centos8 virtualbox change size (like it did for centos7) 
```bash
default_disk_size=37580963840 # 35G
disk_size=$(qemu-img info --output json ${last} | jq '.["virtual-size"]')
if [ $disk_size -lt $default_disk_size ]; then
    disk_size=$default_disk_size
fi
```

jean-edouard: Good idea, I'll change it to that. Thanks!

jean-edouard: @qinqon, done, also re-added growpart to actually use the space.

jean-edouard:
Let's wait until https://github.com/kubevirt/kubevirtci/pull/342 is merged to run test push commands to avoid conflicts.

qinqon:
@jean-edouard awesome ! thanks for the growpart resurrection.

qinqon:
/lgtm
/approve
/hold

Let's waith for shas to be updated

qinqon:
Last hash to update

jean-edouard:
The other lanes also needed updating from previous changes.

qinqon:
@jean-edouard only k8s-1.17 is based on centos8 so no need to update the others.

jean-edouard:
@qinqon, sorry I wasn't very clear. I need to update those lanes to reflect the changes from https://github.com/kubevirt/kubevirtci/pull/340, which already got merged, just to avoid opening yet another PR if that's ok.

rmohr:
/lgtm
/approve
/hold

Waiting for the pending (not required) PRs to pass.

PR #342, opened 2020-04-13
Port different bridge for secondary-nics to centos8

qinqon:


cluster-provision/k8s/1.17/provision.sh
```diff
@@ -21,6 +21,9 @@ yum -y remove firewalld
 # Required for iscsi demo to work.
 yum -y install iscsi-initiator-utils
 
+# To prevent preflight issue realted to tc not found
+dnf install -y tc
```
ormergi, line 25: Is that a new requirement?

cluster-provision/centos8/scripts/dnsmasq.sh
```diff
@@ -18,7 +24,7 @@ for i in $(seq 1 ${NUM_NODES}); do
   for s in $(seq 1 ${NUM_SECONDARY_NICS}); do
     tap_name=stap$(($i - 1))-$(($s - 1))
     ip tuntap add dev $tap_name mode tap user $(whoami)
-    ip link set $tap_name master br0
+    ip link set $tap_name master br${s}
```
ormergi, line 27: Is it ok that those bridges are not connected outside?

qinqon: you can connect from eth1 to eth1 at different nodes (they use the same bridge)
What you cannot do is communicate from eth1 to eth2 and I think that we don't care for that, also we loose dhcp (I think) at knmstate is find and at kubevirt we don't use secondary nics at all, is kubemacpool in need for them ?

cluster-provision/k8s/1.17/provision.sh
qinqon: I found a warning at provisiong from kubeadm missing `tc`

cluster-provision/centos8/scripts/dnsmasq.sh
ormergi: I dont familiar with secondary nics usage for cluster nodes in KMP, 
@RamLavi do you have any idea if we use secondary nics in kmp?

RamLavi: well yeah, kmp allows to allocate mac addresses from a pool to secondary interfaces using Network Plumbing Working Group de-facto standard.

PR #343, opened 2020-04-15
Move k8s-fedora-1.17.0 provider to k8s-fedora-1.17 like the others

qinqon:
project-infra PR https://github.com/kubevirt/project-infra/pull/412


phoracek:
This is not reviewable. It is moving files between directories and also touching existing files. @qinqon would you please split this into multiple commits? So we can go commit by commit in reviewable chunks. Move files is several commits by categories, changes to existing files should be in separate commits.

PR #344, opened 2020-04-23
Enable hugepages again

rmohr:
Hugepages were not enabled anymore. Enabling them now in a less fragile
way and ensure at startup that hugepages are there.

Signed-off-by: Roman Mohr <rmohr@redhat.com>


rmohr:
@dhiller thx
/hold cancel

PR #345, opened 2020-04-27
Make base image ipv6 ready and use it for centos8

oshoval:
Thanks Roman / Edy for your work here.

Update dnsmasq for ipv6 support.

Let dnsmasq advertise ipv6 addresses which can reach each other
over the internal br0 bridge in addition to the ipv4 addresses.

The k8s 1.17/1.18 clusters which are based on centos8, are configured to
automatically ask for an ipv6 address.
The IPv6 addresses will be assigned as follows:
node01: fd00::101
node02: fd00::102
and so forth.

Depends on #380


cluster-provision/k8s/publish.sh
```diff
@@ -4,7 +4,8 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 major_minor="$(basename $(pwd))"
+tag=$(git log -1 --pretty=%h)-$(date +%s)
```
ormergi, line 7: I think you should go for more friendly timestamp format to save complicated conversions
```
echo $(git log -1 --pretty=%h)-$(date +%d%m%y%y_%H%M%S)

1813608-28042020_112614
```

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -2,6 +2,13 @@
 
 set -ex
 
+# Configure IPv6 DHCP
+nmcli connection modify "System eth0" ipv6.method dhcp
```
ormergi, line 6: You can have all these flags in one command, or exporting the NIC name to a variable

cluster-provision/centos8/scripts/dnsmasq.sh
```diff
@@ -34,4 +36,4 @@ iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
 iptables -A FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
 iptables -A FORWARD -i br0 -o eth0 -j ACCEPT
 
-exec dnsmasq -d ${DHCP_HOSTS} --dhcp-range=192.168.66.10,192.168.66.200,infinite
+exec dnsmasq -d ${DHCP_HOSTS} --dhcp-range=192.168.66.10,192.168.66.200,infinite --dhcp-range=::,static
```
ormergi, line 39: --dhcp-range=::
I am almost sure passing `::` to this flag means it will not allocate address

cluster-provision/k8s/publish.sh
oshoval: removed this commit from here

cluster-provision/centos8/scripts/dnsmasq.sh
oshoval: "The optional <mode> keyword may be static which tells dnsmasq to enable 
DHCP for the network specified, but not to dynamically allocate IP 
addresses: only hosts which have static addresses given via dhcp-host or 
from /etc/ethers will be served. A static-only subnet with address all 
zeros may be used as a "catch-all" address to enable replies to all 
Information-request packets on a subnet which is provided with stateless 
DHCPv6, ie --dhcp-range=::,static"

i believe line 25 is the one that gives the ip, this is so it will answer replies,
the nodes do have the ip they should

cluster-provision/k8s/1.17/provision.sh
oshoval: i think its better this way, lets see other opinions

oshoval:
@rmohr 
can you review please (your own 99%)?
thanks

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -2,6 +2,13 @@
 
 set -ex
 
+# Configure IPv6 DHCP
```
EdDev, line 5: What happens if you do not include this change at all?
I am asking to understand the results of the default behavior.

cluster-provision/k8s/1.17/provision.sh
EdDev: I agree with @ormergi , if we need these (see previous question), then something like this is more suited:
```
nmcli connection modify "System eth0" \
    ipv6.method dhcp \
    ipv6.may-fail no \
    ...
&& \
nmcli connection up "System eth0"
```
I do not think the `nmcli connection down` is needed, `up` should be enough.

cluster-provision/gocli/go.mod
```diff
@@ -21,3 +21,5 @@ require (
 )
 
 replace github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.4.1
+
+go 1.13
```
EdDev, line 25: Is this intentional or a leftover?

cluster-provision/k8s/1.17/provision.sh
oshoval: about last comment line
Quique suggested here a reload 

Threre is a nmcli connection reload I think it will do same thing.
which is preferred ? (up only or reload)

cluster-provision/gocli/go.mod
oshoval: will need to ask @rmohr here
(or try and error)

cluster-provision/k8s/1.17/node01.sh
```diff
@@ -47,3 +47,4 @@ done
 
 local_volume_manifest="/tmp/local-volume.yaml"
 kubectl --kubeconfig=/etc/kubernetes/admin.conf create -f "$local_volume_manifest"
+sudo ip -6 route add default dev eth0 metric 1
```
EdDev, line 50: Why this is needed? Is there inter-network communication going on?

oshoval: without this , no ping -6 between the nodes, which is the 3rd point that Calico ipv6 explanation require from the host. 
groomed with Roman its ok and in this file, if its ok

oshoval:
> > `Expose a VNC port for every VM`
> 
> I think this merits a separate PR so it will not be blocked by the IPv6 stuff.
> What do you say about extracting it to a separate PR?
> 
> I guess also the `Update gocli hash to support vnc ports patch` commit needs to move with it.

poped them out of this and moved to #349

cluster-provision/k8s/1.17/provision.sh
oshoval: without it the eth0 wont get this ip
```
    inet6 fd00::101/128 scope global deprecated dynamic noprefixroute 
       valid_lft 7316sec preferred_lft 0sec
```

also the block solve problems like this one
https://github.com/kubevirt/kubevirtci/pull/313#issuecomment-602570373

oshoval:
will updated hashes once lgtm

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -3,11 +3,10 @@
...
 # Configure IPv6 DHCP
-nmcli connection modify "System eth0" ipv6.method dhcp
-nmcli connection modify "System eth0" ipv6.may-fail no
-nmcli connection modify "System eth0" connection.autoconnect yes
-nmcli connection modify "System eth0" ipv6.addr-gen-mode eui64
-nmcli connection down "System eth0" && sudo nmcli connection up "System eth0"
+nmcli connection modify "System eth0" \
+   ipv6.method auto \
+   ipv6.addr-gen-mode eui64
+nmcli connection up "System eth0"
```
EdDev, line 9: Because it is in provisioning stage, this is not needed.
Please correct me if I am wrong.

EdDev:
Actually all 3 commits are not in sync.
Either make one, or split between the dnsmasq and the provisioning of the nodes (i.e. two commits).

cluster-provision/k8s/1.17/provision.sh
oshoval: for example when i tried in the previous config of nmcli just reload, it didnt work
i believe we need it, i can verify if needed

oshoval: Removing it cause the 2nd node to not get ipv6 fd00::102 in cluster-up.

Moreover for the provider of ipv6 we would need it, because the provision would create k8s ipv6 cluster,
unless we will do a split, that just the cluster-up would create ipv6 cluster,
and the provision would create ipv4, which is a possibility.

EdDev: Maybe we are not talking about the same thing?
`nmcli connection up "System eth0"` is activating the NM profile you modified. But as the provisioning is over, this VM does not need this config now, it needs it when it comes up later at `cluster-up`.
What am I missing here?

oshoval: i just tried to remove that exact line `nmcli connection up "System eth0"`
maybe there is an asymetric between nodes.sh and node01.sh or other reason, since its generates according the mac, but without it, node02 and only it doesnt get the fd00::102 address.

EdDev: Maybe this is not running really on the provision step? You can just check that the modification is there when the VM starts.

oshoval: this is the script that runs in the provision step
we can check everything needed in the VM indeed after it starts at the end of cluster-up

EdDev: Or maybe the profile is not saved without the `up` command? Makes no sense to me.

oshoval:
@EdDev 
Can you lgtm so i will update hashes if all good ?
Thanks

EdDev:
The content changes LGTM.

oshoval:
> > Images: Update k8s-1.17 and k8s-1.18 Centos8 base image
> 
> Seems you used the same hash, please quad check.

its the centos, they both use centos8 
i didnt update yet the providers, which would be the different ones

unless i am missing something ?

EdDev:
> > Seems you used the same hash, please quad check.
> 
> its the centos, they both use centos8
> i didnt update yet the providers, which would be the different ones
> 
> unless i am missing something ?

Ohh.. I was wrong.
Where is the output from `/test push-centos8` ?

oshoval:
https://prow.apps.ovirt.org/
search please for `push-centos8`
it was here also but its cleaned after the push because its not a mandatory lane,

EdDev:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/345/push-centos8/1269919934805184512

oshoval:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/345/push-k8s-1.17/1269930378861219840

https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/345/push-k8s-1.18/1269922704140537857

PR #346, opened 2020-04-28
Support time based tags in k8s publish

oshoval:
In order to keep previous images as well,
at least during development and provisioing of new images,
support time based image tag, instead of using latest,
which will cause previous image to be untagged and deleted.
(as was done on ocp providers)

### DIFF
--- a/cluster-provision/k8s/publish.sh
+++ b/cluster-provision/k8s/publish.sh
@@ -4,7 +4,8 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 major_minor="$(basename $(pwd))"
+tag=$(git log -1 --pretty=%h)-$(date +%s)
+destination="docker.io/kubevirtci/k8s-${major_minor}:$tag"
 
-version="$(cat version | tr -d '\n')"
-docker tag kubevirtci/k8s-${major_minor}:latest docker.io/kubevirtci/k8s-${major_minor}:latest
-docker push docker.io/kubevirtci/k8s-${major_minor}:latest
+docker tag kubevirtci/k8s-${major_minor}:latest $destination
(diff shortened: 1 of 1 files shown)


cluster-provision/k8s/publish.sh
```diff
@@ -4,7 +4,8 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 major_minor="$(basename $(pwd))"
+tag=$(git log -1 --pretty=%h)-$(date +%s)
+destination="docker.io/kubevirtci/k8s-${major_minor}:$tag"
 
-version="$(cat version | tr -d '\n')"
-docker tag kubevirtci/k8s-${major_minor}:latest docker.io/kubevirtci/k8s-${major_minor}:latest
-docker push docker.io/kubevirtci/k8s-${major_minor}:latest
+docker tag kubevirtci/k8s-${major_minor}:latest $destination
```
ormergi, line 10: I think you should go for more friendly timestamp format to save complicated conversions
`date +%d%m%y%y_%H%M%S%` :
`1813608-28042020_112614`

cluster-provision/k8s/publish.sh
```diff
@@ -4,7 +4,8 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 major_minor="$(basename $(pwd))"
+tag=$(git log -1 --pretty=%h)-$(date +%s)
```
ormergi, line 7: I think you should go for more friendly timestamp format to save complicated conversions
date +%d%m%y%y_%H%M%S% :
1813608-28042020_112614

oshoval:
closing this, we would need it when the push would be moved to post rule
upon Roman suggestion

PR #347, opened 2020-04-30
Add a kind k8s cluster for 1.17

qinqon:

### DIFF
--- a/cluster-up/cluster/kind-k8s-1.17/README.md
+++ b/cluster-up/cluster/kind-k8s-1.17/README.md
@@ -1,0 +1,33 @@
+# K8S 1.17 in a Kind cluster
+
+Provides a pre-deployed k8s cluster with version 1.17 that runs using [kind](https://github.com/kubernetes-sigs/kind) The cluster is completely ephemeral and is recreated on every cluster restart.
+The KubeVirt containers are built on the local machine and are then pushed to a registry which is exposed at
+`localhost:5000`.
+
+
+## Bringing the cluster up
+
+```bash
+export KUBEVIRT_PROVIDER=kind-k8s-1.17
+export KUBEVIRT_NUM_NODES=2 # master + one node
+make cluster-up
+```
+
+The cluster can be accessed as usual:
+
+```bash
+$ cluster-up/kubectl.sh get nodes
+NAME                        STATUS   ROLES    AGE    VERSION
+kind-1.17-control-plane   Ready    master   105s   v1.17.x
+kind-1.17-worker          Ready    <none>   71s    v1.17.x
+```
+
+## Bringing the cluster down
+
+```bash
+export KUBEVIRT_PROVIDER=kind-k8s-1.17
+make cluster-down
+```
+
+This destroys the whole cluster.
+
--- a/cluster-up/cluster/kind-k8s-1.17/provider.sh
+++ b/cluster-up/cluster/kind-k8s-1.17/provider.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+set -e
+
+DOCKER="${CONTAINER_RUNTIME:-docker}"
+
+export CLUSTER_NAME="kind-1.17"
+export KIND_NODE_IMAGE="kindest/node:v1.17.2"
+
+source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
+
+function up() {
+    cp $KIND_MANIFESTS_DIR/kind.yaml ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
+    kind_up
+
+    # remove the rancher.io kind default storageClass
+    _kubectl delete sc standard
+
+    nodes=$(_kubectl get nodes -o=custom-columns=:.metadata.name | awk NF)
+    for node in $nodes; do
+        # Create local-volume directories, which, on other providers, are pre-provisioned.
+        # For more info, check https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/STORAGE.md
+        for i in {1..10}; do
+            mount_disk $node $i
+        done
+        $DOCKER exec $node bash -c "chmod -R 777 /var/local/kubevirt-storage/local-volume"
+
+        # Create a unique UUID file reference file for each node, that will be mounted in order to support
+        # Migration in Kind providers.
+        $DOCKER exec $node bash -c 'cat /proc/sys/kernel/random/uuid > /kind/product_uuid'
+    done
+
+    # create the `local` storage class - which functional tests assume to exist
+    _kubectl apply -f $KIND_MANIFESTS_DIR/local-volume.yaml
+
+    # Since Kind provider uses containers as nodes, the UUID on all of them will be the same,
+    # and Migration by libvirt would be blocked, because migrate between the same UUID is forbidden.
+    # Enable PodPreset so we can use it in order to mount a fake UUID for each launcher pod.
+    $DOCKER exec kind-1.17-control-plane bash -c 'sed -i \
+    -e "s/NodeRestriction/NodeRestriction,PodPreset/" \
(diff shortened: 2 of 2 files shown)


cluster-up/cluster/kind-k8s-1.17/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Provides a pre-deployed k8s cluster with version 1.14.2 that runs using [kind](https://github.com/kubernetes-sigs/kind) The cluster is completely ephemeral and is recreated on every cluster restart. 
+The KubeVirt containers are built on the local machine and are then pushed to a registry which is exposed at
+`localhost:5000`.
+
+
+## Bringing the cluster up
+
+```bash
+export KUBEVIRT_PROVIDER=kind-k8s-1.14.2
```
oshoval, line 11: version is wrong here and on line 3

cluster-up/cluster/kind-k8s-1.17/README.md
```diff
@@ -0,0 +1,33 @@
...
+make cluster-up
+```
+
+The cluster can be accessed as usual:
+
+```bash
+$ cluster-up/kubectl.sh get nodes
+NAME                        STATUS   ROLES    AGE    VERSION
+kind-1.14.2-control-plane   Ready    master   105s   v1.14.2
+kind-1.14.2-worker          Ready    <none>   71s    v1.14.2
```
oshoval, line 22: label is wrong, it was fixed

cluster-up/cluster/kind-k8s-1.17/README.md
```diff
@@ -0,0 +1,33 @@
...
+Provides a pre-deployed k8s cluster with version 1.14.2 that runs using [kind](https://github.com/kubernetes-sigs/kind) The cluster is completely ephemeral and is recreated on every cluster restart. 
+The KubeVirt containers are built on the local machine and are then pushed to a registry which is exposed at
+`localhost:5000`.
+
+
+## Bringing the cluster up
+
+```bash
+export KUBEVIRT_PROVIDER=kind-k8s-1.14.2
+export KUBEVIRT_NUM_NODES=2 # master + one node
```
oshoval, line 12: this not always work on laptops (at least on ipv6)
but i guess we can just leave it for now ?

cluster-up/cluster/kind-k8s-1.17/README.md
```diff
@@ -0,0 +1,33 @@
...
+$ cluster-up/kubectl.sh get nodes
+NAME                        STATUS   ROLES    AGE    VERSION
+kind-1.14.2-control-plane   Ready    master   105s   v1.14.2
+kind-1.14.2-worker          Ready    <none>   71s    v1.14.2
+```
+
+## Bringing the cluster down
+
+```bash
+export KUBEVIRT_PROVIDER=kind-k8s-1.14.2
```
oshoval, line 28: ditto version

cluster-up/cluster/kind-k8s-1.17/README.md
RamLavi: I think if it's not working then I think it should be removed - I personally wasted time on it not working.. @qinqon  - can you check?

oshoval: it works, just not on laptop of some people , this used to be situation
maybe due to firewall but not sure
on modi it works perfect for ipv6 kind, and also mandatory for migration and such

cluster-up/cluster/kind-k8s-1.17/README.md
```diff
@@ -0,0 +1,33 @@
...
+make cluster-up
+```
+
+The cluster can be accessed as usual:
+
+```bash
+$ cluster-up/kubectl.sh get nodes
+NAME                        STATUS   ROLES    AGE    VERSION
+kind-1.17-control-plane   Ready    master   105s   v1.17.x
+kind-1.17-worker          Ready    <none>   71s    v1.17.x
```
oshoval, line 22: this line is misleading if you want to change it, since this bug is fixed
but no biggie

i just dont understand how it wont give u ipv6 , unless you have other kind config which doesnt config ipv6 ?
saw you reasked review, i will lgtm , feel free to change or not whatever you want , thanks

cluster-up/cluster/kind-k8s-1.17/provider.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -e
+
+export CLUSTER_NAME="kind-1.17"
+export KIND_NODE_IMAGE="kindest/node:v1.17.2"
+
+source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
+
+function up() {
+    cp $KIND_MANIFESTS_DIR/kind.yaml ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
```
phoracek, line 11: In the IPv6 provider.sh we do way more: https://github.com/kubevirt/kubevirtci/blob/ec741ec3a2ad418457c81be260ddf5bc1941f3ca/cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh

I get that this is all we need, but I hate seeing the providers getting out of sync and being implicitly specific to their use-cases.

Would you abstract the IPv6 setup regarding storage and others and use it in both providers?

cluster-up/cluster/kind-k8s-1.17/provider.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -e
+
+export CLUSTER_NAME="kind-1.17"
```
phoracek, line 5: All the other kind providers use X.Y.Z naming. I think we should stay aligned, at least in this PR

cluster-up/cluster/kind-k8s-1.17/provider.sh
oshoval: ah thanks for the explanation , forgot that

hmm about the abstract
i am not sure this is what Roman strived to do ?
i think he separated it by meaning ?
i had a conversion with him on that, maybe worth asking so it wont 
intersect with his new design ?
i might be mistaken here, just recalled for sure we did speak about 
making providers core separated even that it mean less code reuse.

cluster-up/cluster/kind-k8s-1.17/README.md
ormergi: Its not work on laptop with RHEL8 and docker installed, docker does not officially support RHEL8.
on laptop with fedora it works, maybe it worth to mention that the host container-runtime should support IPV6

oshoval: i am not sure at all this is the criteria
moreover with this 
https://www.linuxtechi.com/install-docker-ce-centos-8-rhel-8/
(even without cgroups playing i think)
docker-ce does work
i wouldnt add this info imo unless proven,
Miguel and Yuval K for example had problems with 2 nodes on their laptop if i remember right
sometimes it was firewalld (but not sure about it)
also Roman had problems with 2 nodes

cluster-up/cluster/kind-k8s-1.17/provider.sh
qinqon: @phoracek as the @rmohr effort of untangling providers (meaning we don't re-use we have all at one place) I will copy the storage code at the provider.

cluster-up/cluster/kind-k8s-1.17/provider.sh
qinqon: We are transitioning to major.minor providers new ones has to follow that also is better for CI since you don't have to change project-infra if you upgrade bugfix version at k8s.

phoracek: Ack, but I hate the inconsistency in other kinds. Will not fight on this PR though.

cluster-up/cluster/kind-k8s-1.17/provider.sh
phoracek: Ack

PR #349, opened 2020-05-05
Expose a VNC port for every VM

oshoval:
Ease debugging by exposing a VNC port not just for the first VM. This
makes it easier to see what is going on in VMs if networking does not
work as expected.

Thanks Roman.

### DIFF
--- a/cluster-provision/gocli/go.mod
+++ b/cluster-provision/gocli/go.mod
@@ -21,3 +21,5 @@ require (
 )
 
 replace github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.4.1
+
+go 1.13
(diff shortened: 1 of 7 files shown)


cluster-provision/gocli/go.mod
```diff
@@ -21,3 +21,5 @@ require (
 )
 
 replace github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.4.1
+
+go 1.13
```
EdDev, line 25: Is this a leftover?

oshoval: will reprovision and check,
(also will ask Roman when he is back, but wont block on that)

oshoval:
@rmohr 
I am closing this atm, if needed lets take it in the future please.

PR #350, opened 2020-05-05
check master state

oshoval:
i think master have broken lane
check-provision-k8s-1.17 

lets check


oshoval:
indeed 
provisioning node node01 failed

PR #354, opened 2020-05-09
Implement _port for kind providers

maya-r:
This is used by [CDI's cluster-sync](https://github.com/kubevirt/containerized-data-importer/blob/master/cluster-sync/sync.sh#L32) so we'll have to have an implementation in order to make cluster-sync.

I have some doubts about whether this implementation matches the intention so cc @awels

### DIFF
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -71,6 +71,10 @@ function _configure_registry_on_node() {
     ${NODE_CMD} $1  sh -c "echo $(docker inspect --format '{{.NetworkSettings.IPAddress }}' $REGISTRY_NAME)'\t'registry >> /etc/hosts"
 }
 
+function _port() {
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/kind/common.sh
```diff
@@ -71,6 +71,10 @@ function _configure_registry_on_node() {
     ${NODE_CMD} $1  sh -c "echo $(docker inspect --format '{{.NetworkSettings.IPAddress }}' $REGISTRY_NAME)'\t'registry >> /etc/hosts"
 }
 
+function _port() {
+	docker inspect "${CLUSTER_NAME}"-"$@" --format='{{range $p, $conf := .NetworkSettings.Ports}}{{(index $conf 0).HostPort}}{{end}}'
```
danielBelenky, line 75: Since this function accepts a single mandatory parameter, can you please assign it to a local variable at the beginning of the function? Also, you probably should verify that the variable that this function needs to operate properly was indeed passed or fail if it wasn't.

Something like:

```
function _port() {
	local container_name="${1:?}"
	...
}
```

This will make the function more readable and easier to use by others.

danielBelenky, line 75: And please fix the indentation to match the rest of the file :smile:

PR #356, opened 2020-05-13
sriov: gather sr-iov operator entities

maiqueb:
The sr-iov kind provider is prone to failures; to help debug those,
storing the state of all sr-iov info on the cluster after
deployment is useful.

This PR introduces a stand-alone go program that fetches the
following sriov cluster info:
      - SriovNetwork
      - SriovNetworkNodePolicy
      - SriovNetworkNodeState
      - SriovOperatorConfig
    
Follow-up commits will persist more SR-IOV related info, like logs
of the sr-iov pods, kubelet logs, etc.


maiqueb:
@phoracek do we want to `vendor` the dependencies ? ...

phoracek:
I don't see any harm in it. I'd go for it unless somebody stops us.

rmohr:
Would it make sense to just extend https://github.com/kubevirt/kubevirt/blob/master/tests/reporter/kubernetes.go ? This reporter is already executed on every provider if cluster-up fails, and on the first 10 failing tests locally and in CI.

maiqueb:
> Would it make sense to just extend https://github.com/kubevirt/kubevirt/blob/master/tests/reporter/kubernetes.go ? This reporter is already executed on every provider if cluster-up fails, and on the first 10 failing tests locally and in CI.

I'm OK with having this on that module. Would you be OK with having a separate reporter that would be a member of the `KubernetesReporter` struct ? That way I could pretty much just move this code over.

rmohr:
> I'm OK with having this on that module. Would you be OK with having a separate reporter that would be a member of the `KubernetesReporter` struct ? That way I could pretty much just move this code over.

I would prefer if you create 4 members of the form 

```go
func (r *KubernetesReporter) logVMs(virtCli kubecli.KubevirtClient) {

	f, err := os.OpenFile(filepath.Join(r.artifactsDir, fmt.Sprintf("%d_vms.log", r.failureCount)),
		os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
	if err != nil {
		fmt.Fprintf(os.Stderr, "failed to open the file: %v\n", err)
		return
	}
	defer f.Close()

	vmis, err := virtCli.VirtualMachine(v1.NamespaceAll).List(&metav1.ListOptions{})
	if err != nil {
		fmt.Fprintf(os.Stderr, "failed to fetch vms: %v\n", err)
		return
	}

	j, err := json.MarshalIndent(vmis, "", "    ")
	if err != nil {
		log.DefaultLogger().Reason(err).Errorf("Failed to marshal vms")
		return
	}
	fmt.Fprintln(f, string(j))
}
```
I know a little bit more verbose than your solution, but everything would work as expected and properly integrated.

rmohr:
and of course feel free to take over your generic dumper method, but it should be added to the existing reporter.

maiqueb:
> and of course feel free to take over your generic dumper method, but it should be added to the existing reporter.

I don't mind adding it to the existing reporter, but I think having the ability of running *just* the sriov reporter is a positive thing - if I'm trying to debug sriov deployment issues (on my local deployment), gathering data about the whole thing is unnecessary.

Hence, what I'm proposing is to have an `sriov-reporter` struct and have it as a member of the current `kubernetesReporter` - if we're on an sr-iov lane, also call the `sriov-reporter::Dump()` method. Something like:

```golang
type KubernetesReporter struct {
	failureCount    int
	artifactsDir    string
	maxFails        int
	mux             sync.Mutex
        sriovReporter   *SRIOVReporter
}
...
func (r *KubernetesReporter) Dump(duration time.Duration) {
	virtCli, err := kubecli.GetKubevirtClient()
	if err != nil {
		fmt.Fprintf(os.Stderr, "failed to get client: %v\n", err)
		return
	}
...
	r.logLogs(virtCli, since)
        if(sriovLane) {
                r.sriovReporter.Dump()
        }
}
```

Then have a new `cmd/dump/sriov.go` that would allow users to dump *only* sr-iov data.

rmohr:
We collect all the data anyway already locally. Just look at you _out/artifacts/ folder. If you want to treat it separately you have to think about it on every layer and in the past we always had a lack of data and such issues are very often related. For instance you may have to look on the dumped nodes to see if the sriov devices are exposed, and so forth. Let's try to avoid cognitive overhead. If something goes wrong I just want to look at the logs and follow the traces.

rmohr:
I see little value in an extra dumper. Just a source for confusion.

maiqueb:
Abandoned in favor of: https://github.com/kubevirt/kubevirt/pull/3439

PR #364, opened 2020-05-18
sriov: Make the provider rock solid

phoracek:
While testing the provider in https://github.com/kubevirt/kubevirt/pull/3435, it failed couple of times due to its instabilities. This PR is aiming to resolve all those related to the deployment.

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -22,6 +22,27 @@ function wait_pods_ready {
     done
 }
 
+function wait_for_daemonset {
+  retries=24
+  while [[ $retries -ge 0 ]]; do
+    ready=$(_kubectl -n $1 get daemonset $2 -o jsonpath='{.status.numberReady}')
+    required=$(_kubectl -n $1 get daemonset $2 -o jsonpath='{.status.desiredNumberScheduled}')
+    if [[ $ready -eq $required ]];then
+      break
+    fi
+    sleep 5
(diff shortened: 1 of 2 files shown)


phoracek:
/hold

Work in progress, needs to be tested on KubeVirt first.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,22 +12,46 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+    fi
+    sleep 6
+    ((retries--))
+  done
+}
+
+function wait_for_daemonset {
+  retries=100
+  while [[ $retries -ge 0 ]]; do
+    ready=$(_kubectl -n $1 get daemonset $2 -o jsonpath='{.status.numberReady}')
```
oshoval, line 29: if both fail, then the condition on line 31 will be true
because the strings would match

phoracek: Good catch, thanks. For now I settled with simple check whether the returned value is not empty, but we may need to use one of the mechanisms you've suggested if the provider will be unstable.

oshoval: i think now its good enough, i dont see other issues since you already have a retry + non empty check, am i miss something?

phoracek: Well the only remaining issue is that the lane is still extremely unstable.

PR #366, opened 2020-05-21
Fix k8s provision

oshoval:
Fix support of secondary interfaces, K8s-1.17
    
    NetworkManager-wait-online.service is causing the boot to take
    much longer when secondary interfaces exists,
    as a result, check-up during provision and cluster-up
    would fail connecting to the node as long as the boot isnt finished.
    By installing NetworkManager-config-server,
    existing network profiles wont be changed, and the boot wont
    be stalled by NetworkManager-wait-online.service.

Fix for docker daemon error during provisioning k8s-1.17
    
    docker-ce 19.03.9 cause
    ERROR: Error response from daemon: client version 1.40 is too new.
    Maximum supported API version is 1.39
    
    One of the possible recommended ways is using API 1.39.
    
    See
    https://github.com/kubernetes-sigs/kubespray/issues/6160

Fix NUM_SECONDARY_NICS default value
    
    Wrong default value syntax was used for NUM_SECONDARY_NICS
    

qinqon:
/hold
update hash

oshoval:
@qinqon 
updated hash, and marked ready, thanks
tested cluster-up with prow generated hash, locally, with and without 2nd iinterface
the provision itself checked the cluster-up as well on prow

oshoval:
thanks @qinqon , lgtm was removed, but approve left lol

PR #312, opened 2020-03-18
Bump k8s 1.17 cnao to 0.32.0

qinqon:
It was using a bad multus version that does not support
--cni-version parameter at entrypoint.sh

Signed-off-by: Quique Llorente <ellorent@redhat.com>

PR #368, opened 2020-05-27
macvtap-cni: bump macvtap-cni to v0.2.0

maiqueb:
In order to expose all the host physical devices + bonds for macvtap creation, the CNAO / macvtap-cni versions have to be updated.

This PR takes care of that.


cluster-provision/k8s/1.17/manifests/cnao/operator.yaml
```diff
@@ -153,9 +153,9 @@ spec:
         - name: KUBEMACPOOL_IMAGE
           value: quay.io/kubevirt/kubemacpool:v0.8.3
         - name: MACVTAP_CNI_IMAGE
-          value: quay.io/kubevirt/macvtap-cni:v0.1.0
+          value: quay.io/kubevirt/macvtap-cni:v0.2.0
         - name: OPERATOR_IMAGE
-          value: quay.io/kubevirt/cluster-network-addons-operator:0.32.0
+          value: quay.io/kubevirt/cluster-network-addons-operator:0.35.0
```
maiqueb, line 158: not sure 0.35.0 is a good version (I see 0.35.1, for instance): https://github.com/kubevirt/cluster-network-addons-operator/commits/master

maiqueb:
/hold

haven't tested it locally 0_o

cluster-provision/k8s/1.17/manifests/cnao/operator.yaml
phoracek: I feel you have just replaced values in the yaml directly. Didn't you?

What about bumping to 0.38.0 right away? This PR should not be concerned about macvtap CNI, but about CNAO in the right version.

phoracek: You should make sure to consume released manifests from a certain version. By injecting random changes in the operator yaml may result in mismatch. Like the one you just introduced with the operator image still set to 0.32.0 on line 176

maiqueb: > I feel you have just replaced values in the yaml directly. Didn't you?

Guilty as charged; is it automated ?... If so, how to do it ? 

> 
> What about bumping to 0.38.0 right away? This PR should not be concerned about macvtap CNI, but about CNAO in the right version.

I couldn't care less about whatever is new on 0.35.0 onwards. Having said that, it is exactly the same amount of effort; I can do it.

phoracek: Similar PR by Quique: https://github.com/kubevirt/kubevirtci/pull/312/files

I think you have to do it manually, I can't find a script for it.

Find matching files in https://github.com/kubevirt/cluster-network-addons-operator/releases/tag/0.38.0 and use them as replacement.

cluster-provision/k8s/1.17/manifests/cnao/operator.yaml
```diff
@@ -121,7 +121,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   annotations:
-    networkaddonsoperator.network.kubevirt.io/version: 0.32.0
+    networkaddonsoperator.network.kubevirt.io/version: 0.35.0
```
phoracek, line 124: Please change the commit and PR title+message so it reflects that we bump whole CNAO. You can mention that it is for macvtap in the message, but I don't think it is really needed. Bump is bump, latest greatest.

cluster-provision/k8s/1.17/manifests/cnao/operator.yaml
maiqueb: Yeah, let's first then this to work, then let's decide to which version should we bump this.

cluster-provision/k8s/1.17/manifests/cnao/operator.yaml
maiqueb: I think you're trying to change the scope of this PR in a way that doesn't reflect what I want to do; I want to bump the version of a CNI/ dev plugin. 

CNAO is the means to an end, not the objective. That why I have the message written as is, and that's why I'm grabbing version 0.35.0 and not the latest. I don't care about anything on that container after 0.35.0 . While I said I can do it, since it's basically free, it's not my goal.

phoracek: It is as if you needed newer RPM, bumped OS version for it and commented about the RPM only, while drastically changing many of other dependencies.

If I browse the history to see why the heck did KubeMacPool version changed, I would look for notes about CNAO, not about macvtap-cni.

maiqueb: That's because we apparently bump this by N versions at a time, thus updating N dependencies at once.

Could we start to update CNAO once it is updated, thus preventing these situations ?

I'll rephrase the commit message, since I have no alternative at all; but that doesn't mean I agree with this strategy.

phoracek: > That's because we apparently bump this by N versions at a time, thus updating N dependencies at once.

Single release of CNAO can carry several updated components or none, there is no set relation.

> 
> Could we start to update CNAO once it is updated, thus preventing these situations ?

Yes. Indeed one of many things we should automate.

maiqueb: > > That's because we apparently bump this by N versions at a time, thus updating N dependencies at once.
> 
> Single release of CNAO can carry several updated components or none, there is no set relation.

I had assumed otherwise. What is the advantage of that if I may ask ? 

> 
> > Could we start to update CNAO once it is updated, thus preventing these situations ?
> 

Above I mean *bump* CNAO once it is updated; glad you understood what I meant anyway.
 
> Yes. Indeed one of many things we should automate.

phoracek: > > > That's because we apparently bump this by N versions at a time, thus updating N dependencies at once.
> > 
> > 
> > Single release of CNAO can carry several updated components or none, there is no set relation.
> 
> I had assumed otherwise. What is the advantage of that if I may ask ?

If we know we are going to bump multiple components during a single week, we can release only after it. Note that releasing in CNAO requires a dedicated PR with generated manifest bundles, it requires manual intervention and take a long time (again, needs automation).

CNAO can be seen as a distribution. We collect upstream components and when we think it makes sense (lack of defined rules) to release a new version with the set of components we have, we release.

maiqueb: > > > > That's because we apparently bump this by N versions at a time, thus updating N dependencies at once.
> > > 
> > > 
> > > Single release of CNAO can carry several updated components or none, there is no set relation.
> > 
> > 
> > I had assumed otherwise. What is the advantage of that if I may ask ?
> 
> If we know we are going to bump multiple components during a single week, we can release only after it. Note that releasing in CNAO requires a dedicated PR with generated manifest bundles, it requires manual intervention and take a long time (again, needs automation).

I understand it's a cumbersome process, but that means it should be automated, not 'applied the least possible amount of times'. 

> 
> CNAO can be seen as a distribution. We collect upstream components and when we think it makes sense (lack of defined rules) to release a new version with the set of components we have, we release.

IMHO, the right thing to do is to bump CNAO for each individual upstream component change, as that is the 'unit' here. 

Anyway, I'll sleep on this for a couple of days, and will evaluate opening an issue where we can discuss the 'what' to automate, as this PR is far from the correct forum for this discussion; thanks for the answers.

cluster-provision/k8s/1.17/manifests/cnao/operator.yaml
```diff
@@ -153,13 +153,13 @@ spec:
         - name: KUBEMACPOOL_IMAGE
           value: quay.io/kubevirt/kubemacpool:v0.8.3
         - name: MACVTAP_CNI_IMAGE
-          value: quay.io/kubevirt/macvtap-cni:v0.1.0
+          value: quay.io/kubevirt/macvtap-cni:v0.2.0
```
maiqueb, line 156: why do we have to bump individual components as well ? we are bumping cnao, which already has pinned versions of the upstream components. This feels redundant.

Expanding on that; what if there's a mismatch between the number I set here and the version pinned in CNAO ?

@phoracek @qinqon ^

phoracek: > why do we have to bump individual components as well ? we are bumping cnao, which already has pinned versions of the upstream components. This feels redundant.

The operator definition is the one carrying the image names, so they can be easily overwritten. e.g. on D/S.

> 
> Expanding on that; what if there's a mismatch between the number I set here and the version pinned in CNAO ?

Since manifests are shipped in CNAO image, it can happen that you use manifests meant for component version X with component image version Y.

maiqueb: Thanks.

cluster-up/cluster/images.sh
```diff
@@ -6,8 +6,8 @@ declare -A IMAGES
 IMAGES[gocli]="gocli@sha256:220f55f6b1bcb3975d535948d335bd0e6b6297149a3eba1a4c14cad9ac80f80d"
 if [ -z $KUBEVIRTCI_PROVISION_CHECK ]; then
     IMAGES[k8s-fedora-1.17.0]="k8s-fedora-1.17.0@sha256:aebf67b8b1b499c721f4d98a7ab9542c680553a14cbc144d1fa701fe611f3c0d"
-    IMAGES[k8s-1.18]="k8s-1.18@sha256:f4ea6411610a4f5eb51c43d27177d7450e32240c253883a58fee102c2becffdc"
-    IMAGES[k8s-1.17]="k8s-1.17@sha256:2bb37d65749439d86e53258fb3151697d8339b9e6109aa8d23f577f123380d49"
+    IMAGES[k8s-1.18]="k8s-1.18@sha256:9728f65ab0f4e99657dae528478014aafe3c4a0a232a02e6146c8e699298861d"
```
phoracek, line 9: I should know where do you get this from, but I don't... Would you help me find the source, so I can double-check this?

maiqueb: Would have been good if you had asked this yesterday; it's somewhere on prow, but looks like it's no longer available; I don't know why the reporter doesn't feature the output of both the `/test push-k8s-1.17` jobs.

I'm now looking into [where the push is done](https://hub.docker.com/r/kubevirtci/k8s-1.17/tags) and don't see those images, since there are other jobs that require newer images (https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/373/push-k8s-1.17/1266099932285112320 and https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/373/push-k8s-1.18/1266099932264140800) .

@dhiller looks like we only keep a single image on dockerhub; does this mean I'll have to wait for the other PR to get merged, rebase, generate new images, and update the `sha` ?

maiqueb: Thanks to @qinqon I was able to become a little less prow illiterate; the [PR's job history](https://prow.apps.ovirt.org/?pull=368) shows them; you're looking for [1.17](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/368/push-k8s-1.17/1265914829206458368#1:build-log.txt%3A2071) and [1.18](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/368/push-k8s-1.18/1265914704484634624#1:build-log.txt%3A2086) images shas.

Issue #365, opened 2020-05-20
Support for Vhostuser tests


krsacme:
This issue captures the changes required in order to support Vhostuser (OvS-DPDK) in kubevirtci. PR kubevirt/kubevirt#3208 adds the support for creating a VM with an additional interface based on vhotuser port to a OvS netdev bridge. Following set of issues need to be addressed in order to enable testing of Vhostuser ports.

Provider & OvS RPM Build
------------------------------------
Fedora vagrant cloud image does not have in-built `uio` kernel modules, which is required for `igb_uio` kernel module for using DPDK. Centos 8 vagrant cloud image has `uio` driver. But `OpenvSwitch` with DPDK enabled is not available in either of them. RPM `OpenvSwitch` with DPDK should be built. RPM is required, because the spec file contains the user/group creation for accessing vhostuser ports between OvS and qemu. 

Hugepages for OvS and VM
------------------------------------
OvS-DPDK requires a minimum of 1GB of hugepages memory to initialize DPDK (it can be 2M or 1G sized hugepages). And VM should also be created with hugepages to allow userspace data transfer between OvS-DPDK and VM. The default configuration of 64 2M hugepages of centos provider is not enough.

PMD Driver (DPDK)
---------------------------
`igb_uio` kernel module is required to bind the nic to the PMD to allow DPDK to read and write on the nic. `igb_uio` is part of DPDK source code, which is not build by default. DPDK has to be built inorder to use this driver. Note that, `vfio-pci` could be used with vIOMMU enabled on Q35 machine, but it requires host changes. Using `igb_uio` is less intrusive to run CI on available machines.

ormergi:
Hey, have you try to install `uio` kernel module at provision stage on `k8s-fedora-1.17.0` provider? 
Since we already have a provider with fedora OS nodes that we provision it could solve part of this issue.

Please feel free to reach out if you have question or need help :wink:

krsacme:
> Hey, have you try to install `uio` kernel module at provision stage on `k8s-fedora-1.17.0` provider?
> Since we already have a provider with fedora OS nodes that we provision it could solve part of this issue.
> 
 Yes, I did start with fedora, but fedora is missing the `uio` kernel module, which is a dependency to use `igb_uio`. Here is the log from the local ci cluster:

```
/home/kubevirtci[master]# ./cluster-up/ssh.sh node02
2020/07/27 06:11:52 Waiting for host: 192.168.66.102:22
[...]

PR #371, opened 2020-05-28
[3] Added ovsdpdk scripts to install ovs and pmd driver

krsacme:
OvS-DPDK RPM and igb_uio driver build scripts are added
to the k8s-1.17 kubevirt provider. This script will be
invoked on gocli run command when DPDK is enabled. This
patch only creates the script, the invoking logic is
added in a follow-up patch

Related: #365 

### DIFF
--- a/cluster-provision/k8s/1.17/ovsdpdk.sh
+++ b/cluster-provision/k8s/1.17/ovsdpdk.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+set -ex
+
+setenforce 0
+
+# TODO: Can this be moved to privision script?
+# Having these packages installed does not have impact on the regular deployments
+yum install -y dpdk dpdk-devel
+
+# Container image is creted using the build scripts in repo https://github.com/krsacme/container-builds-scripts
+# By using the prebuilt driver and rpm, the time is reduced in CI to build it
+KUBEVIRT_OVSDPDK_BUILD=${KUBEVIRT_OVSDPDK_BUILD:-"false"}
+KUBEVIRT_OVSDPDK_HELPER_IMAGE=${KUBEVIRT_OVSDPDK_HELPER_IMAGE:-"quay.io/krsacme/kubevirt-ovsdpdk-helpers:latest"}
+if [[ "$KUBEVIRT_OVSDPDK_BUILD" == "false" ]]; then
+    echo "Copying driver and rpm from helper image..."
+    FILES="/tmp/files"
+    mkdir -p $FILES
+    docker run -v $FILES:/host $KUBEVIRT_OVSDPDK_HELPER_IMAGE
+
+    modprobe uio
+    insmod $FILES/igb_uio.ko
+    yum install -y  $FILES/openvswitch-2.13*
+
+    # /proc/self/pagemap permission issue, running as root for now
+    sed -i 's/^OVS_USER_ID/#OVS_USER_ID/g' /etc/sysconfig/openvswitch
+    systemctl enable --now openvswitch
+    exit 0
+fi
+
+###############################################################################
+echo "Building driver and rpm source locally..."
+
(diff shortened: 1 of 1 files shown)


cluster-provision/k8s/1.17/ovsdpdk.sh
```diff
@@ -0,0 +1,91 @@
...
+setenforce 0
+
+# TODO: Can this be moved to privision script?
+# Having these packages installed does not have impact on the regular deployments
+yum install -y dpdk dpdk-devel
+
+# Container image is creted using the build scripts in repo https://github.com/krsacme/container-builds-scripts
+# By using the prebuilt driver and rpm, the time is reduced in CI to build it
+KUBEVIRT_OVSDPDK_BUILD=${KUBEVIRT_OVSDPDK_BUILD:-"false"}
+KUBEVIRT_OVSDPDK_HELPER_IMAGE=${KUBEVIRT_OVSDPDK_HELPER_IMAGE:-"quay.io/krsacme/kubevirt-ovsdpdk-helpers:latest"}
```
dhiller, line 14: Are you considering to move the image into the kubevirtci docker repo at a later point in time?

cluster-provision/k8s/1.17/ovsdpdk.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+set -ex
+
+setenforce 0
```
dhiller, line 5: Why is that necessary? I understood we were making it possible to leave selinux enabled since 1.17? @jean-edouard WDYT?

cluster-provision/k8s/1.17/ovsdpdk.sh
krsacme: > I understand that you are planning to install dpdk libs on the cluster nodes everytime the cluster is spun up, right?
> 
> Wouldn't it make sense to create a separate provider (i.e `k8s-1.17-dpdk`) so to reduce spin up time instead of having to install everything everytime you do `cluster-up`?

Thanks @dhiller for the comments. 

Yes, it would be ideal if we are able to create ``k8s-1.x-dpdk`` images for each release. I am unable to idenify how these images are built (actual scripts to run the ``provision`` command to create the ``k8s.1.17`` image). Can you point how these images are created and pushed?

One drawback for a separate image is, the release process will have to depend on the DPDK version and its compatibility with the CentOS version. Though it is not a riskier option (because we use only LTS version), I am not sure if it is agreeable to include this dependency in the general release process of a new version. If it is acceptable to create a new image, then the same script can be invoked by ``provision`` command instead of ``run`` command in the follow-up patch.

Yes, if we invoke this as part of the ``run`` without creating a new image, it will include additional CI run time, for DPDK jobs. Here are the major steps:
1) Update kernel package
2) Reboot
3) Pull Docker image for kernel module and ovs rpm
4) OvS RPM install

All these steps increse the time around 4-5 mins based on the CentOS repo sync.

> Are you considering to move the image into the kubevirtci docker repo at a later point in time?

If the seprate image creating is acceptable, then we can remove this container and build the image DPDK everytime the ``k8s.1.17-dpdk`` image is created. In case it is not agreeable solution, then yes, I will plan to move this container image and the repo to build this container image [container-build-scripts](https://github.com/krsacme/container-builds-scripts) in to kubevirt umberla.

cluster-provision/k8s/1.17/ovsdpdk.sh
krsacme: There is need to share the vhostuser ports created by libvirt inside the EmptyDir volume to be used by OvS ports, by locally mounting it to ``/var/lib/vhost_sockets`` directory (addtional details [here](https://github.com/kubevirt/kubevirt/pull/3208#issuecomment-607765265)) . It had access failures, it requires update in selinux policies, which can be done in order to remove this. This change will be temporary.

PR #373, opened 2020-05-28
Fix docker version issues, bump k8s-1.17 micro and temporarily disable dontaudit on k8s-1.17

jean-edouard:


dhiller:
/lgtm
/hold

I guess you need to remove the temporary disablement.

jean-edouard:
@dhiller well it's not *that* temporary... It is there to ensure that no flaky test fails because of SELinux. The idea is to run tests with that for a few days to collect artifacts info and then revert the change. If that's not acceptable though, I can remove the commit.

dhiller:
> @dhiller well it's not _that_ temporary... It is there to ensure that no flaky test fails because of SELinux. The idea is to run tests with that for a few days to collect artifacts info and then revert the change. If that's not acceptable though, I can remove the commit.

OK, understood. SGTM. Still holding for pushing and updating image shas.

dhiller:
Meh, jobs are failing just due to quay outage :-(

rmohr:
Do we need the same for 1.18?

jean-edouard:
@rmohr which part?

PR #374, opened 2020-05-29
[5] Create OvS-DPDK operator when enabled

krsacme:
In order to create a VM with vhostuser interface, OvS-DPDK should
be enabled and relevant parameter should be configured.
ovsdpdk-network-operator does the configuration based on the
provided interface configuration. In order to use vhostuser
interface on VMs as additional interface, a secondary
interface should be added to the deployed nodes. And
operator config will specify the additional interface eth1
to configure it for vhostuser socket. Relevant manifests
are added to the provider to perform this operation.

Fixes: #365 

### DIFF
--- a/cluster-up/cluster/k8s-1.17/manifests/ovsdpdk/allinone.yaml
+++ b/cluster-up/cluster/k8s-1.17/manifests/ovsdpdk/allinone.yaml
@@ -1,0 +1,310 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: ovsdpdk-network-operator
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: ovsdpdk-network-operator
+rules:
+- apiGroups: [""]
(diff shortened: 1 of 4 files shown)


cluster-up/cluster/k8s-1.17/manifests/ovsdpdk/allinone.yaml
```diff
@@ -0,0 +1,310 @@
...
+      name: ovsdpdk-network-operator
+  template:
+    metadata:
+      labels:
+        name: ovsdpdk-network-operator
+    spec:
+      serviceAccountName: ovsdpdk-network-operator
+      containers:
+        - name: ovsdpdk-network-operator
+          image: quay.io/krsacme/ovsdpdk-network-operator:latest
```
rmohr, line 191: You can't reference "latest" images. Please also pre-pull the image.

krsacme: Updated the ovsdpdk operator version to a specific version.

PR #301, opened 2020-03-03
Fix OCP-4.4 provision

oshoval:
Update the release-4.4 patch file we apply to fit the checkout version.

Signed-off-by: Or Shoval <oshoval@redhat.com>

PR #378, opened 2020-06-03
Stabilize SRIOV provider

ormergi:
This PR stabilize SRIOV provider as part of stabilizing sriov-lane effort.
Adding extra checks and validations, ensuring:
1. All webhook configurations objects are created
2. Ensuring luster nodes are functional after sriov webhooks caBundle patching
4. SriovNodeNetworkPolicy is created and available.
5. `sriov-cni` and `sriov-device-plugin` pods created and ready after SriovNodeNetworkPolicy creation.
6. Cluster nodes are functional after creating SriovNodeNetworkPolicy.
7. Affected node has sriov nics "allocateable" resource as requested in SriovNodeNetworkPolicy
8. Ensure Multus deployment is ready
9. Re-ordering the script to: nodes configurations, cluster configurations and sriov-operator deployment.
10. Refactoring a few functions

This PR following #364 

This PR depends on #383

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -79,6 +79,44 @@ function wait_k8s_object {
   echo $error_message && return  1
 }
 
+
+function is_taint_absence {
+  local -r taint=$1
+
+  result=$(_kubectl get nodes -o jsonpath="{.items[*].spec.taints[?(@.effect == \"$taint\")].effect}" || echo error)
+  if [[ -z $result ]]; then
+    echo "not-present"
+  fi
+}
+
+function wait_for_taint_absence {
+  local -r taint=$1
+
+  local -r tries=60
+  local -r wait_time=5
+
+  local -r wait_message="Waiting for $taint taint absence"
+  local -r error_message="Taint $taint $name did not removed"
+  local -r action="is_taint_absence $taint"
+
+  retry "$tries" "$wait_time" "$action" "$wait_message" && return 0
+  echo $error_message && return 1
+}
+
+function wait_for_taint {
+  local -r taint=$1
+
+  local -r tries=60
+  local -r wait_time=5
+
+  local -r wait_message="Waiting for $taint taint to present"
+  local -r error_message="Taint $taint $name did not present"
+  local -r action="_kubectl get nodes -o custom-columns=taints:.spec.taints[*].effect --no-headers | grep -i $taint"
+
+  retry "$tries" "$wait_time" "$action" "$wait_message" && return 0
+  echo $error_message && return 1
+}
+
 # not using kubectl wait since with the sriov operator the pods get restarted a couple of times and this is
 # more reliable
 function wait_pods_ready {
@@ -208,5 +246,11 @@ wait_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_DEVICE_PLUGIN_LABEL || exit 1
 # Wait for cni and device-plugin pods to be ready
 _kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l $SRIOV_CNI_LABEL           --for condition=Ready --timeout 10m
 _kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l $SRIOV_DEVICE_PLUGIN_LABEL --for condition=Ready --timeout 10m
+
+# Since SriovNodeNetworkPolicy doesnt have Status to indicate if its
+# configured successfully, it is necessary to wait for the "NoSchedule"
+# taint to present and then absent.
+wait_for_taint "NoSchedule" || true
+wait_for_taint_absence "NoSchedule" || exit  1
 
 ${SRIOV_NODE_CMD} chmod 666 /dev/vfio/vfio
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -34,6 +34,30 @@ function retry {
   return 1
 }
 
+function wait_pod {
+  local namespace=$1
+  local label=$2
+
+  local -r tries=60
+  local -r wait_time=5
+
+  local -r wait_message="Waiting for pods with $label to create"
+  local -r error_message="Pods with  label $label at $namespace namespace found"
+
+  if [[ $namespace != "" ]];then
+    namespace="-n $namespace"
+  fi
+
+  if [[ $label != "" ]];then
+    label="-l $label"
+  fi
+
+  local -r action="_kubectl get pod $namespace $label -o custom-columns=NAME:.metadata.name --no-headers"
+
+  retry "$tries" "$wait_time" "$action" "$wait_message" && return  0
+  echo $error_message && return 1
+}
+
 function wait_k8s_object {
   local -r object_type=$1
   local -r name=$2
@@ -169,10 +193,20 @@ sleep 60
 envsubst < $MANIFESTS_DIR/network_config_policy.yaml | _kubectl create -f -
 
 SRIOV_OPERATOR_NAMESPACE="sriov-network-operator"
+SRIOV_CNI_LABEL="app=sriov-cni"
+SRIOV_DEVICE_PLUGIN_LABEL="app=sriov-device-plugin"
 
 # Ensure SriovNetworkNodePolicy CR is created
 policy_name=$(cat $MANIFESTS_DIR/network_config_policy.yaml | grep 'name:' | awk '{print $2}')
 wait_k8s_object "SriovNetworkNodePolicy" $policy_name $SRIOV_OPERATOR_NAMESPACE  || exit 1
 
+# Wait for sriov-operator to reconcile SriovNodeNetworkPolicy
+# and create cni and device-plugin pods
+wait_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_CNI_LABEL  || exit 1
+wait_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_DEVICE_PLUGIN_LABEL || exit 1
+
+# Wait for cni and device-plugin pods to be ready
+_kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l $SRIOV_CNI_LABEL           --for condition=Ready --timeout 10m
+_kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l $SRIOV_DEVICE_PLUGIN_LABEL --for condition=Ready --timeout 10m
 
 ${SRIOV_NODE_CMD} chmod 666 /dev/vfio/vfio
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -329,3 +329,7 @@ wait_pods_ready
 resource_name=$(cat $policy | grep 'resourceName:' | awk '{print $2}')
 wait_allocatable_resource $SRIOV_NODE "openshift.io/$resource_name" $NODE_PF_NUM_VFS || exit 1
 
+_kubectl get nodes
+_kubectl get pods -n $SRIOV_OPERATOR_NAMESPACE
+echo
+echo "$KUBEVIRT_PROVIDER cluster is ready"
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -12,6 +12,49 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
 
 OPERATOR_GIT_HASH=8d3c30de8ec5a9a0c9eeb84ea0aa16ba2395cd68  # release-4.4
 
+# This function gets a command string and invoke it
+# until the command returns an empty string or until timeout
+function retry {
+  local -r tries=$1
+  local -r wait_time=$2
+  local -r action=$3
+  local -r wait_message=$4
+
+  local result=$(eval $action)
+  for i in $(seq $tries); do
+    if [[ -z $result ]] ; then
+      echo "[$i/$tries] $wait_message"
+      sleep $wait_time
+      result=$(eval $action)
+    else
+      return 0
+    fi
+  done
+
+  return 1
+}
+
+function wait_k8s_object {
+  local -r object_type=$1
+  local -r name=$2
+  local namespace=$3
+
+  local -r tries=60
+  local -r wait_time=3
+
+  local -r wait_message="Waiting for $object_type $name"
+  local -r error_message="$object_type $name at $namespace namespace found"
+
+  if [[ $namespace != "" ]];then
+    namespace="-n $namespace"
+  fi
+
+  local -r action="_kubectl get $object_type $name $namespace -o custom-columns=NAME:.metadata.name --no-headers"
+
+  retry "$tries" "$wait_time" "$action" "$wait_message" && return 0
+  echo $error_message && return  1
+}
+
 # not using kubectl wait since with the sriov operator the pods get restarted a couple of times and this is
 # more reliable
 function wait_pods_ready {
@@ -110,8 +153,10 @@ _kubectl label node $SRIOV_NODE sriov=true
 
 wait_pods_ready
 
-# we need to sleep as the configurations below need to appear
-sleep 30
+# Ensure webook-configuration object created
+wait_k8s_object "validatingwebhookconfiguration" "operator-webhook-config"  || exit 1
+wait_k8s_object "mutatingwebhookconfiguration"   "operator-webhook-config"  || exit 1
+wait_k8s_object "mutatingwebhookconfiguration"   "network-resources-injector-config"  || exit 1
 
 _kubectl patch validatingwebhookconfiguration operator-webhook-config --patch '{"webhooks":[{"name":"operator-webhook.sriovnetwork.openshift.io", "clientConfig": { "caBundle": "'"$(cat $CSRCREATORPATH/operator-webhook.cert)"'" }}]}'
 _kubectl patch mutatingwebhookconfiguration network-resources-injector-config --patch '{"webhooks":[{"name":"network-resources-injector-config.k8s.io", "clientConfig": { "caBundle": "'"$(cat $CSRCREATORPATH/network-resources-injector.cert)"'" }}]}'
@@ -120,8 +165,14 @@ _kubectl patch mutatingwebhookconfiguration operator-webhook-config --patch '{"w
 # we need to sleep to wait for the configuration above the be picked up
 sleep 60
 
+# Substitute NODE_PF and NODE_PF_NUM_VFS then create SriovNetworkNodePolicy CR
 envsubst < $MANIFESTS_DIR/network_config_policy.yaml | _kubectl create -f -
+
+SRIOV_OPERATOR_NAMESPACE="sriov-network-operator"
+
+# Ensure SriovNetworkNodePolicy CR is created
+policy_name=$(cat $MANIFESTS_DIR/network_config_policy.yaml | grep 'name:' | awk '{print $2}')
+wait_k8s_object "SriovNetworkNodePolicy" $policy_name $SRIOV_OPERATOR_NAMESPACE  || exit 1
 
 
 ${SRIOV_NODE_CMD} chmod 666 /dev/vfio/vfio
-
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -231,6 +231,40 @@ function deploy_sriov_operator {
   return 0
 }
 
+function apply_sriov_node_policy {
+  policy_file=$1
+
+  SRIOV_OPERATOR_NAMESPACE="sriov-network-operator"
+  SRIOV_DEVICE_PLUGIN_LABEL="app=sriov-device-plugin"
+  SRIOV_CNI_LABEL="app=sriov-cni"
+
+  echo "Applying SriovNetworkNodeConfigPolicy:"
+  cat $policy_file
+  # Substitute $NODE_PF and $NODE_PF_NUM_VFS and create SriovNetworkNodePolicy CR
+  envsubst < $policy_file | _kubectl create -f -
+
+  # Ensure SriovNetworkNodePolicy CR is created
+  policy_name=$(cat $policy_file | grep -Po '(?<=name:) \S*')
+  wait_k8s_object "SriovNetworkNodePolicy" "$policy_name" "$SRIOV_OPERATOR_NAMESPACE"  || return 1
+
+  # Wait for sriov-operator to reconcile SriovNodeNetworkPolicy
+  # and create cni and device-plugin pods
+  wait_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_CNI_LABEL  || return 1
+  wait_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_DEVICE_PLUGIN_LABEL || return 1
+
+  # Wait for cni and device-plugin pods to be ready
+  _kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l $SRIOV_CNI_LABEL           --for condition=Ready --timeout 10m
+  _kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l $SRIOV_DEVICE_PLUGIN_LABEL --for condition=Ready --timeout 10m
+
+  # Since SriovNodeNetworkPolicy doesnt have Status to indicate if its
+  # configured successfully, it is necessary to wait for the "NoSchedule"
+  # taint to present and then absent.
+  wait_for_taint "NoSchedule" || true
+  wait_for_taint_absence "NoSchedule" || return  1
+
+  return 0
+}
+
 # The first worker needs to be handled specially as it has no ending number, and sort will not work
 # We add the 0 to it and we remove it if it's the candidate worker
 WORKER=$(_kubectl get nodes | grep $WORKER_NODE_ROOT | sed "s/\b$WORKER_NODE_ROOT\b/${WORKER_NODE_ROOT}0/g" | sort -r | awk 'NR==1 {print $1}')
@@ -287,32 +321,11 @@ deploy_sriov_operator || exit 1
 wait_pods_ready
 
 # Substitute NODE_PF and NODE_PF_NUM_VFS then create SriovNetworkNodePolicy CR
-envsubst < $MANIFESTS_DIR/network_config_policy.yaml | _kubectl create -f -
-
-SRIOV_OPERATOR_NAMESPACE="sriov-network-operator"
-SRIOV_CNI_LABEL="app=sriov-cni"
-SRIOV_DEVICE_PLUGIN_LABEL="app=sriov-device-plugin"
-
-# Ensure SriovNetworkNodePolicy CR is created
-policy_name=$(cat $MANIFESTS_DIR/network_config_policy.yaml | grep 'name:' | awk '{print $2}')
-wait_k8s_object "SriovNetworkNodePolicy" $policy_name $SRIOV_OPERATOR_NAMESPACE  || exit 1
-
-# Wait for sriov-operator to reconcile SriovNodeNetworkPolicy
-# and create cni and device-plugin pods
-wait_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_CNI_LABEL  || exit 1
-wait_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_DEVICE_PLUGIN_LABEL || exit 1
-
-# Wait for cni and device-plugin pods to be ready
-_kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l $SRIOV_CNI_LABEL           --for condition=Ready --timeout 10m
-_kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l $SRIOV_DEVICE_PLUGIN_LABEL --for condition=Ready --timeout 10m
-
-# Since SriovNodeNetworkPolicy doesnt have Status to indicate if its
-# configured successfully, it is necessary to wait for the "NoSchedule"
-# taint to present and then absent.
-wait_for_taint "NoSchedule" || true
-wait_for_taint_absence "NoSchedule" || exit  1
+policy="$MANIFESTS_DIR/network_config_policy.yaml"
+apply_sriov_node_policy "$policy" || exit 1
+wait_pods_ready
 
 # Verify that sriov node has sriov VFs allocatable resource
-resource_name=$(cat $MANIFESTS_DIR/network_config_policy.yaml | grep 'resourceName:' | awk '{print $2}')
+resource_name=$(cat $policy | grep 'resourceName:' | awk '{print $2}')
 wait_allocatable_resource $SRIOV_NODE "openshift.io/$resource_name" $NODE_PF_NUM_VFS || exit 1
 
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -34,6 +34,29 @@ function retry {
   return 1
 }
 
+function wait_for_daemonSet {
+  local name=$1
+  local namespace=$2
+  local required_replicas=$3
+
+  if [[ $namespace != "" ]];then
+    namespace="-n $namespace"
+  fi
+
+  if (( required_replicas < 0 )); then
+      echo "DaemonSet $name ready replicas number is not valid: $required_replicas" && return 1
+  fi
+
+  local -r tries=30
+  local -r wait_time=10
+  wait_message="Waiting for DaemonSet $name to have $required_replicas ready replicas"
+  error_message="DaemonSet $name did not have $required_replicas ready replicas"
+  action="_kubectl get daemonset $namespace $name -o jsonpath='{.status.numberReady}' | grep -w $required_replicas"
+
+  retry "$tries" "$wait_time" "$action" "$wait_message" && return  0
+  echo $error_message && return 1
+}
+
 function wait_pod {
   local namespace=$1
   local label=$2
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -110,8 +130,10 @@ _kubectl label node $SRIOV_NODE sriov=true
 
 wait_pods_ready
 
-# we need to sleep as the configurations below need to appear
-sleep 30
+# Ensure webook-configuration object created
+ensure_cr "validatingwebhookconfiguration" "operator-webhook-config"
```
phoracek, line 204: None of these object types are custom, all are part of Kubernetes. Please rename the function to `ensure_object` or something more generic.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -139,8 +157,10 @@ _kubectl patch validatingwebhookconfiguration operator-webhook-config --patch '{
 _kubectl patch mutatingwebhookconfiguration network-resources-injector-config --patch '{"webhooks":[{"name":"network-resources-injector-config.k8s.io", "clientConfig": { "caBundle": "'"$(cat $CSRCREATORPATH/network-resources-injector.cert)"'" }}]}'
 _kubectl patch mutatingwebhookconfiguration operator-webhook-config --patch '{"webhooks":[{"name":"operator-webhook.sriovnetwork.openshift.io", "clientConfig": { "caBundle": "'"$(cat $CSRCREATORPATH/operator-webhook.cert)"'" }}]}'
 
-# we need to sleep to wait for the configuration above the be picked up
-sleep 60
+# Ensure caBundle is configured with new certificates
+ensure_caBundle "validatingwebhookconfiguration" "operator-webhook-config"           "$(cat $CSRCREATORPATH/operator-webhook.cert)"
```
phoracek, line 213: I'm not sure this will be enough. This only waits for the caBundle to be set, that should happen as soon as we _patch. The timeout was here mostly to wait until the changed caBundle is picked up and used by the webhook. I think this change is not needed. That being said, I don't have a good way how to check for webhook readiness to serve.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -166,10 +189,21 @@ ensure_caBundle "mutatingwebhookconfiguration"   "network-resources-injector-con
 envsubst < $MANIFESTS_DIR/network_config_policy.yaml | _kubectl create -f -
 
 sriov_operator_namespace="sriov-network-operator"
```
phoracek, line 225: These are not configurable nor dynamically created. Would you please use UPPER_CASE to signalize those are constants?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -166,10 +189,21 @@ ensure_caBundle "mutatingwebhookconfiguration"   "network-resources-injector-con
...
+sriov_cni_label="app=sriov-cni"
+sriov_device_plugin_label="app=sriov-device-plugin"
 
 # Ensure SriovNetworkNodePolicy CR is created
 policy_name=$(cat $MANIFESTS_DIR/network_config_policy.yaml | grep 'name:' | awk '{print $2}')
 ensure_cr "SriovNetworkNodePolicy" $policy_name $sriov_operator_namespace
 
+# Wait for sriov-operator to reconsile SriovNodeNetworkPolicy
```
phoracek, line 233: This comment does not really describe what is done on following lines. It just waits for pods to be created. Could we do that sooner? Right deploying the operator? And fix the comment

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: In line 217 we check if caBundle is picked up by waiting for node taint `NoShcedule` to be removed,
as soon as you patch those webhooks the taint is there, I think its enough because once the taint is removed all the pods are ready, what you think?

ensure_caBundle verifies that caBundle got the right hash, I understand why it my seem redundant

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I elaborated on the commit message, should I document it in the code?
`sriov-cn` and `sriov-device-plugin` pods are rescheduled even though there are already on a node
that should be affected by applying a policy, so its not enough to check it right after operator deployment.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: np

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: ack

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
phoracek: Please drop it, let's just trust `kubectl patch` and reduce the amount of code that can break (and we need to maintain)

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
phoracek: ah, makes sense

phoracek:
Let's drop the unneeded code and verify this change on kubevirt.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: ack

ormergi:
> Let's drop the unneeded code and verify this change on kubevirt.

done,
I just realized we use flannel in kind providers, 
I rebased this PR on top of  another branch with changes to deploy calico.

cluster-up/cluster/kind/common.sh
```diff
@@ -154,13 +154,13 @@ function setup_kind() {
 
     echo "ipv6 cni: $IPV6_CNI"
     if [ -z ${IPV6_CNI+x} ]; then
-        echo "no ipv6, safe to install flannel"
-        _kubectl apply -f $KIND_MANIFESTS_DIR/kube-flannel.yaml
+        echo "no ipv6, safe to install calico"
```
phoracek, line 157: This change has merit on its own and is not necessarily connected with SR-IOV. Please open a separate PR for it

ormergi: #383

ormergi:
> > Let's drop the unneeded code and verify this change on kubevirt.
> 
> done,
> I just realized we use flannel in kind providers,
> I rebased this PR on top of another branch with changes to deploy calico.

PR for that branch https://github.com/kubevirt/kubevirtci/pull/383

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,81 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+
+function wait_for_taint {
+  taint=$1
+
+  intervals=30
+  timeout=10
+  let "total = ($timeout * $intervals) / 60"
+
+  count=0
+  until [[ $(_kubectl  get nodes -ocustom-columns=taints:.spec.taints[*].effect --no-headers | grep -i $taint) != "" ]]; do
```
oshoval, line 66: this and line 82 are almost the same, beside the exit and break,
maybe worth a function ?

also there is a double space before `get nodes`

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,81 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+      sleep $timeout
+  done
+}
+
+function ensure_k8s_object {
+  object_type=$1
+  name=$2
+  namespace=$3
+
+  intervals=30
```
oshoval, line 43: the retry is happening on more than one function
maybe worth to move the retry logic outside as 
https://gist.github.com/dublx/e99ea94858c07d2ca6de

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -22,6 +97,28 @@ function wait_pods_ready {
...
+  timeout=10
+
+  count=0
+  resource_name=$(echo $resource_name | sed s/\\./\\\\\./g)
+  current_value=$(_kubectl get node $node -o "custom-columns=CAPACITY:.status.allocatable.$resource_name" --no-headers)
+  until [[ $current_value == $expected_value ]];do
+    ((count++)) && ((count == intervals)) && "node $node doesnt have allocatabale resource $resource_name:$expected_value" && exit 1
+    echo "$node node should have allocateable resource:
+      $resource_name: $expected_value 
+      got:
```
oshoval, line 115: ginko style

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I think this should be done in another PR because there are more code we can export to a helper functions file

oshoval: Its a new code that have this issue couple time, but whatever you want.

ormergi: I understand, I checked the example you sent and in order to implement "retry" function 
It is necessary to pass bash command string to that function, 
I find it a bit risky and prune to bugs and malicious code (even-though this script used by cluster-up), this is why I did not do it in the first place.
What you think?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,26 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+  object_type=$1
+  name=$2
+  namespace=$3
+
+  intervals=30
+  timeout=3
+
+  if [[ $namespace != "" ]];then
+    namespace="-n $namespace"
+  fi
```
EdDev, line 48: I think you should use `: ${namespace:=$3}`

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,26 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+
+  count=0
+  until _kubectl get $object_type $name $namespace; do
```
EdDev, line 51: I think this format is better for such cases:
`for i in {1..$timeout}; do <operations> && break || sleep $interval; done`
You may need to save a return value so you can return it, it depends what is the return value of the `for` when it finishes without `break`.

You should not use `exit` in the function, it exists the whole script and that is to be determined by the caller not by this specific function.

The naming also needs to fit some reasonable well knows naming.
`timeout` is the time in total you are willing to wait for the operation to succeed.
`interval` is the poll accuracy where you re-check the status to exit earlier.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,26 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
 
 OPERATOR_GIT_HASH=8d3c30de8ec5a9a0c9eeb84ea0aa16ba2395cd68  # release-4.4
 
+function ensure_k8s_object {
+  object_type=$1
```
EdDev, line 39: All these variables better be set as `local` with `-r` to make then read-only, otherwise they are shared all over.
Example: `local -r object_type=$1`.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: (we talked on other media)

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,26 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
 
 OPERATOR_GIT_HASH=8d3c30de8ec5a9a0c9eeb84ea0aa16ba2395cd68  # release-4.4
 
+function ensure_k8s_object {
```
EdDev, line 38: How about `wait_for_k8s_object` ?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -110,18 +207,45 @@ _kubectl label node $SRIOV_NODE sriov=true
...
+policy_name=$(cat $MANIFESTS_DIR/network_config_policy.yaml | grep 'name:' | awk '{print $2}')
+ensure_k8s_object "SriovNetworkNodePolicy" $policy_name $SRIOV_OPERATOR_NAMESPACE
+
+# Wait for sriov-operator to reconsile SriovNodeNetworkPolicy
+# and deploy cni and device-plugin pods
+ensure_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_CNI_LABEL
+ensure_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_DEVICE_PLUGIN_LABEL
+
+# Wait for cni and device-plugin pods to be ready
+_kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l app=sriov-cni           --for condition=Ready --timeout 300s
```
phoracek, line 241: You can reuse SRIOV_CNI_LABEL here

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -110,18 +207,45 @@ _kubectl label node $SRIOV_NODE sriov=true
...
+_kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l app=sriov-device-plugin --for condition=Ready --timeout 300s
```
phoracek, line 242: ditto

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: ack

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: done

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I will change all 'ensure' to 'wait' because those functions dont  act when what need to be ensured is missing.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Can you please elaborate? (on which line, what is the difference)

EdDev: Instead of:
```
  local namespace=$3
  if [[ $namespace != "" ]];then
    namespace="-n $namespace"
  fi
```

You do:
```
: ${namespace:=$3}
```

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,102 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
 
+# This function gets a command string and invoke it
+# until the command returns an empty string or until timeout
+function retry {
+  local -r tries=$1
+  local -r wait_time=$2
+  local -r action=$3
+  local -r wait_message=$4
+
+  local result=$(eval $action)
```
EdDev, line 23: I think this needs to be in the for loop:
```
for i in ..; do
    result=$(...)
    if [ -z result ] ...
        echo ...
        sleep ...
    else  // probably swapping by negating the condition will make more sense.
        break
    fi
done
return result
```

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,49 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+    if [[ -z $result ]] ; then
+      echo "[$i/$tries] $wait_message"
+      sleep $wait_time
+      result=$(eval $action)
+    else
+      return 0
+    fi
+  done
+
+  return 1
```
EdDev, line 34: You should return with the `result` code, not mask it.

phoracek:
/approve cancel

Until all comments are resolved.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,100 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+
+function wait_for_taint_absence {
+  local -r taint=$1
+
+  local -r tries=60
+  local -r wait_time=5
+
+  local -r wait_message="Waiting for $taint taint absence"
+  local -r error_message="Taint $taint $name did not removed"
+  local -r action='[[ -z $(_kubectl get nodes -o custom-columns=taints:.spec.taints[*].effect --no-headers | grep -i $taint) ]] && echo false || echo '
```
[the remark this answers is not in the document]
oshoval: would you please inject a hiccup, for this command
```
 local -r action='[[ -z $(_kubectl get nodes -o custom-columns=taints:.spec.taints[*].effect --no-headers | grep -i $taint) ]] && echo false || echo '
````
its meant to be echo false ? or should be false ?
because the condition for retry is -z, and echo false print the string false
(unless there is a double eval that i miss)
just want to make sure please it actually retry on hiccups, unless you decide its not part of this PR.
thanks

File: cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh [hunk 18]
```diff
@@ -12,6 +12,102 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+
+function wait_for_taint_absence {
+  local -r taint=$1
+
+  local -r tries=60
+  local -r wait_time=5
+
+  local -r wait_message="Waiting for $taint taint absence"
+  local -r error_message="Taint $taint $name did not removed"
+  local -r action='[[ -z $(_kubectl get nodes -o custom-columns=taints:.spec.taints[*].effect --no-headers | grep -i $taint) ]] && echo taint-removed || echo '
```
oshoval, line 91: see please the question about inject an error here
i believe that hiccup wont retry
since it will just think there arent taints

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: The logic here is simple, checking if the command you passed returns an empty string, if it is we loop continues,
if it dosent return 0 (success).
`retry` mask the result with '1' and '0' so when you call it you wont need to check whether the return value is empty string or not, the code will be cleaner and more straightforward.
Because of how `kubectl` return values are, it is necessary for the command to return an empty string if it fails (using grep for example)
This script is complex already, I am trying not to complicate it even more :)

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: how should I inject an error or simulate a "hiccup"?

I checked this function like this:
I added a taint to one of the nodes and ran `wait_for_taint_absence` function (manually no through the script)
and it keep running and checking if there is a taint (according to the action passed to retry)
then in another session I removed the taint and the function returned.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,102 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+
+function wait_for_taint {
+  local -r taint=$1
+
+  local -r tries=60
+  local -r wait_time=5
+
+  local -r wait_message="Waiting for $taint taint to present"
+  local -r error_message="Taint $taint $name did not present"
+  local -r action="_kubectl get nodes -o custom-columns=taints:.spec.taints[*].effect --no-headers | grep -i $taint"
```
oshoval, line 105: same here, hiccup will give empty string, it doesnt mean there arent taints
no retry would occur, and taints may exists

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -110,18 +226,50 @@ _kubectl label node $SRIOV_NODE sriov=true
...
 _kubectl patch mutatingwebhookconfiguration network-resources-injector-config --patch '{"webhooks":[{"name":"network-resources-injector-config.k8s.io", "clientConfig": { "caBundle": "'"$(cat $CSRCREATORPATH/network-resources-injector.cert)"'" }}]}'
 _kubectl patch mutatingwebhookconfiguration operator-webhook-config --patch '{"webhooks":[{"name":"operator-webhook.sriovnetwork.openshift.io", "clientConfig": { "caBundle": "'"$(cat $CSRCREATORPATH/operator-webhook.cert)"'" }}]}'
 
-# we need to sleep to wait for the configuration above the be picked up
-sleep 60
+# Since sriov-operator doesnt have a condition or Status to indicate if
+# 'operator-webhook' and 'network-resources-injector' webhooks  certificates are
+# configured, in order to check if caBundle reconcile is finished it is necessary
+# to wait for the "NoSchedule" taint to present and then absent.
+wait_for_taint "NoSchedule" || true
```
oshoval, line 242: can there be a race that lines 242 243 would finish before the cluster managed to create taints and remove them ?
hence we would lose the barrier

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: just change the "oc get" to "oc ge"
command error will play as error occur (timeout and so on)

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: you mean?
```
  if [[ : ${namespace:=$3} != "" ]];then
    namespace="-n $namespace"
  fi
```

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I tried it and the function stops immediately with an error from `kubectl, 
which is fine because `wait_for_taint_absence` is dedicated function for waiting for a taint to remove,
the `action` inside it should not be changed.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: if a hiccup is `kube-api` not reachable of bad args passed to `kubectl` (ge insted of get),
`wait_for_taint` will keep running, if it fail we the script will be terminated.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: i dont think its fine, an hiccup should have retry
hiccup means for example etcd isnt reachable
the answer you have when it happens doesnt represnt the real state

i mentioned hiccup becuase hiccup need retry
see please the PR Petr did for this reason
https://github.com/kubevirt/kubevirtci/pull/364
or see #301 
where i fixed that hiccups due to 3 masters, which takes lots of IO, would have retry

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: There will be no race here.
The assumption should be that a taint **might** be created on the node after creating SriovPolicy,
there is no guarantee that there will be a taint, its depends on the environment and the sriov card hardware.
If there is we want it in the logs, if there isnt `wait_for_taint` will finish after the timeout and `wait_for_absence`will return immediately.

oshoval: lets say line 242 + 243 runs on t4, no taint exists on t4
but taint does comes on t5 (t5 > t4),
script is already on other advanced place on t5, the question is would it suffer from it that now there are taints when it checks for other things ?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
EdDev: No, I mean exactly what I wrote.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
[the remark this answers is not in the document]
EdDev: What hiccups? Why is it needed?
You should retry only if you can reason for it, and then I want to see it explained in the commit.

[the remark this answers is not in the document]
oshoval: based on real story (ask Petr)
once in a while a kubectl get command has timeout
and the logic right now, when it happens would thing that no taints exists
(no stdout of kubectl when the command fails | grep taint == nothing)

the code wont be a barrier for waiting for the taints to be removed,
unless it will do the kubectl again few times until it gets the actual state

see please the PR i mentioned that Petr did, also i saw it couple times happening both on prow and locally
again i wont block, but its a true error from what i understand so far

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Thanks for the explanations, 
it is necessary to append the `-n` flag.

I prefer to leave it as it is, because its easier to understand.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: gotcha, I will fix it.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Just to check that we are on the same page,
line 251 (was 242) run until there is a taint or timeout.
then 243 (was 243) run until taint is removed or a timeout
(no simultaneously)
As I said in earlier comment, after patching sriov webhooks, a taint may be created but it is not guaranteed 
thus we wait (about 5 minutes which is enough to verify caBundle picked up)
(next step is crating a policy which may also cause a taint, we look for it in line 277)

oshoval: ye we are sync on that, my question is other, but we can take it offline thanks, you did great enough ;]

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: https://github.com/kubevirt/kubevirtci/pull/378#discussion_r438910900
same issue can happen here, i would suggest to check each one of the actions.
hiccup is not "get versus ge", hiccup is a transient event that need retry to reflect the real answer,

thanks

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,134 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+}
+
+
+function is_taint_absence {
+  local -r taint=$1
+
+  result=$(_kubectl get nodes -o jsonpath="{.items[*].spec.taints[?(@.effect == \"$taint\")].effect}")
+  if [[ -z $result ]]; then 
+    echo "not-presnt"
+  fi
```
oshoval, line 112: dont you need else echo $result ?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,134 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+  echo $error_message && return  1
+}
+
+
+function is_taint_absence {
+  local -r taint=$1
+
+  result=$(_kubectl get nodes -o jsonpath="{.items[*].spec.taints[?(@.effect == \"$taint\")].effect}")
+  if [[ -z $result ]]; then 
+    echo "not-presnt"
```
oshoval, line 111: s/presnt/present/

ormergi: done

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -12,6 +12,135 @@ WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
...
+function wait_for_taint_absence {
+  local -r taint=$1
+
+  local -r tries=60
+  local -r wait_time=5
+
+  local -r wait_message="Waiting for $taint taint absence"
+  local -r error_message="Taint $taint $name did not removed"
+  # '-v'  flag propagates the string that echo'ed by 'is_taint_absence'
+  local -r action="is_taint_absence $taint | grep -i $taint -v"
```
oshoval, line 124: $ function bla() {
> x=$(echo 123)
> }

$ bla
$ bla | grep 123 -v

how come -v propagate the var that is encapsulated in result var ?
what do i miss ?

btw the right use is the cause the return code in the function, but i wont stall on it

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: `is_taint_absence`should return empty string if the taint is presented,
If that case `$result=<taint effect name>`

oshoval: The comment was about the previous situation. since then the logic was adapted.
lets look please on the new one.

if result is empty its of course not "error", this means i would double check the condition to see if thats what you actually meant please. even that it works, the conditions arent ortogonal in their current AND semantic.

i checked by heart that it works on those 3 cases. hope we both arent wrong, 
if the condition above will do change, we need to recheck of course.

1. taint present - we need to retry because we want to wait taints to be removed
2. error due to hiccup - retry
3. no taint present and no error - we are done

Thanks

ormergi: Got it, I see what is the problem now thanks.
I pushed new commit to fix it.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I see what is the problem now thanks.
I pushed new commit with different approach to solve it.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: > dont you need else echo $result ?
I see the problem not, thanks for raising that and for previous comments
I pushed new commit and changed how it works

I checked all those cases on a cluster and it works as we expect :)

oshoval: great
i meant mostly that
`if [[ -z $result ]] && [[ $result != error ]];`
even if it works, some stuff might be redudent there, 
because if result is empty its of course not error (which should be "error", again even if it works)
food for thought

ormergi: In case of a hiccup or kubectl error `kubectl get nodes -o jsonpath..` return an error with error return code
casing the called function `wait_for_taint_absence` to terminate immediately.
With adding  `|| error` we handle that error by setting `result="error"` (with `ehco error`, could be any string)
thus that error wont terminate the caller function and `wait_for_taint_absence` will keep running until timeout.

Thanks for the feedback much appreciated ;)

oshoval: this i understood, its not what i meant, i mean only about the condition itself where the right part (comparing to error string), might be redundant.
i wont stall about it, just food for thought

ormergi: I checked again and you were right, the second literal ` [[ $result != error ]];` is redundant, I will remove it
Thanks.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Consider this example:
`retry "3" "1" "kubectl get po -l app=app" "trying to get pods with label app=app" && echo "success"`

If there are no pods with the label `app=app` the kubectl output would be "No resource found" without raising any error,
then instead of terminating the loop continues and eventually the output would be "success".

oshoval:
/lgtm
Thanks

oshoval:
there is a problem here imo that it doesnt run the cluster-up of sriov ?

ormergi:
> there is a problem here imo that prow doesnt run the cluster-up of sriov ?
> is the cluster-up folder doesn't listen to changes ?

You right, I created a PR to add new check-up job for sriov provider
https://github.com/kubevirt/project-infra/pull/481

phoracek:
@ormergi would you do the bump in kubevirtci?

PR #380, opened 2020-06-04
Machine id and nm cleanup

EdDev:
k8s/1.17-8 provision: Replace /tmp with /provision directory
k8s/1.17-8 provision: Clean the /etc/machine-id
k8s/1.17-8 provision: Cleanup NetworkManager profiles

---

The tmp folder is not assured to keep content after a reboot and should
not be used for content that is shared between the provision stage and
the running stage.

Moving the relevant content from /tmp to /provision.

---
The machine-id contains the unique machine ID of the local system.
Multiple applications depend on it as a seed.

The VM node provision should not keep it on the system so the node
instances at cluster-up will not have an identical machine-id.

The issue has been identified when IPv6 link-local addresses have been
created with identical mac addresses and later when dynamic IPv6 has
been attempted.

Note: The seed of the machine-id comes from the KVM UUID which is either
auto-generated by qemu or inputted by the user.

---

The base VM (CentOS 8) comes with non-default profiles (e.g.
ipv6.method=ignore).
The default profiles should fit well with the network needs, therefore
remove the persisted configuration and allow NM to auto-create them when
the nodes will boot next.


cluster-provision/k8s/1.17/provision.sh
```diff
@@ -329,5 +329,12 @@ echo "tmpfs /tmp/kubevirt.io/tests tmpfs rw,context=system_u:object_r:container_
 
 dnf install -y NetworkManager-config-server
 
+# Cleanup the existing NetworkManager profiles so the VM instances will come
+# up with the default profiles. (Base VM image includes non default settings)
+rm -f /etc/sysconfig/network-scripts/ifcfg-*
```
oshoval, line 334: We need it i guess also on 1.18 ?

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -329,5 +329,13 @@ echo "tmpfs /tmp/kubevirt.io/tests tmpfs rw,context=system_u:object_r:container_
...
+nmcli dev con eth0
```
oshoval, line 335: Better to not use acronyms in code imo

EdDev:
/hold
Something is wrong.

qinqon:
@EdDev `cli/cli` copy stuff to /tmp/ you will have to copy first things from /tmp/ to /provision/ or whatever you need

from cli/cli
```
docker exec ${VM_CID} /bin/bash -c "scp -r -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i vagrant.key -P 22 /scripts/manifests/* vagrant@192.168.66.101:/tmp"
```

qinqon:
@EdDev all good let's port to k8s-1.18, there also some comments from @oshoval

EdDev:
> @EdDev all good let's port to k8s-1.18, there are also some comments from @oshoval

Done.

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -323,11 +330,19 @@ cp -rf /tmp/cnao/ /opt/
 for i in $(grep -A 2 "IMAGE" /opt/cnao/operator.yaml |grep value | awk '{print $2}'); do docker pull $i; done
 
 # Create a properly labelled tmp directory for testing
-mkdir -p /tmp/kubevirt.io/tests
-chcon -t container_file_t /tmp/kubevirt.io/tests
-echo "tmpfs /tmp/kubevirt.io/tests tmpfs rw,context=system_u:object_r:container_file_t:s0 0 1" >> /etc/fstab
+mkdir -p /provision/kubevirt.io/tests
```
oshoval, line 333: line 330 also have /tmp (cant put there a comment)
cp -rf /tmp/cnao/ /opt/

/lgtm cancel

EdDev: Sorry, I do not understand your comment.
- What "real place"?
- The cnao copy is moving files from /tmp to /opt, so there is no need to share `/tmp/cnao` between the provision stage and the run stage, it is done there with that copy already.

oshoval: Right, disregard
/lgtm

oshoval:
imo, the 6 commits can be made 3, better to not separate 1.17/1.18

EdDev:
> imo, the 6 commits can be made 3, better to not separate 1.17/1.18
> just generally speaking, not blocking.

As these are different files on different providers, I think they should be separated.
Consider that one can argue this also is needed for 1.16 as well.

qinqon:
@EdDev can you update shas ?

oshoval:
/lgtm
Thanks

oshoval:
Its problematic that it didnt let the tests finish, or it did ?
when i gave lgtm they were still running

vladikr:
@EdDev @qinqon @oshoval This PR had to replace /tmp with /provision directory on all current providers, not just 1.17 and 1.18

EdDev:
@vladikr Originally I did not expect anyone to depend on `/tmp` to be persisted over boot.
Now that it seems tests are dependent on provisioning leftovers I am more than surprised.

I guess the need to change it to all other lanes is due to the tests dependency on it.

vladikr:
@EdDev I've pushed https://github.com/kubevirt/kubevirtci/pull/388/

PR #382, opened 2020-06-05
update centos to the latest kernel

vladikr:
KubeVirt has recently updated virt-launchers to use libvirt 6.0 and qemu 4.2 [1]
However, this exposed a KVM bug that affects nested virtualization. This mainly manifests itself when kubevirtci runs on baremetal hosts.
Virtual machines will then sometimes fail with `qemu-kvm: error: failed to set MSR 0x48d to 0xff00000016`
However, this issue has been fixed in kernel `kernel-4.18.0-179`, but the currently available centos image has an older kernel and needs to be updated.

[1] https://github.com/kubevirt/kubevirt/pull/3268


cluster-provision/k8s/1.17/provision.sh
```diff
@@ -2,6 +2,9 @@
 
 set -ex
 
+# Need to have the latest kernel
```
oshoval, line 5: We would need it i think also on 1.18 ?
Do we want to make it part of the base image provisioning so we wont need to do it on
every provision ?
unless its fast.

Thanks

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -2,6 +2,9 @@
...
+dnf update -y --nobest kernel
```
EdDev, line 6: Why `--nobest` is used?
Could you explain this and the reasoning for all this change in the commit message?

EdDev: There seem to be a missing step in `./cluster-provision/centos8`, where one can add generic, OS related changes (and not lane related). Such changes seem to fit there more than here. I think this is what @oshoval mentioned above.

oshoval: Yep, thanks

vladikr: Where do we make any modifications to the centos image itself in cluster-provision/centos8?

EdDev: > Where do we make any modifications to the centos image itself in cluster-provision/centos8?

We do not at the moment, that is the “missing step”.
I guess it is an action for a separate PR, but this one reveals the need.

oshoval: cluster-provision/centos8/Dockerfile
what about adding it in this file or in the scripts it runs ?
(either on other PR)

vladikr: Let's fix the current problem right now and then think about how can we expand this to a proper solution.

oshoval: sgtm, just rebase is needed, there are couple PRs on fire today that changing hashes

oshoval:
lgtm, but Edy left some comments so i didnt use slash
Thanks

vladikr:
> It is also worth mentioning in the commit message that this may unintended update the kernel every-time one updates these providers. This may be surprising.

Which is fine as we are using a centos stream anyway.

vladikr:
> Could you please add the reasoning to the commit message and also explain why you used `-nobest` there as well?

I don't mind this `-nobest` setting. it is used all over these scripts and I was just trying to be consistent. if you really insist I'll remove this. 
This issue is currently destabilizing local development so I don't want this PR to stale.

EdDev:
> > It is also worth mentioning in the commit message that this may unintended update the kernel every-time one updates these providers. This may be surprising.
> 
> Which is fine as we are using a centos stream anyway.

How is that related?
If one changes a provision script, I do not expect the kernel to get updated behind my back.
It is a side-effect (an implicit one).

> I don't mind this -nobest setting. it is used all over these scripts and I was just trying to be consistent. if you really insist I'll remove this.
> This issue is currently destabilizing local development so I don't want this PR to stale.

I do not mind if you leave it or not either, I just want an explanation on why it is needed in this case.

PR #385, opened 2020-06-08
Make provision less sensitive to quay hiccups

oshoval:
Add retry on docker pull, because quay might have connection hiccups,
that a simple retry solves instead of failing the whole provision.

Seen for example on
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/345/push-k8s-1.17/1269922704140537856
and few times locally in the last week as part of #345 development.

Inspired by
https://gist.github.com/dublx/e99ea94858c07d2ca6de


rmohr:
I did not see the need for that so far. Was there an incident here?

oshoval:
> I did not see the need for that so far. Was there an incident here?

yes today on prow, at on the last week couple time locally
prow link is this

oshoval:
> I did not see the need for that so far. Was there an incident here?

yes today on prow, at on the last week couple time locally

https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/345/push-k8s-1.17/1269922704140537856

net/http: TLS handshake timeout  - for example on pull image
```
error pulling image configuration: Get https://cdn.quay.io/sha256/ea/eacd9559dc3d5dda44e61daedcc070308ccb082381dbdb1a3b95c19a0c6d1be6?Expires=1591610309&Signature=PrVFDXahy9CXOzoR7oQzTUridLI2ngB93GCd-LDJWAD6JAqbGsR3BC6~kufPwRPVCgC1u~WPKsK1G63jDrZe~dojyvaXGRtGtfMUIW6wbYmC3b7v5dWfPuqN-AnBbRgz1lQibJ0s4F6BITNUrZ8wxbIaJzHX6VP0y6fhG2LDr194BpSbFtIKTrHVuux0-h2cBeLOZi7nyMmhDvovpr8a1~yMkodt93N1BYxMBNPY9e2jf5bPoBnZBa8odXZppBGFPPU6fg86xCOjdpWBgWcAHRKqzGADF4Q6ecgUgQA5X1i1VeVhW8VWjTBMYyTZy1P69zCqaTTp7eliwL7oJFvQMQ__&Key-Pair-Id=APKAJ67PQLWGCSP66DGA: net/http: TLS handshake timeout
+ finish
```

oshoval:
> I think we need to trust the services we work with, contribute to make them better (by reporting service outages) or just replace them with better alternatives.
> 
> The commit itself is also not convincing me, it lacks numbers (how many times it happens, how much time was lost, etc).
> 
> Also, retrying on every single pull attempt is bit extreme. Extracting to a separate script and retrying it sounds better to me. But as mentioned earlier, I question if this is even a correct path.

locally when i was debugging in order to develop #345 
it happened 1 out of 3 time,
its not reasonable imo to just run the whole process again and again,
and on prow it happened as well,
until quay will fix it, (and even they fix it, network issues happens for every cloud provider),
we should be robust imo,
i can report to quay but in a parallel mode

EdDev:
Unfortunately, I do not favor defensive programming or scripting. If this would have been costing us hours of sufferings on a weekly basis I would have consider it reasonable as a temporary fix.
IMO it needs to happen daily multiple times to reach a point of not trusting the quay service.

oshoval:
> Unfortunately, I do not favor defensive programming or scripting. If this would have been costing us hours of sufferings on a weekly basis I would have consider it reasonable as a temporary fix.
> IMO it needs to happen daily multiple times to reach a point of not trusting the quay service.

Locally, without it, it wasted me hours, people that will run provision more locally will suffer from it,
i really don't think its that defensive, the other option is to rerun failing jobs with is tedious and causing wear and tear, and btw, when there will be automation it would kill the whole chain,
can we find a way so we both like it please?
I think its needed, we cant waste time because hiccups, and we shouldn't imo.

oshoval:
> I can only speak from my experience with this "retry" pattern.
> Although it seems useful for the immediate usage, it is usually a bad habit that causes one to think that every single operation needs to be retry because there may be network issues, service issues and slow responsive hosts.
> There are many users out there that need to use docker/podman pull, it is unreasonable to ask them all to retry like this... this is what this change implies.
> 
> If there is such a problem with quay, we have production issues and not only infra ones.
> 
> I do not think we need to take into account network issues, quay service outages or hiccups and so on... if it benefits your local work, add this to you local version.
> I prefer to trust my network and services and I personally did not experience these so far.
> If all experience this often as you said (30%??), then we have a bigger problem than this little project of ours.

There are lots of emails lately on quay outage, and as part of #345 i had to provision frequently,
i added retry to just this function, i can add a comment above it why its needed,
personally i don't think that we should be afraid of malformed use, and pay with time, especially the ones that need to re provision.
even using local hacks, when one can solve for the others, isn't the right way imo.

Its a fact that it might happen for people, and they will need to check the log either on locally and on prow, and decide to retry, code should be bullet proofed and network hiccups and IO is something that occur, even when everything is good,
again i didnt bullet proof everything, just what is indeed needed. and without hiccups it wont do anything crucial that it didnt do before.

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -2,6 +2,24 @@
 
 set -ex
 
+function docker_pull_retry {
+    retry=0
+    maxRetries=5
+    retryInterval=3
```
dhiller, line 8: nit: Maybe rename to `retryAfterSeconds` to be clear what this is about.

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -2,6 +2,24 @@
 
 set -ex
 
+function docker_pull_retry {
```
dhiller, line 5: nit: I see mixed indentation in this file, do we have sth like [`shfmt`](https://github.com/mvdan/sh) in place? I guess not? This way we could avoid the discussion about what bracket style etc. we should be using and just use the formatter.

In the meantime we could just make it obligatory by applying it in a presubmit check.

cluster-provision/k8s/1.18/provision.sh
```diff
@@ -2,6 +2,24 @@
 
 set -ex
 
+function docker_pull_retry {
```
dhiller, line 5: Same goes here. See 1.17/provision comments.

cluster-provision/k8s/1.17/provision.sh
oshoval: in Kubevirt, we have it indeed as you meant i think,
hack/bazel-fmt.sh

i will to align the style manually ? 
or once we integrate shfmt it would do it automatically?
thanks

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -206,7 +222,7 @@ done
 # Wait until k8s pods are running
 while [ -n "$(kubectl --kubeconfig=/etc/kubernetes/admin.conf get pods -n kube-system --no-headers | grep -v Running)" ]; do
     echo "Waiting for k8s pods to enter the Running state ..."
-    kubectl --kubeconfig=/etc/kubernetes/admin.conf get pods -n kube-system --no-headers | >&2 grep -v Running || true
+    kubectl --kubeconfig=/etc/kubernetes/admin.conf get pods -n kube-system --no-headers | grep >&2 -v Running || true
```
oshoval, line 225: this is weird , shfmt changed it

cluster-provision/k8s/1.17/provision.sh
oshoval: done

oshoval:
@dhiller can you please take another look ?
(didn't create hashes, will create once there is lgtm)
Thanks

oshoval:
provision lanes are failing, 
https://github.com/kubevirt/kubevirtci/issues/394
suggested fix
https://github.com/kubevirt/kubevirtci/pull/395

oshoval:
@dhiller can you please take another look ?

oshoval:
> The retry logic looks good so far.
> 
> What I am wondering about is whether we should really retry every pull here as this will only solve the small hiccups that a service sometimes has, but will not solve the problem if there is a real outage. For a real outage it would make more sense to actually fail fast, as there's nothing we can do about the outage itself (we could mitigate the impact by having cached the images in a docker cache inside our build system, but that would be a completely different topic).
> 
> Furthermore we would then have to decide whether it is a real outage (what would define this? Normally we'd use circuit breaker, but within a script I don't think we have the ability to do so) or not and then either fail fast or retry.
> 

> @oshoval did you actually observe that i.e. quay is having problems with small hiccups and therefore the image build often fails on this?

Totally yes,  both on prow and locally, several times.
Example:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/345/push-k8s-1.17/1269922704140537856

This is why i purpose this PR which solved it, especially for people during development which need to provision here and there,
but also to remove wear and tear when it happen when people use provision in prow (or just on each kubevirtci PR that does provision as part of the basic lanes).

This simple retry doesn't even take time in case all is good, and wont take much time on a real outage,
but can hint about it sometimes.

oshoval:
Thanks Daniel
/hold
Will rebase and create hashes, didn't put hold from first place so it wont reduce motivation or something, was nice if we had an intermediate flag for it, or maybe hold is good enough.

oshoval:
/hold cancel
Updated hashes, PR finally ready, thanks

PR #388, opened 2020-06-16
Replace tests dir /tmp with /provision directory on all providers

vladikr:
e6a266e9e (kubevirt/kubevirtci#380 ) replaced the location of the tests directory only on the 1.17 and
1.18 providers while this directory should be able for all providers.

Some tests in KubeVirt relies on /tmp//kubevirt.io/tests to be present. This PR will align all providers to use /provision/kubevirt.io/tests


vladikr:
> You have moved only the parts that are needed by the tests?
> (because compared to the 1.17/8 there are others which were needed when the /tmp got cleaned between reboots)

Yes, exactly. I only moved the directory needed for the tests.

EdDev:
> Some tests in KubeVirt relies on /tmp//kubevirt.io/tests to be present. This PR will align all providers to use /provision/kubevirt.io/tests

@vladikr any chance you can add the text to the commit as well?

PR #389, opened 2020-06-16
kind, automation: detach all loopbacks on provider startup

maiqueb:
When running docker in docker (which is how kind does it's thing)
, the `losetup` command ends up looking for the next available
device in the host's namespace, instead of using the k8s node
namespace.

Eventually, it will grow past the number of statically allocated
loopback devices by the `loopdev` container. When that happens,
the `disks-images-provider` pod enters the crash loop.

By detaching all loopback devices from the host at provider
startup we assure the next available device will be on range.

Fixes: https://github.com/kubevirt/kubevirt/issues/3569


phoracek:
/approve cancel

There are some questions around this approach

maiqueb:
/hold

I may have found a couple of alternatives.

maiqueb:
Abandoning.

PR #392, opened 2020-06-24
[2] k8s1.18: Add provision script to update kernel args

krsacme:
Copy the kargs script from [1.17](https://github.com/kubevirt/kubevirtci/pull/370) to 1.18 provision
scripts.

Related: #365

### DIFF
--- a/cluster-provision/k8s/1.18/kargs.sh
+++ b/cluster-provision/k8s/1.18/kargs.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -ex
+
+# OvS-DPDK requires latest kernel
+yum update -y --nobest kernel
+
+# Increase the Hugepages count so that DPDK can be enabled in OvS which requires 1GB of hugepages
+# Add iommu support to enable PMD drivers
+GRUB_FILE=/etc/default/grub
+MODIFIED=0
+if ! grep -q iommu $GRUB_FILE; then
+    sed -i 's/${GRUB_CMDLINE_LINUX}/${GRUB_CMDLINE_LINUX} iommu=pt intel_iommu=on/g' $GRUB_FILE
+    MODIFIED=1
+fi
+
+if grep -q 'hugepagesz=2M' $GRUB_FILE && ! grep -q 'hugepages=2048' $GRUB_FILE; then
+    sed -i 's/ hugepages=[0-9]*/ hugepages=2048/g' $GRUB_FILE
+    MODIFIED=1
+fi
+
+if [[ $MODIFIED == "1" ]]; then
+    grub2-mkconfig -o /boot/grub2/grub.cfg
+fi


dhiller:
@EdDev a while back ago we had that exact situation where we had a bunch of scripts that were required to support several versions, thus needed to have a lot of case handling for each and every k8s version we supported. This got very tedious and cumbersome, so we resorted to just copying the scripts into each version folder and making the adjustments required, so we got rid of the decision logic.

So, even though it's a redundant copy in this case, let's just stick to what we agreed here.

/lgtm

EdDev:
> @EdDev a while back ago we had that exact situation where we had a bunch of scripts that were required to support several versions, thus needed to have a lot of case handling for each and every k8s version we supported. This got very tedious and cumbersome, so we resorted to just copying the scripts into each version folder and making the adjustments required, so we got rid of the decision logic.
> 
> So, even though it's a redundant copy in this case, let's just stick to what we agreed here.

This approach has introduced multiple bugs since, where changes have not been distributed to other providers (of that same "family").
Using a symlink is a one-liner and if there will a need to have differences, then one can consider to break it.

cluster-provision/k8s/1.18/kargs.sh
```diff
@@ -0,0 +1,24 @@
...
+# Increase the Hugepages count so that DPDK can be enabled in OvS which requires 1GB of hugepages
+# Add iommu support to enable PMD drivers
+GRUB_FILE=/etc/default/grub
+MODIFIED=0
+if ! grep -q iommu $GRUB_FILE; then
+    sed -i 's/${GRUB_CMDLINE_LINUX}/${GRUB_CMDLINE_LINUX} iommu=pt intel_iommu=on/g' $GRUB_FILE
+    MODIFIED=1
+fi
+
+if grep -q 'hugepagesz=2M' $GRUB_FILE && ! grep -q 'hugepages=2048' $GRUB_FILE; then
```
jean-edouard, line 17: Typo in `hugepagesz`? Looks like 1.17 has it too...

krsacme: Can you be more specific? ``hugepagesz`` specifies the size whether 2M or 1G. ``hugepages`` specifies the number of hugepages of specifified size. The check here ensures that when 2M hugepage size is enabled, it should have atleast 2048 number of 2M hugepages, which allows 4GB of total hugepages for the deployment. 

1GB will be allocated to ovs-vswitch when DPDK is enabled. And the rest of the hugepages will be used in the VMs which uses vhostuser interfaces.

jean-edouard: My bad, thanks for the explanation.
/lgtm

krsacme:
> > @EdDev a while back ago we had that exact situation where we had a bunch of scripts that were required to support several versions, thus needed to have a lot of case handling for each and every k8s version we supported. This got very tedious and cumbersome, so we resorted to just copying the scripts into each version folder and making the adjustments required, so we got rid of the decision logic.
> > So, even though it's a redundant copy in this case, let's just stick to what we agreed here.
> 
> This approach has introduced multiple bugs since, where changes have not been distributed to other providers (of that same "family").
> Using a symlink is a one-liner and if there will a need to have differences, then one can consider to break it.

@dhiller I can change it so symlink, until there is an actual difference to make it a copy, let me know if it works out as per the general guidelines.

dhiller:
@krsacme sorry for taking so long to answer :-/ I think it's fine like this.

/approve

PR #395, opened 2020-06-29
Fix K8s 1.17, 1.18 provisioning

oshoval:
Update k8s 1.17/8 install to overcome conflict errors
    
Due to a bug in k8s there is a conflict issue
during provisioning.
    
https://github.com/kubernetes/kubernetes/issues/92463

Fixes: #394

PR #396, opened 2020-06-29
Fix check-cluster-up.sh missing Var init

oshoval:
Since the script runs with -u,
variables must be set also when running locally.

When running locally, the error was:
`./check-cluster-up.sh: line 31: CI: unbound variable`


oshoval:
lanes failing because a bug this PR fixes
https://github.com/kubevirt/kubevirtci/pull/395

PR #400, opened 2020-07-02
Bump CNAO to 0.39.2

qinqon:
It includes a lot of fixes from 0.35


cluster-provision/k8s/1.18/manifests/cnao/network-addons-config.crd.yaml
```diff
@@ -26,6 +26,7 @@ spec:
           type: object
         status:
           type: object
+      type: object
```
maiqueb, line 29: 0_o ... why is this suddenly needed ?

qinqon: No clue, I downloaded from CNAO release pagehttps://github.com/kubevirt/cluster-network-addons-operator/releases/tag/0.39.2

phoracek: It became a requirement for CRDs in recent versions. Previously, the top level object was treated as `object` implicitly, but it was added that it has to be explicit.

phoracek:
/hold

We need to update image hashes

phoracek:
@qinqon would you update hashes?

qinqon:
/hold cancel
shas updated.

PR #401, opened 2020-07-02
Replace bazel with Makefile and bash

qinqon:
The gocli project is not complex enough to require bazel


qinqon:
/check-provision-k8s-1.18

cluster-provision/gocli/hack/install-go.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -xe
+
+destination=$1
```
rmohr, line 3: Why do we need this script?

qinqon: Is a simplified version of go gimme from travis, it install go under `build` directory so we vendor the go compiler.

rmohr: Why do we not use an image where go is installed?

qinqon: The idea is to post gocli binaries at release page at github no gocli container will exists, so we do a plain compilation with a vendored go.

qinqon:
/hold

Prow does not like it.

PR #402, opened 2020-07-03
Add update provider hashes make target

oshoval:
In order to ease provider hash updating and eliminated possible
mixing between provider hashes when updating manually.
    
Can be used also as part of an automation that would create diff
file, or actually update the images list.

### DIFF
--- a/hack/bump.sh
+++ b/hack/bump.sh
@@ -1,0 +1,32 @@
+#!/bin/bash -e
+#
+# This file is part of the KubeVirt project
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2020 Red Hat, Inc.
+#
+
+# Updates cluster-up/cluster/images.sh provider hash to point on new given hash.
+# After usage, commit the changes of cluster-up/cluster/images.sh.
+
+# Usage: ./hack/bump.sh <provider> <hash>
+# Example: ./hack/bump.sh k8s-1.18 c41e3d9adb756b60e1fbce2ffd774c66c99fdde7ee337460c472ee92868e579e
+
+PROVIDER=${1:?}
+HASH=${2:?}
+
+function main() {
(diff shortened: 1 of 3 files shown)


oshoval:
Will create make target if needed

hack/update_hash.sh
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2020 Red Hat, Inc.
+#
+
+# Usage example: ./hack/update_hash.sh k8s-1.18 HASH
+
+LANE=${1:?}
```
RamLavi, line 21: it doesn't look like you're using these env params, and instead using the $1,$2 in the main function... did you mean to use them in the main function?

oshoval: ye you right, i want to use them, it works because those two lines enforce them to exists, but i would change it
thanks

oshoval:
> Will create make target if needed

@RamLavi wdyt ?

RamLavi:
> > Will create make target if needed
> 
> @RamLavi wdyt ?

I think adding a `make ..` is preferable.

Also, is this command used by the developer manually?

oshoval:
> > > Will create make target if needed
> > 
> > 
> > @RamLavi wdyt ?
> 
> I think adding a `make ..` is preferable.
> 
> Also, is this command used by the developer manually?

As of today we are doing it manually, it will ease it, as i wrote it can be used, in the future, automatically, by capturing push output hash, bumping the hash and create a PR, if desired.

RamLavi:
> As of today we are doing it manually, it will ease it.
> As i wrote it can be used, in the future, automatically, by capturing push output hash, bumping the hash and create a PR, if desired, but just when the push job are post submit jobs, not part of this PR.

thanks, I plan to use something similar in CNAO automation.

hack/update_hash.sh
oshoval: done

hack/bump.sh
```diff
@@ -0,0 +1,32 @@
...
+# After usage, commit the changes of cluster-up/cluster/images.sh.
+
+# Usage: ./hack/bump.sh <provider> <hash>
+# Example: ./hack/bump.sh k8s-1.18 c41e3d9adb756b60e1fbce2ffd774c66c99fdde7ee337460c472ee92868e579e
+
+PROVIDER=${1:?}
+HASH=${2:?}
+
+function main() {
+    sed -i "s/IMAGES\["$PROVIDER"\].*/IMAGES\["$PROVIDER"\]=\""$PROVIDER"@sha256:"$HASH"\"/g" cluster-up/cluster/images.sh
```
RamLavi, line 29: noob bash question - shouldn't there be && between `sed` and `cluster-up/cluster/images.sh`?

oshoval: nope, the filename is the file that sed acts on, its a parameter of sed

RamLavi: ack

PR #403, opened 2020-07-08
Create a kubevirtci release pipeline

qinqon:
The current release method for kubevirtci involves a lot of manual steps that also have to be repeated if a conflict at container digest appear, this PR automatize this with the following approach:
- A prow postsubmit jobs `release-kubevirtci`  start after kubevirtci PR is merged (https://github.com/kubevirt/project-infra/pull/546).
- Detect what changes have being done and run `release-centosX` and `release-k8s-x.y` according to those changes, those changes will be calculated from latest release done at github instead of using prow BASE env variables, making easier to re-play a release job if something is broken.
- Build `cluster-provision/gocli` passing the digest from `:latest` as `ldflags` so binary contains the hashes
- Create a release tarball `kubevirtci.tar.gz` that contains at root dir `cluster-up/*` and `cluster-provision/gocli/build/cli`
- Tag the current kubevirtci master commit with a epoch tag
- Do a github release of that epoch tag and add a the release tarball as an asset.

The projects that will consume kubevirtci will need to download the kubevirtci.tar.gz and call the `up.sh`, `down.sh` and friends from there the environment variables like KUBEVIRT_PROVIDER and the like are exactly the same, an alternative would be to just do `./cli run k8s-x.y [args]` directly at projects since the digest are now inside the binary from the tarball.

Also noted that the `cluster-provision/k8s/x.y/base` are now just pointing to `:latest` since `:latest` will be only pushed by automation this is safe, same applies for gocli default ldflags instead of digest is pointing to `:latest`

Developing kubvertci is simpler with this PR since calling `cluster-up` will compile `gocli` so changes done there are also tested too and providers are pointing to `:latest`.

Since cluster-up may endup compiling gocli kubevirtci depends on golang 1.13 for that the kubekins prow image has to be updated that's done at the project-infra PR.

Right now I am testing it towards my github org `qinqon` results from `release-kubevirtci` job can be seen here https://github.com/qinqon/kubevirtci/releases and this is the prow job that generated it https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/logs/release-kubevirtci/1291685881299603456,

TODO:
* Fix git http server unit test for tagRepository.
* Add unit tests for https git repository and github release.
* Run base jobs. 
* Detect changes from latest release to orchestrate jobs.
* If one of the provision job fails delete the others to consume less resources.
* Add a root `kubevirtci` at the tarball.
* Remove harcoded providers versions from gocli, one approach is to pass just one ldflag with digests separated by spaces like `"k8s-1.18=adfas k8s-1.17=asdfasf"`


cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -2,13 +2,16 @@
...
-source ${KUBEVIRTCI_PATH}/cluster/images.sh
-
 if [ "${KUBEVIRTCI_RUNTIME}" = "podman" ]; then
     _cli="pack8s"
 else
-    _cli_container="${KUBEVIRTCI_GOCLI_CONTAINER:-kubevirtci/${IMAGES[gocli]}}"
-    _cli="docker run --privileged --net=host --rm ${USE_TTY} -v /var/run/docker.sock:/var/run/docker.sock ${_cli_container}"
+    _cli=./cli
+    # If this is not kubevirtci release tarball build gocli and use, it will
+    # Use :latest providers
```
rmohr, line 10: Would you assume to have a pipeline, where a pre-step creates all clusters?

qinqon: That's it only the pipeline would have permission for `:latest` so latest are safe.

automation/release.sh
```diff
@@ -0,0 +1,112 @@
...
+        pushd cluster-provision/k8s/$k8s_provider
+            ../provision.sh
+            ../publish.sh
+        popd
+    done
+}
+
+kubectl get prowjob -o yaml
+
+exit 1
```
oshoval, line 50: ?

cluster-up/cluster/images.sh
```diff
@@ -1,18 +0,0 @@
-#!/usr/bin/env bash
-
-set -e
-
-declare -A IMAGES
-IMAGES[gocli]="gocli@sha256:0cec3163e5f8b050f129fa795b76c20cb1eb27c91564c63663160b095064a1f9"
-if [ -z $KUBEVIRTCI_PROVISION_CHECK ]; then
-    IMAGES[k8s-fedora-1.17.0]="k8s-fedora-1.17.0@sha256:aebf67b8b1b499c721f4d98a7ab9542c680553a14cbc144d1fa701fe611f3c0d"
```
oshoval, line 8: we need please to still support local provision check
maybe you planned it according makefile args?

other question
if someone would add some ocp provider or non k8s, the mechanism in this PR would be still be as friendly at least as before ?

rmohr:
@qinqon would you mind describing in a little detail on how this is supposed to work in the PR description?

qinqon:
/hold

provisioning is broken

jean-edouard:
@kubevirt-bot I'm begging you, stop spamming me with this PR
/uncc jean-edouard
/unassign jean-edouard
/unsubscribe jean-edouard

oshoval:
btw the PR description doesnt list changes of cluster-up , which might affect the post provision check-up, in case you want to handle it as well, thanks

PR #404, opened 2020-07-08
Set default KUBEVIRT_PROVIDER to k8s-1.18.

maya-r:
Older lanes (such as the default k8s-1.15) use an older host OS
and don't run properly on Ryzen CPUs, as described in
https://github.com/kubevirt/kubevirt/issues/3615


rmohr:
/approve

The last test in kubevirt/kubevirt which had issues on 1.18 is no issue anymore.

PR #407, opened 2020-07-14
Pre pull and check images for vanilla k8s test clusters

dhiller:
This PR introduces checks and tools for the user (here: the creator or editor of any vanilla k8s test cluster) to make sure that the images in the provisioned virtualized 1.21 cluster all get pre-pulled during provisioning. The idea here is to be resilient against any image registry outage. Further benefit should be that the cluster-up gets fast.

To achieve that we do the following:

- during provisioning
  - we pre-pull all images from the list additionally to the list of images fetched from the info in provisioning directory (manifests, scripts)
- during cluster-up check
  - we deploy all manifests that are available and wait for all pods to get ready
  - we create a list of images from the pods that the cluster should have
  - we check if this list differs and fail if it does
  - finally we fetch the pull policies from all running pods and check whether it will lead to _always_ pulling that image (**Note: we don't fail yet to give teams time to fix the manifests**)

As a result cluster-up should be fast, as it should have all images pre-pulled that might be required (at the cost of the cluster image being big, but pulling the cluster image should occur less often hopefully)


cluster-provision/k8s/1.18/provision.sh
```diff
@@ -2,19 +2,22 @@
 
 set -ex
 
+[ -f "/tmp/pre-pull-images" ] || exit 1
```
oshoval, line 5: i would suggest to support please an env var that skip the pre pulling
its good when one debug provisioning locally, and no need to pull images
it saves couple minutes of provisioning, when you need to do faster iterations

EdDev:
Commenting on the CDI dependency (and perhaps others that I could not identify):

There are groups of tests (e.g. SRIOV) which do not need the CDI (may need some work to make this statement an actual) and I think we should not enforce having it so we will not need to operate something that is not needed.

Also, kubevirtci is used by other projects, adding resources that they do not need may not be preferable.
Keeping it simple is an advantage, but we could think of preparing additional services/resources per need as optional.

dhiller:
> Commenting on the CDI dependency (and perhaps others that I could not identify):
> 
> There are groups of tests (e.g. SRIOV) which do not need the CDI (may need some work to make this statement an actual) and I think we should not enforce having it so we will not need to operate something that is not needed.
> 
> Also, kubevirtci is used by other projects, adding resources that they do not need may not be preferable.
> Keeping it simple is an advantage, but we could think of preparing additional services/resources per need as optional.

The initial idea was to be resilient against any image registry outage. To achieve that we pre-pull all images that we know of (i.e. those of which we have the manifests for) into the cluster image.

As a result cluster-up should be fast (I tested with CNAO enabled, cluster-up, incl. deploying all manifests, and waiting for all pods ready took 1m45s on my machine), and has all images that might be required pre-pulled in order to be fast to use (at the cost of the cluster image being bigger, but that should occur less often hopefully)

Wouldn't this lead to having to provision several clusters (the ones having a dependency and the ones who don't)? I'm not sure if that is what we want, as this would make the situation more complicated by forcing the user to choose what cluster he wants (i.e. the exact one that has all his dependencies he requires?

> Commenting on the CDI dependency (and perhaps others that I could not identify):
> 
> There are groups of tests (e.g. SRIOV) which do not need the CDI (may need some work to make this statement an actual) and I think we should not enforce having it so we will not need to operate something that is not needed.
> 
> Also, kubevirtci is used by other projects, adding resources that they do not need may not be preferable.
> Keeping it simple is an advantage, but we could think of preparing additional services/resources per need as optional.

The deployment of the manifest is optional. Nothing is deployed out of the box, just the images are pre-pulled.

cluster-provision/k8s/1.18/provision.sh
dhiller: For local debugging you just can remove the file containing the images (1.18/pre-pull-images i.e.)

dhiller:
@awels FYI: the idea is to pre-pull images of cdi here. Not sure what version we should provide per cluster here? Can you help out?

awels:
I am trying to wrap my head around this. So the point is that we want to have some images pre pulled on the host, so that if someone wants to use a particular feature it will be faster. For my personal work flow, I don't I would have any benefit from this because I prefer to have a naked cluster, into which I then install whatever I am working on. Its pretty rare that I need anything besides CDI (Maybe KubeVirt if I am integrating some features).

As for which version of CDI to include. We release a new version every 3 weeks so whatever you pre pull will get stale fairly quickly.

oshoval:
> > Commenting on the CDI dependency (and perhaps others that I could not identify):
> > There are groups of tests (e.g. SRIOV) which do not need the CDI (may need some work to make this statement an actual) and I think we should not enforce having it so we will not need to operate something that is not needed.
> > Also, kubevirtci is used by other projects, adding resources that they do not need may not be preferable.
> > Keeping it simple is an advantage, but we could think of preparing additional services/resources per need as optional.
> 
> The deployment of the manifest is optional. Nothing is deployed out of the box, just the images are pre-pulled.
> 
> The initial idea was to be resilient against any image registry outage. To achieve that we pre-pull all images that we know of (i.e. those of which we have the manifests for) into the cluster image.
> 
> As a result cluster-up should be fast (I tested with CNAO enabled, cluster-up, incl. deploying all manifests, and waiting for all pods ready took 1m45s on my machine), and has all images that might be required pre-pulled in order to be fast to use (at the cost of the cluster image being bigger, but that should occur less often hopefully)
> 
> Wouldn't this lead to having to provision several clusters (the ones having a dependency and the ones who don't)? I'm not sure if that is what we want, as this would make the situation more complicated by forcing the user to choose what cluster he wants (i.e. the exact one that has all his dependencies he requires?

Just please dont mix, in the discussion, kind SRIOV with K8s providers, kind doesnt have provision stage, so even if kind doesnt need CDI or so, it wont affect this PR scope.

dhiller:
> I am trying to wrap my head around this. So the point is that we want to have some images pre pulled on the host, so that if someone wants to use a particular feature it will be faster. For my personal work flow, I don't I would have any benefit from this because I prefer to have a naked cluster, into which I then install whatever I am working on. Its pretty rare that I need anything besides CDI (Maybe KubeVirt if I am integrating some features).
> 
> As for which version of CDI to include. We release a new version every 3 weeks so whatever you pre pull will get stale fairly quickly.

[CDI deployment during testing on kubevirt/kubevirt is still on 1.18.0 IIUC](https://github.com/kubevirt/kubevirt/blob/master/manifests/testing/cdi-v1.18.0.yaml.in#L409). @rmohr suggested to include it here and remove it from kubevirt/kubevirt. I'd be fine with not including it here, but I tend to feel it makes sense to include it here for kubevirt/kubevirt somehow.

EdDev:
If the manifests and binaries are just placed in the image and not deployed, then my main worry is gone.

The note made that this manifest may become obsolete very fast should be taken into account though.
Would you like to provision a new image per dependencies version updates? I guess this may be just fine, as you do not have to use the latest, or at least you want to control it.
If this is automated, like CNAO is now in progress to do, it may reduce this limitation.

dhiller:
Yeah, like I said, kubevirt/kubevirt is not using latest cdi either, so we might just use that for now. The more I think of it the more I feel that adding CDI properly is worthwhile, but should be done in a follow up step. Also looks like only kubevirt/kubevirt will benefit from this. 

For now I'm going to resort to adding the cdi manifests that we use in kubevirt/kubevirt (i.e. 1.18.0) so that the cluster-up there gets resilient. WDYT?

rmohr:
> If the manifests and binaries are just placed in the image and not deployed, then my main worry is gone.

Yes the whole point is having everything in the image which *might* be deployed. We already have some switches to distinguish then what you want to actually deploy and we can make them as fine-grained as we want. But everything which *could* be deployed by the switches has to be in the base-image already.

That we deploy right now CDI and rook in kubevirt/kubevirt without pre-cashing them here already is for instance a not so small risk. It only works good enough right now because most of the things are already cashed properly.

cluster-provision/k8s/1.18/provision.sh
oshoval: but here wont it exit 1 if the file wont be copied to /tmp?
would be glad please if it can be tested end to end

dhiller:
@qinqon do you have time to have a look also? At least to check whether the parts of pre-pulling vs. deploying the cnao stuff etc. is correct like this?

dhiller:
@qinqon updated to only use an extra-pre-pull-images file now. PTAL, thanks!

cluster-provision/k8s/validate-manifest-pull-policies.sh
```diff
@@ -0,0 +1,35 @@
...
+    if [ "$#" -lt 1 ]; then
+        usage
+        exit 1
+    fi
+
+    manifest_dir="$DIR/$1/manifests"
+    echo "Checking $manifest_dir"
+    (
+        cd "$DIR/../tools"
+        docker build -f check-image-pull-policies/Dockerfile -t kubevirtci/check-image-pull-policies .
```
dhiller, line 29: TODO: Move this to a build file where the image gets pushed

cluster-provision/k8s/validate-pod-pull-policies.sh
```diff
@@ -0,0 +1,44 @@
...
+        mkdir -p "$manifest_dir/$namespace"
+        for pod in $(${ksh} get pods --namespace "$namespace" --no-headers --output=custom-columns=:.metadata.name); do
+            (${ksh} get pod --output=yaml --namespace "$namespace" "$pod") > "$manifest_dir/$namespace/$pod.yaml"
+        done
+    done
+
+    echo "Checking $manifest_dir"
+    (
+        cd "$DIR/../tools"
+        docker build -f check-image-pull-policies/Dockerfile -t kubevirtci/check-image-pull-policies .
```
dhiller: TODO: Move this to a build file where the image gets pushed

dhiller:
@rmohr I applied the changes to 1.17 and & 1.18, is this required for 1.16 also? I guess that 1.19 is expected to come out at the beginning of next month?

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -2,19 +2,27 @@
 
 set -ex
 
+if [ ! -f "/tmp/extra-pre-pull-images" ]; then
+    echo "WARNING: extra-pre-pull-images list missing"
```
danielBelenky, line 6: Shouldn't we fail the process if we failed to pre-pull the images?

cluster-provision/k8s/fetch-images.sh
```diff
@@ -0,0 +1,53 @@
...
+    temp_file=$(mktemp)
+    trap 'rm -f "${temp_file}"' EXIT SIGINT SIGTERM
+
+    provision_dir="$1"
+    image_regex='([a-z0-9\_\.]+[/-]?)+:[a-z0-9\_\.\-]+'
+    image_regex_w_double_quotes='"?'"${image_regex}"'"?'
+    (
+        # Avoid bailing out because of nothing found in scripts part
+        set +e
+        find "$provision_dir" -type f -name '*.sh' -print0 | \
```
danielBelenky, line 43: Maybe we should just store a list with all the images that we want to cache instead of trying to grep them from the files. If each provider will have a file with pull urls we can easily iterate over it and source it. I suspect that anything else that we will try to do will be error prone and eventually will fail.

cluster-provision/tools/Makefile
```diff
@@ -0,0 +1,29 @@
...
+container: fmt
+	docker build -f check-image-pull-policies/Dockerfile -t kubevirtci/check-image-pull-policies .
+
+.PHONY: vendor
+vendor:
+	$(GO) mod tidy
+	$(GO) mod vendor
+
+.PHONY: push
+push: container
```
qinqon, line 28: Pushing this is done manually ? or do we need some prow infra like gocli ? 

Also  we may have in the future a tarball with scripts and binaries at kubevirtci release page that tarball can contain this binary too, https://github.com/kubevirt/kubevirtci/pull/403

cluster-provision/tools/check-image-pull-policies/check-image-pull-policies.go
```diff
@@ -0,0 +1,177 @@
+package main
```
danielBelenky, line 1: Can you please provide unit-tests for this code?

cluster-provision/tools/check-image-pull-policies/check-image-pull-policies.go
```diff
@@ -0,0 +1,177 @@
...
+	fmt.Printf("%d files with pull policies detected\n", len(filesWithPullPolicies))
+	hasOffendingPolicies := false
+	for filePath, pullPolicies := range filesWithPullPolicies {
+		for image, pullPolicy := range pullPolicies {
+			imageParts := strings.Split(image, ":")
+			var imageTag string
+			if len(imageParts) > 0 {
+				imageTag = imageParts[1]
+			}
+			offending := pullPolicy == corev1.PullAlways ||
```
danielBelenky, line 72: Can you please make it a proper condition? It's not a one liner anyway and I think that it's hard to understand it that way. I would even move it into a proper function to easily test it.

cluster-provision/tools/check-image-pull-policies/check-image-pull-policies.go
```diff
@@ -0,0 +1,177 @@
...
+			if offending {
+				hasOffendingPolicies = true
+			}
+			if !offending && !options.verbose {
+				continue
+			}
+			fmt.Printf("File: %s\n", filePath)
+			fmt.Printf("\tImage: %s\n", image)
+			// https://kubernetes.io/docs/concepts/containers/images/#updating-images
+			if offending {
```
danielBelenky, line 83: This check is redundant as you already checked it above. Maybe you can move the print there?

cluster-provision/tools/check-image-pull-policies/check-image-pull-policies.go
```diff
@@ -0,0 +1,177 @@
...
+	file, err := os.Open(manifestFile)
+	if err != nil {
+		log.Fatalf("Error on opening file %s: %v", manifestFile, err)
+	}
+	//noinspection GoUnhandledErrorResult
+	defer file.Close()
+
+	pullPolicies := map[string]corev1.PullPolicy{}
+	scanner := bufio.NewScanner(file)
+	var bufferString *bytes.Buffer
```
danielBelenky, line 109: Maybe use a proper YAML parser like https://github.com/go-yaml/yaml?

cluster-provision/tools/Makefile
dhiller: I've pushed it manually for now, was sufficient. Not sure whether we do need updates on a regular basis.

cluster-provision/k8s/fetch-images.sh
dhiller: @qinqon argued against this, as the alternative process would also be error prone. Then, instead  of just putting new manifests into the folder, you also would have to update the list, which the user might forget.

cluster-provision/k8s/1.17/provision.sh
dhiller: @oshoval argued against this as to be able to iterate faster when provisioning clusters locally, which makes sense to me.

oshoval: i meant please if a flag is set SKIP_PULL or something, then it would ?

rmohr: Why do we need a warning if we have no images to pull? Let's either agree on a required empty file, or no warning since the sanity check afterwards would detect missing images. Am I missing something?

cluster-provision/k8s/fetch-images.sh
rmohr: I was also favouring a simple and complete list. But for as long as we fail on the sanity check if we can't extract all images for pre-pulling I don't mind too much.

cluster-provision/k8s/1.17/provision.sh
oshoval: the sanity indeed also need to know what it should check, 
i think that it should enforce the file and the check, unless there is a flag that tell them both, all good pass on

cluster-provision/tools/check-image-pull-policies/check-image-pull-policies.go
dhiller: I don't understand? I'm using the k8s deserializer [here](https://github.com/kubevirt/kubevirtci/pull/407/files/9643c60e6ae92c1ee156a58cabefbb0af4a501a2#diff-1925687890c7a0e40494aa643f0b97c7R33)? What is bad about that?

cluster-provision/k8s/1.17/provision.sh
dhiller: Required empty file makes sense. Then let's fail fast if it's not present.

cluster-provision/tools/check-image-pull-policies/check-image-pull-policies.go
dhiller: @qinqon @danielBelenky you might be aware that we need to scan files with multiple yaml documents of which we don't know what to expect there, right? Now, as we are using a proper k8s yaml decoder later on, I thin we are already good by just fetching just the documents one after another, no?

qinqon: I am ok with manual

cluster-provision/k8s/fetch-images.sh
qinqon: If we go back to having all images at a file thinks like bumping CNAO is more complicated, but if we are able to automatize that we don't care too much.

cluster-provision/tools/Makefile
qinqon: Ack, let's be reactive if we see a lot of updates we put some infra.

cluster-provision/k8s/fetch-images.sh
rmohr: Not necessarily. You would basically locally bump your manifests, run the check, it would print the new file-list (or change the file even), you commit the diff and done.

dhiller: Yeah, this was it initially.

mhenriks:
Can we also add to this (for 1.17+ clusters) the manifests under:

https://github.com/kubevirt/kubevirt/tree/master/manifests/testing/external-snapshotter
https://github.com/kubevirt/kubevirt/tree/master/manifests/testing/rook-ceph


I'd also suggest removing the existing ceph-csi manifest.  It will be replaced by above.

cluster-provision/k8s/check-cluster-up.sh
```diff
@@ -12,21 +12,31 @@ function cleanup {
...
+# repetition cluster-up
 (
   ksh="./cluster-up/kubectl.sh"
   cd "$DIR" && cd ../..
   export KUBEVIRTCI_PROVISION_CHECK=1
   export KUBEVIRT_PROVIDER="k8s-${provision_dir}"
   export KUBEVIRT_NUM_NODES=2
   export KUBEVIRT_NUM_SECONDARY_NICS=2
+  if [[ ! "${provision_dir}" =~ ^1\.1[456]$ ]]; then
+    export KUBEVIRT_WITH_CNAO=true
```
oshoval, line 24: it means that sonobuoy would check with CNAO upon provision,
but usually people use the provider without CNAO installed, in the rare case that
with CNAO there isnt failures with sonobuoy and without it there is, we wont see it upon provision,
if its interesting..

cluster-provision/tools/check-image-pull-policies/testdata/blah.yaml
```diff
@@ -0,0 +1,60 @@
+---
```
oshoval, line 1: Nice file name xD

cluster-provision/tools/check-image-pull-policies/check-image-pull-policies.go
```diff
@@ -0,0 +1,200 @@
...
+			bufferString.WriteString(fmt.Sprintf("ERROR: detected pull policies that will always pull images!\n"))
+		}
+	}
+	return hasOffendingPolicies
+}
+
+// isEffectivelyPullAlways checks by looking at the combination of pullPolicy and imageTag
+// whether this will lead to effectively always pulling the image, as described in
+// https://kubernetes.io/docs/concepts/containers/images/#updating-images
+func isEffectivelyPullAlways(pullPolicy corev1.PullPolicy, imageTag string) bool {
```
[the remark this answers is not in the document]
oshoval: imo, but lets hear other as well on that,  i wouldnt make CNAO as part of the cluster-check-up mandatory gating, and about its pull policy, at least not in the first step.
i would focus on the main lanes and usage which is what you already did.
wdyt ?

dhiller:
@qinqon long time no see on this PR ;-) seriously, could you take a look at this here, please? https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/407/check-provision-k8s-1.21/1400069959014420480#1:build-log.txt%3A14084

dhiller:
We need to update CDI to 1.34.1 as that is more stable.

enp0s3 (changes requested):
@dhiller Great work! please see some minor comments below

cluster-provision/tools/check-image-pull-policies/check-image-pull-policies.go
```diff
@@ -0,0 +1,200 @@
...
+	case "Pod":
+		pod := deserializeManifestAs(bufferString, &corev1.Pod{}).(*corev1.Pod)
+		appendPullPoliciesFromContainers(pod.Spec.Containers, pod.Spec.InitContainers, pullPolicies)
+	default:
+		break
+	}
+}
+
+func deserializeManifestAsObject(bufferString *bytes.Buffer) (object runtime.Object, err error) {
+	object, _, err = deserializer.Decode(bufferString.Bytes(), nil, &appsv1.Deployment{})
```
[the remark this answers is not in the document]
dhiller: Yes, it does. Or do I oversee something here? `err` is declared by name in the return values of the function and assigned right here?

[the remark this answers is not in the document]
enp0s3: @dhiller Oh sorry I missed the usage of named return in this function.

cluster-provision/tools/Makefile
```diff
@@ -0,0 +1,29 @@
+SHELL := /bin/bash
```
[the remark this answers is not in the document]
enp0s3: @dhiller Thinking about it again, perhaps makes little sense, since we can just run `go test .` from inside the `tools/check-image-pull-policies`

enp0s3 (approved):
/lgtm

cluster-provision/tools/Makefile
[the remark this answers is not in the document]
dhiller: Did add `test` and `coverage` targets :)

enp0s3 (approved):
/approve

dhiller:
/override check-provision-k8s-1.17

See https://github.com/kubevirt/project-infra/pull/1337

PR #413, opened 2020-07-21
Remove OKD/OCP cluster provision

qinqon:
As second part of ocp/okd kubevirtci cleanup this commits remove
the provision part the gocli and the pack8s from it.


oshoval:
what about os-3.11 ? we want to keep them ? (there is a client that use it still i think)
its broken from time to time, but simple provision fixed it last time someone needed it

rmohr:
> what about os-3.11 ? we want to keep them ? (there is a client that use it still i think)
> its broken from time to time, but simple provision fixed it last time someone needed it

I see CDI still using it.

qinqon:
@oshoval os-3.11 is used by CDI team.

rmohr:
/lgtm
/hold

Let's run a few check jobs to see if gocli still works as expected.

oshoval:
images.sh need to be updated of course (i see the lane was already finished)
how come all the other lanes arent running, prow knows which lanes to run according the files that are changed?

oshoval:
> /test check-provision-k8s-1.16
> /test check-provision-k8s-1.17
> /test check-provision-k8s-1.18

they should run automaticlly ..
on other PRs if you change provision folder it runs auto, it looks like there is kinda a rule, so the rule should include gocli folder i think ? (and images.sh)

qinqon:
/hold cancel

CI is good.

PR #416, opened 2020-07-22
Remove k8s fedora node provider

qinqon:
It's not being used and it add complexity to the project


qinqon:
@phoracek done

ormergi:
/unlgtm
I appreciate your effort to cleanup kubevirtci and I am always about getting rid of non necessary code that we need to maintain, having that said I have second thoughts about removing this provider. 
I think we should re-consider or at least wait for community feedback after I will point out why we have this provider.

The main reason for creating this provider was to enable developers to create an environment where the nodes OS is Fedora, enabling us to test latest releases of various packages, for example 
we could test latests nmstate releases with latest NetworkManager dependencies by using copr repository that targeted to fedora only at that time (it is available for centos today).
While I was working on this provider we found comparability issues with Flannel CNI that affected all k8s-1.17 providers and lanes, we switched to Calico CNI on the go eventually paved the way to prepare kubevirtci for ipv6 support.

As for today not many developers benefit form this provider or even know it exists due to not publishing it at the time created.
Having second provider with different OS could also help us debugging, and developing new features and capabilities that wont go easy with centos OS (current default OS for nodes),  for example: 
DPDK support for kubevirtci effort #356, https://github.com/kubevirt/kubevirt/issues/3405 
and future VM's based SRIOV provider.

qinqon:
/hold 
we need to update gocli container digest

Issue #412, opened 2020-07-21
Sonobuoy failures doesnt cause lanes to fails


oshoval:
In case Sonobuoy fails, it wont fail the lane
for example push-k8s-1.18 lane, where one of the tests failed, and still the lane passed.

https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/408/push-k8s-1.18/1284827424558682112

PR #417, opened 2020-07-22
Fix sonobuoy error detection

oshoval:
Post provision runs cluster check up which
include running sonobuoy test suite.
Even in case that sononbuoy found an error, the lane
still was reported as passed.
Fix that by checking if any of the sonobuoy suites failed.

Fixes https://github.com/kubevirt/kubevirtci/issues/412

### DIFF
--- a/hack/conformance.sh
+++ b/hack/conformance.sh
@@ -1,8 +1,10 @@
-set -xe
+#!/bin/bash
+
+set -xeuo pipefail
 
 ARTIFACTS=${ARTIFACTS:-${PWD}}
 sonobuoy_version=0.18.2
-config_file=$1
+config_file=${1:-}
 
 if [[ -z "$KUBEVIRT_PROVIDER" ]]; then
     echo "KUBEVIRT_PROVIDER is not set" 1>&2
@@ -12,11 +14,25 @@ fi
 export KUBECONFIG=$(cluster-up/kubeconfig.sh)
 
 teardown() {
+    rv=$?
     ./sonobuoy status --json
     ./sonobuoy logs > ${ARTIFACTS}/sonobuoy.log
     results_tarball=$(./sonobuoy retrieve)
     tar -xvzf $results_tarball plugins/e2e/results/
     cp -f $(find plugins/e2e/results/* -name "*.xml") ${ARTIFACTS}/
+
+    if [ $rv -ne 0 ]; then
+        echo "error found, exiting"
+        exit $rv
+    fi
+
+    passed=$(./sonobuoy status --json | jq  ' .plugins[] | select(."result-status" == "passed")'  | wc -l)
+    failed=$(./sonobuoy status --json | jq  ' .plugins[] | select(."result-status" == "failed")'  | wc -l)
+
+    if [ $passed -eq 0 ] || [ $failed -ne 0 ]; then
+        echo "sonobuoy failed"
+        exit 1
+    fi
 }
 
 curl -L https://github.com/vmware-tanzu/sonobuoy/releases/download/v${sonobuoy_version}/sonobuoy_${sonobuoy_version}_linux_amd64.tar.gz | tar -xz


oshoval:
/cc @qinqon 
would you please review?

hack/conformance.sh
```diff
@@ -12,11 +12,25 @@ fi
 export KUBECONFIG=$(cluster-up/kubeconfig.sh)
 
 teardown() {
+    rv=$?
```
ormergi, line 15: What is rv? could you give this variable more informative name?

hack/conformance.sh
```diff
@@ -12,11 +12,25 @@ fi
...
     results_tarball=$(./sonobuoy retrieve)
     tar -xvzf $results_tarball plugins/e2e/results/
     cp -f $(find plugins/e2e/results/* -name "*.xml") ${ARTIFACTS}/
+
+    if [ $rv -ne 0 ]; then
+        echo "error found, exiting"
+        exit $rv
+    fi
+
+    passed=$(./sonobuoy status --json | jq  ' .plugins[] | select(."result-status" == "passed")'  | wc -l)
```
qinqon, line 27: Can we do a `count_by_status` function and will be use like `cound_by_status passed` and `count_by_status failed`

hack/conformance.sh
```diff
@@ -12,11 +12,25 @@ fi
...
+    rv=$?
     ./sonobuoy status --json
     ./sonobuoy logs > ${ARTIFACTS}/sonobuoy.log
     results_tarball=$(./sonobuoy retrieve)
     tar -xvzf $results_tarball plugins/e2e/results/
     cp -f $(find plugins/e2e/results/* -name "*.xml") ${ARTIFACTS}/
+
+    if [ $rv -ne 0 ]; then
+        echo "error found, exiting"
+        exit $rv
```
qinqon, line 24: Are we sure we need this ? I expect the exit not being calculated by the teardown function called by trap.

hack/conformance.sh
oshoval: return value, its pretty common in bash

hack/conformance.sh
oshoval: this is if the main script failed before the teardown and we dont want to mask this error

hack/conformance.sh
oshoval: since result-status has dash, the command need double ", adding it in a function and using $var
would need playing with escapes and so on, it will be less readable
for just single case of two lines, does it worth it ?

qinqon: fair enough

hack/conformance.sh
qinqon: understood

hack/conformance.sh
```diff
@@ -1,4 +1,4 @@
-set -xe
+#!/bin/bash -xe
```
qinqon, line 1: Can we use `set -euxo pipefail` ? 

So we also fail on pipes failures ?

oshoval: done, thx

oshoval:
/test push-k8s-1.18
running sanity

oshoval:
/test push-k8s-1.18
try again

hack/conformance.sh
```diff
@@ -1,8 +1,10 @@
-set -xe
+#!/bin/bash
+
+set -xeuo pipefail
 
 ARTIFACTS=${ARTIFACTS:-${PWD}}
 sonobuoy_version=0.18.2
-config_file=$1
+config_file=${1:-}
```
ormergi, line 7: Shouldnt we need to check if the file exists, here or at caller function line 44, and only then continue?

hack/conformance.sh
```diff
@@ -1,8 +1,10 @@
...
 
 if [[ -z "$KUBEVIRT_PROVIDER" ]]; then
     echo "KUBEVIRT_PROVIDER is not set" 1>&2
```
ormergi, line 10: By specifying `1>&2` we would get the value of `KUBEVIRT_PROVIDER`?

hack/conformance.sh
ormergi: Please consider renaming thin value to `returnValue` or  `errorCode`,
even though it is common in bash it is not trivial for new comers.

hack/conformance.sh
ormergi: Just to understand, Is there special reason to use `exit` and not `return` inside `teardown()` function?

hack/conformance.sh
```diff
@@ -12,11 +14,25 @@ fi
...
+    rv=$?
     ./sonobuoy status --json
     ./sonobuoy logs > ${ARTIFACTS}/sonobuoy.log
     results_tarball=$(./sonobuoy retrieve)
     tar -xvzf $results_tarball plugins/e2e/results/
     cp -f $(find plugins/e2e/results/* -name "*.xml") ${ARTIFACTS}/
+
+    if [ $rv -ne 0 ]; then
+        echo "error found, exiting"
+        exit $rv
```
ormergi, line 26: Just to understand, Is there special reason to use exit and not return inside teardown() function?

hack/conformance.sh
ormergi: Please consider renaming thin value to returnValue or errorCode,
even though it is common in bash it is not trivial for new comers.

hack/conformance.sh
```diff
@@ -12,11 +14,25 @@ fi
 export KUBECONFIG=$(cluster-up/kubeconfig.sh)
 
 teardown() {
+    rv=$?
```
ormergi, line 17: Please consider renaming thin value to `returnValue` or `errorCode`,
even though it is common in bash it is not trivial for new comers.

oshoval: i think we should stick with the common, new comers will learn it

hack/conformance.sh
oshoval: exit is more explicit in this context, and we do want to exit with a specific ret code, i dont know by heart if return would do here the same, but anyway wanted the known explicit method

hack/conformance.sh
oshoval: i didnt touch this line, out of this PR scope, 
anyhow its for redirecting stdout to stderr

hack/conformance.sh
oshoval: it allows to override the conformance_config env var, 
and the change i did here is because i added set -u

Issue #414, opened 2020-07-22
Suggestion, yaml modifier


oshoval:
Due to the work on dual stack, one sonobuoy conformance test is failing,
and it seems its the portmap CNI,
update to Calico might fix it.
While doing that i saw that the new Calico version changed a lot of things, and we also have some custom changes,
merging manually prone for errors, for example if we forget the scc.

I suggest a method which will do the following,
1. a base official file, pinned to a specific version. i.e 3.12.0 (until the base would change),
will be available in the repo.
2. a script that will take [1] and update our custom fields, generating the actual file we use with provision.
(the script will need strict matching)

This way updating the plugin, would be smooth, and we will just need to refetch [1] and make sure [2] produces the desired changes (which will be reflected in the PR of the updating).

phoracek:
Have you checked whether bump in Calico fixes the issue, so we know it is worth it?

oshoval:
> Have you checked whether bump in Calico fixes the issue, so we know it is worth it?

i am on it, manually,  will report,
on any case, i think it worth considering, since from what i know cni.yaml is in provision without the original one
but lets first focus on the line above.

PR #419, opened 2020-07-23
Refactor calico on K8s 1.18 provider

oshoval:
Align cni.yaml to original 3.12.0 yaml

In order to keep the original calico yaml,
patch it during provision and be able to
update it easily when needed, this commit
align the yaml to the official one.
Next commit would update the mechanism of provision
and introduce the needed patch.

---

Provision updated to be based on the original calico
manifest, and then apply kubevirtci custom changes.
This way it would be clear what changes are applied,
and updating the manifest would be easier and safer.

Fixes: #414


oshoval:
updated hash according
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/419/push-k8s-1.18/1286324196414066688

oshoval:
@phoracek 
the 3 commits here are the first 3 commits of https://github.com/kubevirt/kubevirtci/pull/422
which you lgtmed, 
did this separation because was building it and verifying by steps,
and so we can bisect in case needed,
so can we merge this one before please? (unless not needed this seperation)
thx

phoracek:
The current CNI was not supposed to be edited neither https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/manifests/cni/calico/manifest.yaml#L3, but it was.

Would you rename the CNI yaml so it makes it clear that it should not be edited in place? `cni.do-not-change.yaml`?

oshoval:
> The current CNI was not supposed to be edited neither https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/manifests/cni/calico/manifest.yaml#L3, but it was.
> 
> Would you rename the CNI yaml so it makes it clear that it should not be edited in place? `cni.do-not-change.yaml`?

i think that once we have `patch` in the code it would be known ?
also i think its more official to use cni.yaml and then its clean to just patch on it, without renaming, or having hint that this file is const in the name of it ?

phoracek:
How would I know about the patch? If I will need to change the CNI I will search for CNI and change it in place. We need to make it crystal clear that this is not meant to be changed or to involve a CI blocking us from editing it.

oshoval:
> How would I know about the patch? If I will need to change the CNI I will search for CNI and change it in place. We need to make it crystal clear that this is not meant to be changed or to involve a CI blocking us from editing it.

how about wget the release , untar it and extract the yaml from there?
this is the only official method btw to extract a specific X.Y.Z , z version ,since the release url is just for X.Y 
i.e 3.15 but it point to latest Z

cluster-provision/k8s/1.18/manifests/cni.diff
```diff
@@ -0,0 +1,31 @@
...
+             # Disable file logging so `kubectl logs` works.
+             - name: CALICO_DISABLE_FILE_LOGGING
+               value: "true"
+@@ -762,6 +765,9 @@
+           effect: NoSchedule
+       serviceAccountName: calico-kube-controllers
+       priorityClassName: system-cluster-critical
++      securityContext:
++        seLinuxOptions:
++          type: spc_t
```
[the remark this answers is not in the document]
oshoval: https://github.com/kubevirt/kubevirtci/pull/422#discussion_r460709427

oshoval:
> I like the idea of using `patch`, nice!
> We should use the same version of Calico in all providers to keep maintenance simple as possible,
> this refactoring should be done to k8s 1.17 and k8s 1.16 providers.

will do on a follow PR, once one is stabilized

oshoval:
> The current CNI was not supposed to be edited neither https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/manifests/cni/calico/manifest.yaml#L3, but it was.
> 
> Would you rename the CNI yaml so it makes it clear that it should not be edited in place? `cni.do-not-change.yaml`?

done

cluster-provision/k8s/1.18/manifests/cni.do-not-change.yaml
```diff
@@ -798,3 +792,5 @@ metadata:
 
 ---
 # Source: calico/templates/configure-canal.yaml
+
```
phoracek, line 795: Please drop these trailing lines unless they come from the downloaded source

cluster-provision/k8s/1.18/manifests/cni.diff
```diff
@@ -0,0 +1,31 @@
+--- calico.3.12.0.yaml	2020-07-23 11:14:17.000000000 +0300
```
phoracek, line 1: This old name does not match the sources we have, does it? Shouldn't it match our cni.yaml?

cluster-provision/k8s/1.18/manifests/cni.do-not-change.yaml
oshoval: its came from the download source, all the changes there, the file is identical now

cluster-provision/k8s/1.18/manifests/cni.diff
oshoval: it doesn't need to match it applies anyway according the args give to patch commands
i generate the patch using diff -Naur between src and dst

oshoval: https://github.com/kubevirt/kubevirtci/pull/422/commits/29fe0fa0a11af07935ad687135dfaf63c7ca7972
here it changed, it depends on the file names i had locally when creating the diff
i dont think we should be strict about it

phoracek:
Waiting for @ormergi in case he wants to leave more comments.

PR #421, opened 2020-07-24
kind: minor fixes to down()

jean-edouard:
Fixes the following when taking down a multi-node cluster:
```
cluster/kind/common.sh: line 204: [: kind-1.17: binary operator expected
```
and this when calling `make cluster-down` on an already down cluster:
```
rm: cannot remove '/home/build/kubevirt/_ci-configs/kind-k8s-1.17/kind.yaml': No such file or directory
```

### DIFF
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -201,9 +201,9 @@ function _kubectl() {
 
 function down() {
     _fetch_kind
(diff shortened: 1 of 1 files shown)


oshoval:
Nice
/lgtm

wonder if we should always use double quote as a rule of thumb or its an overkill ?
some consider it as a good practice

PR #422, opened 2020-07-26
Bump calico to 3.15.1 for K8s-1.18

oshoval:
In order to fix sonobuoy tests on dual stack,
a bump of calico is needed, because it updates
the CNI plugin from v0.8.0 to v0.8.6 which has an updated portmap
plugin, which fixes the problem of
`[sig-scheduling] SchedulerPredicates [Serial] validates that there is no conflict between pods with same hostPort but different hostIP and protocol [Conformance]`


oshoval:
No regression was caused on kubevirt 
https://github.com/kubevirt/kubevirt/pull/3894

cluster-provision/k8s/1.18/manifests/cni.diff
```diff
@@ -0,0 +1,22 @@
...
++              "allow_ip_forwarding": true
++          },
+           "policy": {
+               "type": "k8s"
+           },
+@@ -3709,6 +3712,9 @@
+           effect: NoSchedule
+       serviceAccountName: calico-kube-controllers
+       priorityClassName: system-cluster-critical
++      securityContext:
```
phoracek, line 17: I see we had this in the previous cni.yml, but do we still need it with the latest Calico version?

cluster-provision/k8s/1.18/provision.sh
```diff
@@ -204,13 +206,10 @@ spec:
 EOF
 
 
-default_cidr="192.168.0.0/16"
```
phoracek, line 207: Why did you drop this?

oshoval: see please the commit desc
https://github.com/kubevirt/kubevirtci/pull/422/commits/3a0b238baf22e64230238067b7a1cadc8f4bdafc

cluster-provision/k8s/1.18/manifests/cni.diff
oshoval: i didnt try without, i can try if needed, unless something set it, i think we will need, else stuff might fail

cluster-provision/k8s/1.18/provision.sh
phoracek: Gotcha. Missed that

cluster-provision/k8s/1.18/manifests/cni.diff
phoracek: I see it was intentionally added in https://github.com/kubevirt/kubevirtci/pull/321/files, so we likely need to keep it.

phoracek:
/hold

Waiting on dependency #419

cluster-provision/k8s/1.18/manifests/cni.diff
```diff
@@ -10,17 +10,6 @@
...
            },
-@@ -3576,8 +3579,8 @@
-             # The default IPv4 pool to create on startup if none exists. Pod IPs will be
-             # chosen from this range. Changing this value after installation will have
-             # no effect. This should fall within `--cluster-cidr`.
--            # - name: CALICO_IPV4POOL_CIDR
--            #   value: "192.168.0.0/16"
-+            - name: CALICO_IPV4POOL_CIDR
-+              value: "10.244.0.0/16"
-             # Disable file logging so `kubectl logs` works.
```
ormergi: I think we need this, do we still get logs?

oshoval: i didnt touch the logs, i touched the 2 above it
and see why reason here please
https://github.com/kubevirt/kubevirtci/pull/422#discussion_r460696582

ormergi: I understand why you drop CIDR variable, my question is why removing lines 22-23
since there is `-` at the start of the lines I assume it will remove them, right?

Do we get logs with `kubectl logs` on the new version?

oshoval: its a diff file with -C 3 (means it shows 3 lines before and after), it actually had zero effect before and also after
it didnt change those line, it just showed the scope

since its a diff file that was changed in non + / - lines , there is zero effect, 
i took just the changes compared to the original and applied only them , removing the unneeded CIDR

the - in the beginning is the diff file of this PR
is a diff of a diff this is why there are 2 columns

if the change of kubectl logs would be touched the diff wont be able to apply

phoracek:
@oshoval I want to make sure we don't break KubeVirt e2e by this change. Could we run those tests against this change?

oshoval:
> @oshoval I want to make sure we don't break KubeVirt e2e by this change. Could we run those tests against this change?

did that already before the name change of the file, can do it again if needed
https://github.com/kubevirt/kubevirtci/pull/422#issuecomment-664002821

creating a new provider first

oshoval:
> > @oshoval I want to make sure we don't break KubeVirt e2e by this change. Could we run those tests against this change?
> 
> did that already before the name change of the file, can do it again if needed
> [#422 (comment)](https://github.com/kubevirt/kubevirtci/pull/422#issuecomment-664002821)
> 
> creating a new provider first - done
> 
> running test
> [kubevirt/kubevirt#3902](https://github.com/kubevirt/kubevirt/pull/3902)

@phoracek 
passed :+1:

oshoval:
i see some problem with dual stack, but i think the bug is there and not on this PR, anyway, lets wait a moment that i will verify and then just release the hold please

phoracek:
What is the problem?

oshoval:
> What is the problem?

More to come, i did couple steps in one shot and there are some nits needed to be fixed)
in glance , tried to refactor provision so it would use kubeadm config instead of parameters + dual stack + 3.15.1 new one and the patching mode, so i will update when i am concrete info please.

phoracek:
/approve cancel

Let's make sure we are not breaking anything. Waiting for your findings.

oshoval:
@phoracek 
All safe, found the problem and have a potential fix for it, its not related to this PR.
it just because in the provision stage itself (before the final cleanup and the reboot of cluster-up) the nm isnt configured well,
so it doesnt have ipv6, and since i refactor it to use the same kubeadm init method on both cluster-up and provision, and there is dual stack in the config, it must have ipv6

please approve and unhold if its ok
thanks

PR #428, opened 2020-08-03
vm-image-builder: Automate creating customized Kubevirt VM's images

ormergi:
This PR add tool for automating the process of creating new container-disk image for KubeVirt VMs
based on  http://github.com/ormergi/vm-image-builder

It is targeted for developers or anyone who would like to create customized ephemeral container disk image for Kubevirt VM's
As were presented at kubevirt/kubevirt#3834

This tool should be used on the user host, integrating this to CI will be presented in following PR.

This PR:
Fixes kubevirt/kubevirt#3903
Is needed to continue with kubevirt/kubevirt#3834 and  kubevirt/kubevirt#3850

Release note:
```
Automate creating customized KubeVirt container disk images.
```

### DIFF
--- a/cluster-provision/images/vm-image-builder/OWNERS
+++ b/cluster-provision/images/vm-image-builder/OWNERS
@@ -1,0 +1,6 @@
+filters:
+  ".*":
+    reviewers:
+      - ormergi
+    approvers:
+      - ormergi
--- a/cluster-provision/images/vm-image-builder/README.md
+++ b/cluster-provision/images/vm-image-builder/README.md
@@ -1,0 +1,162 @@
+# Customize ephemeral container-disk images for KubeVirt VM's
+
+This tool is targeted for KubeVirt developers, users or anyone who
+would like to create KubeVirt VM's with customized container-disk,
+without the hassle of installing and customizing the OS manually.
+
+Use cases:
+- VM's for Demos
+- Pre-configured Kubevirt VM image for tests, for example:
+    - Image with qemu-guest-agent installed and configured.
+    - Image with sriov drivers for sriov-lane images.
+    - Image with DPDK package for testing DPDK applications on KubeVirt.
+
+## Create customized container-disk
+
+In order to create customized  container-disk for KubeVirt VM,
+use `create-containderdisk.sh` script.
+
+This script automates the process of customizing VM image and build Kubevirt VM container-disk-image.
+Soon there will be used for creating and image automatically using CI.
+
+What this script does:
+- Download VM image from given URL
+
+- Customize VM image using `customize-image.sh` script, according to given cloud-config file.
+  Keeping customizing image method loosely-coupled, so it will be easy to maintain.
+  You can use any image customizing script by exporting `CUSTOMIZE_IMAGE_SCRIPT=<script path>`.
+
+- Building container-disk image using `build-containerdisk.sh`
+  according to the doc:
+  https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md
+
+- Store the image in local registry and exports the container image to .tar file, so it will be easier to store or send.
+
+This script expected variables:  
+- `VM_IMAGE_URL` - URL to download the image
+- `CLOUD_CONFIG_PATH` - cloud-conifg file path used  for customizing the image with cloud-init
+- `IMAGE_NAME`, `TAG` - container-disk image tag that will be created tag details `IMAGE_NAME:TAG`
+- `OS_VARIANT`- required by customize-image.sh
+
+Notes:
+- If no variables passed, the script will create container-disk image according to the files at `example/`.
+- Currently, image customization process done by spinning up live VM with cloud-init disk to provision the image with the changes described by the cloud-config file.
+- It is necessary to add `sudo shutdown` at the end of the cloud-config`runcmd` block, so the script will not wait for user input.
+- You can use your own image customizing script by exporting the path to `CUSTOMIZE_IMAGE_SCRIPT`.
+
+This script requires the following packages:
+- cloud-utils
+- docker-ce
+- libvirt
+- libguestfs
+- qemu-img
+
+## How to use:
+```bash
+cd cluster-provision/images/container-disk-images/
+# Export image name and tag
+$ export IMAGE_NAME="example-fedora"
+$ export TAG="32"
+
+# Export VM image URL:
+$ export IMAGE_URL=$(cat example/image-url)
+$ export OS_VARIANT=fedora31
+
+# Export cloud-config file
+$ export CLOUD_CONFIG_PATH="example/cloud-config"
+
+# Run script
+$ ./create-containerdisk.sh
+...
+Successfully tagged example-fedora:32
+...
+Container image saved as tar file at: example-fedora_build/example-fedora-32.tar
+...
+
+# You can send / store the image .tar file:
+$ ls example-fedora_build/example-fedora-32.tar
+example-fedora_build/example-fedora-32.tar
+```
+
+## Publish customized container-disk image
+
+Push the new container-disk image to a registry with `publish-containerdisk.sh` script.
+
+This script exports:
+REGISTRY - the registry the image will be stored at.
+REPOSITORY - image repository.
+
+This script expects:
+IMAGE_NAME - container disk image name.
+TAG - container disk image tag.
+
+### Example:
+```bash
+$ export IMAGE_NAME=example-fedora
+$ export TAG=32
+
+$ ./publish-containerdisk.sh
+```
+
+### Push the new image to local cluster registry:
+```bash
+$ export IMAGE_NAME=example-fedora
+$ export TAG=32
+
+# From kubevirtci / kubevirt directory
+$ export REGISTRY="localhost:$(./cluster-up/cli.sh ports registry | tr -d '\r')"$
+
+$ ./publish-containerdisk.sh
+```
+
+### Create Kubevirt VM's with the new image
+
+In the VMI / VM yaml file, change `spec.volumes[].containerDisk.image` to the new image path.
+
+It is possible to use an image from local cluster registry.
+
+```bash
+$ kubectl apply -f <VMI yaml file>
+$ kubectl wait --for=condition=AgentConnected vmi $VMI_NAME --timeout 5m
+$ virtctl console testvm1
+```
+
+
+## Build container-disk image
+
+To build container-disk image form qcow2 image file, use `build-containerdisk.sh` script.
+
+This script converts `qcow2` image to container-disk image that can be consumed by KubeVirt VM's.
+
+What this script does:
+- Creates temp directory with a copy of the source VM image.
+
+- Build container image with kubevirt/container-disk-v1alpha layer, 
+  using `Dokcerfile.template` according to:
+  https://github.com/kubevirt/kubevirt/blob/master/docs/container-register-disks.md  
+  The image is stored in the local registry.
+  
+- Exports container image as .tar file
+
+Arguments:
+- IMAGE_NAME - container-disk image name
+- TAG - container-disk image tag
+- VM_IMAGE_FILE - source VM image path to convert
+
+### Example:
+```bash
+$ image_name='example-fedora'
+$ tag='32'
+$ vm_image='customized-image.qcow2'
+
+$ ./build-containerdisk.sh $image_name $tag $vm_image
+...
+Successfully tagged example-fedora:32
+...
+Container image saved as tar file at: example-fedora_build/example-fedora-32.tar
+...
+
+# Generated image .tar file
+$ ls example-fedora_build/example-fedora-32.tar
+example-fedora_build/example-fedora-32.tar
+```
--- a/cluster-provision/images/vm-image-builder/build-containerdisk.sh
+++ b/cluster-provision/images/vm-image-builder/build-containerdisk.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+set -exuo pipefail
+
+SCRIPT_PATH=$(dirname "$(realpath "$0")")
+
+function cleanup() {
+  if [ $? -ne 0 ]; then
+    rm -f "${tar_file}"
+    docker image rm -f "${full_image_tag}"
+  fi
+
+  rm -rf "${temp_dir}"
+}
+
+IMAGE_NAME=$1
+TAG=$2
+VM_IMAGE_FILE=$3
+
+template_dockerfile="${SCRIPT_PATH}/Dockerfile.containerdisk"
+
+parent_dir=$(dirname "$VM_IMAGE_FILE")
+tar_file="${parent_dir}/${IMAGE_NAME}-${TAG}.tar"
+full_image_tag="${IMAGE_NAME}:${TAG}"
+
+trap 'cleanup' EXIT SIGINT
+
+temp_dir=$(mktemp -d -p /tmp -t build.container-disk.XXXX)
+cp "${template_dockerfile}" "${temp_dir}/Dockerfile"
+cp "${VM_IMAGE_FILE}" "${temp_dir}/image"
+
+pushd "${temp_dir}"
+  docker build -t "${full_image_tag}" .
+popd
+
+docker save --output "${tar_file}" "${full_image_tag}"
+
+echo "Container image saved as tar file at: ${tar_file}"
--- a/cluster-provision/images/vm-image-builder/create-containerdisk.sh
+++ b/cluster-provision/images/vm-image-builder/create-containerdisk.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+set -exuo pipefail
+
+SCRIPT_PATH=$(dirname "$(realpath "$0")")
+
+export CUSTOMIZE_IMAGE_SCRIPT=${CUSTOMIZE_IMAGE_SCRIPT:-"${SCRIPT_PATH}/customize-image.sh"}
+
+function customize_image() {
+  local source_image=$1
+  local os_variant=$2
+  local customized_image=$3
+  local cloud_config=$4
+
+  # Backup the VM image and pass copy of the original image
+  # in case customizing script fail.
+  vm_image_copy="$(dirname "${customized_image}")/copy-${source_image}"
+  cp "${source_image}" "${vm_image_copy}"
+
+  # TODO: convert this script and its dependencies to container
+  ${CUSTOMIZE_IMAGE_SCRIPT} "${vm_image_copy}" "${os_variant}" "${customized_image}" "${cloud_config}"
+
+  # Backup no longer needed.
+  rm -f "${vm_image_copy}"
+}
+
+function cleanup() {
+  if [ $? -ne 0 ]; then
+    rm -rf "${build_directory}"
+  fi
+
+  rm -f "copy-${VM_IMAGE}"
+}
+
+export IMAGE_NAME=${IMAGE_NAME:-example-fedora}
+export TAG=${TAG:-32}
+export OS_VARIANT=${OS_VARIANT:-fedora31}
+export CLOUD_CONFIG_PATH=${CLOUD_CONFIG_PATH:-"${SCRIPT_PATH}/example/cloud-config"}
+export VM_IMAGE_URL=${VM_IMAGE_URL:-$(cat "${SCRIPT_PATH}/example/image-url")}
+
+readonly VM_IMAGE="source-image.qcow2"
+readonly build_directory="${IMAGE_NAME}_build"
+readonly new_vm_image_name="provisioned-image.qcow2"
+
+trap 'cleanup' EXIT SIGINT
+
+pushd "${SCRIPT_PATH}"
+  cleanup
+
+   if ! [ -e "${VM_IMAGE}" ]; then
+    # Download base VM image
+    curl -L "${VM_IMAGE_URL}" -o "${VM_IMAGE}"
+  fi
+
+  mkdir "${build_directory}"
+
+  customize_image "${VM_IMAGE}" "${OS_VARIANT}" "${build_directory}/${new_vm_image_name}" "${CLOUD_CONFIG_PATH}"
+
+  ${SCRIPT_PATH}/build-containerdisk.sh "${IMAGE_NAME}" "${TAG}" "${build_directory}/${new_vm_image_name}"
+
+popd
--- a/cluster-provision/images/vm-image-builder/customize-image.sh
+++ b/cluster-provision/images/vm-image-builder/customize-image.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+set -exuo pipefail
+
+function cleanup() {
+  if [ $? -ne 0 ]; then
+    rm -f "${CUSTOMIZE_IMAGE_PATH}"
+  fi
+
+  rm -rf "${CLOUD_INIT_ISO}"
+  virsh destroy "${DOMAIN_NAME}" || true
+  virsh undefine "${DOMAIN_NAME}" || true
+}
+
+SOURCE_IMAGE_PATH=$1
+OS_VARIANT=$2
+CUSTOMIZE_IMAGE_PATH=$3
+CLOUD_CONFIG_PATH=$4
+
+readonly DOMAIN_NAME="provision-vm"
+readonly CLOUD_INIT_ISO="cloudinit.iso"
+
+trap 'cleanup' EXIT SIGINT
+
+# Create cloud-init user data ISO
+cloud-localds "${CLOUD_INIT_ISO}" "${CLOUD_CONFIG_PATH}"
+
+echo "Customize image by booting a VM with
+ the image and cloud-init disk
+ press ctrl+] to exit"
+virt-install \
+  --memory 2048 \
+  --vcpus 2 \
+  --name $DOMAIN_NAME \
+  --disk "${SOURCE_IMAGE_PATH}",device=disk \
+  --disk "${CLOUD_INIT_ISO}",device=cdrom \
+  --os-type Linux \
+  --os-variant "${OS_VARIANT}" \
+  --virt-type kvm \
+  --graphics none \
+  --network default \
+  --import
+
+# Stop VM
+virsh destroy $DOMAIN_NAME || true
+
+# Prepare VM image
+virt-sysprep -d $DOMAIN_NAME --operations machine-id,bash-history,logfiles,tmp-files,net-hostname,net-hwaddr
+
+# Remove VM
+virsh undefine $DOMAIN_NAME
+
+# Convert image
+qemu-img convert -c -O qcow2 "${SOURCE_IMAGE_PATH}" "${CUSTOMIZE_IMAGE_PATH}"
--- a/cluster-provision/images/vm-image-builder/example/image-url
+++ b/cluster-provision/images/vm-image-builder/example/image-url
@@ -1,0 +1,1 @@
+https://download.fedoraproject.org/pub/fedora/linux/releases/32/Cloud/x86_64/images/Fedora-Cloud-Base-32-1.6.x86_64.qcow2
--- a/cluster-provision/images/vm-image-builder/Dockerfile.containerdisk
+++ b/cluster-provision/images/vm-image-builder/Dockerfile.containerdisk
@@ -1,0 +1,2 @@
+FROM kubevirt/container-disk-v1alpha
+ADD image /disk
--- a/cluster-provision/images/vm-image-builder/example/cloud-config
+++ b/cluster-provision/images/vm-image-builder/example/cloud-config
@@ -1,0 +1,11 @@
+#cloud-config
+password: fedora
+chpasswd: { expire: False }
+ssh_pwauth: yes
+runcmd:
+  - sudo dnf install -y qemu-guest-agent stress
+  - sudo dnf clean all
+  - sudo hostnamectl set-hostname ""
+  - sudo hostnamectl set-hostname "" --transient
+  - sudo systemctl start qemu-guest-agent
+  - sudo shutdown
--- a/cluster-provision/images/vm-image-builder/fedora-sriov-lane/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-sriov-lane/cloud-config
@@ -1,0 +1,26 @@
+#cloud-config
+password: fedora
+chpasswd: { expire: False }
+ssh_pwauth: yes
+write_files:
+    - path: /etc/modules-load.d/mlx5.conf
+      content: |
+        mlx5_core
+        mlx5_ib
+    - path: /etc/modules-load.d/mlx4.conf
+      content: |
+        mlx4_core
+        mlx4_ib
+    - path: /etc/modules-load.d/i40e.conf
+      content: |
+        i40e
+    - path: /etc/modules-load.d/igb.conf
+      content: |
+        igb
+runcmd:
+  - sudo dnf install -y kernel-modules-$(uname -r) qemu-guest-agent
+  - sudo dnf clean all
+  - sudo hostnamectl set-hostname ""
+  - sudo hostnamectl set-hostname "" --transient
+  - sudo systemctl start qemu-guest-agent
+  - sudo shutdown
--- a/cluster-provision/images/vm-image-builder/fedora-sriov-lane/image-url
+++ b/cluster-provision/images/vm-image-builder/fedora-sriov-lane/image-url
@@ -1,0 +1,1 @@
+https://download.fedoraproject.org/pub/fedora/linux/releases/32/Cloud/x86_64/images/Fedora-Cloud-Base-32-1.6.x86_64.qcow2
--- a/cluster-provision/images/vm-image-builder/publish-containerdisk.sh
+++ b/cluster-provision/images/vm-image-builder/publish-containerdisk.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+set -exuo pipefail
+
+export REGISTRY=${REGISTRY:-docker.io}
+export REPOSITORY=${REPOSITORY:-kubevirt}
+
+full_image_tag="${REGISTRY}/${REPOSITORY}/${IMAGE_NAME}:${TAG}"
+
+docker tag "${IMAGE_NAME}:${TAG}" "${full_image_tag}"
+docker push "${full_image_tag}"


cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -0,0 +1,59 @@
...
+
+function cleanup() {
+  echo "[customize-image] cleanup"
+  rm -rf "$CLOUD_INIT_ISO"
+  virsh destroy "$DOMAIN_NAME" || true
+  virsh undefine "$DOMAIN_NAME" || true
+
+  if [ $? -ne 0 ]; then
+    rm -f "$NEW_IMAGE"
+  fi
```
EdDev, line 12: This is a common bash pattern:
```
[ $? -ne 0 ] && rm -f "$NEW_IMAGE"
```

But this will always run because you do `|| true`

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -0,0 +1,59 @@
...
+}
+
+IMAGE_PATH=$1
+NEW_IMAGE=$2
+CLOUD_CONFIG_PATH=$3
+
+export OS_VARIANT=${OS_VARIANT:-fedora31}
```
EdDev, line 19: Why 31 and not 32? Maybe even without a version? It is just a label, nothing more.

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -0,0 +1,59 @@
...
+
+readonly DOMAIN_NAME="provision-vm"
+readonly CLOUD_INIT_ISO="cloudinit.iso"
+
+if ! [ -f "${CLOUD_CONFIG_PATH}" ]; then
+  CLOUD_CONFIG_PATH="user-cloud-config"
```
EdDev, line 25: I do not understand this. You use the same variable to provide the file instead of the provided path?

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -0,0 +1,59 @@
...
+fi
+
+trap 'cleanup' EXIT SIGINT
+
+# Create cloud-init user data ISO
+cloud-localds $CLOUD_INIT_ISO $CLOUD_CONFIG_PATH
+
+echo "Customize image by booting a VM with
+ the image and cloud-init disk
+ press ctrl+] to exit"
```
EdDev, line 35: > press ctrl+] to exit

But you do not need this as you will not stay on the console, right?

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -0,0 +1,59 @@
...
+  --import
+
+virt-sysprep -d $DOMAIN_NAME --operations machine-id,bash-history,logfiles,tmp-files,net-hostname,net-hwaddr,customize --hostname ""
+
+# Remove VM
+virsh destroy $DOMAIN_NAME || true
+virsh undefine $DOMAIN_NAME
+
+# Remove cloud-init image
+rm -rf $CLOUD_INIT_ISO
```
EdDev, line 56: Won't this happen on exit anyway (based on the trap)?

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -0,0 +1,59 @@
...
+  virsh destroy "$DOMAIN_NAME" || true
+  virsh undefine "$DOMAIN_NAME" || true
+
+  if [ $? -ne 0 ]; then
+    rm -f "$NEW_IMAGE"
+  fi
+}
+
+IMAGE_PATH=$1
+NEW_IMAGE=$2
```
EdDev, line 16: Is this also a path, right?

cluster-provision/images/vm-image-builder/build-containerdisk.sh
```diff
@@ -0,0 +1,36 @@
...
+)"
+
+function cleanup() {
+  echo "[build-container-disk] cleanup"
+  rm -f Dockerfile
+
+  if [ $? -ne 0 ]; then
+    rm -f "$tar_file"
+    docker image rm "$full_image_tag"
+  fi
```
EdDev, line 16: I do not understand this one. If the file removal done earlier fails, you do not remove the rest?
It also uses "-e" so it is expected to fail anyway on first error, I see no point to check for the returned value.

cluster-provision/images/vm-image-builder/build-containerdisk.sh
```diff
@@ -0,0 +1,36 @@
...
+}
+
+image_name=$1
+tag=$2
```
EdDev, line 20: Why a tag is needed? Isn't there a default? (latest)

cluster-provision/images/vm-image-builder/build-containerdisk.sh
```diff
@@ -0,0 +1,36 @@
...
+vm_image_file=$3
+
+export TEMPLATE_DOCKERFILE=${TEMPLATE_DOCKERFILE:-$SCRIPT_DIR/Dockerfile.containerdisk}
```
EdDev, line 23: Why do you need to export this? Why is using it later?

cluster-provision/images/vm-image-builder/build-containerdisk.sh
```diff
@@ -0,0 +1,36 @@
...
+
+parent_dir=$(basename "$(dirname "$vm_image_file")")
+tar_file="${parent_dir}/${image_name}-${tag}.tar"
+full_image_tag="${image_name}:${tag}"
+
+trap 'cleanup' EXIT
+
+export IMAGE=${vm_image_file}
```
EdDev, line 31: Why do you need to export this?

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: so should I move it to line 6?

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: Yes this the path of the new qcow2 file that will be created

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: If the file path of `CLOUD_CONFIG_PATH` does not exist its sets `CLOUD_CONFIG_PATH` default file path.
There is an error here because the file "user-cloud-config" is not there I forgot to add it,
I should set `CLOUD_CONFIG_PATH` default to `example/example-cloud-config` file

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: You can run this script only, and you didnt add `sudo shutdown` in cloud-config file at `runCmd`
you will stay connected to the console until you close it manually.
This is good if you want to check what is the state of the image on live VM after cloud init provision.

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: Yes it will, at first I though it should be removed so it wont collide with `qemu-img convert`
but now I dont see how this could happen, this like should be removed.

cluster-provision/images/vm-image-builder/build-containerdisk.sh
ormergi: The idea is to  delete the exported image tar file and the image stored in local docker registry, if there is was error during  he script.
I will move lines 13-16 to line 10 so it will actually catch errors from the script and not from cleanup().

cluster-provision/images/vm-image-builder/build-containerdisk.sh
ormergi: To enable users to pick their own tag

cluster-provision/images/vm-image-builder/build-containerdisk.sh
ormergi: You right it should not be exported, I will remove `export` in following commits

cluster-provision/images/vm-image-builder/build-containerdisk.sh
ormergi: You right it should not be exported, I will remove export in following commits

cluster-provision/images/vm-image-builder/customize-image.sh
EdDev: Just run it without any condition, the `-f` will always cause it to succeed.

cluster-provision/images/vm-image-builder/customize-image.sh
EdDev: So if be consistent and add "PATH" to this name as well.
Maybe you should call them SRC and DST (instead of NEW) for name consistency.

cluster-provision/images/vm-image-builder/build-containerdisk.sh
EdDev: I'm not sure I understand what you just mentioned but please make sure to test your conditions to make sure they do what you indented them to do.

cluster-provision/images/vm-image-builder/build-containerdisk.sh
EdDev: Lets continue the "why"... Why would the users pick their own tag?
I fail to understand the scenario.

cluster-provision/images/vm-image-builder/test-containerdisk.sh
```diff
@@ -0,0 +1,67 @@
...
+        make cluster-down
+      popd
+      rm -rf $kubevirt_dir
+      rm -rf "${IMAGE_NAME}_build"
+    fi
+}
+
+readonly KUBEVIRT_URL="https://github.com/kubevirt/kubevirt"
+
+trap 'cleanup' EXIT SIGINT
```
EdDev, line 28: I do not think you need SIGINT, EXIT should cover exit reasons. (please confirm)

cluster-provision/images/vm-image-builder/test-containerdisk.sh
```diff
@@ -0,0 +1,67 @@
...
+)"
+
+SCRIPT_DIR="$(
+  cd "$(dirname "$BASH_SOURCE")"
+  pwd
+)"
+
+function cleanup() {
+    echo "[test-containerdisk] cleanup"
+    if [ -z $DEBUG ]; then
```
EdDev, line 16: "DEBUG" merits an explanation in the commit or here.

cluster-provision/images/vm-image-builder/build-containerdisk.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+set -exuo pipefail
+
+SCRIPT_DIR="$(
```
EdDev, line 4: Not sure what you are doing here, but sub-shelling is wasteful and looks strange.
Maybe this will help:
Getting the current folder (of where the script exists): `EXEC_PATH=$(dirname "$(realpath "$0")")`
Getting one folder above: `PARENT_PATH="$(dirname $EXEC_PATH)"`
Getting a folder under the current one: `FOO_PATH="$PWD/foo"`

EdDev:
> vm-image-builder: add test example

This one fetched kubevirt and tests against it? That cannot happen under kubevirtci.
Why would it even depend in kubevirt or kubevirtci? It needs to depend on the minimum, e.g. run it using "virsh" and check it is working (maybe create an artifact and check it is there). There is no need to raise a cluster to confirm the image is good.

cluster-provision/images/vm-image-builder/build-containerdisk.sh
ormergi: Sorry I mislead you, `IMAGE` should be exported because we use `envsubst` to replace the placeholder with the image file path

cluster-provision/images/vm-image-builder/test-containerdisk.sh
ormergi: I will add it to the commit message
Basically DEBUG lets you keep the cluster created by this script, the VMI with the customized image and the customized image tar file,
So you could debug the image/VM if there was an error during the script invocation, for example:
 if the VMI that created with the new image wont be ready, you could debug what happen to the VMI and console to the VMI and check.

cluster-provision/images/vm-image-builder/build-containerdisk.sh
ormergi: I will simplify it in following commits.

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: I want to delete  NEW_IMAGE only if there was an error during the script invocation
this is why I added the condition in the first place, it should be moved to line 6
so `|| true` will not interfere with `$?`

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: How about `SOURCE_IMAGE` and `CUSTOMIZED_IMAGE`?

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: I will remove it in following commits

cluster-provision/images/vm-image-builder/build-containerdisk.sh
ormergi: You could create few container-disk images and test them instead of overriding the default `latest` tag, for example: 
creating container-disk with the name "fedora" and the tag will be "32"
then you could test this image by creating a VM and use it:
on VM yaml file:
```...
  image: localhost:<port>/fedora:32
...```
You could create another version of the image without loosing `fedora:32` image, or changing its tag manually
let say we created another version with the tag `fedora:tests`, you could use it in a VM like so:
```...
  image: localhost:<port>/fedora:tests
```

ormergi:
> I do not understand the first commit.
> The fact that you fork the original repo should not be expressed in the code itself, not on the commit.
> 
> I do not think we can have such a dependency or to maintain it here.
> I guess the original repo can be kept in sync manually (or with such scripts) until we will decide to extract this code to its own repo (at which time we can just move your original repo into kubevirt).

I will remove this commit

ormergi:
> vm-image-builder: publish script:
> 
> I think this script does not deserve existence. Whoever needs to do this can call docker with the correct commands, we do not need to wrap these two lines.
> 
> Or do I miss something else here?

I will remove it and add what it does to create-push script

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: After few tests apparently this flag is necessary for cloud-init to work properly with virt-install
I will drop `export` and document this in the commit and readme

ormergi:
@EdDev I addressed most of the changes requested in new commits and updated the commit messages

cluster-provision/images/vm-image-builder/create-containerdisk.sh
```diff
@@ -0,0 +1,60 @@
...
+export TAG=${TAG:-32}
+export OS_VARIANT=${OS_VARIANT:-fedora31}
+export CLOUD_CONFIG_PATH=${CLOUD_CONFIG_PATH:-"${SCRIPT_PATH}/example/cloud-config"}
+export VM_IMAGE_URL=${VM_IMAGE_URL:-$(cat "${SCRIPT_PATH}/example/image-url")}
+
+readonly VM_IMAGE="source-image.qcow2"
+readonly build_directory="${IMAGE_NAME}_build"
+readonly new_vm_image_name="provisioned-image.qcow2"
+
+trap 'cleanup' EXIT SIGINT
```
oshoval, line 44: just Exit is needed since it will run all the time

another thing is that -e might wont catch errors in functions according what read and saw once,
i would suggest to inject an error for example if VM_IMAGE doesnt exists 
and see if the cleanup is triggered
if it doesnt 
the -e should be replaced to -E and then it should catch it

ormergi: I will remove `SIGINT`

If VM_IMAGE is removed, since `-u` flag is up the script stops with `unbound variable` error
If `-u` is removed as well `cleanup` still invokes
I am not familiar with `-E` what it does?

oshoval: dont remove the var, remove the file the var point to.
trap might not catch errors that happen in functions even there is -e
-E should fix it if it indeed happen when u inject an error

ormergi: `VM_IMAGE` is path to downloaded image, if you set it to bad path it still works and cleanup is invoked
I saw in man page that `-E` works with trap that catch `ERR` interrupts, I dont think we need it in this script

oshoval: question is if the first error that happened was in a function, 
can you just add please `false` in a function and make sure it still runs indeed?

cluster-provision/images/vm-image-builder/fedora-extended/cloud-config
```diff
@@ -0,0 +1,26 @@
...
+    - path: /etc/modules-load.d/igb.conf
+      content: |
+        igb
+runcmd:
+  - sudo dnf install -y kernel-modules-$(uname -r) qemu-guest-agent
+  - sudo dnf clean all
+  - sudo hostnamectl set-hostname ""
+  - sudo hostnamectl set-hostname "" --transient
+  - sudo systemctl start qemu-guest-agent
+  - sudo shutdown -h now
```
oshoval, line 26: nit: newline

cluster-provision/images/vm-image-builder/fedora-extended/image-url
```diff
@@ -0,0 +1 @@
+https://download.fedoraproject.org/pub/fedora/linux/releases/32/Cloud/x86_64/images/Fedora-Cloud-Base-32-1.6.x86_64.qcow2
```
oshoval, line 1: nit: newline

cluster-provision/images/vm-image-builder/example/image-url
```diff
@@ -0,0 +1 @@
+https://download.fedoraproject.org/pub/fedora/linux/releases/32/Cloud/x86_64/images/Fedora-Cloud-Base-32-1.6.x86_64.qcow2
```
oshoval, line 1: nit: newline

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -0,0 +1,53 @@
...
+echo "Customize image by booting a VM with
+ the image and cloud-init disk
+ press ctrl+] to exit"
+virt-install \
+  --memory 2048 \
+  --vcpus 2 \
+  --name $DOMAIN_NAME \
+  --disk "$SOURCE_IMAGE_PATH",device=disk \
+  --disk "$CLOUD_INIT_ISO",device=cdrom \
+  --os-type Linux \
```
oshoval, line 36: some os have a more specific os-type package, so maybe worth to parameterize it

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -0,0 +1,53 @@
...
+
+SOURCE_IMAGE_PATH=$1
+OS_VARIANT=$2
+CUSTOMIZE_IMAGE_PATH=$3
+CLOUD_CONFIG_PATH=$4
+
+readonly DOMAIN_NAME="provision-vm"
+readonly CLOUD_INIT_ISO="cloudinit.iso"
+
+trap 'cleanup' EXIT SIGINT
```
oshoval, line 22: remove here plz SIGINT as well

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+set -exuo pipefail
+
+function cleanup() {
+  if [ $? -ne 0 ]; then
+    rm -f "$CUSTOMIZE_IMAGE_PATH"
+  fi
+
+  rm -rf "$CLOUD_INIT_ISO"
+  virsh destroy "$DOMAIN_NAME" || true
```
oshoval, line 10: all over, please keep one format ie
"${VAR}" and not just double quote, since you did use with `"{` in some places and indeed its the cleanest

cluster-provision/images/vm-image-builder/create-containerdisk.sh
```diff
@@ -0,0 +1,60 @@
...
+
+pushd "${SCRIPT_PATH}"
```
oshoval, line 46: please use '(' instead of the pushd + popd

cluster-provision/images/vm-image-builder/create-containerdisk.sh
```diff
@@ -0,0 +1,60 @@
...
+  cleanup
+
+  if ! [ -e "$VM_IMAGE" ]; then
+    # Download base VM image
+    curl -L $VM_IMAGE_URL -o $VM_IMAGE
+  fi
+
+  mkdir "${build_directory}"
```
oshoval, line 54: mkdir -p , in case its re entrant

oshoval, line 54: -p , in case the previous run was fine

cluster-provision/images/vm-image-builder/build-containerdisk.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+set -exuo pipefail
+
+SCRIPT_PATH=$(dirname "$(realpath "$0")")
+
+function cleanup() {
+  if [ $? -ne 0 ]; then
+    rm -f "$tar_file"
+    docker image rm -f "$full_image_tag"
```
oshoval, line 9: if image doesnt exists , it wont fail ?
if it can fail add || true please

cluster-provision/images/vm-image-builder/README.md
```diff
@@ -0,0 +1,170 @@
...
+## Publish customized container-disk image
+
+Push the new container-disk image to a registry with `publish-containerdisk.sh` script.
+
+This script exports 
+REGISTRY - the registry the image will be stored at
+REPOSITORY - image repository
+
+This script expects: 
+IAMGE_NAME - container disk image name
```
oshoval, line 150: typo

cluster-provision/images/vm-image-builder/README.md
```diff
@@ -0,0 +1,170 @@
...
+$ image_name='example-fedora'
+$ tag='tests'
+$ vm_image='customized-image.qcow2'
+
+$ ./build-containerdisk.sh $image_name $tag $vm_image
+$ Container image saved as tar file at: ./fedora-32.tar
+
+# Generated image .tar file
+$ ls "example-fedora_build/example-fedora-tests.tar"
+
```
oshoval, line 138: space

cluster-provision/images/vm-image-builder/README.md
```diff
@@ -0,0 +1,170 @@
...
+## Build container-disk image
+
+To build container-disk image form qcow2 image file, use `build-containerdisk.sh` script.
+
+This script converts qcow2 image to container-disk image that can be consumed by KubeVirt VM's.
+
+What this script does:
+- Creates temp directory and copy the source
+  VM image and '
```
oshoval, line 112: `and '` ?

cluster-provision/images/vm-image-builder/README.md
```diff
@@ -0,0 +1,170 @@
...
+$ VMI_YAML="vmi-test-image.yaml"
+$ envsubst < "$VMI_YAML" | kubectl apply -f -
+
+$ kubectl wait --for=condition=AgentConnected vmi $VMI_NAME --timeout 5m
+$ virtctl console testvm1
+...
+
+# You could send / store the image .tar file: 
+$ ls example_build/example-tests.tar
+
```
oshoval, line 101: space

cluster-provision/images/vm-image-builder/README.md
```diff
@@ -0,0 +1,170 @@
...
+- libvirt
+- libguestfs
+- qemu-img
+
+## How to use:
+```bash
+
+# Create kubernetes cluster with Kubvirt
+git clone https://github.com/kubevirt/kubevirt --depth=1 --branch=master --single-branch kubevirt
+$ pushd kubevirt
```
oshoval, line 61: add ``` please

cluster-provision/images/vm-image-builder/README.md
```diff
@@ -0,0 +1,170 @@
...
+This script requires the following packages:
+- cloud-utils
+- docker-ce
+- libvirt
+- libguestfs
+- qemu-img
+
+## How to use:
+```bash
+
```
oshoval, line 58: space

cluster-provision/images/vm-image-builder/create-containerdisk.sh
ormergi: I checked and it works

cluster-provision/images/vm-image-builder/create-containerdisk.sh
ormergi: I did not added `-p` so if a user run it again it wont delete the qcow2 image that was already created
What you think?

cluster-provision/images/vm-image-builder/build-containerdisk.sh
ormergi: Since `-f` is mentioned `echo $?` returns `0` so we good

cluster-provision/images/vm-image-builder/create-containerdisk.sh
oshoval: but wont it recreate it on the same name ?
i would then fail fast if the folder exists, if the case is what is above, or just `rm -rf` the folder if the above answer is yes

cluster-provision/images/vm-image-builder/create-containerdisk.sh
oshoval: thx

cluster-provision/images/vm-image-builder/create-containerdisk.sh
ormergi: Using `( ` `)` cuses all command inside to run in a subshell and it is not necessay in this script
Also I prefer using `pushd` and `popd` because it is more verbose commands

oshoval: i dont see any reason to not use a sub shell,
its more nature here to use '(' without the push pop, and the ident will be more natural, now its not natural imo.
lets see what other reviewers think please.

cluster-provision/images/vm-image-builder/fedora-extended/cloud-config
ormergi: Done

cluster-provision/images/vm-image-builder/fedora-extended/image-url
ormergi: Done

cluster-provision/images/vm-image-builder/example/image-url
ormergi: Done

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: Done

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: Done
All scripts aligned to the format `"${VAR}"`

cluster-provision/images/vm-image-builder/create-containerdisk.sh
ormergi: By verbose I meant to the commands name, so instead of `(` we use `pushd` it make it more readable in my opinion.
If you dont mind lets wait for other reviewers

cluster-provision/images/vm-image-builder/README.md
ormergi: Fixed

cluster-provision/images/vm-image-builder/README.md
ormergi: Done

cluster-provision/images/vm-image-builder/README.md
ormergi: Fixed
I also arranged this file and removed redundant explanations

cluster-provision/images/vm-image-builder/README.md
ormergi: Done

cluster-provision/images/vm-image-builder/README.md
ormergi: Done

cluster-provision/images/vm-image-builder/create-containerdisk.sh
oshoval: First time i had this dilemma i also thought like that, but then was asked to use `(`
and got used to it, we are adaptive, and now its self explanatory, i believe we should stick to a common practice,
it will save us precious time.

ormergi:
Changes:
- Resolved all nits, formats and typos at README file, also rearranged it and rephrased few explanations
- Changed the format of using variables in all scripts to `"${VAR}"`
- Renamed `fedora-extended` to `fedora-sriov-lane` to describe better the propose of the files
- Changed `sudo shutdown -h now` to `sudo shutdown` in `example/cloug-config` `fedora-sriov-lane/cloud-config` files so cloud-init will finish its operation with no errors.
Using `shutdown` w/o `-h now` will shutdown a host after one minute

cluster-provision/images/vm-image-builder/create-containerdisk.sh
ormergi: It is like you described, if you already created an image next time you run this script you will fail with an error saying directory already exists

cluster-provision/images/vm-image-builder/create-containerdisk.sh
ormergi: BTW I tried it and since using `( )` creating subshell, it is necessary to catch `ERR` interupts with `trap` and adding `-E` flag in order to invoke `cleanup` on errors
Maybe this explains why you saw scripts with `-e` and `trap` that wont catch errors

cluster-provision/images/vm-image-builder/create-containerdisk.sh
ormergi: Sorry for the noise, at first the idea was to keep files if something went wrong but it had no use
so i changed to delete file that were created in each phase is there is an error

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: I checked it again and `SIGINT` is necessary to invoke `cleanup` if the user interupt the script with CTRL+C

oshoval: if so, consider please SIGQUIT as well
like ctrl c but with core dump, we dont use it most of the time, but its a good practice i believe to have a full picture

cluster-provision/images/vm-image-builder/create-containerdisk.sh
oshoval: thx, ok lets then leave it as you did, just i am not sure the ident there is right to use.

cluster-provision/images/vm-image-builder/create-containerdisk.sh
oshoval: i saw before that it clean just if error occour, but what if no error ?
i think that even if you choose to fail the test , i would prefer please something like
[ -d dir ] && echo "dir exist" && exit 1
(dont remember the exact check in the first one)

cluster-provision/images/vm-image-builder/OWNERS
```diff
@@ -0,0 +1,6 @@
+filters:
+  ".*":
+    reviewers:
+      - ormergi
```
oshoval, line 4: i think we should use something like network-reviewers
(on both lines, as other OWNER files are doing please)

oshoval (changes requested):

cluster-provision/images/vm-image-builder/OWNERS
EdDev: This adds ormergi to the reviewers and approvers of this specific folder. I do not see why it is worth creating a special group for this at the moment, especially when only one is included in it.

oshoval: no problem, just saying what i saw on other files in some kubevirt repos, and on other habits from the past.

EdDev (commented):
Commit "vm-image-builder: build script": typo, `Dacording`.

cluster-provision/images/vm-image-builder/build-containerdisk.sh
```diff
@@ -0,0 +1,37 @@
...
+
+template_dockerfile="${SCRIPT_PATH}/Dockerfile.containerdisk"
+
+parent_dir=$(dirname "$VM_IMAGE_FILE")
+tar_file="${parent_dir}/${IMAGE_NAME}-${TAG}.tar"
+full_image_tag="${IMAGE_NAME}:${TAG}"
+
+trap 'cleanup' EXIT
+
+temp_dir=$(mktemp -d -p /tmp -t build.container-disk.XXXX)
```
EdDev, line 27: Who cleans the temp directory at the end?

EdDev (commented):
Commit "vm-image-builder: Automate VM customization flow": Typo, `containes`.

cluster-provision/images/vm-image-builder/build-containerdisk.sh
oshoval: there is a trap but look on 
https://github.com/kubevirt/kubevirt/pull/4115#discussion_r484034330
bottom line we should return SIGTERM

cluster-provision/images/vm-image-builder/README.md
```diff
@@ -0,0 +1,162 @@
...
+Notes:
+- If no variables passed, the script will create container-disk image according to the files at `example/`.
+- Currently, image customization process done by spinning up live VM with cloud-init disk to provision the image with the changes described by the cloud-config file.
+- It is necessary to add `sudo shutdown` at the end of the cloud-config`runcmd` block, so the script will not wait for user input.
+- You can use your own image customizing script by exporting the path to `CUSTOMIZE_IMAGE_SCRIPT`.
+
+This script requires the following packages:
+- cloud-utils
+- docker-ce
```
EdDev, line 49: You should consider changing to podman in a following change.

cluster-provision/images/vm-image-builder/create-containerdisk.sh
```diff
@@ -0,0 +1,60 @@
...
+export TAG=${TAG:-32}
+export OS_VARIANT=${OS_VARIANT:-fedora31}
+export CLOUD_CONFIG_PATH=${CLOUD_CONFIG_PATH:-"${SCRIPT_PATH}/example/cloud-config"}
+export VM_IMAGE_URL=${VM_IMAGE_URL:-$(cat "${SCRIPT_PATH}/example/image-url")}
+
+readonly VM_IMAGE="source-image.qcow2"
+readonly build_directory="${IMAGE_NAME}_build"
+readonly new_vm_image_name="provisioned-image.qcow2"
+
+trap 'cleanup' EXIT
```
oshoval, line 44: you said you need SIGINT right ?
also please consider adding SIGTERM, see
https://github.com/kubevirt/kubevirt/pull/4115#discussion_r484034330

EdDev: But we use `bash` here, not anything else.

oshoval: wouldnt it might run on prow one day ?
better be safe and robust i think, in any case

EdDev: It will probably run in a container and there is little chance we will not be fedora/centos base.

cluster-provision/images/vm-image-builder/build-containerdisk.sh
EdDev: ohh I missed line 12.
SIGTERM is discusses in a different thread, no need to mention it here.

cluster-provision/images/vm-image-builder/create-containerdisk.sh
oshoval: imho until it has one click to run in a container i would suggest to be robust,
the change which is needed is minor and harmless from my point of view

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -0,0 +1,53 @@
...
+
+SOURCE_IMAGE_PATH=$1
+OS_VARIANT=$2
+CUSTOMIZE_IMAGE_PATH=$3
+CLOUD_CONFIG_PATH=$4
+
+readonly DOMAIN_NAME="provision-vm"
+readonly CLOUD_INIT_ISO="cloudinit.iso"
+
+trap 'cleanup' EXIT
```
oshoval, line 22: there are multi traps, please see the discussion about SIGTERM and SIGINT 
(we will continue discussion on the other threads)

(wont write it on each one, please consider on each trap occurence)

cluster-provision/images/vm-image-builder/create-containerdisk.sh
EdDev: @oshoval , there is no robustness here, the script is limited to bash and it says so on the first line of the file.

cluster-provision/images/vm-image-builder/README.md
ormergi: podman can be used as well I will add it to README

ormergi: I will test it and consider a following PR

cluster-provision/images/vm-image-builder/create-containerdisk.sh
ormergi: If no error occurs and the `...build` directory exists it will report that the directory exists and delete it, then on the second run it will run normally

cluster-provision/images/vm-image-builder/create-containerdisk.sh
ormergi: No need for `SIGINT`/`SIGTERM` for this PR scripts, I checked and it works as expected with `EXIT`
See this [disscution](https://github.com/kubevirt/kubevirtci/pull/428#discussion_r476229164)

oshoval: https://github.com/kubevirt/kubevirtci/pull/428#discussion_r476598843
you said the SIGINT is needed, if you double checked and its not needed, then remove but please update if possible so we know the behaviour.

SIGTERM is not needed because you have shebang of bash as Edy said.

ormergi: Alright I will add it , it was two weeks ago and all those threads are confusing :confused:

ormergi:
Changes:
- fixed typos in commit messages
- added `SIGINT` where there is `trap` usage

EdDev:
> Changes:
> 
>     * fixed typos in commit messages
> 
>     * added `SIGINT` where there is `trap` usage

You have both rebased and added the changes, making it very hard now to see what changed.
Please do not mix both when pushing.

EdDev (approved):
Thank you, looks great.

ormergi:
> > Changes:
> > ```
> > * fixed typos in commit messages
> > 
> > * added `SIGINT` where there is `trap` usage
> > ```
> 
> You have both rebased and added the changes, making it very hard now to see what changed.
> Please do not mix both when pushing.

Would rebase first then add changes be easier to review?

EdDev:
> Would rebase first then add changes be easier to review?

The order does not matter, only that fact that two steps exist:
- rebase, push
- add changes, push
(or vice versa)

PR #429, opened 2020-08-05
Add deploy kubevirt script

ormergi:
This PR add simple script to deploy latest stable release of Kubevirt.

This will be used by Kubevirt component projects such as 
`cluster-network-addons-operator`, `kubemacpool` or any other 
project that uses kubevirtci.


oshoval:
can we extend `cluster-deploy` target to use a specific remote version instead creating an additional script ?

PR #430, opened 2020-08-06
Cargo cult NAT forwarding for IPv6

maya-r:
I wanted to ssh.sh node01 and be able to ping6 google.com. I couldn't.
Running (some) of these commands locally allowed me do so.

---
Question: why are we using NAT at all? wouldn't it be enough to bridge tap01 and eth0?

### DIFF
--- a/cluster-provision/centos8/scripts/dnsmasq.sh
+++ b/cluster-provision/centos8/scripts/dnsmasq.sh
@@ -35,5 +35,8 @@ done
 iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
 iptables -A FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
 iptables -A FORWARD -i br0 -o eth0 -j ACCEPT
+ip6tables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
+ip6tables -A FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
+ip6tables -A FORWARD -i br0 -o eth0 -j ACCEPT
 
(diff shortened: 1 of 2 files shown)


maya-r:
after I add to 'kubectl edit ippools default-ipv6-ippool':
````yaml
spec:
  natOutgoing: true
````

my pods can ping the outer world

phoracek:
@maya-r do we need to do any changes in iptables then?

maya-r:
the ip6tables additions were necessary for the virtual machines to be able to reach the outer world, the ippools change was for the pods within the cluster within the virtual machines to reach the outer world

phoracek:
Understood. Would it make sense to apply the change to ippools in kubevirtci directly? So then IPv6 works out of the box?

And this maybe a stupid question, but what is "Cargo cult"?

maya-r:
Cargo cult: mimic without understanding why something is working. Although I understand things better now :)

It's an [idiom coined by Richard Feynman](https://en.wikipedia.org/wiki/Cargo_cult_science#Feynman's_speech) in reference to people in remote island communities associating religious attributes to cargo landing from planes during world war 2, and attempting to cause it to happen again by building things resembling airport runways.

phoracek:
Oh dear :D Gotcha. Adding it to my bag of "skills", next to shotgun debugging.

maya-r:
I'll try to add the ippool too.

oshoval:
> /test check-provision-k8s-1.18

wont work
because you need to provide centos since 
cluster-provision/centos8/scripts/dnsmasq.sh
is changed, and then to update hash and do this (unless this flow was changed)

phoracek:
I'm following up on this with https://github.com/kubevirt/kubevirtci/pull/432. We can close this one. Thanks a lot @maya-r

PR #431, opened 2020-08-08
WIP reverse service order

oshoval:


oshoval:
doesnt work out of the box, calico and core dns stuck
i guess need to add ipv6 as adv address, if we decide we want it, we can debug and investigate it

oshoval:
doesnt work out of the box, calico and core dns stuck
i guess need to add ipv6 as adv address, if we decide we want it, we can debug and investigate it

oshoval:
doesnt work out of the box, calico and core dns stuck
i guess need to add ipv6 as adv address, if we decide we want it, we can debug and investigate it

PR #432, opened 2020-08-13
NAT cluster's IPv6 network to the outside

phoracek:
The first commit enables IPv6 NAT between cluster hosts and the outside network.

The following commit configures Calico CNI to NAT from the pod network to the outside.

Kudos to @maya-r for figuring out how to configure all mentioned above.


oshoval:
> /test push-k8s-1.18

you changed dns, so you need to provide please first centos, unless the flow was changed.

phoracek:
Gotcha

phoracek:
TODO: Still have to enable IPv6 forwarding in dnsmasq container

oshoval:
Sw33t
/lgtm

phoracek:
/hold cancel

Thanks!

cluster-provision/k8s/1.18/provision.sh
```diff
@@ -87,7 +87,7 @@ cat << EOF > /etc/docker/daemon.json
   "log-driver": "json-file",
   "exec-opts": ["native.cgroupdriver=systemd"],
   "ipv6": true,
-  "fixed-cidr-v6": "2001:db8:1::/64",
+  "fixed-cidr-v6": "2001:db9:1::/64",
```
oshoval, line 90: if we have 2 nodes. do we need to change it for the 2nd node ?
(after cluster-up of course)

phoracek: It should be "db9" on all the nodes. Have I missed a place to change?

oshoval: you didn't miss, just in case it was need to be different, then other files would need to be updated.

PR #436, opened 2020-09-03
Change the pre-provisioned path on the K8s nodes

danielBelenky:
This is done in order to enable testing on CoreOS, where the root path
is read only.

Signed-off-by: Daniel Belenky <dbelenky@redhat.com>


enp0s3 (approved):
/lgtm

PR #440, opened 2020-09-07
GoCli: Expose ports 80 and 443

gabrielecerami:
When launching the cluster to be used as a testing environment it is
sometimes necessary to access the services inside the cluster through an
ingress configuration.
This change allows gocli to expose http and https and redirect any
request to node01 which, if configured as ingress controller, allow
access to the services in the cluster.


rmohr:
/lgtm
/approve
/hold

@gabrielecerami it is working for you already, right? What configuration did you do to make node01 ready to act as an ingress machine?

We need to upload the updated binary before merging. Is the PR ready from your side?

gabrielecerami:
> /lgtm
> /approve
> /hold
> 
> @gabrielecerami it is working for you already, right? What configuration did you do to make node01 ready to act as an ingress machine?

The main configuration is in these two parts
https://github.com/gabrielecerami/project-infra/blob/prow-deploy/github/ci/prow-deploy/templates/manifests/ingress.yaml.j2
https://github.com/gabrielecerami/project-infra/blob/prow-deploy/github/ci/prow-deploy/molecule/default/prepare.yml#L43-L56
 
> We need to upload the updated binary before merging. Is the PR ready from your side?

Yes, it is working for me when I run it locally. Of course since the image is not updated, I have to jump on the dnsmasq container and add the two iptables rules manually. After that it works correctly.

Thanks,

cluster-provision/cli/cli
```diff
@@ -76,6 +78,18 @@ if [ "$COMMAND" = "port" ] ; then
     docker port $dnsmasq_container 5000 | cut -d':' -f2
     exit
   fi
+  if [ "$1" = "vnc" ]; then
```
dhiller, line 81: So `cli port vnc` has never worked, right? Nice catch!

dhiller (commented):
/lgtm

So IIUC correctly what we have to do to incorporate the changes is this:
* update the centos8 image -> `push-centos8`
* update the `base` file with new sha in the k8s providers (I suggest only updating >= 1.17)
* push the k8s providers (same here)
* update the image.json with new shas for k8s provider images
* build gocli -> `push-gocli`

@qinqon did I forget anything?

rmohr:
> * update the `base` file with new sha in the k8s providers (I suggest only updating >= 1.17)

Sounds good.

qinqon (commented):
/lgtm

qinqon:
@dhiller that's exactly it, PR looks fine let's wait for digests to land and we approve.

dhiller:
> /test push-centos8

@gabrielecerami FYI

dhiller (approved):
/lgtm
/hold cancel

jean-edouard:
I didn't realize this PR meant ports 80 and 443 had to be free on the build machine!
That will probably bother more than one dev, since it's pretty common to run web servers on build machines...
This requirement (ports 80 and 443 must be free) should at least be highlighted in developer documentation.

dhiller:
> I didn't realize this PR meant ports 80 and 443 had to be free on the build machine!
> That will probably bother more than one dev, since it's pretty common to run web servers on build machines...
> This requirement (ports 80 and 443 must be free) should at least be highlighted in developer documentation.

Good point! @gabrielecerami do you see a chance to make this into an opt-in mechanism somehow so that we can enable this if needed?

jean-edouard:
More info: I just built a 1.19 lane with the latest kubevirtci and tried to use it for a `make cluster-up` on the latest kubevirt and got:
```
Error response from daemon: driver failed programming external connectivity on endpoint k8s-1.19-dnsmasq (6742e9aedb974761889f0f34e9162269eb73c7c5a333d59faf10f1350e26f366): Error starting userland proxy: listen tcp 127.0.0.1:80: bind: address already in use
```
Stopping Apache on the machine fixed the issue.

gabrielecerami:
Hi,
the cause is clear, kubevirtci now uses standard ports. Unfortunately usability bugs surface only then users are hit.
Regarding an opt-in/opt-out system, it may be possible.
The implementation is quite transparent, there is no hardcoding of ports in the go code, but there's a bit of
hard coding in the bash part.
i.e. here
https://github.com/kubevirt/kubevirtci/pull/440/files#diff-3be8fc8f7b90e2b2674108eb553d4045R137
and here
https://github.com/kubevirt/kubevirtci/pull/440/files#diff-2cf0053fb2f864065221202cc5a0faecR28
is where the ports are activated. If we remove the ports from those lines, the dnsmasq container will not listen to those ports.
Making the exposure of those ports optional, requires a few adjustments.
For all the ports, there are two states: fixed binding or randomized bindings, there is currently no option to "turn off" any port. So
we may need to add two options to explicitly add the to the base port configuration.

dhiller:
We might just use some other port for connecting from outside instead of the defaults in case the arguments are not used?
I.e 8080:80 8043:443 ?

dhiller:
I mean the args are here, right? https://github.com/kubevirt/kubevirtci/blob/b87ef18cd10ed8e98da58c8b0f92620e34407481/cluster-provision/cli/cli#L49

dhiller:
In the KubeVirt community meeting it was suggested to expose thrt ports using `-P` instead of `-p`, see https://docs.google.com/document/d/1kyhpWlEPzZtQJSjJlAqhPcn3t0Mt_o0amhpuNPGs1Ls/edit#heading=h.2xsziwp3d9j for reference.

rmohr:
Yes, we normally use for all ports random ports if none are specified. There is then a `gocli ports` command to look them up. They should definitely not use these host ports by default.

gabrielecerami:
Ok, I'll revert the line at https://github.com/kubevirt/kubevirtci/pull/440/files#diff-2cf0053fb2f864065221202cc5a0faecR28, then I'll change the tests to lookup the correct ports for http tests.

dhiller:
Should be addressed with https://github.com/kubevirt/kubevirtci/pull/451

PR #443, opened 2020-09-11
Disable endpointslices

qinqon:
Disable endpointslices it's activated by default at k8s-1.19
    
At k8s-1.17 endpointslices was introduced with as an opt-in but at k8s-1.19
is opt-out with a feture gate a kube-proxy, what we observe at kubevirt is that
virt-operator is re-create services very fast so the endpointslices does not get
re-created by the k8s controller, it's possible to reproduces this with at simple
http pod server and service.
    
k8s issue: https://github.com/kubernetes/kubernetes/issues/94720


qinqon:
Looks like endpointslices may have some raices and that affects at kubevirt https://github.com/kubernetes/kubernetes/issues/94720

qinqon:
/cc @dhiller 
This could be it.

dhiller (commented):
/lgtm
/approve
/hold let's wait what the job is saying

dhiller (approved):

PR #445, opened 2020-09-14
dnsmasq: Mount /lib/modules to use ip6tables

qinqon:
After [1] ip6tables is used at dnsmasq.sh script this command
depends on a some kernel modules to be installed so
ip6tables try to load them but since the container has not
mounted the /lib/modules dir it cannot load them.


maiqueb:
You're missing the link in the PR's description.

oshoval:
Nice
but please consider it doesn't solve provision, the dnsmasq container there is triggered from bash script,
not the go cli, if i remember right.

oshoval (changes requested):
Nice
but please consider it doesn't solve provision, the dnsmasq container there is triggered from bash script,
not the go cli, if i remember right.

qinqon:
@oshoval right, let me add it to the `cli/cli`

oshoval:
/lgtm
(just fill the [1] as Miguel said please)

oshoval (approved):

PR #446, opened 2020-09-17
Mount /lib/modules at gocli if exists

qinqon:
Right now it tries always to mount /lib/modules at gocli but this brings problems
with prow jobs since they are not mounted on the job's pod, let's make this optional,
normal users will use scripts here and we mount /lib/modules at those scripts.

### DIFF
--- a/cluster-provision/cli/cli
+++ b/cluster-provision/cli/cli
@@ -173,7 +173,13 @@ if [ "$COMMAND" = "update" ] ; then
   exit 0
 fi
 
-DNSMASQ_CID=$(docker run -d ${PORTS} -v /lib/modules:/lib/modules -e NUM_NODES=${NODES} --name ${PREFIX}dnsmasq --privileged ${BASE} /bin/bash -c /dnsmasq.sh)
+# gocli will try to mount /lib/modules to make it accessible to dnsmasq in
+# in case it exists
+if [ -d /lib/modules ]; then
+        dnsmasq_mounts="-v /lib/modules/:/lib/modules/"
+fi
+
+DNSMASQ_CID=$(docker run -d ${PORTS} ${dnsmasq_mounts} -e NUM_NODES=${NODES} --name ${PREFIX}dnsmasq --privileged ${BASE} /bin/bash -c /dnsmasq.sh)
 CONTAINERS=${DNSMASQ_CID}
 
 if [ "$COMMAND" == "provision" ] ; then
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -194,6 +194,21 @@ func run(cmd *cobra.Command, args []string) (err error) {
 		}
 	}
 
+	// Mount /lib/modules at dnsmasq if it's there since sometimes
+	// some kernel modules may be mounted
+	dnsmasqMounts := []mount.Mount{}
+	_, err = os.Stat("/lib/modules")
+	if err == nil {
+		dnsmasqMounts = []mount.Mount{
+			{
+				Type:   mount.TypeBind,
+				Source: "/lib/modules",
+				Target: "/lib/modules",
+			},
+		}
+
+	}
+
 	// Start dnsmasq
 	dnsmasq, err := cli.ContainerCreate(ctx, &container.Config{
 		Image: cluster,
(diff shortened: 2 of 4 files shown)


cluster-provision/gocli/cmd/run.go
```diff
@@ -194,6 +194,21 @@ func run(cmd *cobra.Command, args []string) (err error) {
 		}
 	}
 
+	// Mount /lib/modules at dnsmasq if it's there since sometimes
+	// some kernel modules may be mounted
+	dnsmasqMounts := []mount.Mount{}
+	_, err = os.Stat("/lib/modules")
```
oshoval, line 200: you can combine to one liner
if .... ; err == nil {

oshoval (commented):
what about cli/cli bash ? (since its also have this -v)

rmohr:
/approve

The approach works for me. Did not review the code in detail, therefore no lgtm.

cluster-provision/gocli/cmd/run.go
qinqon: I know I can but I am not going to do it, every line should have as less responsibility as possible.

oshoval: i honestly dont understand why, in this specific situation when we just need both for one thing, but nvm.

qinqon:
@oshoval cli/cli is done already by previous PR.

oshoval:
> @oshoval cli/cli is done already by previous PR.

i know, i mean do you want to protect it as well ?

qinqon:
@oshoval Ahh wait you right the condition ! thanks, let me prepare it.

qinqon:
Pushed conditional check at `cli/cli` too, thanks @oshoval.

cluster-provision/gocli/cmd/run.go
qinqon: Error handling is one thing and opening a file/dir is other, so they belong to different lines as specified at effective golang, I am going to resolve this.

cluster-provision/cli/cli
```diff
@@ -173,7 +173,13 @@ if [ "$COMMAND" = "update" ] ; then
   exit 0
 fi
 
-DNSMASQ_CID=$(docker run -d ${PORTS} -v /lib/modules:/lib/modules -e NUM_NODES=${NODES} --name ${PREFIX}dnsmasq --privileged ${BASE} /bin/bash -c /dnsmasq.sh)
+# gocli will try to mount /lib/modules to make it accessible to dnsmasq in
+# in case it exists
+if [ -f /lib/modules ]; then
+        dnsmasq_mounts="-v /lib/modules/:/lib/modules/"
```
oshoval, line 179: i would check if ip6tables command can be executed (which means the module already installed), and if it doesnt i will exit 1
because else the provision would failed on the errors that dnsmasq cant execture the ip6tables commands

qinqon:
Third would be the one :-)

/test check-provision-k8s-1.16
/test check-provision-k8s-1.17
/test check-provision-k8s-1.19

qinqon:
Another nit

/test check-provision-k8s-1.16
/test check-provision-k8s-1.17
/test check-provision-k8s-1.19

cluster-provision/cli/cli
qinqon: In case /lib/modules is not there it's suppose to be already loaded (that's the case for CI) we can document that but that's exactly what we want for CI since CI is not going to mount /lib/modules as this is a security issue.

qinqon:
Checked mounting dummy module on dnsmasq container

```bash
[ellorent@localhost kubevirtci]$ docker exec -it k8s-1.18-dnsmasq /bin/bash
[root@fac09a4a94d0 /]# lsmod |grep dummy
[root@fac09a4a94d0 /]# modprobe dummy
[root@fac09a4a94d0 /]# lsmod |grep dummy
dummy                  16384  0

```

cluster-up/up.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -xe
```
oshoval, line 3: intentional to stay?
or a debug

qinqon: debug I have to remove, need not to hurry things up.

oshoval:
/lgtm
/hold

remove the hold please when all the links are passed, because else it wont wait for them.

PR #447, opened 2020-09-21
Emulate krew integration for virtctl.sh in our dev environment

rmohr:
virtctl is available as krew plugin. In order to train the muscle memory of people the right way, emulate the krew invocation.

The following will work now:

```
cluster-up/kubectl.sh console myvm # just to keep working what was there before
cluster-up/kubectl.sh virt console myvm # like with krew
```


dhiller:
Cool idea! Thanks for that!

/lgtm
/approve
/hold for others who might be interested

dhiller:
A little background: https://github.com/kubevirt/kubectl-virt-plugin/

dhiller (approved):

PR #448, opened 2020-09-28
Install iptables at k8s provision for docker-ce providers

qinqon:
Looks like latest docker-ce RPM for centos8 stream is missing the iptables dependency
needed for docker to create rules:

"failed to start daemon: Error initializing network controller:
error obtaining controller instance:
failed to create NAT chain DOCKER: Iptables not found"

This PR install iptables before installing docker-ce


dhiller (approved):
/approve

oshoval:
> /hold I don't think we need to update the image shas here, right?

We done, since the k8s-X/provision.sh were changed,
unless there is a mechanism that does it automatically

PR #449, opened 2020-09-30
Bump k8s-1.19 nm to 1.22.8

qinqon:
Latest centos8 stream installs nm 1.28 we should pin with the version that we know
it's working for us.


phoracek (commented):
/approve

dhiller (approved):
/lgtm

PR #452, opened 2020-10-08
Remove all manual steps from kubevirtci

rmohr:
**Remove all unsupported providers**

* All k8s versions older than 1.17 are gone now
* All openshift versions are gone now

**Use gocli for provisioning**

Remove the bash script which still did the provisioning and further introduce a package concept which allows gocli to automatically discover all parameters based on our already established package structure.

To provision a cluster, this is now sufficient:

```
gocli provision cluster-provision/k8s/1.19
```

The resulting provisioned cluster will be named `kubevirtci/k8s-1.19`.

**Base images are now always built with the cluster**

Remove the need to publish base images. the PR jobs will now also always build the base image. The publish script will build the base images once and they will be used for all clusters which it provisions. The base image will therefore never be explicitly released but still shared by all published  clusters of a commit.

**How to vendor kubevirtci after this change**

There are two main difference to before. 
1. One is now supposed to vendor in tags, created from a postsubmit job. Vendoring a commit directly will not work anymore
1. The tag which was used to checkout the repo needs to be exported in `KUBEVIRTCI_TAG`, so that the checked out code can pull the right `gocli` container.

**Additional notes**

Until https://github.com/kubevirt/project-infra/pull/666 is merged, no longer needed jobs will fail. We only care about:

* check-provision-k8s-1.19
* check-provision-k8s-1.18
* check-provision-k8s-1.17

PR #453, opened 2020-10-12
Make KUBEVIRTCI_TAG requirement explicit

gabrielecerami:
when the KUBEVIRTCI_TAG env variable is missing, cluster-up fails with a
cryptic error on docker command line format.
This cnange makes it explicit that the variable is mandatory and fails
the script with an explanation

### DIFF
--- a/cluster-up/cluster/ephemeral-provider-common.sh
+++ b/cluster-up/cluster/ephemeral-provider-common.sh
@@ -2,6 +2,15 @@
 
 set -e
 
+if [ -z "${KUBEVIRTCI_TAG}" ] && [ -z "${KUBEVIRTCI_GOCLI_CONTAINER}" ]; then
+    echo "FATAL: either KUBEVIRTCI_TAG or KUBEVIRTCI_GOCLI_CONTAINER must be set"
+    exit 1
+fi
+
+if [ -n "${KUBEVIRTCI_TAG}" ] && [ -n "${KUBEVIRTCI_GOCLI_CONTAINER}" ]; then
+    echo "WARNING: KUBEVIRTCI_GOCLI_CONTAINER is set and will take precedence over the also set KUBEVIRTCI_TAG"
+fi
+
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -2,6 +2,11 @@
 
 set -e
 
+if [ -z "${KUBEVIRTCI_TAG}" ]; then
```
rmohr, line 5: I think we should check that either `KUBEVIRTCI_GOCLI_CONTAINER` or `KUBEVIRTCI_TAG` is provided.

Because below we have

```
${KUBEVIRTCI_GOCLI_CONTAINER:-kubevirtci/gocli:${KUBEVIRTCI_TAG}}
```

rmohr (changes requested):
Nice change, one request.

cluster-up/cluster/ephemeral-provider-common.sh
gabrielecerami: Done, with an extra.

rmohr:
Very weird. These checks should be required and should have started. Will investigate that asap.

maya-r:
this means that KUBEVIRTCI_TAG is now required when even running just kubectl.sh

gabrielecerami:
> this means that KUBEVIRTCI_TAG is now required when even running just kubectl.sh

Yes, but it was already required after #452 . This patch just helps users understand what is the prerequisite for running kubectl.sh

rmohr:
> Yes, but it was already required after #452 . This patch just helps users understand what is the prerequisite for running kubectl.sh

Indeed. This is a little bit inconvenient, but it allowed us in a very simple way to  get rid of all these manual push steps. 

In kubevirt we simply write after loading kubevirtci the KUBEVIRTCI_TAG into cluster-up/hack/common.sh and it is present always: https://github.com/kubevirt/kubevirt/blob/master/cluster-up/hack/common.sh.

maya-r:
thanks. looks like mhenriks has done the same.

PR #454, opened 2020-10-13
Use latest gocli if non specified at `make cluster-up`

qinqon:
The Makefile from kubevirtci is only consumed by kubevirtci itself
to test it, this commit use latest KUBEVIRTCI_TAG if not specified.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+
+export KUBEVIRTCI_TAG ?= $(shell curl -L https://storage.googleapis.com/kubevirt-prow/release/kubevirt/kubevirtci/latest)
+
 cluster-up:
 	./cluster-up/check.sh
 	./cluster-up/up.sh


ormergi:
How it used to work until now? how KUBEVIRTCI_TAG was initialized?

rmohr:
> How it used to work until now? how KUBEVIRTCI_TAG was initialized?

We don't use the makefile targets for testing and KUBEVIRTCI_TAG was just introduce last week. I am a little bit uncertain what to do here. I think the Makefile targets are very misleading (I would expect that they only use things which belong to the code base). Maybe we can rework this somehow differently.

qinqon:
I don't like it too much either but I was expecting doing `make cluster-up` to just work here.

rmohr:
> I don't like it too much either but I was expecting `make cluster-up` to just work here.

Ok, will approve it, but we need to think about something here. :/

/lgtm
/approve

PR #455, opened 2020-10-19
Remove unwanted traces from older clusters

rmohr:
The unwanted traces of the old clusters make the postsubmit job fail.


rmohr:
/cc @vladikr 

FYI

rmohr:
Merging to run the postsubmit without these traces of the old cluster.

PR #459, opened 2020-10-25
Fix SRIOV lane: configure CNI MTU according to the CRI bridge MTU on KIND infra 

ormergi:
Since on CI environments the CRI (currently docker) MTU could vary, it is necessary to configure the CNI (currently calico) MTU according to the current host CRI bridge MTU.
This PR checks the host CRI bridge MTU and patches the CNI manifest MTU accordigly.

Since we use calico we should calculate the MTU according [calico documentation](https://docs.projectcalico.org/networking/mtu#:~:text=MTU%20and%20Calico%20defaults,IP%20and%2For%20VXLAN).).

This PR fixes SRIOV lane failures on CI.


cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,28 @@ function _fix_node_labels() {
     done
 }
 
+function _get_cri_bridge_mtu() {
+  docker network inspect bridge  | grep mtu | awk '{print $2}' | sed "s/\"//g"
```
EdDev, line 145: Please try `docker network inspect -f '{{index .Options "com.docker.network.driver.mtu"}}' bridge`

cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,28 @@ function _fix_node_labels() {
...
+  local -r calico_manifest="$1"
+  local -r mode="$2"
+  local -r cri_mtu="$3"
+
+  local -r default_calico_mtu="1440"
+
+  if [ $mode == "ipip" ]; then
+    overhead=$((20))
+  elif [ $mode == "vxlan" ];then
+    overhead=$((50))
```
EdDev, line 158: This is dead code at the moment, there is no way to test this and therefore I prefer it out until we really need it.
I would prefer you read the mode from the manifest, confirm it is ipip and then patch.

cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,28 @@ function _fix_node_labels() {
     done
 }
 
+function _get_cri_bridge_mtu() {
+  docker network inspect bridge  | grep mtu | awk '{print $2}' | sed "s/\"//g"
+}
+
+function _patch_calico_manifest_mtu() {
```
EdDev, line 148: It is not common to mutate config file in-place.
I suggest you change the original file to `kube-calico.yaml.in` and place instead of the mtu a placeholder `@CNI_MTU@`, that you later can `sed` and create a `kube-calico.yaml` which you can apply.

EdDev (changes requested):
The usage of "CRI" here is misleading.

This has nothing to do with the CRI, it is related to the deployment setup. Accessing the network part (which is not CRI related as far as I know, as it does not define networking) is an implementation detail indeed, related to docker in this case. So the function reads the setup mtu set on the container network (the abstract) by using docker (the implementation).

EdDev:
Nice work!!!

cluster-up/cluster/kind/common.sh
ormergi: I couldn't find how to read the inner map inside Options with `--format`, very nice! 
It works I will change this

cluster-up/cluster/kind/common.sh
oshoval: consider please envsubst if you put a place holder

cluster-up/cluster/kind/common.sh
```diff
@@ -152,15 +174,9 @@ function setup_kind() {
         docker exec $node /bin/sh -c "curl -L https://github.com/containernetworking/plugins/releases/download/v0.8.5/cni-plugins-linux-amd64-v0.8.5.tgz | tar xz -C /opt/cni/bin"
     done
 
-    echo "ipv6 cni: $IPV6_CNI"
```
oshoval, line 155: not related to this PR imo

cluster-up/cluster/kind/common.sh
ormergi: We can do the same we did for qemu based providers, use diff file and a manifest w/o any change

oshoval: this we done for calico
but for kubeadm we used envsubst
cluster-provision/k8s/1.18/provision.sh:envsubst < /tmp/kubeadm.conf > $kubeadm_manifest

cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,28 @@ function _fix_node_labels() {
...
+
+  if [ $mode == "ipip" ]; then
+    overhead=$((20))
+  elif [ $mode == "vxlan" ];then
+    overhead=$((50))
+  fi
+
+  calico_mtu=$((cri_mtu - overhead))
+
+  sed -e "s?veth_mtu: \"$default_calico_mtu\"?veth_mtu: \"$calico_mtu\"?g" "$calico_manifest" > "$calico_manifest"
```
qinqon, line 163: You can replace in-place with `sed` and also make a backup with `-i.bak` so this would be 

```bash
sed -i.bak -e "s?veth_mtu: \"$default_calico_mtu\"?veth_mtu: \"$calico_mtu\"?g" $calico_manifest 
```

qinqon (changes requested):

cluster-up/cluster/kind/common.sh
ormergi: Done

ormergi:
Changes:

-  Now will have clalico manifest template (kube-calico.yaml.in) and diff file template (kube-calico.diff.in).
   The MTU is calculated buy checking what is configured in calico manifest using `yk` (we use it inside a container so no 
   installation needed) then the calculated MTU is substituted in the diff template file.
   Finally we patch calico manifest with the new diff and generate patched manifest with the correct MTU.
-  Changed the command that get the CRI MTU
-  Remove flannel manifest

ormergi:
> The usage of "CRI" here is misleading.
> 
> This has nothing to do with the CRI, it is related to the deployment setup. Accessing the network part (which is not CRI related as far as I know, as it does not define networking) is an implementation detail indeed, related to docker in this case. So the function reads the setup mtu set on the container network (the abstract) by using docker (the implementation).

By mentioning CRI I mean the container runtime, I am not quite understand why it is misguiding.
Currently we use docker, which has `bridge` network, for other container runtimes (e.g: podman, crio) we would use a standard CNI which also configures MTU.
Changing CRI to container runtime will be better?

EdDev:
> By mentioning CRI I mean the container runtime, I am not quite understand why it is misguiding.
> Currently we use docker, which has `bridge` network, for other container runtimes (e.g: podman, crio) we would use a standard CNI which also configures MTU.
> Changing CRI to container runtime will be better?

I think that the network part is not included in the cri, it is external to it. Especially the bridge you are treating here.

cluster-up/cluster/kind/common.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 NODE_CMD="docker exec -it -d "
-export KIND_MANIFESTS_DIR="${KUBEVIRTCI_PATH}/cluster/kind/manifests"
+export KIND_MANIFESTS_DIR="cluster-up/cluster/kind/manifests"
```
EdDev, line 6: Why this changed? Relative paths are risky.

cluster-up/cluster/kind/common.sh
```diff
@@ -153,9 +182,11 @@ function setup_kind() {
     done
 
     echo "ipv6 cni: $IPV6_CNI"
-    if [ -z ${IPV6_CNI+x} ]; then
+     if [ -z ${IPV6_CNI+x} ]; then
```
EdDev: Why this changed?

cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,35 @@ function _fix_node_labels() {
     done
 }
 
+function _yq() {
```
EdDev, line 144: From where this `_` convention is coming from?

cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,35 @@ function _fix_node_labels() {
...
+  docker run --rm -i -v ${PWD}:/workdir mikefarah/yq yq "$@"
```
EdDev, line 145: How big is this container?

cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,35 @@ function _fix_node_labels() {
...
+  docker network inspect -f '{{index .Options "com.docker.network.driver.mtu"}}' bridge
+}
+
+function _patch_calico_manifest() {
+  local -r calico_manifest="$1"
+  local -r calico_diff="$2"
+  local -r destination="$3"
+
+  local -r ipip_mode=$(_yq read -d"*" "$calico_manifest" "spec.template.spec.containers.*.env.(name==CALICO_IPV4POOL_IPIP).value" | grep -Po "(Always|Never)")
+  local -r vxlan_mode=$(_yq read -d"*" "$calico_manifest" "spec.template.spec.containers.*.env.(name==CALICO_IPV4POOL_VXLAN).value" | grep -Po "(Always|Never)")
```
EdDev, line 158: Seems like a leftover.

cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,35 @@ function _fix_node_labels() {
...
+
+  local -r cri_mtu=$(_get_cri_bridge_mtu)
+  if [ $ipip_mode == "Always" ]; then
+    overhead=$((20))
+    calico_mtu=$((cri_mtu - overhead))
+  else
+    calico_mtu=$(_yq read -d"*" "$calico_manifest" "data.veth_mtu" | grep -P "\d")
+  fi
+
+  CNI_MTU=$calico_mtu envsubst < "$calico_diff" > "/tmp/calico.diff"
```
EdDev, line 168: Congratulation, now we use all 3 options... sed, envsubst and patch.

Sed is used here: `cluster-provision/manifests/cni/calico/script.d/003-change-cidr.sh`.

cluster-up/cluster/kind/common.sh
```diff
@@ -152,15 +181,10 @@ function setup_kind() {
...
-        _kubectl apply -f $KIND_MANIFESTS_DIR/kube-calico.yaml
-    else
-        echo "ipv6 enabled, using kindnet"
-        # currently kind does not fully support ipv6 or ipv6-DualStack,
-        # when using diffrent CNI's.
-    fi
+    # Deploy CNI
+    cni_patched_manifest="/tmp/calico.yaml"
+    _patch_calico_manifest "$KIND_MANIFESTS_DIR/kube-calico.yaml.in" "$KIND_MANIFESTS_DIR/kube-calico.diff.in" "$cni_patched_manifest"
+    _kubectl apply -f "$cni_patched_manifest"
```
qinqon, line 187: I don't know if it's worth it but you can use stdin with `kubectl apply -f -` so no need for /tmp/calico.yaml

qinqon (commented):
Just proposal

cluster-up/cluster/kind/common.sh
ormergi: 16M

cluster-up/cluster/kind/common.sh
ormergi: I almost sure that we can delete all files under `cluster-provision/manifests/cni/calico/script.d`, 
we used it before we started to use diff.

cluster-up/cluster/kind/common.sh
ormergi: I noticed that on this file function that being used only in that file starts with `_`

cluster-up/cluster/kind/common.sh
ormergi: My bad, I removed this check on the next commit anyway

cluster-up/cluster/kind/common.sh
ormergi: Sorry, I removed it.

cluster-up/cluster/kind/common.sh
ormergi: When using `yq` inside a container we mount kubevirtci path as the working dir inside the container.
Now wen we pass a file with a full path it fails because `yq` cant find that file, its out of the container scope.
So I changed `KIND_MANIFESTS_DIR` to relative path and checked that it dosent break anything (all check provision passed as well) then `yq` worked

cluster-up/cluster/kind/common.sh
oshoval: please see
https://github.com/kubevirt/kubevirtci/pull/463

i removed the whole provider there, but didnt touch this code since you already remove it,
again i think its not belong to this PR context imho

ormergi: Hey you can remove it anyway and if your PR will merge first I will change mine

oshoval: sure will do thx

EdDev (approved):

ormergi:
Changes:
- Dropped `yq` usage, use builtin standard tools instead
- Use `mktemp` for diff and calico manifest patched files

ormergi:
Removed the commit that removed the check if the cluster is IPv6 before CNI deployment

cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,35 @@ function _fix_node_labels() {
...
+
+  # Substitute MTU placeholder with the calculated MTU
+  patched_diff=$(mktemp "$destination_dir/calico.XXXX.diff")
+  CNI_MTU=$calico_mtu envsubst < "$calico_diff" > "$patched_diff"
+
+  # Patch the manifest with the patched diff file
+  patched_manifest=$(mktemp "$destination_dir/calico.XXXX.yaml")
+  patch "$calico_manifest" "$patched_diff" -o "$patched_manifest"
+
+  echo "$patched_manifest"
```
qinqon, line 170: To dump directly to stdout from `patch` command and patching using stdin
```suggestion
  CNI_MTU=$calico_mtu envsubst < $calico_diff | patch $calico_manifest $patched_diff -o -
```

EdDev (approved):
Thanks!

cluster-up/cluster/kind/common.sh
```diff
@@ -155,7 +184,7 @@ function setup_kind() {
     echo "ipv6 cni: $IPV6_CNI"
     if [ -z ${IPV6_CNI+x} ]; then
         echo "no ipv6, safe to install calico"
-        _kubectl apply -f $KIND_MANIFESTS_DIR/kube-calico.yaml
+        _kubectl apply -f "$(_patch_calico_manifest "$KIND_MANIFESTS_DIR/kube-calico.yaml.in" "$KIND_MANIFESTS_DIR/kube-calico.diff.in" "/tmp")"
```
qinqon, line 187: I think this is wrong, if you are going to use stdin you have to use `kubectl apply -f -` after a pip, also remove all those arguments to `_patch_calico_manifest` we don't need a generic function there
```suggestion
        _patch_calico_manifest | _kubectl apply -f -
```

cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,35 @@ function _fix_node_labels() {
...
 }
 
+function _get_cri_bridge_mtu() {
+  docker network inspect -f '{{index .Options "com.docker.network.driver.mtu"}}' bridge
+}
+
+function _patch_calico_manifest() {
+  local -r calico_manifest="$1"
+  local -r calico_diff="$2"
+  local -r destination_dir="$3"
```
qinqon, line 151: We don't need a generic function here, just put the details here.

cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,35 @@ function _fix_node_labels() {
...
+
+  local -r cri_mtu=$(_get_cri_bridge_mtu)
+  local -r ipip_mode=$(cat "$calico_manifest" | grep CALICO_IPV4POOL_IPIP -A 1 | grep -Po "(Always|Never)")
```
qinqon, line 154: Using `sed` capturing groups you don't need piping.

cluster-up/cluster/kind/common.sh
```diff
@@ -141,6 +141,35 @@ function _fix_node_labels() {
...
+  if [ $ipip_mode == "Always" ]; then
+    overhead=$((20))
+    calico_mtu=$((cri_mtu - overhead))
+  else
+    calico_mtu=$(cat "$calico_manifest" | grep -Po "veth_mtu: \"\K\d*")
```
qinqon, line 159: Using `sed` capturing groups you don't need piping.

qinqon (changes requested):
We can simplify stuff, also we use double quotes too much.

cluster-up/cluster/kind/common.sh
ormergi: Done

cluster-up/cluster/kind/common.sh
ormergi: Done

cluster-up/cluster/kind/common.sh
ormergi: I removed them

cluster-up/cluster/kind/common.sh
ormergi: Done

cluster-up/cluster/kind/common.sh
ormergi: Done

PR #460, opened 2020-10-26
run: Move container create to the correct place

qinqon:
The provider container node was created before setting qemu-args correctly

After provisioning the script check-cluster-up.sh is run, this PR
add some logic to check the node and secondary nics existence.

### DIFF
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -432,18 +432,6 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 		}
 		volumes <- vol.Name
 
-
-		vmContainerConfig := &container.Config{
-			Image: clusterImage,
-			Env: []string{
-				fmt.Sprintf("NODE_NUM=%s", nodeNum),
-			},
-			Volumes: map[string]struct{}{
-				"/var/run/disk/": {},
-			},
-			Cmd: []string{"/bin/bash", "-c", fmt.Sprintf("/vm.sh -n /var/run/disk/disk.qcow2 --memory %s --cpu %s %s", memory, strconv.Itoa(int(cpu)), nodeQemuArgs)},
-		}
-
 		// assign a GPU to one node
 		var deviceMappings []container.DeviceMapping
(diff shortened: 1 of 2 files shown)


cluster-provision/k8s/check-cluster-up.sh
```diff
@@ -27,9 +28,18 @@ export KUBEVIRTCI_GOCLI_CONTAINER=kubevirtci/gocli:latest
   bash -x ./cluster-up/up.sh
   ${ksh} wait --for=condition=Ready pod --timeout=200s --all
   ${ksh} wait --for=condition=Ready pod --timeout=200s -n kube-system --all
-  ${ksh} get nodes
+  ${ksh} get nodesç
```
rmohr, line 31: This line probably causes some issues.

qinqon: Done

ormergi:
Nice!
/lgtm

cluster-provision/gocli/cmd/run.go
```diff
@@ -526,25 +524,25 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 		}
 
 		if success {
-            // move the VM sound card to a vfio-pci driver to prepare for assignment
-            err = prepareDeviceForAssignment(cli, nodeContainer(prefix, nodeName), soundcardPCIID, "")
-            if err != nil {
-                return err
-            }
+			// move the VM sound card to a vfio-pci driver to prepare for assignment
```
rmohr, line 527: We should add go-formatting. :/ Can we keep whitespaces?

rmohr: Never mind. This was go-fmt changing things :+1:

rmohr (approved):

cluster-provision/gocli/cmd/run.go
qinqon: tab is the standard for go, let's try to add those checks at follow up.

rmohr: Definitely. :+1:

PR #461, opened 2020-10-26
Bump SRIOV operator to release 4.6

oshoval:
Bump sriov operator version to release 4.6.
The previous hash that was used didn't point actually to 4.4 but couple commits below it.

Admin MAC set by the CNI is added on release 4.6, previous versions didn't have this feature.

Changes:
* Force ENABLE_ADMISSION_CONTROLLER in Makefile, since we need the resource-injector which comes with the webhooks, see Note at https://github.com/openshift/sriov-network-operator/blob/release-4.6/doc/quickstart.md
(unless we expend the makefile target).
* Disable webhook so we can create policies on unsupported nic types (see https://bugzilla.redhat.com/show_bug.cgi?id=1850505)
* Replace skopeo and jq with echo in `hack/env.sh` so we dont need to install them and not fail the script as its mandator.
We already override the values and don't need the use of it (until https://github.com/kubevirt/kubevirtci/pull/466 is merged).
* Support bumping by replacing the RELEASE_VERSION only.
* Works for both cases if the previous SRIOV operator was already downloaded or not.

Known issue to fix:
_kubectl not found

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -10,7 +10,12 @@ KUBECONFIG_PATH="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig"
 MASTER_NODE="${CLUSTER_NAME}-control-plane"
 WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
 
-OPERATOR_GIT_HASH=8d3c30de8ec5a9a0c9eeb84ea0aa16ba2395cd68  # release-4.4
+# In order to bump, update this env var
+# Tested with:
+# 4.4, 4.6
+export RELEASE_VERSION=4.6
+
+OPERATOR_GIT_HASH=$(git ls-remote --heads https://github.com/openshift/sriov-network-operator.git | grep release-$RELEASE_VERSION | cut -f1)
 
 # This function gets a command string and invoke it
 # until the command returns an empty string or until timeout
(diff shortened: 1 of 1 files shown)


oshoval:
We might want to remove the cert creator, as from what i saw there arent cert anymore ?
the cluster worked without it

The PR is because Zenghui Shi
said that we set the admin MAC in 4.6, previous versions didnt have it,
and this was done as part of investigating other task.

oshoval:
Tests passed when tried locally,
one note i saw is that if i created a pod manually, i had to inject to it manually the requests and limits

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -211,16 +211,6 @@ function deploy_sriov_operator {
     go run . -namespace sriov-network-operator -secret network-resources-injector-secret -hook network-resources-injector -kubeconfig $KUBECONFIG_PATH || return 1
   popd
 
-  echo 'Setting caBundle for SR-IOV webhooks'
```
oshoval, line 214: bug here will work on it
name changed

oshoval:
Locally it passsed, here it failed, checking

oshoval:
@ormergi 
please see here, that we must disable the webhooks, so even thought we might want the fix the certificates as they cause flaky cluster-up on CI, at the end we will need to disable them.
"Disable webhook so we can create policies on unsupported nic types (see https://bugzilla.redhat.com/show_bug.cgi?id=1850505)"

ormergi:
> @ormergi
> please see here, that we must disable the webhooks, so even thought we might want the fix the certificates as they cause flaky cluster-up on CI, at the end we will need to disable them.
> "Disable webhook so we can create policies on unsupported nic types (see https://bugzilla.redhat.com/show_bug.cgi?id=1850505)"

Currently we dont encounter those issues on CI we use the supported hardware we dont have to disable the webhooks.
Having the webhooks enable is important for preventing miss-configurations with the object they handle

oshoval:
> > @ormergi
> > please see here, that we must disable the webhooks, so even thought we might want the fix the certificates as they cause flaky cluster-up on CI, at the end we will need to disable them.
> > "Disable webhook so we can create policies on unsupported nic types (see https://bugzilla.redhat.com/show_bug.cgi?id=1850505)"
> 
> Currently we dont encounter those issues on CI we use the supported hardware we dont have to disable the webhooks.
> Having the webhooks enable is important for preventing miss-configurations with the object they handle

We would need soon to bump to 4.6, and then we would must disable the webhooks, see the bugzilla please.

I am suggesting you to use this fact in order to disable the webhooks either currently or when we move to 4.6
because then you will be able to fix CI flakiness upon cscreator, you are free to fix it if you want.
This PR is orthogonal for that, but i wanted to save you time that will cause CI stable meanwhile.

You dont see the problem because you use 4.4 which doesn't have this problem, we will not be on 4.4 forever.
When we move to 4.6 we wouldn't have a choice, this is also what the readme.md of SRIOV operator says.

oshoval:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/461/check-up-kind-1.17-sriov/1324360562389291008

Passed with 4.6 (locally i checked it also with kubevirt)
I have some more stuff i want to fix here still but, in order to make it interchangeable, and robust even if the sriov operator exists.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -192,16 +197,26 @@ function deploy_sriov_operator {
     curl -L https://github.com/openshift/sriov-network-operator/archive/${OPERATOR_GIT_HASH}/sriov-network-operator.tar.gz | tar xz -C ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/
   fi
 
+  if [ ! $RELEASE_VERSION = "4.4.0" ]; then
```
oshoval, line 200: Note
* This if can be removed as the code here is generic for all versions i saw.
* We can also use diff file, but apply it only in the first time the folder is created.
* Another method is to expend the make target, to the one it uses, skipping the ENABLE_ADMISSION_CONTROLLER=false in it.
(it wouldn't solve the need of the other lines for now).

oshoval:
using a fix that was merged to 4.8
effort moves to https://github.com/kubevirt/kubevirtci/pull/500
it now supports unsupported nics (creating a PR that will make it official)

PR #463, opened 2020-10-27
Remove kind ipv6 provider

oshoval:
We use k8s 1.18+ as dual stack, so no need for kind ipv6
provider anymore.


EdDev (approved):

oshoval:
Rebased to solve conflict, PTAL

cluster-up/cluster/kind/common.sh
```diff
@@ -180,19 +176,12 @@ function setup_kind() {
...
-        patched_diff=$(_patch_calico_manifest_diff $calico_manifest)
-        echo "Log Calico manifest diff:"
-        echo "$patched_diff"
-        _patch_calico_manifest "$calico_manifest" "$patched_diff" | _kubectl apply -f -
-    else
-         echo "ipv6 enabled, using kindnet"
-         #currently kind does not fully support ipv6 or ipv6-DualStack,
-         #when using diffrent CNI's.
-    fi
+    echo "Installing Calico"
```
ormergi, line 179: I would write "Deploy CNI" since its the place where you deploy the cni plugin and the plugin that we use could vary

ormergi:

This lane is not mandatory yet, lets verify that nothing breaks sriov provider

oshoval:
> /test check-up-kind-1.17-sriov
> 
> This lane is not mandatory yet, lets verify that nothing breaks sriov provider

tested locally, but np lets test

cluster-up/cluster/kind/common.sh
oshoval: for now we install calico, i am not sure its important to make it generic, as its specific for the time being,
if you insist i will change

ormergi: We can also add "CNI plugin" or somthing to indicate this is what handels the network and should be done right after the nodes are up, so it will be easier to understand by newcomers, what you think?

oshoval: yep i also wondered if to add the word CNI, and then thought its might be redundant because calico here is known as the CNI, i will add those words.

oshoval: Done

ormergi: Thanks!

ormergi:
/hold 
Only for verifying sriov lane passes

oshoval:
@ormergi 
why the lane looks like it started but doesnt do anything? is it ok ?
i saw it in the past from time to time, does it wait for a machine to be available ? (if so wonder if we need to fix, so it wont looks like started already)
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/463/check-up-kind-1.17-sriov/1321390766941016064

ormergi:
> @ormergi
> why the lane looks like it started but doesnt do anything? is it ok ?
> i saw it in the past from time to time, does it wait for a machine to be available ? (if so wonder if we need to fix, so it wont looks like started already)
> https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/463/check-up-kind-1.17-sriov/1321390766941016064

Yep it waits for capable node on CI cluster to be ready

oshoval:
> > @ormergi
> > why the lane looks like it started but doesnt do anything? is it ok ?
> > i saw it in the past from time to time, does it wait for a machine to be available ? (if so wonder if we need to fix, so it wont looks like started already)
> > https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/463/check-up-kind-1.17-sriov/1321390766941016064
> 
> Yep it waits for capable node on CI cluster to be ready

it smells that we would need more machines of prow SRIOV ?
(unless already requested)
once the lane will be mandatory, it might be a new bottleneck on CI.

ormergi:
> > > @ormergi
> > > why the lane looks like it started but doesnt do anything? is it ok ?
> > > i saw it in the past from time to time, does it wait for a machine to be available ? (if so wonder if we need to fix, so it wont looks like started already)
> > > https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/463/check-up-kind-1.17-sriov/1321390766941016064
> > 
> > 
> > Yep it waits for capable node on CI cluster to be ready
> 
> it smells that we would need more machines of prow SRIOV ?
> (unless already requested)
> once the lane will be mandatory, it might be a new bottleneck on CI.
 
We sure need more, currently there are two.
This is why I am not sure if we should to set `check-up-kind-1.17-sriov` mandatory as well

PR #465, opened 2020-10-28
Try to cope with initial unresponsiveness of ssh

rmohr:
Retry establishing a connection to the VM a few times when we try to
connect to the VM the first time.

This may resolve some flakes which manifest themselves with

```
binding device to vfio driver failed
```

coming from the first script which gets executed.

### DIFF
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -12,7 +12,9 @@ import (
 	"strings"
 	"sync"
 	"text/template"
+	"time"
 
+	"github.com/Sirupsen/logrus"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/mount"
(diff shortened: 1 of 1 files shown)


rmohr:
/cc @ormergi 
/cc @qinqon 
/cc @vladikr 

I have at least right now no better idea. Maybe you have?

cluster-provision/gocli/cmd/run.go
```diff
@@ -502,6 +504,20 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
...
+			err = _cmd(cli, nodeContainer(prefix, nodeName), "ssh.sh echo VM is up", "waiting for node to come up")
+			if err == nil {
+				break
+			}
+			logrus.WithError(err).Warningf("Could not establish a ssh connection to the VM, retrying ...")
+			time.Sleep(1 * time.Second)
+		}
+
+		if err != nil {
+			return fmt.Errorf("could not establish a connectdion to the node after a generous timeout: %v", err)
```
vladikr, line 518: s/connectdion/connection

vladikr:
Great!
/approve

vladikr:
> I have at least right now no better idea. Maybe you have?

At least this will log the ssh failure now, so we will know the root cause

cluster-provision/gocli/cmd/run.go
rmohr: Done.

PR #471, opened 2020-11-04
SRIOV provider: Add Initial Example SriovNetwork

ormergi:
This PR adds an example SriovNetwork creation after sriov-operaor-deployment for SRIOV provider.
It enable one to create VM without worrying about creating SriovNetwork for initial results.

This is necessary for kubevirt/kubevirt#4356

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/manifests/sriov-network-example.yaml
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/manifests/sriov-network-example.yaml
@@ -1,0 +1,10 @@
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetwork
+metadata:
+  name: sriov-network-example
+  namespace: sriov-network-operator
+spec:
+  ipam: '{}'
+  vlan: 0
+  resourceName: sriov_net
+  networkNamespace: default
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -267,6 +267,16 @@ function apply_sriov_node_policy {
   return 0
 }
 
+function apply_sriov_network {
+  local -r sriov_network_file="$1"
+
+  _kubectl apply -f $sriov_network_file || return 1
+  wait_k8s_object "sriovnetwork" "sriov-network-example" "sriov-network-operator" || return 1
+  wait_k8s_object "net-attach-def" "sriov-network-example" "default" || return 1
+
+  echo "sriov network is ready 'sriov-network-example'" || return 0
+}
+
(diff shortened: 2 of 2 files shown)


cluster-up/cluster/kind-k8s-sriov-1.17.0/manifests/sriov-network-example.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: sriovnetwork.openshift.io/v1
```
oshoval, line 1: why having the yaml in the folder isnt enough?
as you did on your other PR, unless this is what the community reached conclusion ?

i wouldn't create please wet example resources, the cluster should have the minimal things needed,
both for the user that inspect it, and also for scripts / tests that iterate it

oshoval (changes requested):

cluster-up/cluster/kind-k8s-sriov-1.17.0/manifests/sriov-network-example.yaml
ormergi: My initial intent it to make better expiriance with SRIOV VM's espascially for new commers
having all necessary objects in place for creating simple SRIOV VM could simplify it.

Also I verified and creating the example SriovNetwork has minimal footprint on resources and the time it takes for the cluster to be ready.

For the community side, it was raised that we can also create it on the cluster deployment phase.

Let wait for other reviewers thoughts

oshoval: i will just add that when i debugged sriov policies, if i had another policy that was just an example 
it would cause noise to me as the one that debugs, (policy here is just an example),
because i had to backup them and create them manually,
also one that would create with the same name and different config would fail

i find it very odd and not needed to install examples

ormergi: The default SriovNetworkNodePolicy is created by the sriov-network-operator by default, I am quite sure that if you delete it the operator will recreate it.
Why would you need to backup and create them?
I dont see a problem if one would create an object with same name and get an error, I mean its not that it will beak something.

I find it more odd to print the example when cluster-up finish.
Since `examples` directory is part of kubevirt it is not in the scope of kubevirtci to change.

oshoval: i meant the not default that is created but not auto created it you delete it, its named policy-1 something like that.
and cluster should be clean, it shouldnt have installed some examples, no one will look on them,
people look on docs, and want their cluster clean, from what i know.

lets wait for other reviews as well to see this question please.
you can create examples on kubevirtci as well, or to just add it to kubevirt examples

ormergi:
/close

The conclusion from the discussion is that  this change should be presented on kubevirt/kubevirt

PR #472, opened 2020-11-04
SRIOV provider, csrcreator: switch to self-sign certificates

ormergi:
As part of SRIOV provider deployment (make cluster-up) we download, build and deploy sriov-network-operator.
The operator uses validating and mutating webhooks and in order to use them it is necessary to sign them
with certificates.

During SRIOV provider deployment we request for a certificate from the cluster API using [CSR](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/).
Once we get the certificates we create secrets that are expected to be there by the webhooks pods 
and patch the caBundle of the webhooks with the certificates.
This is done by the csrcreator code.

The problem is that during the CSR negotiation with the cluster API,
we encounter etcdserver timeout errors often that make this process flaky, for example:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/4195/pull-kubevirt-e2e-kind-1.17-sriov/1323574884390932480
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/4195/pull-kubevirt-e2e-kind-1.17-sriov/1323298566893998080
Those errors causing the cluster deployment to fail.

This PR removes the CSR mechanism and presets new code for generating self-signed 
certificates for sriov-network-operator webhooks.
With that we avoid CSR negotiation with the cluster API, and reduce the chance for cluster-up to fail.
Also using self-signed certificates is faster and less prune to errors.

PR #473, opened 2020-11-09
Suggestion - sriov provider, Let cluster stabilize after deployed

oshoval:
Since the tests start right after cluster is deployed,
and some pods still work with etcd extensively
we should let the cluster stabilize in order
to avoid `etcdserver: request timed out` which is caused by IO pressure
typically when cluster starts, before it stabilized.

SKIP_SRIOV_SLEEP can be set in order to skip the
waiting.

Fixes:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/4511/pull-kubevirt-e2e-kind-1.17-sriov/1325727568359329793

Reference that can help more:
https://github.com/kubernetes-sigs/kind/issues/717#issuecomment-512507692
https://github.com/kubernetes-sigs/kind/issues/717#issuecomment-513083096 (unless already bumped high enough)
https://github.com/etcd-io/etcd/issues/11809#issuecomment-621004811

See https://github.com/kubevirt/kubevirt/issues/4519

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -334,4 +334,10 @@ wait_allocatable_resource $SRIOV_NODE "openshift.io/$resource_name" $NODE_PF_NUM
 _kubectl get nodes
 _kubectl get pods -n $SRIOV_OPERATOR_NAMESPACE
 echo
+
+# Upon cluster start we might see "etcdserver: request timed out"
+# when accessing API, due to IO pressure.
+# Let the transient pass away (skip it by export SKIP_SRIOV_SLEEP=1)
+[ -z $SKIP_SRIOV_SLEEP ] && echo "sleeping 120 seconds to stabilize etcd, export SKIP_SRIOV_SLEEP=1 to skip"  && sleep 120
+
 echo "$KUBEVIRT_PROVIDER cluster is ready"


oshoval:
/cc @EdDev 
Else we would need to add a stabilizer in the tests

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -334,4 +334,10 @@ wait_allocatable_resource $SRIOV_NODE "openshift.io/$resource_name" $NODE_PF_NUM
 _kubectl get nodes
 _kubectl get pods -n $SRIOV_OPERATOR_NAMESPACE
 echo
+
+# Upon cluster start we might see "etcdserver: request timed out"
+# when accessing API, due to IO pressure.
+# Let the transient pass away (skip it by export SKIP_SRIOV_SLEEP=1)
+[ -z $SKIP_SRIOV_SLEEP ] && echo "sleeping 120 seconds to stabilize etcd, export SKIP_SRIOV_SLEEP=1 to skip"  && sleep 120
```
dhiller, line 341: Isn't there a better condition we can check besides just an arbitrary sleep?

oshoval: I am not sure yet beside the info in the PR description, unless we will "wait" for more specific conditions on each pod,
some of the pods or SRIOV operator are known to missing few conditions already, this is why the SRIOV waits for taint addon and then absent.

Please see the PR description 2 references that shows its a known issue that can be caused by IO stress, or if there aren't enough inotify in the system.

ormergi:
Notice that we encounter ectd time out errors in two places:
On CSR request
On the the test [test_id:1754]should create a virtual machine with sriov interface [It] tests/vmi_multus_test.go:697

I am monitoring CI for a while now and from what I saw there are no disk IO pressure on the CI nodes that schedule sriov lane jobs,
disk IO operation peeks at 250 (read/write) so I think its not our case.

We also increased inotify watchers count on all nodes to prevent more problems regarding IOP's and ectd. 

About adding a sleep arbitrary time is not the way to go.
- We try to reduce the time it takes for cluster-up to finish, so eventually we could run more CI jobs or at least reduce the pressure.
- Any peek in resource consumption on the node can reproduce the problem.
- We try to avoid sleeps as much as we can.

Do you know which pods works with etcd extensively right after cluster is up?
Did you saw "what goes under the hood" of etcd that cause these timeouts?
How could you know what the cluster is stabilized and ready to go?

oshoval:
> Notice that we encounter ectd time out errors in two places:
> On CSR request
> On the the test [test_id:1754]should create a virtual machine with sriov interface [It] tests/vmi_multus_test.go:697
> 
> I am monitoring CI for a while now and from what I saw there are no disk IO pressure on the CI nodes that schedule sriov lane jobs,
> disk IO operation peeks at 250 (read/write) so I think its not our case.
> 
> We also increased inotify watchers count on all nodes to prevent more problems regarding IOP's and ectd.
> 
> About adding a sleep arbitrary time is not the way to go.
> 
> * We try to reduce the time it takes for cluster-up to finish, so eventually we could run more CI jobs or at least reduce the pressure.
> * Any peek in resource consumption on the node can reproduce the problem.
> * We try to avoid sleeps as much as we can.
> 
> Do you know which pods works with etcd extensively right after cluster is up?
> Did you saw "what goes under the hood" of etcd that cause these timeouts?
> How could you know what the cluster is stabilized and ready to go?

See please the discussion started with Daniel Hiller, it suggests what can be tried empirically,
as this phenomenon happen almost just on the start of the cluster, i would focus on researching this transient period, but meanwhile we can have a stabilized CI, parallel to research.
Moreover we can allow the user / CI project-infra to control which timeout to use according this variable,
a lane which is flake is much more expansive that a lane that wait for the transient to go away up to 2 minutes.
A random peak in the IO can cause it, even task switch or kernel page fault, just a bit after creating the pods, until when they reach their stable state, so i would suggest to research it, even if the monitors shows disk IO which looks fine,
see please the refs in the PR description.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
dhiller: So what exactly is the issue here? That the etcd cannot get reached by the sriov operator? Or that the etcdserver takes a long time to get up? Or that accessing the cluster API times out, even though the etcd is okay?

If the etcd itself is the problem, can't we rather try to check that? I guess that the operator is missing some kind of state then?

oshoval: Issue is that after cluster is freshly created, first test might experience etcd server timeout when trying to reach the API ep.
etcd is ok, its legit that it happens (see please the Ref links, as its a known issue)

https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/4511/pull-kubevirt-e2e-kind-1.17-sriov/1325727568359329793

I saw it several times on different providers in the past, always it was a transient state on a freshly created cluster,
because all the pods starting, we check the conditions that are enough for us (unless we find another specific condition we should check), but then there are couple pods that still access relatively extensively the etcd, causing this pressure.

oshoval: One more thing to make sure is that we dont have a leader change in SRIOV, 
leader change can cause it, and sriov operator has leader election mechanism

oshoval:
```
2020-11-09 10:02:48.075278 W | etcdserver: timed out waiting for read index response (local node might have slow network)
2020-11-09 10:02:48.075402 W | etcdserver: read-only range request "key:\"/registry/namespaces/default\" " with result "error:etcdserver: request timed out" took too long (7.184676107s) to execute
2020-11-09 10:02:48.075444 W | etcdserver: read-only range request "key:\"/registry/services/endpoints/kubevirt/virt-operator\" " with result "error:etcdserver: request timed out" took too long (7.489442621s) to execute
2020-11-09 10:02:48.075493 W | etcdserver: read-only range request "key:\"/registry/kubevirt.io/virtualmachineinstances/kubevirt-test-default1/testvmicvlq7tmjcblw5bhhtqpzvcjrc87z494xqxnlks9jxc7b4mpb\" " with result "error:etcdserver: request timed out" took too long (8.490783071s) to execute
```
from the time of the error in the test

rmohr: @oshoval what would be creating pressure on the API at this stage? It looks that exactly at the point where you add the  wait, the cluster is already stabilized (all pods ready, all conditions met, and so forth). This means that the pressure should be lower now than during all the deploy steps. This in turn means that all our `kubectl` calls from above worked without a single failure on etcd. This wait looks a little bit questionable.

oshoval: I saw in several providers, that even when the provider finishes, we might still have some transient symtoms
and only on a fresh created provider.
For example https://github.com/kubevirt/kubevirt/issues/4158
Which fails only when runs on check-for-flakes lane, which run the test as the first test and not one in the middle of all the e2e ones.

I believe that even cluster-up is finished, there are some processes like leader election and so on, when all the new pods are started, they are operational but they little more busy at the begining
for example the logs that are mentioned here
https://github.com/etcd-io/etcd/issues/11809#issuecomment-621004811

see please that the link mentioned on the PR description always fails on the first test when it fails,
its a freshly cluster.

What about trying it empirically by merging the PR, see if it removes the flakes and then rethink ?
I already started digging a bit, but don't have the exact solution yet.

ormergi:
> See please the discussion started with Daniel Hiller, it suggests what can be tried empirically,
> as this phenomenon happen almost just on the start of the cluster, i would focus on researching this transient period, but 
> meanwhile we can have a stabilized CI, parallel to research.
> Moreover we can allow the user / CI project-infra to control which timeout to use according this variable,
> a lane which is flake is much more expansive that a lane that wait for the transient to go away up to 2 minutes.
> A random peak in the IO can cause it, even task switch or kernel page fault, just a bit after creating the pods, until when they reach their stable state, so i would suggest to research it, even if the monitors shows disk IO which looks fine,
> see please the refs in the PR description.

This should be verified to actually affect before merging such peculiar timeout.
Notice that it happens right after sriov-operator deployment and then again on the first test that going to start (no matter what is the order of the tests)
 
> EDIT btw, i saw this behavior on more lanes, all were in the transient state in the beginning, where there is more etcd pressure for example during election and so on going on.

On which lane you saw it can you refere to the job logs? did it happen on cluster-up phase?

> so about your 3rd question, i would suggest, lets see that it solves empirically and then think again, since i did saw it couple times in the past and the ref is relevant that a pressure can cause it.

Do you know for sure that there is disk IO pressure when this happens?

> Moreover we see it only on cluster-start related stuff, (tests, part of the cluster-up).

Its not on cluster start when we encounter those those errors on tests

oshoval:
> > See please the discussion started with Daniel Hiller, it suggests what can be tried empirically,
> > as this phenomenon happen almost just on the start of the cluster, i would focus on researching this transient period, but
> > meanwhile we can have a stabilized CI, parallel to research.
> > Moreover we can allow the user / CI project-infra to control which timeout to use according this variable,
> > a lane which is flake is much more expansive that a lane that wait for the transient to go away up to 2 minutes.
> > A random peak in the IO can cause it, even task switch or kernel page fault, just a bit after creating the pods, until when they reach their stable state, so i would suggest to research it, even if the monitors shows disk IO which looks fine,
> > see please the refs in the PR description.
> 
> This should be verified to actually affect before merging such peculiar timeout.
> Notice that it happens right after sriov-operator deployment and then again on the first test that going to start (no matter what is the order of the tests)
> 

The best way to verify it, unless i will manage to simulate it locally (didnt try yet) is by prow itself,
This what i suggested.

> > EDIT btw, i saw this behavior on more lanes, all were in the transient state in the beginning, where there is more etcd pressure for example during election and so on going on.
> 
> On which lane you saw it can you refer to the job logs? did it happen on cluster-up phase?

After cluster-up, (see also the last inline comment as its the interesting stuff only)
https://github.com/kubevirt/kubevirt/issues/4158

> 
> > so about your 3rd question, i would suggest, lets see that it solves empirically and then think again, since i did saw it couple times in the past and the ref is relevant that a pressure can cause it.
> 
> Do you know for sure that there is disk IO pressure when this happens?

No, but i know for sure the etcd transaction takes long time, and its pretty enough, a sudden spike might happen,
we can use metrics monitoring as suggested in one of the links in the PR.

> 
> > Moreover we see it only on cluster-start related stuff, (tests, part of the cluster-up).
> 
> Its not on cluster start when we encounter those those errors on tests

I mean on a fresh cluster, on e2e directly after a fresh cluster-up is "done", as the mentioned prow runs shows,
which is what PR cares about, only.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: > I saw in several providers, that even when the provider finishes, we might still have some transient symtoms
> and only on a fresh created provider.

On which providers you saw that happen? could you refer to build log?

> For example [kubevirt/kubevirt#4158](https://github.com/kubevirt/kubevirt/issues/4158)
> Which fails only when runs on check-for-flakes lane, which run the test as the first test and not one in the middle of all the e2e ones.

The job that this issue refers to does not have ectd time out errors, I am not sure it is good example

> I believe that even cluster-up is finished, there are some processes like leader election and so on.
> When all the new pods are started, they are operational but they little more busy at the beginning,
> for example the logs that are mentioned here
> [etcd-io/etcd#11809 (comment)](https://github.com/etcd-io/etcd/issues/11809#issuecomment-621004811)
> 

Can you back this with reproducer or logs you saw?

> see please that the link mentioned on the PR description always fails on the first test when it fails,
> its a freshly cluster.
> 

Note that it could happens on earlier stage, on [CSR for creating a certificate](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/4444/pull-kubevirt-e2e-kind-1.17-sriov/1326121036692852736) or when creating a secret which happens right after the certificate is ready.

> What about trying it empirically by merging the PR, see if it removes the flakes and then rethink ?
> I already started digging a bit, but don't have the exact solution yet.

This should be verified to actually affect before merge

> EDIT:
> about "This in turn means that all our kubectl calls from above worked without a single failure on etcd. This wait looks a little bit questionable."
> Just a note - See please that in some cases we had to add retry (see for example the CSR WIP PR #472 that suffer from etcd timeout), but we indeed care from the point after the cluster-up is "done", and its affect on e2e tests.

About Sriov operator leader election, I verified and we use only one replica so the leader election shouldn't affect

oshoval: 1. See please the issue i enclosed in the above message, 
https://github.com/kubevirt/kubevirt/issues/4158,
even that pods do exists, there is a hiccup and the node list is empty, i dont have the specific reason, but its a hiccups, i saw those on more providers in the past, for example ocp with 3 masters and there it was exactly the "and" of those:
A. on startup
B. due to disk pressure
C. etcd timeout
And saw more, i think i created once a PR that fixed cluster-sync due to hiccups on other lane beside the 3 masters PR.

2. please see 1, its an hiccup, and etcd is the natural reason for it usually from all the stuff i saw in the past.
3. not have at hand.
4. in the tests its after CSR is done, if CSR had error it wont finish cluster-up
5. see please https://github.com/kubevirt/kubevirtci/pull/473#issuecomment-724808265 the first answer here.

ormergi:
> After cluster-up, (see also the last inline comment as its the interesting stuff only)
> [kubevirt/kubevirt#4158](https://github.com/kubevirt/kubevirt/issues/4158)
>
The job that this issue refers to does not describe the problem we see, 
we cant conclude form this that there was ectd timeout error.

oshoval:
> > After cluster-up, (see also the last inline comment as its the interesting stuff only)
> > [kubevirt/kubevirt#4158](https://github.com/kubevirt/kubevirt/issues/4158)
> 
> The job that this issue refers to does not describe the problem we see,
> we cant conclude form this that there was ectd timeout error.

See 2 here please
https://github.com/kubevirt/kubevirtci/pull/473#discussion_r520715236
lets keep one thread at a time, or if 2, each should concentrate on some stuff, preferably in a thread, then on the PR root

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: see this one that had the same problem (see PR description) https://github.com/kubevirt/kubevirt/pull/3095
at the end of kubevirt `make cluster-sync`

oshoval: see this one that had the same problem (see PR description) kubevirt/kubevirt#3095
at the end of kubevirt make cluster-sync
(referred to it in [1] above).
We dont need to care if its the same string, hiccup is hiccup thats what matters, most of them have things in common, we should understand if its the family of them imo.

Moreover << request time out itself is mentioned in the comments there.

oshoval: about 2 
opened this draft to try and catch the error that happens
https://github.com/kubevirt/kubevirt/pull/4523

oshoval: btw its interesting (above message), it might be after cluster-sync what we see on e2e.....

about 2
opened this draft to try and catch the error that happens
kubevirt/kubevirt#4523

oshoval:
Closing this PR
The idea i would suggest to check is 
https://github.com/kubevirt/kubevirt/issues/4519#issuecomment-724913645

more info at
https://github.com/kubevirt/kubevirtci/pull/473#discussion_r520739730

since it happens after cluster-sync, the culprit is there, and using the metrics we will be able to see
if the disk is too slow

PR #475, opened 2020-11-10
sriov, create sriov OWNERS file

oshoval:
Create OWNERS file in sriov folder, so specific people would get
notification when the folder is changed.

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/OWNERS
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/OWNERS
@@ -1,0 +1,10 @@
+filters:
+  ".*":
+    reviewers:
+      - qinqon
+      - oshoval
+      - phoracek
+      - ormergi
+    approvers:
+      - qinqon
+      - phoracek


cluster-up/cluster/kind-k8s-sriov-1.17.0/OWNERS
```diff
@@ -0,0 +1,31 @@
+filters:
+  ".*":
+    reviewers:
```
phoracek, line 3: I believe that for reviewers, Prow would always try to prefer this file over the root one. For approvers, all root approvers would have rights here, but the ones listed below would be preferred.

Could we limit the reviewers only to people who recently worked on this subdirectory? And for approvers, filter out only those who recently worked on the network part as well.

oshoval: Done, thanks

oshoval:
Answered comments

phoracek (commented):
/approve

@EdDev please take a look.

oshoval:
@qinqon please take a look

phoracek (commented):
@oshoval would you please make this match the root level OWNERS file? Just so it contains only contributors to this package. I believe it means dropping Edy out from approvers.

If we do that, there is no change in privileges, we would just prioritize sending notifications to people involved here.

oshoval:
> @oshoval would you please make this match the root level OWNERS file? Just so it contains only contributors to this package. I believe it means dropping Edy out from approvers.
> 
> If we do that, there is no change in privileges, we would just prioritize sending notifications to people involved here.

Done

phoracek (approved):
/lgtm
/approve

This is just a subset of the root OWNERS who worked on this provider. The only change is that prow will prioritize these users when asking for reviews.

Thanks Or.

oshoval:
The sriov failure here has a previous open issue, happens just once in a while.

/retest

PR #478, opened 2020-11-15
KIND infra, SRIOV proivder,  Run in-memory etcd

ormergi:
Currently we encounter bad performance of KIND cluster on DinD setup,
we get 'etcdserver: timeout errors' on SRIOV lane jobs causing jobs to fail often.

In such cases it is recommended to run in-memory etcd
kubernetes-sigs/kind#1922
kubernetes-sigs/kind#845
Running etcd in memory should improve performance and make sriov lane more stabilized

This PR patch the kind-config yaml for creating a cluster to run etcd from a specified path
which will be backed by in memory volume https://github.com/kubevirt/project-infra/pull/709

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
@@ -18,18 +18,8 @@ function up() {
     docker run --rm --cap-add=SYS_RAWIO quay.io/phoracek/lspci@sha256:0f3cacf7098202ef284308c64e3fc0ba441871a846022bb87d65ff130c79adb1 sh -c "lspci | egrep -i 'network|ethernet'"
 
     cp $KIND_MANIFESTS_DIR/kind.yaml ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
-    _fetch_kind
-    prepare_workers
-    # adding mounts to control plane, need them for sriov
-    cat >> ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml << EOF 
-  extraMounts:
-  - containerPath: /lib/modules
-    hostPath: /lib/modules
-    readOnly: true
-  - containerPath: /dev/vfio/
-    hostPath: /dev/vfio/
-EOF
 
-    setup_kind
+    kind_up
+
     ${KUBEVIRTCI_PATH}/cluster/$KUBEVIRT_PROVIDER/config_sriov.sh
 }
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
@@ -16,6 +16,7 @@ function up() {
     # print hardware info for easier debugging based on logs
     echo 'Available NICs'
     docker run --rm --cap-add=SYS_RAWIO quay.io/phoracek/lspci@sha256:0f3cacf7098202ef284308c64e3fc0ba441871a846022bb87d65ff130c79adb1 sh -c "lspci | egrep -i 'network|ethernet'"
+    echo ""
 
     cp $KIND_MANIFESTS_DIR/kind.yaml ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
 
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -192,6 +192,19 @@ function setup_kind() {
     prepare_config
 }
 
+function _add_worker_extra_mounts() {
+    if [[ "$KUBEVIRT_PROVIDER" =~ sriov.* ]]; then
+        cat <<EOF >> ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
+  extraMounts:
+  - containerPath: /lib/modules
+    hostPath: /lib/modules
+    readOnly: true
+  - containerPath: /dev/vfio/
+    hostPath: /dev/vfio/
+EOF
+  fi
+}
+
 function _add_worker_kubeadm_config_patch() {
     cat << EOF >> ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
   kubeadmConfigPatches:
@@ -213,6 +226,7 @@ function _add_workers() {
 - role: worker
 EOF
     _add_worker_kubeadm_config_patch
+    _add_worker_extra_mounts
     done
 }
 
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -91,24 +91,6 @@ function _configure_network() {
     done
 }
 
-function prepare_workers() {
-    # appending eventual workers to the yaml
-    for ((n=0;n<$(($KUBEVIRT_NUM_NODES-1));n++)); do
-        cat << EOF >> ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
-- role: worker
-  kubeadmConfigPatches:
-  - |
-    kind: JoinConfiguration
-    nodeRegistration:
-      kubeletExtraArgs:
-        "feature-gates": "CPUManager=true"
-        "cpu-manager-policy": "static"
-        "kube-reserved": "cpu=500m"
-        "system-reserved": "cpu=500m"
-EOF
-    done
-}
-
 function _get_nodes() {
     _kubectl get nodes --no-headers
 }
@@ -210,9 +192,40 @@ function setup_kind() {
     prepare_config
 }
 
+function _add_worker_kubeadm_config_patch() {
+    cat << EOF >> ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
+  kubeadmConfigPatches:
+  - |-
+    kind: JoinConfiguration
+    nodeRegistration:
+      kubeletExtraArgs:
+        "feature-gates": "CPUManager=true"
+        "cpu-manager-policy": "static"
+        "kube-reserved": "cpu=500m"
+        "system-reserved": "cpu=500m"
+EOF
+}
+
+function _add_workers() {
+    # appending eventual workers to the yaml
+    for ((n=0;n<$(($KUBEVIRT_NUM_NODES-1));n++)); do
+        cat << EOF >> ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
+- role: worker
+EOF
+    _add_worker_kubeadm_config_patch
+    done
+}
+
+function _prepare_kind_config() {
+    _add_workers
+
+    echo "Final KIND config:"
+    cat ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
+}
+
 function kind_up() {
     _fetch_kind
-    prepare_workers
+    _prepare_kind_config
     setup_kind
 }
 
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -15,6 +15,9 @@ REGISTRY_NAME=${CLUSTER_NAME}-registry
 MASTER_NODES_PATTERN="control-plane"
 WORKER_NODES_PATTERN="worker"
 
+KUBEVIRT_WITH_KIND_ETCD_IN_MEMORY=${KUBEVIRT_WITH_KIND_ETCD_IN_MEMORY:-"true"}
+ETCD_IN_MEMORY_DATA_DIR="/tmp/kind-cluster-etcd"
+
 function _wait_kind_up {
     echo "Waiting for kind to be ready ..."
     while [ -z "$(docker exec --privileged ${CLUSTER_NAME}-control-plane kubectl --kubeconfig=/etc/kubernetes/admin.conf get nodes --selector=node-role.kubernetes.io/master -o=jsonpath='{.items..status.conditions[-1:].status}' | grep True)" ]; do
@@ -154,6 +157,17 @@ function setup_kind() {
     docker cp ${CLUSTER_NAME}-control-plane:/kind/bin/kubectl ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl
     chmod u+x ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl
 
+    if [ $KUBEVIRT_WITH_KIND_ETCD_IN_MEMORY == "true" ]; then
+        for node in $(_get_nodes | awk '{print $1}' | grep control-plane); do
+            echo "[$node] Checking KIND cluster etcd data is mounted to RAM: $ETCD_IN_MEMORY_DATA_DIR"
+            docker exec $node df -h $(dirname $ETCD_IN_MEMORY_DATA_DIR) | grep -P '(tmpfs|ramfs)'
+            [ $(echo $?) != 0 ] && echo "[$node] etcd data directory is not mounted to RAM" && return 1
+
+            docker exec $node du -h $ETCD_IN_MEMORY_DATA_DIR
+            [ $(echo $?) != 0 ] && echo "[$node] Failed to check etcd data directory" && return 1
+        done
+    fi
+
     for node in $(_get_nodes | awk '{print $1}'); do
         docker exec $node /bin/sh -c "curl -L https://github.com/containernetworking/plugins/releases/download/v0.8.5/cni-plugins-linux-amd64-v0.8.5.tgz | tar xz -C /opt/cni/bin"
     done
@@ -230,8 +244,25 @@ EOF
     done
 }
 
+function _add_kubeadm_config_patches() {
+    if [ $KUBEVIRT_WITH_KIND_ETCD_IN_MEMORY == "true" ]; then
+        cat <<EOF >> ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
+kubeadmConfigPatches:
+- |
+  kind: ClusterConfiguration
+  metadata:
+    name: config
+  etcd:
+    local:
+      dataDir: $ETCD_IN_MEMORY_DATA_DIR
+EOF
+        echo "KIND cluster etcd data will be mounted to RAM on kind nodes: $ETCD_IN_MEMORY_DATA_DIR"
+    fi
+}
+
 function _prepare_kind_config() {
     _add_workers
+    _add_kubeadm_config_patches
 
     echo "Final KIND config:"
     cat ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml


cluster-up/cluster/kind/common.sh
```diff
@@ -172,8 +156,10 @@ function setup_kind() {
     docker cp ${CLUSTER_NAME}-control-plane:/kind/bin/kubectl ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl
     chmod u+x ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl
 
+    mkdir -p ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/cni-plugins
+    curl -L https://github.com/containernetworking/plugins/releases/download/v0.8.5/cni-plugins-linux-amd64-v0.8.5.tgz | tar xz -C ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/cni-plugins
     for node in $(_get_nodes | awk '{print $1}'); do
-        docker exec $node /bin/sh -c "curl -L https://github.com/containernetworking/plugins/releases/download/v0.8.5/cni-plugins-linux-amd64-v0.8.5.tgz | tar xz -C /opt/cni/bin"
+        docker cp ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/cni-plugins/. $node:/opt/cni/bin
     done
```
qinqon, line 163: This is not related and leaves a tarball and a directory there, also the number of nodes we usually have is quite small so is like 2 curl + unpacks vs 1 curl + unpack.

cluster-up/cluster/kind/common.sh
```diff
@@ -210,9 +196,91 @@ function setup_kind() {
     prepare_config
 }
 
+function ensure_memory_dir_for_etcd_data() {
+    if [ "$ETCD_DATA_DIR" == none ];then
+        echo "ETCD_DATA_DIR is not defined" && exit 1
```
qinqon, line 201: Better two lines
```bash
echo "ETCD_DATA_DIR is not defined"
exit 1
```

cluster-up/cluster/kind/common.sh
```diff
@@ -15,6 +15,8 @@ REGISTRY_NAME=${CLUSTER_NAME}-registry
 MASTER_NODES_PATTERN="control-plane"
 WORKER_NODES_PATTERN="worker"
 
+ETCD_DATA_DIR=${ETCD_DATA_DIR:-none}
```
qinqon, line 18: You don't need this, just check if the variable set or not before use it
```bash
[  -z "$ETCD_DATA_DIR" ]
```

cluster-up/cluster/kind/common.sh
```diff
@@ -210,9 +196,91 @@ function setup_kind() {
     prepare_config
 }
 
+function ensure_memory_dir_for_etcd_data() {
+    if [ "$ETCD_DATA_DIR" == none ];then
```
qinqon, line 200: ditto: `[ -z "$ETCD_DATA_DIR" ]`

cluster-up/cluster/kind/common.sh
```diff
@@ -210,9 +196,91 @@ function setup_kind() {
...
+            rm -rf $ETCD_DATA_DIR/*
+        fi
+    else
+        echo "creating directory for ectd data"
+        mkdir -p $ETCD_DATA_DIR
+        [ "echo $?" != 0 ] && echo "failed to create etcd data directory" && exit 1
+        ls -lsah $ETCD_DATA_DIR
+
+        echo "mount etcd data directory as tempfs"
+        mount -t tmpfs tmpfs $ETCD_DATA_DIR
```
qinqon, line 218: Maybe I am wrong but when we run this on prow the tmpfs is done by k8s when we crate the volume of typoe `Memory`, right, so we don't need to do this ourselves.

qinqon (changes requested):
Looks great, have some nits and a big question.

cluster-up/cluster/kind/common.sh
ormergi: My intent was to reduce as much time possible for cluster-up to finish so I prefer to curl and unpack one time
and only then copy the content
I can add remove the tar and the directory right after the for loop, is that ok with you?

cluster-up/cluster/kind/common.sh
ormergi: Great question :) 
I added this logic to support development environments flow where you dont have k8s to 
create this mount for you.
On local dev env this function will create a directory and mount it, 
if the mount already exists it will delete the mount content to prevent contradicts with new data.
On CI since we ask from the k8s to create a mount for us,
this function will check that the mount exists, delete the mount content to prevent contradicts with new data
and prints its details

cluster-up/cluster/kind/common.sh
qinqon: Well but this is really not related to run etcd in memory, we can do that at different PR.

cluster-up/cluster/kind/common.sh
ormergi: I added this to common.sh for better visibility of that variable, so if one will review for example `ensure_memory_dir_for_etcd_data` function it will be easier to understand from where this variable comes form
What you think?

cluster-up/cluster/kind/common.sh
qinqon: at local env we have to also remove it after we do cluster-down right ?

cluster-up/cluster/kind/common.sh
ormergi: No problem I will remove this commit

cluster-up/cluster/kind/common.sh
ormergi: You right I will add a clean up for this on cluster-down

cluster-up/cluster/kind/common.sh
qinqon: If you want to do that, then just throw an error here is the variable is not set.

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
```diff
@@ -18,18 +18,11 @@ function up() {
...
-  - containerPath: /lib/modules
-    hostPath: /lib/modules
-    readOnly: true
-  - containerPath: /dev/vfio/
-    hostPath: /dev/vfio/
-EOF
-
-    setup_kind
+
+    export ETCD_DATA_DIR="/mnt/etcd"
```
qinqon, line 22: What will happend if we open more than one kind cluster ? they cannot share the etcd directory. Let's better add something that is unique per kind cluster.

ormergi: Currently we deploy each job at a time on a CI node which means that there could be only one kind cluster at a given time.
I think we should tackle this when we would need to support (if ever) multiple sriov lane jobs on single CI job.
Also it should be changed together with  changes we might what to to do such as handling the cluster name (should be unique), attaching PFs to each cluster (check which is available) and any other change we would need to do.
What you think?

EdDev: But you also have assumptions here which could be avoided.
1. That the prev deployment was properly cleaned.
2. That someone experiments on the node and runs once manually and once through the CI.
3 .The above unintentionally... i.e. bugs.

As the option of this to occur does exist, it makes sense to do it right from the beginning. I would argue that not accounting for this in the first place in the original deployment scripts was the wrong approach, so we should try and not repeat it now.

Of course, you can reason not to do it now because it requires too much effort.
I would also not attempt to do/fix this until all works.

ormergi: I think that we should see if running etcd in memory works for us and actually solve the problem, 
If it will and the lane will be green most of the time we could start to think about more optimizations (running more then one sriov lane job on a single CI node)

Note that the amount of sriov lane jobs  that could run on single CI node is bound to the number of PF's sriov card provides which is two at the current state.

ormergi:
Changes:
- Removed first commit that address cni-plugins downloads
- Added teardown login for ectd data mount that will be invoked during `make cluster-down`

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
oshoval: I am working already working on this, POC soon.

cluster-up/cluster/kind/common.sh
ormergi: Done

ormergi:
The mount is created
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/709/rehearsal-check-up-kind-1.17-sriov/1328628056557883392#1:build-log.txt%3A95

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
qinqon: Now that I think of, if two sriov lanes run at same node they will have different pods so the are not going to share the `/mnt/etcd` but still we have beneficts from using different etcd directories, if it's not too much effort.

Also let's add more info to the name and use something like `/mnt/kubevirt-kind...-[unique name]-etcd`. One thing you can use is the is the container of the first master or the like or the one from the registry container.

qinqon (changes requested):

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
ormergi: I think we should address this when we will work on supporting running multiple jobs on the same CI node
especially when we dont know in what way we are going to do it.
The way I see it, we need to manipulate sriov-cni on the CI cluster so it will attach single PF for each prow job pod
than we dont need to change the current code and  the whole process will be transparent.
And like you said since each job run on the different pod there shouldn't be a problem with the directories
What you think?

I can add another directory for visability so the path will be like `/mnt/kubevirt-kind/etcd`
is that ok?

EdDev: I think @qinqon mentioned that it is not about running multiple lanes in parallel, so I do not understand why you refer to that in your answer.

Please reason why there is no need for an unique, per-run folder name in this case, regardless of usage by other features we may encounter in the future.

ormergi: On CI unique names is not a must since we create in memory directory using emptyDir volume  API (on CI cluster) on the prow job pod https://github.com/kubevirt/project-infra/pull/709

After discussing about it with Quique I realized we unique name for local environments in case cluster-up is stopped in the middle for any reason.
I pushed new commit to resolve this

ormergi:
Changes:
- Use unique names when creating etcd data directory for the running kind cluster
- Add trap to teardown etcd in memory dir mount

cluster-up/cluster/kind/common.sh
```diff
@@ -210,9 +195,120 @@ function setup_kind() {
...
+    if [ "$BASE_ETCD_DATA_DIR" == none ];then
+        echo "BASE_ETCD_DATA_DIR is not defined"
+        exit 1
+    fi
+    if [ "$ETCD_DATA_DIR" == none ];then
+        echo "ETCD_DATA_DIR is not defined"
+        exit 1
+    fi
+    
+    trap cleanup_dir_for_etcd_data ERR SIGINT SIGTERM
```
qinqon, line 208: We can only trigger this is we create it ourselves.

qinqon (changes requested):
I don't see the cluster-down cleanup.

ormergi:
> I don't see the cluster-down cleanup.

This function https://github.com/kubevirt/kubevirtci/pull/478/files#diff-051c8b9fe674143cdeb3ddbd375235788b1f081c7ba111f2f73b55f621fe2d95R319 will be invoked by `make cluster-down`
so if `BASE_ETCD_DATA_DIR` or `ETCD_DATA_DIR` were defined we will call https://github.com/kubevirt/kubevirtci/pull/478/files#diff-051c8b9fe674143cdeb3ddbd375235788b1f081c7ba111f2f73b55f621fe2d95R235-R249
that teardown the mount.

cluster-up/cluster/kind/common.sh
ormergi: I push new commit to skip this if we running on CI

cluster-up/cluster/kind/common.sh
```diff
@@ -324,7 +324,7 @@ function down() {
     $KIND delete cluster --name=${CLUSTER_NAME}
     rm -f ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
 
-    if [ "$ETCD_DATA_DIR" != none ] && [ "$BASE_ETCD_DATA_DIR" != none ];then
+    if [ -d "$BASE_ETCD_DATA_DIR" ];then
```
qinqon, line 327: This is going to be true always

ormergi: Right, I pushed new commit to run teardown only on local environments

ormergi:
Changes:
  Run teardown only on local environments

qinqon:
/hold

Let's simplify this, the issue is only at CI so let's fix only this, it means that for local env we don't mount tmpfs and configure different directory for etcd, but for CI somehow we do the customization, this way we are sure that we don't have to do cleanups or anything like that.

ormergi:
> /hold
> 
> Let's simplify this, the issue is only at CI so let's fix only this, it means that for local env we don't mount tmpfs and configure different directory for etcd, but for CI somehow we do the customization, this way we are sure that we don't have to do cleanups or anything like that.

Actually we encounter the same issues on local environments (with ssd disks) but the symptoms does not affect
tests, we do see `operation took too long` on etcd logs and from time to time we enocunter`etcdserver: timout error` 
but it doesn't cause kubevirt tests fail so we never cared about it.
I guess it is fine for now, If we will encounter worse performance and tests failures on local environment 
we would need to consider mount and teardown in place.

For the sake of keeping things simple I will push new changes to support it on CI.

ormergi:
Changes:
- Remove the logic for setup and teardown tmpfs mount for etcd data.
  Now only if the mount already exists it will configure kind to use it for etcd data.

cluster-up/cluster/kind/common.sh
```diff
@@ -210,9 +194,68 @@ function setup_kind() {
...
+    for ((n=0;n<$(($KUBEVIRT_NUM_NODES-1));n++)); do
+        cat << EOF >> ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
+- role: worker
+EOF
+    _add_worker_kubeadm_config_patch
+    _add_worker_extra_mounts
+    done
+}
+
+function _add_control_plane_extra_mounts() {
```
qinqon, line 235: On local dev we don't need `extraMounts` it just uses /var/lib/etcd from container.

For CI we just mount /var/lib/etcd from host to container without unique name 
```yaml
  extraMounts:
  - containerPath: /var/lib/etcd
    hostPath: /var/lib/etcd
```

qinqon (changes requested):
We can make even simpler we don't need ETCD_DATA_MOUNT at all since local env does not need `extraMounts` and prow job will mount it at `/var/lib/etcd`

ormergi:
> We can make even simpler we don't need ETCD_DATA_MOUNT at all since local env does not need `extraMounts` and prow job will mount it at `/var/lib/etcd`

I tried it but it seems that we don't have access to kind node `var/vib/etcd` from the prow job pod, this is why we need another mount it.

qinqon:
@ormergi don't understand, we cannot mount `Memory` at prow into `/var/lib/etcd` or that we cannot `containerPath: /var/lib/etcd` ?

ormergi:
> @ormergi, what do you mean ? we cannot mount `Memory` at prow into `/var/lib/etcd` or that we cannot use `containerPath: /var/lib/etcd` since kind node container has already the directory in place?

We can mount `/var/lib/etcd` from prow job pod to kind containers only through `extraMount`, 
inside kind containers `/var/lib/etcd` is directory inside the container and does not shard or mounted from the host

But it dosent matter anymore because I found a way to run etcd in memory w/o the need to create a mount! :)

ormergi:
Changes:
- Remove the need to create tmpfs mount to back etcd data directory and teardown and reladed commits 
- Adding kubeadm ClusterConfiguration to kind config yaml that points etcd data dir to `/tmp/...` inside **kind 
  cluster nodes**, which is already mounted as `tmpfs` :).
  To enable it on any kind provider simply add `RUN_ETCD_IN_MEMORY=true` to provider.sh

ormergi:
Tests are passing on kubevirt
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/4583/pull-kubevirt-e2e-kind-1.17-sriov/1330469917769601024

qinqon:
@ormergi this means that we don't need to mount a volume at the prow job ?

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
```diff
@@ -16,20 +16,12 @@ function up() {
...
-  - containerPath: /lib/modules
-    hostPath: /lib/modules
-    readOnly: true
-  - containerPath: /dev/vfio/
-    hostPath: /dev/vfio/
-EOF
-
-    setup_kind
+
+    RUN_ETCD_IN_MEMORY="true"
```
qinqon, line 23: Remove this and leave true as default

cluster-up/cluster/kind/common.sh
```diff
@@ -15,6 +15,8 @@ REGISTRY_NAME=${CLUSTER_NAME}-registry
 MASTER_NODES_PATTERN="control-plane"
 WORKER_NODES_PATTERN="worker"
 
+RUN_ETCD_IN_MEMORY=${RUN_ETCD_IN_MEMORY:-"false"}
```
qinqon, line 18: All the variables that can be set outside of providers are prefixed with KUBEVIRT, let's rename this KUBEVIRT_WITH_KIND_ETCD_IN_MEMORY, also set the default to true at the end is what we want, if there is a problem we can go back to false changing just the prow job.

qinqon (changes requested):

ormergi:
> @ormergi this means that we don't need to mount a volume at the prow job ?
> 
> I understand that what /tmp means inside the kind node depends what the kind project want to do with it, are we sure they are not going to change it ? In the future they can mount a container volume that is writing to disk and then we are at square one.
> 
> Although the solution is better Thant what we have now and is self contained.

Sorry I didnt mean to say the obvious about /tmp I am sure you familiar with how it works.
This was suggested by KIND it also suggested by kind https://github.com/kubernetes-sigs/kind/issues/845#issuecomment-616202792 like we already know etcd will write to disk by default
as far as I know there is no intent to change this API, if they will so like any other API change we will adapt.
I also understand that Kind is also proposed for production and should survive restart (at least)
Even if they will come up with container volume /tmp should be mounted to RAM in any case 
so I think we good

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
ormergi: Done

cluster-up/cluster/kind/common.sh
ormergi: Done

ormergi:
Changes:
- Squashed last two commits and updated the commit message accordingly
- Address comments:
  - Present `KUBEVIRT_WITH_KIND_ETCD_IN_MEMORY` env var to control etcd in memory form kubevirt
  - Add a check at cluster-up (at `kind_up`) that verifies the mount actually exists

cluster-up/cluster/kind/common.sh
```diff
@@ -15,6 +15,9 @@ REGISTRY_NAME=${CLUSTER_NAME}-registry
 MASTER_NODES_PATTERN="control-plane"
 WORKER_NODES_PATTERN="worker"
 
+KUBEVIRT_WITH_KIND_ETCD_IN_MEMORY=${KUBEVIRT_WITH_KIND_ETCD_IN_MEMORY:-"true"}
+ETCD_DATA_DIR="/tmp/kind-cluster-etcd"
```
qinqon, line 19: Rename this to `ETCD_IN_MEMORY_DATA_DIR`.

qinqon, line 19: Rename this to ETCD_IN_MEMORY_DATA_DIR.

cluster-up/cluster/kind/common.sh
```diff
@@ -172,6 +157,13 @@ function setup_kind() {
     docker cp ${CLUSTER_NAME}-control-plane:/kind/bin/kubectl ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl
     chmod u+x ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl
 
+    if [ $KUBEVIRT_WITH_KIND_ETCD_IN_MEMORY == "true" ]; then
+        echo "Checking KIND cluster etcd data is mounted to RAM: $ETCD_DATA_DIR"
+        docker exec "$CLUSTER_NAME-control-plane" df -h $(dirname $ETCD_DATA_DIR)
+        docker exec "$CLUSTER_NAME-control-plane" du -h $ETCD_DATA_DIR
+        [ $(echo $?) != 0 ] && echo "falid to check etcd data directory" && return 1
```
qinqon, line 164: typo: failed

cluster-up/cluster/kind/common.sh
```diff
@@ -172,6 +157,13 @@ function setup_kind() {
     docker cp ${CLUSTER_NAME}-control-plane:/kind/bin/kubectl ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl
     chmod u+x ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl
 
+    if [ $KUBEVIRT_WITH_KIND_ETCD_IN_MEMORY == "true" ]; then
+        echo "Checking KIND cluster etcd data is mounted to RAM: $ETCD_DATA_DIR"
+        docker exec "$CLUSTER_NAME-control-plane" df -h $(dirname $ETCD_DATA_DIR)
+        docker exec "$CLUSTER_NAME-control-plane" du -h $ETCD_DATA_DIR
```
qinqon, line 163: This will check that it's mounted but not that it's mounted in memory, can you check that it's using `tmpfs` ?

qinqon (changes requested):

cluster-up/cluster/kind/common.sh
ormergi: Done

cluster-up/cluster/kind/common.sh
ormergi: fixed

cluster-up/cluster/kind/common.sh
ormergi: No problem

ormergi:
Changes: 
- Addresed comments:
  - Typos
  - Rename ETCD_DATA_DIR
  - Add check that etcd data directory is mounted to RAM memory
- Preform etcd data directory checks on all control-plane nodes

PR #479, opened 2020-11-16
jumbo frames: use an env var to set jumbo frames on the cluster

maiqueb:
Configure the cluster's MTU to use jumbo frames

It will apply jumbo frames on:
  - bridge connecting the nodes (only primary iface)
  - tap device connecting the nodes to the bridge (only primary iface)
  - calico CNI mtu

### DIFF
--- a/cluster-up/cluster/k8s-provider-common.sh
+++ b/cluster-up/cluster/k8s-provider-common.sh
@@ -43,4 +43,12 @@ function up() {
         $kubectl create -f /opt/cnao/network-addons-config-example.cr.yaml
         $kubectl wait networkaddonsconfig cluster --for condition=Available --timeout=200s
     fi
+
+    if [ "$KUBEVIRT_WITH_JUMBO_FRAMES" == "true" ];then
+      client_kubectl=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl
+      $client_kubectl patch configmap/calico-config -n kube-system --type merge \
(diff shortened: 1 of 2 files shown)


cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -43,4 +43,12 @@ function up() {
         $kubectl create -f /opt/cnao/network-addons-config-example.cr.yaml
         $kubectl wait networkaddonsconfig cluster --for condition=Available --timeout=200s
     fi
+
+    if [ "$KUBEVIRT_WITH_JUMBO_FRAMES" == "true" ];then
```
ormergi, line 47: Hey note that according to calico docs the patching new MTU would affect new workloads https://docs.projectcalico.org/networking/mtu#configure-mtu
We need to verify it also change all other pods MTU that created before the patch.
If this is the case you can add another change to calico manifest 
 https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/k8s/1.18/manifests/cni.diff

ormergi (commented):
Hi I though about it and in order to make this work we will probably need to configure docker0 interface inside the prow job pod.
Does kubevirt e2e tests pass on local environment?

rmohr:
> Hi I though about it and in order to make this work we will probably need to configure docker0 interface inside the prow job pod.
> Does kubevirt e2e tests pass on local environment?

If you don't need to reach out to the internet, just changingthe bridge should be good enough. The bridge is in a common network namespae for all pods. Maybe the wiring needs to be adjusted somehow.

PR #481, opened 2020-11-19
sriov, Fix waiting for pod's condition "Ready" logic

oshoval:
Using "wait pods" is prone to wait it's whole timeout
in case the pod, which it is waiting for, restarted.
The reason is that the pod name is changed in case of pod restart,
and the previous waiting condition is outdated
because `wait pods` command samples the pod name only once.

Therefore the changes are:
* By using smaller timeout, and iterations, we bound the time consumed on waiting for dead pods,
allowing the cluster-up to be around 10 minutes shorter when this scenario happens,
because each iteration has a chance to sample the updated pod name.
* Updated the logic to stop cluster-up with an error, in case the pod waiting condition didn't meet eventually.
Otherwise, if the mandatory condition wasn't met, cluster-up would still continue, hiding this fact,
and prone for errors.
* Reduce waiting time to a more reasonable value (changed from 10m to 5m).
Since usually waiting up to 1m is enough.

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -150,6 +150,23 @@ function wait_pods_ready {
     done
 }
 
+function wait_pods_condition_by_label {
+  local -r tries=10
+  local -r wait_time=1
+  local -r namespace=$1
+  local -r label=$2
+  local -r condition=$3
+
+  local -r wait_message="waiting for pod label $label at namespace $namespace to meet condition $condition"
+  local -r error_message="pod label $label at namespace $namespace did not meet condition $condition"
+  local -r action="_kubectl wait pods -n $namespace -l $label --for condition=$condition --timeout 30s >/dev/null && echo succeed"
+
+  if ! retry "$tries" "$wait_time" "$action" "$wait_message"; then
+    echo $error_message
+    return 1
+  fi
+}
+
 function wait_allocatable_resource {
   local -r node=$1
   local resource_name=$2
@@ -238,6 +255,7 @@ function apply_sriov_node_policy {
   SRIOV_OPERATOR_NAMESPACE="sriov-network-operator"
   SRIOV_DEVICE_PLUGIN_LABEL="app=sriov-device-plugin"
   SRIOV_CNI_LABEL="app=sriov-cni"
+  CONDITION="Ready"
 
   echo "Applying SriovNetworkNodeConfigPolicy:"
   cat $policy_file
@@ -254,8 +272,8 @@ function apply_sriov_node_policy {
   wait_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_DEVICE_PLUGIN_LABEL || return 1
 
   # Wait for cni and device-plugin pods to be ready
-  _kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l $SRIOV_CNI_LABEL           --for condition=Ready --timeout 10m
-  _kubectl wait pods -n $SRIOV_OPERATOR_NAMESPACE -l $SRIOV_DEVICE_PLUGIN_LABEL --for condition=Ready --timeout 10m
+  wait_pods_condition_by_label $SRIOV_OPERATOR_NAMESPACE $SRIOV_CNI_LABEL $CONDITION || return 1
+  wait_pods_condition_by_label $SRIOV_OPERATOR_NAMESPACE $SRIOV_DEVICE_PLUGIN_LABEL $CONDITION || return 1
 
   # Since SriovNodeNetworkPolicy doesnt have Status to indicate if its
   # configured successfully, it is necessary to wait for the "NoSchedule"


cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -150,6 +150,18 @@ function wait_pods_ready {
...
+function wait_pods_ready_by_label {
+    local -r tries=10
+    local -r namespace=$1
+    local -r label=$2
+
+    ITERATION=0
+    while [ ! "$(_kubectl wait pods -n $namespace -l $label --for condition=Ready --timeout 1m)" ] && [ $ITERATION -lt $tries ]; do
+        echo "Waiting for $label (namespace $namespace) to become ready"
+        ITERATION=$((ITERATION+1))
+    done
```
oshoval, line 162: need to fix here return 1 in case of error

oshoval:
Not related

qinqon:
/hold

Let's not merge this by mistake

qinqon:
@oshoval since we are waitting by label are we sure that if pods restart kubectl wait on label is not going to view those new.

oshoval:
> @oshoval since we are waitting by label are we sure that if pods restart kubectl wait on label is not going to view those new.

If i understood the question right,
In case of pod restart the bug was:

```
t0 - Wait for pod according label.
t1 - pod get restarted, and gets new name
... After 10 minutes
t10 - timeout on the old name.
```

With the fix, we will timeout after 1 minute, and let it 10 times instead 1 time waiting 10m.
So in case of restart of even couple restart, triggering the command again, will resample the new pod name,
and it will be totally fine.

```
t0 - Wait for pod according label.
t1 - Pod get restarted, and gets new name
... After 1 minutes
t2 - Wait for pod according label .... get the new name
t3 - Condition met.
```

Usually 1 minute is enough from what i see locally, but in case there is a restart it will just wait for 10 minutes before the fix.

oshoval:
> /hold
> 
> Let's not merge this by mistake

Can you please add the hold reason for visibility and so we remember easier ?
From what i understand, the reason is because the lane is broken globally right now.

oshoval:
@qinqon 
check-provision-k8s-1.17 passed

oshoval:
@qinqon 
check-up-kind-1.17-sriov passed

ormergi:
wait pods is kubectl standard command that waits for a condition and we use labels to get the pods as you saw.
Can you refer to a job where we encounter sriov-cni or sriov-device-plugin pods get restarted and the job fail due to that?

oshoval:
> wait pods is kubectl standard command that waits for a condition and we use labels to get the pods as you saw.
> Can you refer to a job where we encounter sriov-cni or sriov-device-plugin pods get restarted and the job fail due to that?

It happens locally, all the time, even the code itself says that in other function that doesn't use labels [1].
I can send you in slack a screen capture that it happened, or you can if you want just kill a pod directly after u restart the wait
and see this phenomena please.

[1]
```
# not using kubectl wait since with the sriov operator the pods get restarted a couple of times and this is
# more reliable
function wait_pods_ready {
    while [ -n "$(_kubectl get pods --all-namespaces -o'custom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers | grep false)" ]; do
        echo "Waiting for all pods to become ready ..."
        _kubectl get pods --all-namespaces -o'custom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers
        sleep 10
    done
}
```

oshoval:
/hold cancel
Correct me if i am wrong please, the hold reason is outdated, since the lane passed.

EdDev (approved):
Thanks!

EdDev:
Wat is [this](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/481/check-provision-k8s-1.18/1329363780320104448#1:build-log.txt%3A3713) error?

oshoval:
> Wat is [this](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/481/check-provision-k8s-1.18/1329363780320104448#1:build-log.txt%3A3713) error?

`error: timed out waiting for the condition on pods/calico-node-bpcht`
on other lane (k8s-1.18), and i touched just `config_sriov.sh`

oshoval:
/hold 
i want to validate something, injecting an error for the loop

ormergi:
> > wait pods is kubectl standard command that waits for a condition and we use labels to get the pods as you saw.
> > Can you refer to a job where we encounter sriov-cni or sriov-device-plugin pods get restarted and the job fail due to that?
> 
> It happens locally, couple times.
> The job doesn't fail because there isn't `return 1`, but it stalls for 10 minutes during cluster-up.
> This is why we didn't saw it on CI, unless we will look closely and might see it.
> The `wait` doesn't refresh the pod name until timeout, in case of pod restart.
> 
> * Even the code itself says that in other function that doesn't use labels [1] (see comment please "not using kubectl wait ...").
> * I can send you in slack a screen capture that it happened if you want.
> * You can as well just kill a pod directly after you start the wait and see this phenomena please.
> 
> [1]
> 
> ```
> # not using kubectl wait since with the sriov operator the pods get restarted a couple of times and this is
> # more reliable
> function wait_pods_ready {
>     while [ -n "$(_kubectl get pods --all-namespaces -o'custom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers | grep false)" ]; do
>         echo "Waiting for all pods to become ready ..."
>         _kubectl get pods --all-namespaces -o'custom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers
>         sleep 10
>     done
> }
> ```

The title of this PR is misleading because it is not reduce the time, it presents new function to check for pods to be ready,
I am saying this because one could be biased by the title of how to approach this PR.
Before we created the `wait_for_..` kind of functions, wait_pod used to ensure in each step that the pod are ready.
Also after applying the policy where new pods are created, and if a node is restarted then pods will be restarted (not only sriov operator pods)
Now we didnt encounter on CI any case where sriov-cni and device-plugin pods restarted and there was no job that failed due to that.
From my side I can confirm I never saw this happen locally this is why I dont find it comes in hand.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -150,6 +150,25 @@ function wait_pods_ready {
     done
 }
 
+function wait_pods_ready_by_label {
+    local -r tries=10
+    local -r namespace=$1
+    local -r label=$2
+
+    ITERATION=0
+    while [ $ITERATION -lt $tries ] && [ ! "$(_kubectl wait pods -n $namespace -l $label --for condition=Ready --timeout 1m)" ]; do
```
ormergi, line 159: All other polling functions in this file make use of `retry` for example [wait_for_k8s_object](https://github.com/kubevirt/kubevirtci/pull/481/files#diff-499cff045d8a875e8b39f4a9fd59354eb99e8f8d91a95ebc70fbf42d2d5b5beaR84) 
this was the idea of presenting that function and when we added it iirc you asked for it on your review
Why you dont make use of it here?

ormergi (changes requested):

oshoval:
/hold cancel
Error injecting is safe

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: Its true that i asked it, but not for "wait pods", just for the other things.
I wanted to preserve the elegant use of "wait pods" logic here.
Every thing can be done in several methods, and i did as close as `wait_pods_ready` which already exists,
but preserved the "wait pods".

oshoval:
> > > wait pods is kubectl standard command that waits for a condition and we use labels to get the pods as you saw.
> > > Can you refer to a job where we encounter sriov-cni or sriov-device-plugin pods get restarted and the job fail due to that?
> > 
> > 
> > It happens locally, couple times.
> > The job doesn't fail because there isn't `return 1`, but it stalls for 10 minutes during cluster-up.
> > This is why we didn't saw it on CI, unless we will look closely and might see it.
> > The `wait` doesn't refresh the pod name until timeout, in case of pod restart.
> > 
> > * Even the code itself says that in other function that doesn't use labels [1] (see comment please "not using kubectl wait ...").
> > * I can send you in slack a screen capture that it happened if you want.
> > * You can as well just kill a pod directly after you start the wait and see this phenomena please.
> > 
> > [1]
> > ```
> > # not using kubectl wait since with the sriov operator the pods get restarted a couple of times and this is
> > # more reliable
> > function wait_pods_ready {
> >     while [ -n "$(_kubectl get pods --all-namespaces -o'custom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers | grep false)" ]; do
> >         echo "Waiting for all pods to become ready ..."
> >         _kubectl get pods --all-namespaces -o'custom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers
> >         sleep 10
> >     done
> > }
> > ```
> 
> The title of this PR is misleading because it is not reduce the time, it presents new function to check for pods to be ready,
> I am saying this because one could be biased by the title of how to approach this PR.
> Before we created the `wait_for_..` kind of functions, wait_pod used to ensure in each step that the pod are ready.
> Also after applying the policy where new pods are created, and if a node is restarted then pods will be restarted (not only sriov operator pods)
> Now we didnt encounter on CI any case where sriov-cni and device-plugin pods restarted and there was no job that failed due to that.
> From my side I can confirm I never saw this happen locally this is why I dont find it comes in hand.

I see it almost every day locally, and suggested a method for you to simulate it.
There is an explanation why we might missed it on CI, as it didn't have `return 1` and was gone silenty.
The title is the core change of this PR, we waited for no reason in case of a pod restart.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: Please see that `retry` doesn't suit the use i need according the comment:

# This function gets a command string and invoke it
# until the command returns an empty string or until timeout

If it returns an empty string i do need to continue.
It shows that we should keep it simple when we can, and stick to existing standard commands (such as "wait").

EdDev: @oshoval I do not think you answered the question presented here.
I did not remember that `retry` function, which seems a total fit here.

You have re-implemented a retry with a `while`. Please try to use the existing `retry` function or explain why it does not fit here technically.

EdDev (changes requested):
Let me start with a request to stop quoting the whole history and focus on the specific section you answer.

> Now we didnt encounter on CI any case where sriov-cni and device-plugin pods restarted and there was no job that failed due to that.
> From my side I can confirm I never saw this happen locally this is why I dont find it comes in hand.

You are saying it never failed on waiting for these pods? That is strange by itself.
But it does no matter, if the wrong operation was used, it needs to be fixed. Using `wait` does not work with restarted pods if I understand correctly, so unless you claim we should fail if a reset happen, this change logically looks fine to me.

> There is an explanation why we might missed it on CI, as it didn't have return 1 and was gone silenty.
> The title is the core change of this PR, we waited for no reason in case of a pod restart.

The subject is a bit misleading because when you read it, you may think that you reduced the time in case of a pod restart, but in practice, if that ever happened, it would have just failed. Now it will retry and actually may continue on to succeed instead of failing after 10min. I hope you understand what I was trying to say.

You can just put in the title that actual fix and in the body explain the scenario and intent.

EDIT: Are you saying that we may have waiting 10min and still continued?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
EdDev: At the same time, we should not have multiple "retry" implementations, that is wrong.
Try to see if you can adjust one or the other to fit here.

oshoval: Please see retry docs
```
# This function gets a command string and invoke it
# until the command returns an empty string or until timeout
```
The function doesn't fit naturally, unless the command will be (added the `|| echo error` in order for it to fit)
`_kubectl wait pods -n $namespace -l $label --for condition=Ready --timeout 1m || echo error`

Moreoever we already have `wait_pods_ready` which use a while,
If you still think i should use `retry` i will do that, but i think it doesn't suit here.

oshoval:
@EdDev 
About your question here at the end
https://github.com/kubevirt/kubevirtci/pull/481#pullrequestreview-534481940
See please the answer in the PR description, a `return 1` was missing, this why this 10m waiting was not visible.

> Updated `wait_pods_ready_by_label` to stop cluster-up with an error,
> in case of failure.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
EdDev: You should either generalize `retry` better or adjust your function to fit it.
The comment about not creating another "retry" method is valid.

oshoval:
Adapted to use `ready`

oshoval:
/hold 
Policy not found, will make sure first its not this PR.

oshoval:
`2020/11/19 16:44:17 failed to create secret 'operator-webhook-service': etcdserver: request timed out`

`2020/11/19 16:44:26 failed to create secret 'operator-webhook-service': secrets "operator-webhook-service" already exists
`
Various errors about creating secret that result in a failure.
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/481/check-up-kind-1.17-sriov/1329459762047225856

oshoval:
Removed redundant grep, as the STDOUT would be already non empty in case of success,
and in case a retry needed it will be empty, and `retry` would trigger a retry.

Removing hold
as so far the errors i saw on CI were due to other errors.
and locally it passes.
/hold cancel

oshoval:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/481/check-up-kind-1.17-sriov/1329471968495276032

Passed

oshoval:
Changes:
* Improve retry action:
In order to support waiting multi pods matching a label, we should check the error code and not the STDOUT,
else some pods might success and some fail, but STDOUT would still be non empty and the `retry` would stop
too early.

---

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -150,6 +150,21 @@ function wait_pods_ready {
     done
 }
 
+function wait_pods_condition_by_label {
+  local -r tries=20
+  local -r wait_time=3
```
EdDev, line 155: It costs now 20*30sec + 20*3sec = 660sec = 11min
You added another minute in case of failure.

oshoval: Just in case of absolute failure, which rarely happens, but now is protected as well.
In case of success, even with pod restart it saves a lot of time.

Usually it takes just 1 minute actually locally, because it continue once condition is met
i just didn't want to play with the times
I will change it to wait_time=1, and tries=15 i believe it will be totally safe.

oshoval:
Update timings

oshoval:
Changes:
Updated timings
---

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
EdDev: Well, changing the time diverges from the main objective of this change.
But 10m is also a very long time, so it makes sense to reduce it.

Anyway, whatever you do, you need to mention this in the commit message and PR description so it will be clear this is intentional and you need to reason for it.

oshoval:
Changes: Update timing constrains to 5m
Updated PR description.

````

**Comment by oshoval** [comment]
Changes: Update timing constrains to 5m
Updated PR description.

**Comment by oshoval** [comment]
Changes: Update timing constrain to 5m
Updated PR description, and commit message.

---

**Review by EdDev — APPROVED** [review]

**Comment by oshoval** [comment]

File: cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh [hunk 4]
```diff
@@ -150,6 +150,21 @@ function wait_pods_ready {
...
+  local -r namespace=$1
+  local -r label=$2
+  local -r condition=$3
+
+  local -r wait_message="waiting for pod label $label at namespace $namespace to meet condition $condition"
+  local -r error_message="pod label $label at namespace $namespace did not meet condition $condition"
+  local -r action="_kubectl wait pods -n $namespace -l $label --for condition=$condition --timeout 30s >/dev/null && echo succeed"
+
+  retry "$tries" "$wait_time" "$action" "$wait_message" && return 0
+  echo $error_message && return 1
```
qinqon, line 165: Let's use a if statement to make things less "oneliner"

```bash
if ! retry "$tries" "$wait_time" "$action" "$wait_message"; then
    echo $error_message
    return 1
fi
```

qinqon (changes requested):
Looks good just one nit

oshoval:
Changes: use one liner
---

oshoval:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/481/check-up-kind-1.17-sriov/1331202623604264960
Passed

EdDev (approved):
Thanks!

PR #486, opened 2020-11-22
sriov, Support running multi clusters side by side

oshoval:
Add support for running two sriov clusters side by side, locally, with few known limitations.

The feature allows:
* Utilize all the available sriov PFs, since `Kubevirtci` lane require only one PF.
* Context switch easier between clusters in case needed, 
without the need to tear down, recreate cluster and re-sync for example `Kubevirt`.

General considerations:

- Each cluster should be created on its own git clone folder, i.e
`/root/project/kubevirtci1`
`/root/project/kubevirtci2`

This feature with it's default values doesn't conflict with prow sriov CNI.
Once prow sriov CNI will allocate one PF to each job, Kubevirtci would be naturally robust to this change.

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
@@ -4,7 +4,7 @@ Provides a pre-deployed k8s cluster with version 1.17.0 that runs using [kind](h
 The KubeVirt containers are built on the local machine and are then pushed to a registry which is exposed at
 `localhost:5000`.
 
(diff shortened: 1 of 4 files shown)


oshoval:
@EdDev 
FYI, this PR works locally with 2 clusters, and on prow it works on all the nodes
(1 cluster on prow at a time, because the under cloud CNI allocate them all to the single job that exists).
Based on #500 (bump to 4.8) and fixing also the reset cloud using the WIP PR of sriov-operator.

oshoval:
Failed due to dev7 ConnectX-5 bug, will take care of it on other PR, not related just to this PR.

oshoval:
Rebased

PR #490, opened 2020-11-24
sriov, Fix certsecret backward compatibility

oshoval:
Use local imports so we can run certsecret
even on machines with older go (1.12.14 for example).

This method is also better, because it uses explicitly
the local certlib, and doesn't point to an external folder,
where the certlib might be different.

Without this change, updating go to 1.13.x was a must
Tried the following successful update:
go1.12.14 to go1.13.15.

Tested on kubevirt as well locally.


oshoval:
go.mod need to be changed as well

PR #492, opened 2020-11-25
SRIOV Provider fine tunes: Simplify Readiness Checks

ormergi:
On SRIOV provider we deploy sriov-network-opertator, this deployment has two phases
- Operator deployment;
  - Build and create all necessary objects for the operator.
  - Create secrets, certificats and patch webhooks configurations for the operator webhooks.
- Creating SriovNetworkNodePolicy, configuring PF and VF's according to the policy, deploys `sriov-cni` and `sriov-device-plugin` pods
At the currents state there are many asserts during the operator deployment.

After a while we monitor and test the process of deploying sriov-network-operator 
there are few asserts that we can get rid of:
- Remove checks for taints being placed and removed in both phases
- Remove checking for pods being create and become ready after policy creation (phase 2)
- Remove checking SriovNetworkNodePolicy object is created (phase 2)

This PR intent is to simplify how we deploy and check for readiness of sriov operator
 w/o compromising on the provider reliability making it less procedural then it is now.

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -144,14 +144,40 @@ function wait_for_taint {
   echo $error_message && return 1
 }
 
+function _check_all_pods_ready() {
+  all_pods_ready_condition=$(_kubectl get pods -A --no-headers -o custom-columns=':.status.conditions[?(@.type == "Ready")].status')
+  if [ "$?" -eq 0 ]; then
+    pods_not_ready_count=$(grep -cw False <<< "$all_pods_ready_condition")
+    if [ "$pods_not_ready_count" -eq 0 ]; then
+      return 0
+    fi
+  fi
+
+  return 1
+}
+
 # not using kubectl wait since with the sriov operator the pods get restarted a couple of times and this is
 # more reliable
 function wait_pods_ready {
-    while [ -n "$(_kubectl get pods --all-namespaces -o'custom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers | grep false)" ]; do
-        echo "Waiting for all pods to become ready ..."
-        _kubectl get pods --all-namespaces -o'custom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers
-        sleep 10
-    done
+  local -r tries=30
+  local -r wait_time=10
+
+  local -r wait_message="Waiting for all pods to become ready.."
+  local -r error_message="Not all pods were ready after $(($tries*$wait_time)) seconds"
+
+  local -r get_pods='_kubectl get pods --all-namespaces'
+  local -r action="_check_all_pods_ready"
+
+  set +x
+  trap "set -x" RETURN
+
+  if ! retry "$tries" "$wait_time" "$action" "$wait_message" "$get_pods"; then
+    echo $error_message
+    return 1
+  fi
+
+  echo "all pods are ready"
+  return 0
 }
 
 function wait_pods_condition_by_label {
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -11,6 +11,7 @@ MASTER_NODE="${CLUSTER_NAME}-control-plane"
 WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
 
 OPERATOR_GIT_HASH=8d3c30de8ec5a9a0c9eeb84ea0aa16ba2395cd68  # release-4.4
+SRIOV_OPERATOR_NAMESPACE="sriov-network-operator"
 
 # This function gets a command and invoke it repeatedly
 # until the command return code is zero
@@ -61,30 +62,6 @@ function wait_for_daemonSet {
   echo $error_message && return 1
 }
 
-function wait_pod {
-  local namespace=$1
-  local label=$2
-
-  local -r tries=60
-  local -r wait_time=5
-
-  local -r wait_message="Waiting for pods with $label to create"
-  local -r error_message="Pods with  label $label at $namespace namespace found"
-
-  if [[ $namespace != "" ]];then
-    namespace="-n $namespace"
-  fi
-
-  if [[ $label != "" ]];then
-    label="-l $label"
-  fi
-
-  local -r action="_kubectl get pod $namespace $label -o custom-columns=NAME:.metadata.name --no-headers"
-
-  retry "$tries" "$wait_time" "$action" "$wait_message" && return  0
-  echo $error_message && return 1
-}
-
 function wait_k8s_object {
   local -r object_type=$1
   local -r name=$2
@@ -104,44 +81,6 @@ function wait_k8s_object {
 
   retry "$tries" "$wait_time" "$action" "$wait_message" && return 0
   echo $error_message && return  1
-}
-
-
-function is_taint_absence {
-  local -r taint=$1
-
-  result=$(_kubectl get nodes -o jsonpath="{.items[*].spec.taints[?(@.effect == \"$taint\")].effect}" || echo error)
-  if [[ -z $result ]]; then
-    echo "not-present"
-  fi
-}
-
-function wait_for_taint_absence {
-  local -r taint=$1
-
-  local -r tries=60
-  local -r wait_time=5
-
-  local -r wait_message="Waiting for $taint taint absence"
-  local -r error_message="Taint $taint $name did not removed"
-  local -r action="is_taint_absence $taint"
-
-  retry "$tries" "$wait_time" "$action" "$wait_message" && return 0
-  echo $error_message && return 1
-}
-
-function wait_for_taint {
-  local -r taint=$1
-
-  local -r tries=60
-  local -r wait_time=5
-
-  local -r wait_message="Waiting for $taint taint to present"
-  local -r error_message="Taint $taint $name did not present"
-  local -r action="_kubectl get nodes -o custom-columns=taints:.spec.taints[*].effect --no-headers | grep -i $taint"
-
-  retry "$tries" "$wait_time" "$action" "$wait_message" && return 0
-  echo $error_message && return 1
 }
 
 function _check_all_pods_ready() {
@@ -180,35 +119,17 @@ function wait_pods_ready {
   return 0
 }
 
-function wait_pods_condition_by_label {
-  local -r tries=10
-  local -r wait_time=1
-  local -r namespace=$1
-  local -r label=$2
-  local -r condition=$3
-
-  local -r wait_message="waiting for pod label $label at namespace $namespace to meet condition $condition"
-  local -r error_message="pod label $label at namespace $namespace did not meet condition $condition"
-  local -r action="_kubectl wait pods -n $namespace -l $label --for condition=$condition --timeout 30s >/dev/null && echo succeed"
-
-  if ! retry "$tries" "$wait_time" "$action" "$wait_message"; then
-    echo $error_message
-    return 1
-  fi
-}
-
 function wait_allocatable_resource {
   local -r node=$1
   local resource_name=$2
   local -r expected_value=$3
 
-  local -r tries=30
+  local -r tries=48
   local -r wait_time=10
 
   local -r wait_message="wait for $node node to have allocatable resource: $resource_name: $expected_value"
   local -r error_message="node $node doesnt have allocatable resource $resource_name:$expected_value"
 
-  # In order to project spesific resource name with -o custom-columns
   # it is necessary to add '\' before '.' in the resource name.
   resource_name=$(echo $resource_name | sed s/\\./\\\\\./g)
   local -r action='_kubectl get node $node -ocustom-columns=:.status.allocatable.$resource_name --no-headers | grep -w $expected_value'
@@ -272,37 +193,13 @@ function deploy_sriov_operator {
 }
 
 function apply_sriov_node_policy {
-  policy_file=$1
+  local -r policy_file=$1
 
-  SRIOV_OPERATOR_NAMESPACE="sriov-network-operator"
-  SRIOV_DEVICE_PLUGIN_LABEL="app=sriov-device-plugin"
-  SRIOV_CNI_LABEL="app=sriov-cni"
-  CONDITION="Ready"
-
-  echo "Applying SriovNetworkNodeConfigPolicy:"
-  cat $policy_file
   # Substitute $NODE_PF and $NODE_PF_NUM_VFS and create SriovNetworkNodePolicy CR
-  envsubst < $policy_file | _kubectl create -f -
-
-  # Ensure SriovNetworkNodePolicy CR is created
-  policy_name=$(cat $policy_file | grep -Po '(?<=name:) \S*')
-  wait_k8s_object "SriovNetworkNodePolicy" "$policy_name" "$SRIOV_OPERATOR_NAMESPACE"  || return 1
-
-  # Wait for sriov-operator to reconcile SriovNodeNetworkPolicy
-  # and create cni and device-plugin pods
-  wait_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_CNI_LABEL  || return 1
-  wait_pod $SRIOV_OPERATOR_NAMESPACE $SRIOV_DEVICE_PLUGIN_LABEL || return 1
-
-  # Wait for cni and device-plugin pods to be ready
-  wait_pods_condition_by_label $SRIOV_OPERATOR_NAMESPACE $SRIOV_CNI_LABEL $CONDITION || return 1
-  wait_pods_condition_by_label $SRIOV_OPERATOR_NAMESPACE $SRIOV_DEVICE_PLUGIN_LABEL $CONDITION || return 1
-
-  # Since SriovNodeNetworkPolicy doesnt have Status to indicate if its
-  # configured successfully, it is necessary to wait for the "NoSchedule"
-  # taint to present and then absent.
-  taint="NoSchedule"
-  wait_for_taint "$taint" || echo "Taint $taint did not present on nodes after creating SriovNodeNetworkPolicy"
-  wait_for_taint_absence "$taint" || return  1
+  local -r policy=$(envsubst < $policy_file)
+  echo "Applying SriovNetworkNodeConfigPolicy:"
+  echo "$policy"
+  _kubectl create -f - <<< "$policy"
 
   return 0
 }
@@ -362,14 +259,13 @@ wait_pods_ready
 deploy_sriov_operator
 wait_pods_ready
 
-# Substitute NODE_PF and NODE_PF_NUM_VFS then create SriovNetworkNodePolicy CR
 policy="$MANIFESTS_DIR/network_config_policy.yaml"
-apply_sriov_node_policy "$policy" || exit 1
-wait_pods_ready
+apply_sriov_node_policy "$policy"
 
 # Verify that sriov node has sriov VFs allocatable resource
-resource_name=$(cat $policy | grep 'resourceName:' | awk '{print $2}')
-wait_allocatable_resource $SRIOV_NODE "openshift.io/$resource_name" $NODE_PF_NUM_VFS || exit 1
+resource_name=$(sed -n 's/.*resourceName: *//p' $policy)
+wait_allocatable_resource $SRIOV_NODE "openshift.io/$resource_name" $NODE_PF_NUM_VFS
+wait_pods_ready
 
 _kubectl get nodes
 _kubectl get pods -n $SRIOV_OPERATOR_NAMESPACE
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -145,10 +145,8 @@ function deploy_multus {
   echo 'Waiting for Multus deployment to become ready'
   daemonset_name=$(cat $MANIFESTS_DIR/multus.yaml | grep -i daemonset -A 3 | grep -Po '(?<=name:) \S*amd64$')
   daemonset_namespace=$(cat $MANIFESTS_DIR/multus.yaml | grep -i daemonset -A 3 | grep -Po '(?<=namespace:) \S*$' | head -1)
-  wait_k8s_object "daemonset" $daemonset_name $daemonset_namespace || return 1
-
-  required_replicas=$(_kubectl get daemonset kube-multus-ds-amd64 -n  kube-system -o jsonpath='{.status.desiredNumberScheduled}')
-  wait_for_daemonSet $daemonset_name $daemonset_namespace $required_replicas || return 1
+  required_replicas=$(_kubectl get daemonset $daemonset_name -n $daemonset_namespace -o jsonpath='{.status.desiredNumberScheduled}')
+  wait_for_daemonSet $daemonset_name $daemonset_namespace $required_replicas
 
   return 0
 }
@@ -253,7 +251,7 @@ ${SRIOV_NODE_CMD} mount -o remount,rw /sys     # kind remounts it as readonly wh
 ${SRIOV_NODE_CMD} chmod 666 /dev/vfio/vfio
 _kubectl label node $SRIOV_NODE sriov=true
 
-deploy_multus || exit 1
+deploy_multus
 wait_pods_ready
 
 deploy_sriov_operator
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -236,7 +236,7 @@ function deploy_sriov_operator {
   echo 'Downloading the SR-IOV operator'
   operator_path=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/sriov-network-operator-${OPERATOR_GIT_HASH}
   if [ ! -d $operator_path ]; then
-    curl -L https://github.com/openshift/sriov-network-operator/archive/${OPERATOR_GIT_HASH}/sriov-network-operator.tar.gz | tar xz -C ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/
+    curl -LSs https://github.com/openshift/sriov-network-operator/archive/${OPERATOR_GIT_HASH}/sriov-network-operator.tar.gz | tar xz -C ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/
   fi
 
   echo 'Installing the SR-IOV operator'
@@ -254,27 +254,19 @@ function deploy_sriov_operator {
 
   echo 'Generating webhook certificates for the SR-IOV operator webhooks'
   pushd "${CERTCREATOR_PATH}"
-    go run . -namespace sriov-network-operator -secret operator-webhook-service -hook operator-webhook -kubeconfig $KUBECONFIG_PATH || return 1
-    go run . -namespace sriov-network-operator -secret network-resources-injector-secret -hook network-resources-injector -kubeconfig $KUBECONFIG_PATH || return 1
+    go run . -namespace sriov-network-operator -secret operator-webhook-service -hook operator-webhook -kubeconfig $KUBECONFIG_PATH
+    go run . -namespace sriov-network-operator -secret network-resources-injector-secret -hook network-resources-injector -kubeconfig $KUBECONFIG_PATH
   popd
 
   echo 'Setting caBundle for SR-IOV webhooks'
-  wait_k8s_object "validatingwebhookconfiguration" "operator-webhook-config" || return 1
+  wait_k8s_object "validatingwebhookconfiguration" "operator-webhook-config"
   _kubectl patch validatingwebhookconfiguration operator-webhook-config --patch '{"webhooks":[{"name":"operator-webhook.sriovnetwork.openshift.io", "clientConfig": { "caBundle": "'"$(cat $CERTCREATOR_PATH/operator-webhook.cert)"'" }}]}'
 
-  wait_k8s_object "mutatingwebhookconfiguration"   "operator-webhook-config" || return 1
+  wait_k8s_object "mutatingwebhookconfiguration"   "operator-webhook-config"
   _kubectl patch mutatingwebhookconfiguration operator-webhook-config --patch '{"webhooks":[{"name":"operator-webhook.sriovnetwork.openshift.io", "clientConfig": { "caBundle": "'"$(cat $CERTCREATOR_PATH/operator-webhook.cert)"'" }}]}'
 
-  wait_k8s_object "mutatingwebhookconfiguration"   "network-resources-injector-config" || return 1
+  wait_k8s_object "mutatingwebhookconfiguration"   "network-resources-injector-config"
   _kubectl patch mutatingwebhookconfiguration network-resources-injector-config --patch '{"webhooks":[{"name":"network-resources-injector-config.k8s.io", "clientConfig": { "caBundle": "'"$(cat $CERTCREATOR_PATH/network-resources-injector.cert)"'" }}]}'
-
-  # Since sriov-operator doesnt have a condition or Status to indicate if
-  # 'operator-webhook' and 'network-resources-injector' webhooks certificates are
-  # configured, in order to check if caBundle reconcile is finished it is necessary
-  # to wait for the "NoSchedule" taint to present and then absent.
-  taint="NoSchedule"
-  wait_for_taint "$taint" || echo "Taint $taint did not present on nodes after setting caBundle for sriov webhooks"
-  wait_for_taint_absence "$taint" || return 1
 
   return 0
 }
@@ -367,7 +359,7 @@ _kubectl label node $SRIOV_NODE sriov=true
 deploy_multus || exit 1
 wait_pods_ready
 
-deploy_sriov_operator || exit 1
+deploy_sriov_operator
 wait_pods_ready
 
 # Substitute NODE_PF and NODE_PF_NUM_VFS then create SriovNetworkNodePolicy CR
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -1,5 +1,5 @@
-#!/bin/bash -e
-set -x
+#!/bin/bash
+set -xe
 
 source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
 
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -49,7 +49,8 @@ function wait_for_daemonSet {
   fi
 
   if (( required_replicas < 0 )); then
-      echo "DaemonSet $name ready replicas number is not valid: $required_replicas" && return 1
+      echo "DaemonSet $name ready replicas number is not valid: $required_replicas"
+      return 1
   fi
 
   local -r tries=30
@@ -58,8 +59,12 @@ function wait_for_daemonSet {
   error_message="DaemonSet $name did not have $required_replicas ready replicas"
   action="_kubectl get daemonset $namespace $name -o jsonpath='{.status.numberReady}' | grep -w $required_replicas"
 
-  retry "$tries" "$wait_time" "$action" "$wait_message" && return  0
-  echo $error_message && return 1
+  if ! retry "$tries" "$wait_time" "$action" "$wait_message";then
+    echo $error_message
+    return 1
+  fi
+
+  return  0
 }
 
 function wait_k8s_object {
@@ -79,8 +84,12 @@ function wait_k8s_object {
 
   local -r action="_kubectl get $object_type $name $namespace -o custom-columns=NAME:.metadata.name --no-headers"
 
-  retry "$tries" "$wait_time" "$action" "$wait_message" && return 0
-  echo $error_message && return  1
+  if ! retry "$tries" "$wait_time" "$action" "$wait_message";then
+    echo $error_message
+    return  1
+  fi
+
+  return 0
 }
 
 function _check_all_pods_ready() {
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -103,69 +82,19 @@ function wait_k8s_object {
...
+  local -r wait_message="Waiting for all pods to become ready.."
+  local -r error_message="Not all pods were ready after $(($tries*$wait_time)) seconds"
+  local -r get_pods_status="cluster-up/kubectl.sh get pods --all-namespaces -ocustom-columns=status:status.conditions[?(@.type==\"Ready\")].status,metadata:metadata.name --no-headers"
+  local -r action='[ -n $($get_pods_status | grep -i false) ] && echo all-pods-ready'
 
-  if ! retry "$tries" "$wait_time" "$action" "$wait_message"; then
-    echo $error_message
-    return 1
-  fi
+  retry "$tries" "$wait_time" "$action" "$wait_message" "$get_pods_status" && return  0
```
qinqon, line 96: Better keep the if statement, oneliners are error prone.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -190,14 +118,14 @@ function wait_allocatable_resource {
 
 function deploy_multus {
   echo 'Deploying Multus'
-  _kubectl create -f $MANIFESTS_DIR/multus.yaml
+  _kubectl create -f $MANIFESTS_DIR/multus.yaml || return 1
```
qinqon, line 121: Isn't this script running with -e?

qinqon (changes requested):
Looks pretty good, I am all for removing code, although let's see if we can remove oneliners, a lot of them are related to not using -e or the like.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: It appears that  adding `|| exit 1`to helper function calls interfere with `-e` flag and somehow the error that occurred on the helper wont propagate.
I will remove `|| exit 1` when we call to helpers and remove`|| return 1` inside helpers since we dont need it.

ormergi:
Changes:
- Remove `|| return 1` usage inside `wait_for..` function and `|| exit 1` usage over the script
- Changes conditional return one-liners to if statements
- Updated commit messages

ormergi:
/hold 

Lets test it few more times to be sure it is stabilized

qinqon:
@ormergi CI is failing
```
+++ grep -w 16
+++ /home/prow/go/src/github.com/kubevirt/kubevirtci/_ci-configs/kind-k8s-sriov-1.17.0/.kubectl --kubeconfig=/home/prow/go/src/github.com/kubevirt/kubevirtci/_ci-configs/kind-k8s-sriov-1.17.0/.kubeconfig get node sriov-worker '-ocustom-columns=:.status.allocatable.openshift\.io/sriov_net' --no-headers
+ result=
+ return 1
+ echo node sriov-worker doesnt have allocatable resource openshift.io/sriov_net:16
node sriov-worker doesnt have allocatable resource openshift.io/sriov_net:16
+ return 1
Makefile:5: recipe for target 'cluster-up' failed
make: *** [cluster-up] Error 1
teardown
./cluster-up/down.sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
 
```

ormergi:
> node sriov-worker doesnt have allocatable resource openshift.io/sriov_net:16

On it

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -99,93 +83,54 @@ function wait_k8s_object {
...
+  local -r tries=30
+  local -r wait_time=10
 
-  local -r wait_message="waiting for pod label $label at namespace $namespace to meet condition $condition"
-  local -r error_message="pod label $label at namespace $namespace did not meet condition $condition"
-  local -r action="_kubectl wait pods -n $namespace -l $label --for condition=$condition --timeout 30s >/dev/null && echo succeed"
+  local -r wait_message="Waiting for all pods to become ready.."
+  local -r error_message="Not all pods were ready after $(($tries*$wait_time)) seconds"
+  local -r get_pods_status="cluster-up/kubectl.sh get pods --all-namespaces -ocustom-columns=status:status.conditions[?(@.type==\"Ready\")].status,metadata:metadata.name --no-headers"
+  local -r action='[[ -z $($get_pods_status | grep -i false) ]] && echo all-pods-ready'
```
oshoval, line 103: In case `$($get_pods_status | grep -i false)` fails, the condition will be still true, since STDOUT might be empty. 
The right method is the look also on the return code, and only if it succeed, to check its stdout + grep please

oshoval (changes requested):

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -99,93 +83,54 @@ function wait_k8s_object {
...
 
-  if ! retry "$tries" "$wait_time" "$action" "$wait_message"; then
+  if ! retry "$tries" "$wait_time" "$action" "$wait_message" "$get_pods_status"; then
```
oshoval, line 105: It runs the get_pods_status twice, one as action and one as the new waiting_action,
I think the design should be different, so i am not going into questions like what happen if `waiting_action` fails due to an error.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -99,93 +83,54 @@ function wait_k8s_object {
...
     echo $error_message
     return 1
   fi
+  
+  return 0
```
oshoval, line 110: Redundant, all over, without explicit return, the function will return the ret code of the last command (which is not "hidden").
Unless you want to be explicit.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -225,63 +168,31 @@ function deploy_sriov_operator {
...
+    go run . -namespace sriov-network-operator -secret network-resources-injector-secret -hook network-resources-injector -kubeconfig $KUBECONFIG_PATH
   popd
 
   echo 'Setting caBundle for SR-IOV webhooks'
-  wait_k8s_object "validatingwebhookconfiguration" "operator-webhook-config" || return 1
+  wait_k8s_object "validatingwebhookconfiguration" "operator-webhook-config"
   _kubectl patch validatingwebhookconfiguration operator-webhook-config --patch '{"webhooks":[{"name":"operator-webhook.sriovnetwork.openshift.io", "clientConfig": { "caBundle": "'"$(cat $CERTCREATOR_PATH/operator-webhook.cert)"'" }}]}'
-
-  wait_k8s_object "mutatingwebhookconfiguration"   "operator-webhook-config" || return 1
+
```
oshoval, line 178: Please drop

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -225,63 +168,31 @@ function deploy_sriov_operator {
...
-
-  # Since SriovNodeNetworkPolicy doesnt have Status to indicate if its
-  # configured successfully, it is necessary to wait for the "NoSchedule"
-  # taint to present and then absent.
-  taint="NoSchedule"
-  wait_for_taint "$taint" || echo "Taint $taint did not present on nodes after creating SriovNodeNetworkPolicy"
-  wait_for_taint_absence "$taint" || return  1
+  local -r policy=$(envsubst < $policy_file)
+  echo "Applying SriovNetworkNodeConfigPolicy:"
+  echo "$policy"
```
oshoval, line 194: with set -x that we have, it will already show the policy in the next line

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: When `kubectl` fails on `get` operations it throws an error and a message to STDOUT
We shouldn't add another operation to that action

oshoval: It throw it to STDERR, and the grep doesn't catch this.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: This is exactly what should happen, invoking `$action` as the conditional for the retry loop 
and `$get_pods_status` as an action to run during the retry loop along with `$wait_message`.
Before those changes we did the same on `wait_pods_ready` but it ran forever..

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I prefer it to be explicit so in case one of the operations inside the function will print a message to STDOUT
the function will return a value that is only 1 or 0

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: Lets first solve this one please
https://github.com/kubevirt/kubevirtci/pull/492#discussion_r533975306
and then rethink, 

The design isn't right imo.
It shouldn't run it twice, one good which cover all the cases is enough.
and even if it did, we need to inject a `false` as the `waiting_action` to see what happens in case there is a kubectl error.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: The output from your example is 
```
$ [[ -z $(cluster-up/kubectl.sh get pods | grep configuration) ]] && echo bla
No resources found in default namespace.
bla
```
in fact if you try to get pods from a namespace that has no pods you would expect to get an error 
but acctially the return code will be '0' and the output will be
```
$ kubectl get po
No resources found in default namespace.
$ echo $?
0
```

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: if will surely return 0, otherwise the `if` above it would return an error
```
#/bin/bash -ex

foo ()
{
  echo bla
  false
  if ! true; then
    echo 11
  fi
}

foo
echo $?
```
This one return 0, as it look on the last command, even that it had `false` above it.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -225,63 +168,31 @@ function deploy_sriov_operator {
 
   echo 'Generating webhook certificates for the SR-IOV operator webhooks'
   pushd "${CERTCREATOR_PATH}"
-    go run . -namespace sriov-network-operator -secret operator-webhook-service -hook operator-webhook -kubeconfig $KUBECONFIG_PATH || return 1
```
oshoval, line 228: please inject here a false, i think that it won't return (according what i tried, unless i am wrong)

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: Please use my command, with real error, without KUBECONFIG and without cluster.
and see the bottom line, you can't relay on a failed command and then `grep -i words`

ormergi: This is also correct if you try to get pod from non existing namespace
```
$ kubectl get po -n blabla
No resources found in blabla namespace.
$ echo $?
0
```

oshoval: I don't want the command to success, i shown an example of a failed command, where the grep doesn't catch STDERR.
and asking please to focus on the bottom line only,
the PR as its now relays on a command that might fail, without taking this possible situation into account, but proceeding without even retry as it should.

ormergi: I am not aware to a scenario where we get to config_sriov.sh script to `wait_pods_ready` w/o a cluster or a kubeconfig
if we do, `wait_pods_ready` will return an error.
```
# kubectl does not exists
$ wait_pods_ready 
bash: ./kubectl: No such file or directory

# no kubeconfig
$ wait_pods_ready 
W1202 11:43:04.521749   11169 loader.go:223] Config not found: asdThe connection to the server localhost:8080 was refused - did you specify the right host or port?

# not valid kubeconfig
$ wait_pods_ready 
error: error loading config file "/root/go/src/kubevirt.io/kubevirt/_ci-configs/kind-k8s-sriov-1.17.0/.kubeconfig": v1.Config.Contexts: []v1.NamedContext: Clusters: []v1.NamedCluster: v1.NamedCluster.Name: Cluster: v1.Cluster.Server: CertificateAuthorityData: decode base64: illegal base64 data at input byte 1371, error found in #10 byte of ...|0tLQoasd=","server":|..., bigger context ...|p4R2pmST0K

# no cluster
$  wait_pods_ready 
The connection to the server 127.0.0.1:32767 was refused - did you specify the right host or port?
```
Those changes already been reviews and approved, CI is green and the cases you mention tested manually on local setup.

If you insist this should be change lets hear other reviewers thoughts.

oshoval: > I am not aware to a scenario where we get to config_sriov.sh script to `wait_pods_ready` w/o a cluster or a kubeconfig
> if we do, `wait_pods_ready` will return an error.

It was just for the example that a failed command will pass silently on the refereed line.

> Those changes already been reviews and approved, CI is green and the cases you mention already been tested manually on a local setup.
> If you insist this should be change lets hear other reviewers thoughts.

I totally insist, its a bug.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: The issues you mention on this comment already resloved from my side and we cover the cases you mention there.
https://github.com/kubevirt/kubevirtci/pull/492#discussion_r533975306

This design enable running another action during the retry loop.
In `wait_pods_ready` function case, we can get the state of the pods during the retry loop 
which is crucial debugging.

oshoval: In case the command fail it will pass silently on the action, without even retrying (see the ref discussion please).
and also silently according what i tested locally on the `waiting_action` (please inject an error and verify)
see here what i tested https://github.com/kubevirt/kubevirtci/pull/492#discussion_r534007650
the best is to inject it in the real one

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: config_sriov.sh runs with `-e` flag, if `certcreator` will raise an error this script will be stopped,
this is already been tested and the script actually stops.

using `|| return 1` after operations inside helpers and `|| exit 1` on calling for helpers, interfere with bash catching the errors although `-e` flag is up, this is why it used to continue on errors.

oshoval: Please see here
https://github.com/kubevirt/kubevirtci/pull/481/files#diff-499cff045d8a875e8b39f4a9fd59354eb99e8f8d91a95ebc70fbf42d2d5b5beaR275

This one is a function, in the same file, and when there was the bug that that PR solved (locally is saw it),
it continued even that the command failed, so i would suggest to inject error in all the changed places and understand exactly what happens.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I will elaborate on how `wait_pods_ready` works 
`wait_pods_ready` make use of `retry` and as you already know it will stop and return 0 once the action we pass does not write to STDOUT
We cant  relay on the return code when working with `kubectl` sometimes it doesn't return error code as you would expect it to return (discussed on other threads)

`[[ -z $($get_pods_status | grep -i false) ]] && echo all-pods-ready` passed as the `action` to `retry`
if there is a pods that is not ready, it will print `all-pods-ready` to STDOUT and `retry` will continue.
It will stop once all pods are ready, no 'false' found by grep, print 'all-pods-ready' to STDOUT and `retry` stop.

The way I see it `wait_pods_ready` acts as expected and there is no actual problem here
in case there is no kubeconfig or a cluster, cluster-up would stop before we get to this script.

@qinqon what you think?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: keep the discussion on the `action` part at the thread you mention the comment.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I tested it on a local environment and saw that the script stops if there is an error inside or by the helpers functions

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: This is a nit that we can leave with I prefer not to rebase for a single tab on a bash script

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: there are 2 action items here, which have in common but have also unique.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I prefer to stay explicit in this script helper functions at least for functions that use `retry`

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: imo please test each of them please.
as i also saw problems locally for sure, at least on some of them.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: i don't agree, its up to the other reviewers.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: It break the one liner approach which was suggested by Quique in the same context imo.
there is a different between explicit and redundant, it might be subjective.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: You complain about the design that you dont like, is there is something actionable here is should be dont on another PR

oshoval: I complain about the design that was added in this PR, `waiting_action`
and about the correctness of it.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Notice that git diff shows that `wait_pods_condition_by_label` is changed but it actually removed and `wait_pods_ready`
is changed.
I was talking about `wait_pods_ready`

This PR intent is to simplify the way we indicate that the cluster is ready after sriov-operator installation, see PR description
long story short we relay on checking that all pods are ready and we cant just get into infinity loop.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Quique asked to remove the conditional return one liners because they were prune to errors.
We already in a script so we dont benefit much from one liners

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: I totally in the loop and understand what PR does,
i am just saying that 
`  local -r action="_kubectl wait pods -n $namespace -l $label --for condition=$condition --timeout 30s >/dev/null && echo succeed"`
is exactly what will solve the all you need in this thread.

instead of those
```
  local -r get_pods_status="cluster-up/kubectl.sh get pods --all-namespaces -ocustom-columns=status:status.conditions[?(@.type==\"Ready\")].status,metadata:metadata.name --no-headers"
  local -r action='[[ -z $($get_pods_status | grep -i false) ]] && echo all-pods-ready'
```
and the problems and the refactor of retry which has problems imo (on the other thread).

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: As I said it already been tested, if you have other insights that suggest a change please share it.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: I am talking about https://github.com/kubevirt/kubevirtci/pull/481#discussion_r529459786

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: one time that `waiting_action` fails, will fail the whole script, and won't retry, even that its in `retry` function

```
kubevirt (rand) $ cat foo.sh 
#!/bin/bash -e
set -x

get_pods_status="false"
function foo {
  echo bla
  $get_pods_status
  echo end
}

foo
echo $?
```

```
kubevirt (rand) $ ./foo.sh 
+ get_pods_status=false
+ foo
+ echo bla
bla
+ false
kubevirt (rand) $ echo $?
1
kubevirt (rand) $ 
```

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: I just re-simulated it locally (see diff in the PR please), created this PR to see if it happens on CI
https://github.com/kubevirt/kubevirtci/pull/494

oshoval: please see that it continue even that the command failed, also on CI
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/494/check-up-kind-1.17-sriov/1334144011677470720#1:build-log.txt%3A1860

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Yeah I am familiar with that comment Quique asked for the same on this PR 
I understand what you see it redundant , I mean that since we use `return 1` we should also use `return 0` to be explicit

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Now that you put it like that I understand
I also tested `wait_pods_ready` and saw it dosnt propagates the errors, good catch!

I like what you suggest, I tested it and it but seems that using 
`kubectl wait pods -A --all --for condition=Ready --timeout 0s` takes about 2~3 seconds to finish while using `kubectl pods get` takes less then a second, it means that it will take for `wait_pods_ready` extra 45\~75 seconds to finish.

I started to investigate how to do it w/o using `kubectl wait .. --for` and it seems that there is no simple kubectl command to get all pods that are not ready, where not ready means at least one container is not ready.
I have managed to make use of `kubectl get pods .. ` combined with `grep` by extracting it to another function and handle the errors there, so no extra time added and now `wait_pods_ready` propagates the errors properly.
it was tested with and w/o kubeconfig, when all pods ready and when there is a pod that not all of its containers are ready.
I will push it with next commits.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I tried to reproduce what you suggest with setting `waiting_action` `df -h /asd`, `kubectl` w/o kubeconfig and `kubectl1`
it dosent reproduce and `retry` loop continues
You can try to copy `retry` and `wait_pods_ready` functions to the example script and see that its not terminate

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I will remove it on next rebase

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: its because for some reason the functions which have error doesnt quit,
but it means that you cant remove the `|| return 1` / `exit 1`
as stated on the other comments please.
else it would be a regression

Moreover the design here is not good, its not needed to retest this command in the middle of the retry,
if there is a timeout you wont see it (as you said now), and if if succeed it will succeed also on the action itself.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: thanks, btw see that git diff shows it with a red marker,
i think i saw at least two of those when tried to apply it manually

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: This is what i also shown an example of that happens,
please dont remove the `return 1`
as it will be a regression, you can either leave them or understand the root cause imo,
since function indeed should obey the `set -x`

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: I don't understand why it would take that much, and i prefer to use please the most keep it simple approach.
we see here that using the `wait pods` can solve at once the two threads,
this one and the one that changed the retry function, and has some problems imo (see the thread of that please).

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: see https://github.com/kubevirt/kubevirtci/pull/492#discussion_r537452325

oshoval: Since you said you checked and it stops if there is an error in a function,
but i saw the two behaviors, i raised it again with an example.

Lets be synced on the action items please.
We can't just remove `return 1` and the `exit 1`,
Agree ?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I noticed that if you set the script with `#!/bin/bash -e` it has no affect can it considered as a comment
you need to set it like any other flag using `set`.
On `config_sriov.sh` script `-e` takes affect because it is configured on king/common.sh that is sourced at the strat of the script. 
I tested this script and if you run a command that will fails the script stops.
Whats unique with `retry` is that we invoke it from `if` closure, 
or have the following addition before this PR change:
```
retry ... && return 0
return 1
``` 
`if` does not propagate the errors the condition may raise,
so errors that raised by `action` and `waiting_action` will not terminate `retry` or the script.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Gotcha, please take another look on the thread that I refer to I answered about the `return 1` and `exit 1`,

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: There are couple issues here:
About retry function:
1. The design is not good, the new waiting_action isn't needed see please reasons above.
Once you fix the wait_pods function as on the other thread lets discuss it again because it depends how you fix it.

About the rest of the code:
See please
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/494/check-up-kind-1.17-sriov/1334144011677470720#1:build-log.txt%3A1860
We can't remove the guards (return 1 / exit 1)

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: I took and answered.
My comment stands
here is a proof on the non retry functions:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/494/check-up-kind-1.17-sriov/1334144011677470720#1:build-log.txt%3A1860

ormergi:
Changes:
- Change `retry` loop condition (check return code instead of STDOUT)
- Export `wait_pods_ready` action to a function for better error handling
- Remove redundant tabs all over
- Set `-e` flag explicitly on `config_sriov.sh`

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: `If` doesn't have anything to do with it, just if its in the if line itself, not on the body,
nor eval would help.

Aligned this example to your latest commits,
Run this locally please and see, that here it does stop (at least locally when i tried),
and you can't allow yourself for it to stop on retry,
and on the other functions it should stop (or at least to have guard as we had).

so this is the situation we should get
```
#!/bin/bash
set -ex

get_pods_status="false"
function foo {
  echo bla
  if true; then
    eval $get_pods_status
  fi
  echo end
}

foo
echo $?
```
and it will stop on the eval line and not echo end.

I am putting aside for a moment that issue of the design of the waiting_action,
i looked only on this change.

ormergi: This PR changes are not reflected at the job you refer to it doesnt prove what you trying to, try with the latest changes.
1. retry, why you think waiting_action is not necessary?
2. `wait_pods_ready` it is fixed now.
3. At the job you refer to the script continues because there is `|| exit 1` right after the call for `apply_sriov_node_policy` 
    It was a mistake to add `|| return/exit 1` in the first place since we use `-e` flag.
4. Using `if retry .. then` or `retry || return 1`, it does have affect on `retry` whether it will stopping or not 
     you can try `if failing_command; then ..` on a script with `-e` and you will see the script doesnt stop
5. about the example, I am not sure what you try to test here I assume it is retry, 
   and if it does you need to run `foo` with `&& return 0 \n return 1`.
   This is how we use `retry` on this script.

ormergi:
Changes: 
- removed `./.kubectl` leftovers

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: 1. I found the change far from keep it simple, i suggested please to use the previous wait pods with `>/dev/null && echo succeed` and then you wont need  any refactor that have several levels as you did now.
The approach suggested right now with the multi level functions, won't get my lgtm.
If you will want later to change the retry to use retcode it merits a standalone focused PR please, not in this PR, 
with too many moving elements.
2. It even didn't pass yet on CI (just not you pushed a change that might fix it).
3. I shown an example that it continue and it should not continue.
i dont understand how you say there is exit 1 on the example i shown that do continue, its a contradiction,
see please only what i added in that POC PR.
The issue there that the functions must not continue, but the retry should not stop.
Please inject errors as i did in the proof and see if it works once its stabilized, else its not a good enough imo.
I will not lgtm without injecting errors.
4. i just tried above please see, again i am saying that it matter if the command is in the if line or in the body,
following your last change, its in the body.
5. Lets just fallback to the action item, please inject errors all over (as 3 says), it was a idea for you that the if body / eval wont help.

ormergi: 1. Changing retry condition to relay on return code actually make things more simple,
    its harder to grasp what using `> /dev/null && echo succeed mean`
    Using `kubectl wait` takes more time and we are about reducing the time it takes to spin up a cluster, 
    Unfortunately there is no simple kubectl command to get all pods that are not ready,
    adding another function that does that actually making it simpler to understand.
3. I am talking about the job you refer to which is the PR you created.
    It doesnt contain this PR changes this is why you see it continues where it shouldn't, try it with this PR changes and you 
    will see it stops.
    Also I tested the latest changes of this PR locally and it behave as expected
4. Well there are two ways to do it, using proper `if` or what I mentioned in the previous comment
    it acts  the same.
5. this section is about the example 

I tested it locally and the script stops as we expect

ormergi: 1. Changing retry condition to relay on return code actually make things more simple,
    its harder to grasp what using `> /dev/null && echo succeed mean`
    Using `kubectl wait` takes more time and we are about reducing the time it takes to spin up a cluster, 
    Unfortunately there is no simple kubectl command to get all pods that are not ready,
    adding another function that does that actually making it simpler to understand.
3. I am talking about the job you refer to which is the PR you created.
    It doesnt contain this PR changes this is why you see it continues where it shouldn't, try it with this PR changes and you 
    will see it stops.
    Also I tested the latest changes of this PR locally and it behave as expected
4. Well there are two ways to do it, using proper `if` or what I mentioned in the previous comment
    it acts  the same.
5. this section is about the example 

I tested it locally and the script stops as we expect

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: It dosent proof what you say, try it again with this PR changes and you will see it stops.
Since we are using `-e` flag there is no need to add `|| return/exit 1` (unless you want to consider only a return code of a function) it was a mistake to add it in the first place.

oshoval: This is after you fixed, before that it was broken.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: about 1
i find this
https://github.com/kubevirt/kubevirtci/compare/7c4c152fe237945e0654b55e8b350d617ea32201..3229c3e908c267e13576224de7583645dec652da#diff-499cff045d8a875e8b39f4a9fd59354eb99e8f8d91a95ebc70fbf42d2d5b5beaR95
much harder to grasp

ormergi: It gets all pods condition Ready status, then checks 
1. there was no error getting the pods 
2. passing the pods statuses to grep with `<<<` (piping `echo` to grep is also an option), 
   `grep -c` counts how many times 'False' appears in the string
   then we check if it equals zero, it means all pods are ready and return 0
   otherwise return 1

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I removed `|| return/exit 1` from the beginning so it would work for you if you would check with the PR changes.

oshoval: This was the fix that fixed it from what i know, and it was after the example of this thread.
https://github.com/kubevirt/kubevirtci/compare/7c4c152fe237945e0654b55e8b350d617ea32201..3229c3e908c267e13576224de7583645dec652da

The removal of the return/exit didnt fix it, the changing of the set fixed it.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: Thanks but the code should be self explanatory, i do understand the code if i want,
but i think it should be written more readable and faster to understand.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: The commit that adds explicit `set -e` actually has no affect, since it is already configured by sourcing `kind/common.sh` at the beginning of the script (it is configured there).
Setting that flag with `#!/bin/bash -e` will not take affect.

I added this commit so it will be explicit that we configure it on that script instead of relaying on other files.
I also mentioned it in the commit message.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Well bash has its drawback and it is readability, but I understand what you say I will try to make it more readable.

ormergi:
changes: 
- Make `check_all_pod_ready` more readable

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I tested it and saw it takes more time, actually the time related to the number of pods you have in the cluster.

ormergi:
check-up-kind-1.17-sriov

PR #496, opened 2020-12-09
Add k8s provider for 1.20

dhiller:


dhiller:
@rmohr I have a strange error in provisioning [here](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/496/check-provision-k8s-1.20/1336598312324698112#1:build-log.txt%3A2678), mind taking a look?

I think it fails [here](https://github.com/dhiller/kubevirtci/blob/ea2c81d6c36442ba6b7ec30da8cfbd3405335cb0/cluster-provision/gocli/cmd/provision.go#L146-L145), but I don't get why...

rmohr:
> I think it fails [here](https://github.com/dhiller/kubevirtci/blob/ea2c81d6c36442ba6b7ec30da8cfbd3405335cb0/cluster-provision/gocli/cmd/provision.go#L146-L145), but I don't get why...

Looks to me like it actually tries to start the qemu process. You are seeing the logs from the started container. Does the generated commandline look wrong, or why do you think it this place?

dhiller:
> > I think it fails [here](https://github.com/dhiller/kubevirtci/blob/ea2c81d6c36442ba6b7ec30da8cfbd3405335cb0/cluster-provision/gocli/cmd/provision.go#L146-L145), but I don't get why...
> 
> Looks to me like it actually tries to start the qemu process. You are seeing the logs from the started container. Does the generated commandline look wrong, or why do you think it this place?

I was of the impression that as we were seeing the output of the vm script and the volume configuration there would be happening something strange here. However, I can't make sense of this.

rmohr:
> I was of the impression that as we were seeing the output of the vm script and the volume configuration there would be happening something strange here. However, I can't make sense of this.

@dhiller I have no better idea than running it locally and checking all container logs there too.

cluster-provision/k8s/1.20/provision.sh
```diff
@@ -0,0 +1,327 @@
...
+net.ipv6.conf.all.forwarding = 1
+net.bridge.bridge-nf-call-ip6tables = 1
+EOF
+sysctl --system
+
+echo bridge >> /etc/modules
+echo br_netfilter >> /etc/modules
+
+# Bump NetworkManager to 1.22.8
+NM_VERSION=1.22.8
```
rmohr, line 155: THis is the issue:

```
No match for argument: NetworkManager-1.22.8
Error: Unable to find a match: NetworkManager-1.22.8
```

you will have to bump it to `1.30.0`.

qinqon: Can we make it 1.26 ? if not kubernetes-nmstate has to downgrade to it.

rmohr: What do you mean with downgrade? The suggested 1.30 is newer and I can't see a 1.26 in the centos  stream: http://mirror.23media.com/centos/8-stream/BaseOS/x86_64/os/Packages/. Only 1.30.

qinqon: ok, let's just remove this check, if needed at kubernetes-nmstate CI we can ping to 1.26 or the like using copr.

ormergi: Currently check-provision-1.19 lane is failing and blocking PR's due to that update.
I think we should drop this NM update or add different repository to get the pinned version,
because we bound to what is provided with Cetos Stream

rmohr: I am not yet removing the check, but this bump PR #497 solves that.

dhiller:
@rmohr I've found the reason. Due to the [deprecation of `kubeadm init --experimental-kustomize`](https://github.com/kubernetes/kubernetes/pull/94871) the patches for selinux didn't get applied. I changed [to use --experimental-patches with strategic merge patches](https://github.com/kubevirt/kubevirtci/pull/496/files#diff-45436c2820ee8fc4393329609ec40c376d2834d1493f54628fa467f7f3313efeR181) as recommended, seems to work. Not sure, is there a better way?

rmohr:
/lgtm

Let's make it required once this PR is in.

rmohr:
Also let's make a follow-up where we remove the networkmanager hardcode.

PR #497, opened 2020-12-10
Bump Z versions of k8s clusters

rmohr:


rmohr:
/cc @qinqon 

I had to bump networkmanager, do we still have to pin it?

cluster-provision/k8s/1.19/provision.sh
```diff
@@ -152,7 +152,7 @@ echo bridge >> /etc/modules
 echo br_netfilter >> /etc/modules
 
 # Bump NetworkManager to 1.22.8
-NM_VERSION=1.22.8
+NM_VERSION=1.30.0
 dnf install -y NetworkManager-$NM_VERSION
```
ormergi, line 156: Hey as I mentioned on other thread, k8s-1.19 lane is failing due to missing NM pinned version.
This change does solve it, but I afraid that eventually the pinned version will not be provided anymore 
by Centos Stream after few NM releases, and we will have those failures again.
@qinqon @rmohr what you think?

rmohr: yes, we should remove it. I just want this to get in.

dhiller (approved):
/lgtm
/approve
/hold

This looks pretty complete, right? Thanks for that!

rmohr:
> This looks pretty complete, right? Thanks for that!

The e2e tests for 1.19 failed. Try to find out why. The tests ran into a timeout.

PR #498, opened 2020-12-11
Move etcd on k8s-1.19 to in-memory storage

rmohr:
Following the example of #478, move etcd on one of the providers to in-memory. Locally running the kubevirt e2e tests show that 300M are more than enough for etcd. After this is running for a while in kubevirt/kubevirt, we can decide if we move all clusters over.


fgimenez (approved):
/lgtm
/approve

qinqon:
/lgtm
/approve

Let's test this on a kubevirt/kubevirt bump.

PR #501, opened 2020-12-16
sriov, Refactor movement of sriov PFs to the network namespace

oshoval:
The logic that moves the sriov PFs to the containerized network namespace
of the kind node, was part of the main flow.
Extracting it to a function, in order to improve readability
and modularity.
    
Modularity is improved by adding parameters to the function,
such as the target node.

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -205,14 +205,40 @@ function deploy_sriov_operator {
 
 function apply_sriov_node_policy {
   local -r policy_file=$1
+  local -r node_pf=$2
+  local -r num_vfs=$3
 
   # Substitute $NODE_PF and $NODE_PF_NUM_VFS and create SriovNetworkNodePolicy CR
-  local -r policy=$(envsubst < $policy_file)
+  local -r policy=$(NODE_PF=$node_pf NODE_PF_NUM_VFS=$num_vfs envsubst < $policy_file)
   echo "Applying SriovNetworkNodeConfigPolicy:"
   echo "$policy"
   _kubectl create -f - <<< "$policy"
 
   return 0
+}
+
+function move_sriov_pfs_netns_to_node {
+  local -r node=$1
+  local -r pid="$(docker inspect -f '{{.State.Pid}}' $node)"
+  local pf_array=()
+
+  mkdir -p /var/run/netns/
+  ln -sf /proc/$pid/ns/net "/var/run/netns/$node"
+
+  local -r sriov_pfs=( /sys/class/net/*/device/sriov_numvfs )
+  for ifs in "${sriov_pfs[@]}"; do
+    local ifs_name="${ifs%%/device/*}"
+    ifs_name="${ifs_name##*/}"
+
+    if [ $(echo "${PF_BLACKLIST[@]}" | grep "${ifs_name}") ]; then
+      continue
+    fi
+
+    ip link set "$ifs_name" netns "$node"
+    pf_array+=($ifs_name)
+  done
+
+  echo ${pf_array[@]}
 }
 
 # The first worker needs to be handled specially as it has no ending number, and sort will not work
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -215,6 +215,37 @@ function apply_sriov_node_policy {
...
+    # to inject into the Network Configuration manifest. We need to move all pfs to
+    # the node's namespace and for that reason we do not interrupt the loop.
+    if [ -z "$NODE_PF" ]; then
+      # These values are used to populate the network definition policy yaml.
+      # We just use the first suitable pf
+      # We need the num of vfs because if we don't set this value equals to the total, in case of mellanox
+      # the sriov operator will trigger a node reboot to update the firmware
+      export NODE_PF="$ifs_name"
+      export NODE_PF_NUM_VFS=$(cat /sys/class/net/"$NODE_PF"/device/sriov_totalvfs)
+    fi
```
EdDev, line 244: This part is a side effect in this new function.
It has nothing to do with the movement of the PF to the netns of the node.

One option is for the function to return the list of interfaces that moved and from there to continue at the caller.

Also, except this variable, the others seem to be used only in this function so why do we need the "exported"? Declaring them local will make it clearer that they are used only here.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -215,6 +215,37 @@ function apply_sriov_node_policy {
   return 0
 }
 
+function setns_sriov_interfaces {
```
EdDev, line 218: How about this name + args: `move_sriov_pfs_netns_to_node`, using as an argument the node name.

With this the comment in the next line can be dropped.

EdDev (commented):
This change refactors the script by extracting the movement of the SRIOV PF/s to the containerized node network namespace.
Please try to express this in the commit message and add your reasoning to do so (e.g. readability, modularity).

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: > This part is a side effect in this new function.
> It has nothing to do with the movement of the PF to the netns of the node.
> 
> One option is for the function to return the list of interfaces that moved and from there to continue at the caller.
> 
> Also, except this variable, the others seem to be used only in this function so why do we need the "exported"? Declaring them local will make it clearer that they are used only here.

Both `NODE_PF NODE_PF` ,`NODE_PF` need to be exported, since they are used with envsubst
https://github.com/kubevirt/kubevirtci/pull/501/files#diff-499cff045d8a875e8b39f4a9fd59354eb99e8f8d91a95ebc70fbf42d2d5b5beaL209
Will change the others to be local.

EdDev: Now that this is a function, I hope you see why this is problematic.
Please try to leave this function with no side-effects. As this code needs to be taken out now, I hope it will sort everything out.

oshoval: >This part is a side effect in this new function.
It has nothing to do with the movement of the PF to the netns of the node.

We will need deeper refactor, because once we moved the pf, in case we need to get NODE_PF_NUM_VFS,
we will need to find it via `/sys/devices/pci` instead of `/sys/class/net/`.
I will try to do that in case you think it deserves it?

> Also, except this variable, the others seem to be used only in this function so why do we need the "exported"? Declaring them local will make it clearer that they are used only here.

Both `NODE_PF NODE_PF`, `NODE_PF` need to be exported, since they are used with envsubst
```
  # Substitute $NODE_PF and $NODE_PF_NUM_VFS and create SriovNetworkNodePolicy CR
  local -r policy=$(envsubst < $policy_file)
```
Will change the others to be local.

EdDev: > We will need deeper refactor, because once we moved the pf, we will able to get NODE_PF_NUM_VFS,
> only via `/sys/devices/pci` instead of `/sys/class/net/`.
> (unless we peek in the container)
> I will try to do that in case you think it deserves it.

I thought the this suggestion should handle it:
>> One option is for the function to return the list of interfaces that moved and from there to continue at the caller.

> > Also, except this variable, the others seem to be used only in this function so why do we need the "exported"? Declaring them local will make it clearer that they are used only here.
> 
> Both `NODE_PF NODE_PF`, `NODE_PF` need to be exported, since they are used with envsubst
> 
> ```
>   # Substitute $NODE_PF and $NODE_PF_NUM_VFS and create SriovNetworkNodePolicy CR
>   local -r policy=$(envsubst < $policy_file)
> ```
> 
> Will change the others to be local.

I think I answered this already. Now that this is a function, doing this from here is a side effect and seems to me like a blocker to extracting this section into its own function.
The prev point hopefully should resolve this. If it does not, a new strategy is needed here.

oshoval:
Answered comments

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -205,7 +205,14 @@ function deploy_sriov_operator {
...
 function apply_sriov_node_policy {
   local -r policy_file=$1
-
+  local -r node=$2
+  local -r node_pf=$3
+
+  # We use just the first suitable pf, for the policy manifest.
+  # We need the num of vfs because if we don't set this value equals to the total, in case of mellanox
+  # the sriov operator will trigger a node reboot to update the firmware
+  export NODE_PF=$node_pf
```
oshoval, line 214: i can remove the exports if i declare those on the envsubst line, will try that

oshoval:
Get rid totally from the `exports` that belong to the function in scope.

EdDev (approved):
Looks nice, thanks.

oshoval:
Thanks Edy
/cc @qinqon 
Would you please take a look?

PR #505, opened 2020-12-22
Use cluster container images from quay.io by default

fgimenez:
Requires https://github.com/kubevirt/project-infra/pull/789

cc/ @rmohr 

### DIFF
--- a/cluster-provision/centos8/build.sh
+++ b/cluster-provision/centos8/build.sh
@@ -4,4 +4,4 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 centos_version="$(cat $DIR/version | tr -d '\n')"
 
-docker build --build-arg centos_version=$centos_version . -t kubevirtci/centos8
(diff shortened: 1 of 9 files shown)


rmohr (changes requested):
I think that we also need to adjust the publish.sh script in the root folder to push to `quay.io/kubevirtci` instead of `kubevirtci`.

rmohr:
/lgtm
/approve
/hold

Let's merge this in a coordinated fashion. Once all lanes are green, we should first merge the new job definitions. Then this one, so that the postsubmit can succeed.

rmohr (approved):

fgimenez:
> Let's merge this in a coordinated fashion. Once all lanes are green, we should first merge the new job definitions. Then this one, so that the postsubmit can succeed.

@rmohr ok lanes are green but they required a few more changes, PTAL. Ready to go ahead with https://github.com/kubevirt/project-infra/pull/789

rmohr:
/lgtm
/hold

Once the other PR is merged we should give it a few minutes since the jobs are not immediately updated in the running components.

PR #500, opened 2020-12-16
Bump sriov operator to 4.8

oshoval:
Signed-off-by: Or Shoval <oshoval@redhat.com>

**Notes**
Known issues:
1. Need to freeze versions.
2. Point to U/S.

PR #510, opened 2020-12-24
sriov, Assign only one PF for the worker

oshoval:
Currently sriov operator manifest actually uses just one of the PFs,
but allocates all available whitelisted PFs.
Therefore this commit update logic
to allocate just one PF for the node, in default.

Modularity improvement:
Currently the PFs allocation function gets
the node it should allocate PFs to.
In order to increase modularity,
add a parameter which determines how many PFs it should allocate to the node.
This way we would be able to use it in a modular
way in case we need to control the number of PFs
we allocate to that node instead allocating
all the PFs to that node.

Additional change:

* In case there are no sriov PFs available,
current sriov PFs scan logic would not detect it.
Bash would keep the array non empty,
and the only element would not be a valid PF
but the path itself.

  Fix logic by using find, which returns an empty
  output in case no valid matches found.

  Since we use local keyword, the command would not
  fail the script, even that we use -e.
  Therefore add a check if the array is empty,
  and fail fast in this case.
  This will detect the problems as soon as possible
  without trailing errors.

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -228,19 +228,19 @@ function move_sriov_pfs_netns_to_node {
   local -r sriov_pfs=( $(find /sys/class/net/*/device/sriov_numvfs) )
   [ "${#sriov_pfs[@]}" -eq 0 ] && echo "FATAL: Could not find available sriov PFs" >&2 && return 1
 
-  for ifs in "${sriov_pfs[@]}"; do
-    local ifs_name="${ifs%%/device/*}"
-    ifs_name="${ifs_name##*/}"
+  for pf in "${sriov_pfs[@]}"; do
+    local pf_name="${pf%%/device/*}"
+    pf_name="${pf_name##*/}"
 
-    if [ $(echo "${PF_BLACKLIST[@]}" | grep "${ifs_name}") ]; then
+    if [ $(echo "${PF_BLACKLIST[@]}" | grep "${pf_name}") ]; then
       continue
     fi
 
-    ip link set "$ifs_name" netns "$node"
-    pf_array+=($ifs_name)
+    ip link set "$pf_name" netns "$node"
+    pf_array+=("$pf_name")
   done
 
-  echo ${pf_array[@]}
+  echo "${pf_array[@]}"
 }
 
 # The first worker needs to be handled specially as it has no ending number, and sort will not work
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -9,9 +9,12 @@ KUBECONFIG_PATH="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig"
 
 MASTER_NODE="${CLUSTER_NAME}-control-plane"
 WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
+PF_COUNT_PER_NODE=${PF_COUNT_PER_NODE:-1}
 
 OPERATOR_GIT_HASH=8d3c30de8ec5a9a0c9eeb84ea0aa16ba2395cd68  # release-4.4
 SRIOV_OPERATOR_NAMESPACE="sriov-network-operator"
+
+[ $PF_COUNT_PER_NODE -le 0 ] && echo "FATAL: PF_COUNT_PER_NODE must be a positive integer" >&2 && exit 1
 
 # This function gets a command and invoke it repeatedly
 # until the command return code is zero
@@ -219,6 +222,7 @@ function apply_sriov_node_policy {
 
 function move_sriov_pfs_netns_to_node {
   local -r node=$1
+  local -r pf_count_per_node=$2
   local -r pid="$(docker inspect -f '{{.State.Pid}}' $node)"
   local pf_array=()
 
@@ -238,7 +242,12 @@ function move_sriov_pfs_netns_to_node {
 
     ip link set "$pf_name" netns "$node"
     pf_array+=("$pf_name")
+    [ "${#pf_array[@]}" -eq "$pf_count_per_node" ] && break
   done
+
+  [ "${#pf_array[@]}" -lt "$pf_count_per_node" ] && \
+    echo "FATAL: Not enough PFs allocated, PF_BLACKLIST (${PF_BLACKLIST[@]}), PF_COUNT_PER_NODE ${PF_COUNT_PER_NODE}" >&2 && \
+    return 1
 
   echo "${pf_array[@]}"
 }
@@ -257,7 +266,7 @@ if [[ "$SRIOV_NODE" == "${WORKER_NODE_ROOT}0" ]]; then
   SRIOV_NODE=${WORKER_NODE_ROOT}
 fi
 
-NODE_PFS=($(move_sriov_pfs_netns_to_node $SRIOV_NODE))
+NODE_PFS=($(move_sriov_pfs_netns_to_node "$SRIOV_NODE" "$PF_COUNT_PER_NODE"))
 
 SRIOV_NODE_CMD="docker exec -it -d ${SRIOV_NODE}"
 ${SRIOV_NODE_CMD} mount -o remount,rw /sys     # kind remounts it as readonly when it starts, we need it to be writeable
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -225,7 +225,9 @@ function move_sriov_pfs_netns_to_node {
   mkdir -p /var/run/netns/
   ln -sf /proc/$pid/ns/net "/var/run/netns/$node"
 
-  local -r sriov_pfs=( /sys/class/net/*/device/sriov_numvfs )
+  local -r sriov_pfs=( $(find /sys/class/net/*/device/sriov_numvfs) )
+  [ "${#sriov_pfs[@]}" -eq 0 ] && echo "FATAL: Could not find available sriov PFs" >&2 && return 1
+
   for ifs in "${sriov_pfs[@]}"; do
     local ifs_name="${ifs%%/device/*}"
     ifs_name="${ifs_name##*/}"


oshoval:
/cc @ormergi 
Please take a look

alonSadan (approved):
Thanks Or.

oshoval:
Thanks @alonSadan 

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -9,10 +9,21 @@ KUBECONFIG_PATH="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig"
...
 
+decimal_re='^[0-9]+$'
+if ! [[ "$NUM_PFS_REQUIRED" =~ ${decimal_re} ]]; then
+  echo "FATAL: Wrong value of NUM_PFS_REQUIRED, must be numeric, less or equal to actual PF available"
+  exit 1
+fi
+
+if [ "$NUM_PFS_REQUIRED" -eq "0" ]; then
+  echo "NUM_PFS_REQUIRED equals zero, allocating all available PFs to the node"
+fi
```
ormergi, line 25: This is wired behavior it could mislead what one would expect to happen
In case `NUM_PFS_REQUIRED` is zero the script should exit, please change it.
Also drop the surrounding quotes there is no need for them when you comparing integers.

ormergi (changes requested):

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: This is in case we want to give all PFs available
without the need to check how many PFs actually available (some nodes might have 2 some might have 4).

bash should use quotes even if we compare integers, this is the right convention, else there are globbing and word splitting.
(for the var, and for the 0 itself which both ways work for the scalar)

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -9,10 +9,21 @@ KUBECONFIG_PATH="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig"
...
+NUM_PFS_REQUIRED="${NUM_PFS_REQUIRED:-1}"
 
 OPERATOR_GIT_HASH=8d3c30de8ec5a9a0c9eeb84ea0aa16ba2395cd68  # release-4.4
 SRIOV_OPERATOR_NAMESPACE="sriov-network-operator"
 
+decimal_re='^[0-9]+$'
+if ! [[ "$NUM_PFS_REQUIRED" =~ ${decimal_re} ]]; then
+  echo "FATAL: Wrong value of NUM_PFS_REQUIRED, must be numeric, less or equal to actual PF available"
+  exit 1
+fi
```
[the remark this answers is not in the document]
oshoval: i prefer to keep it, there is no reason to drop it, i want it to catch erroneous inputs, and that the script will fail as soon as possible.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -225,7 +225,12 @@ function move_sriov_pfs_netns_to_node {
   mkdir -p /var/run/netns/
   ln -sf /proc/$pid/ns/net "/var/run/netns/$node"
 
-  local -r sriov_pfs=( /sys/class/net/*/device/sriov_numvfs )
+  local -r sriov_pfs=( $(find /sys/class/net/*/device/sriov_numvfs) )
+  if [ ${#sriov_pfs[@]} -eq 0 ]; then
+    echo "FATAL: Could not find available sriov PFs"
+    exit 1
```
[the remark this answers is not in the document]
oshoval: it wont fail, try it, there is local -r that cause it to not fail.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: The assumption that 0 means all PFs is misleading, this need to be changed.

At this script abstraction level we shouldn't worry about how many PF's there are on a node, 
it should move only what is requested.

About the quotes, why there will be word splinting its a flag that you expect to have an integer
if it doesn't it will fail before.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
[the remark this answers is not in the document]
ormergi: This check is redundant because right after this one there is an if that checks `NUM_PFS_REQUIRED` 
is zero, that if validates for you.
And if the latter check fails this script stops so it wont deploy anything.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: Its not an assumption, its a feature for debugging.
See please the code, and what happens if you put 0 there.
In case of a problem we might do care how many we want to move, and to be agnostic about which baremetal we run on, and how many PFs there are.

About the quotes, its a convention, if you use bash vars you should use quotes, not just if there can be problems.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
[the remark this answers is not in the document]
oshoval: The later just echo, it doesn't fail.
Check again please.
https://github.com/kubevirt/kubevirtci/pull/510/files#diff-499cff045d8a875e8b39f4a9fd59354eb99e8f8d91a95ebc70fbf42d2d5b5beaR24

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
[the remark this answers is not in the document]
ormergi: So why using `find` at all if you masks the error? 
I mean there is no different then how it was before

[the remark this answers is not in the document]
oshoval: there is a difference.
The previous line would have an item even if there are no PF
and then to check if its empty, it will be awkward:
check if the first element is `/sys/class/net/*/device/sriov_numvfs` 
Please try and see it.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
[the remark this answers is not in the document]
ormergi: Sorry I wasn't clear let me explain myself
You can drop the first check, 
and change the second check to `if [ $NUM_PFS_REQUIRED -le 0 ]`
if its true then there is nothing to do.
This way you dont need to validate `NUM_PFS_REQUIRED`.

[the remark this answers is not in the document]
oshoval: But i think otherwise about blocking zero. as we spoke on other thread, lets keep it there please:
https://github.com/kubevirt/kubevirtci/pull/510#discussion_r549399014

Its a feature imo

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: If I wouldn't review the code I wouldn't say anything :)

My point is by doing that you forcing one flag to do two things which is wrong in my opinion.
At the current state we move all PF's but use only one,
with this change we would move one PF which I understand.
But how is moving all PF's to a single namespace will help debugging (or anything)?

About the convention its not craved in stone at least on kubevirtci, 
if you sure a variable will be single "word" there is no need to add quotes.
Anyway this is not the place to discuss what convention to use.

oshoval: > But how is moving all PF's to a single namespace will help debugging (or anything)?

Lets say we there is a bug on CI, that only on one node there is a failure,
and one want to check if its related that we move just one PF to that node,
so just setting it to 0 will allow this to auto detect how many PFs there are and move them, in order to see the behavior.
I saw other problems (not due to this PR), that a feature like this would help to debug if they happened on CI.

There is an echo that explains it, and sometimes a special value like 0 means special things,
for example in timeouts it sometimes infinity, on a non optional fields.

Lets hear another opinion please, and take the majority.
Maybe they will agree with you, but i want to give it a chance.
Same for the quotes.

oshoval:
Going to refactor a bit, please wait for the next push.

oshoval:
Changes:
Splitted to smaller commits.
Update PR description and title.
Addressed comments.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -225,7 +225,12 @@ function move_sriov_pfs_netns_to_node {
   mkdir -p /var/run/netns/
   ln -sf /proc/$pid/ns/net "/var/run/netns/$node"
 
-  local -r sriov_pfs=( /sys/class/net/*/device/sriov_numvfs )
+  local -r sriov_pfs=( $(find /sys/class/net/*/device/sriov_numvfs) )
```
ormergi, line 235: Recently we removed all the error masking (removing `||return/exit`) and relay on `-e` flag
This change actually masking the error from `find` in order to echo more informative one,
the intention is good I like informative error message.
I dont like that we getting back to masking error even for better error message, where we can achive that w/o masking the error.
Instead of this check I prefer that we echo before this line "Discovering PF's" or something like that.
In case of an error the script will stop, and in the logs we could see where it failed 
and why because the error from `find` will reflect in the logs.
What you think?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -9,10 +9,16 @@ KUBECONFIG_PATH="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig"
 
 MASTER_NODE="${CLUSTER_NAME}-control-plane"
 WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
+NUM_PFS_REQUIRED=${NUM_PFS_REQUIRED:-1}
```
ormergi, line 12: I would change this variable name to `PFS_COUNT_PER_NODE` 
so it will reflect better what it affects

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -241,8 +248,16 @@ function move_sriov_pfs_netns_to_node {
...
+    if [ "${#pf_array[@]}" -eq "$num_pfs_required" ]; then
+      break
+    fi
   done
 
+  if [ "${#pf_array[@]}" -lt "$num_pfs_required" ]; then
+    echo "FATAL: Could not allocate enough PFs, please check PF_BLACKLIST, NUM_PFS_REQUIRED, and how many PF actually available or used already"
+    exit 1
+  fi
+
```
ormergi: Instead of this check I would add check if there are enough PF's before the loop starts by comparing between `sriov_pfs length` and `num_pfs_required`, if false stop the script with informative message.
And same principal as before, echo "Moving PF's to netns" so in case of an error during moving the PF to a netns the script would stop and the error from iplink would reflect in the logs.

ormergi (changes requested):

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: This is not this change, its the local that mask it, and we do need local imo.
This change is just about changing to find instead the bash globbing.
I dont think we should echo and search, code should do it for us.

Another option is remove the local here, but someone might add it back.
(we will need to keep the find in both cases, as the bash glob won't error in case of an error)

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: It will need to take into account the black list ones,
I find it more complex, and imo not needed.
I think the current method is straightforward and simple.
The whole operation takes a jiffy.

Again i am against putting an anchor echo.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: It allocates only for the worker not to the master, so should i use `PFS_COUNT_PER_NODE` or 
`PFS_COUNT_PER_WORKER_NODE` ?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
EdDev: @ormergi , IIUC you refer to the `find` execution as if it can fail and we should care.
I think that in this specific case, it is used to just fetch the list of PF/s and for this context it does not matter if it passes or fails.. the only thing that matters is the return value. Just think of it as if it fails, it returns an empty list.

If you look at it this way, we do not need to catch any errors here. And the error is only about the list content (if it is empty, we exit early).

Is this perspective better fits what is going on here?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -225,7 +225,12 @@ function move_sriov_pfs_netns_to_node {
   mkdir -p /var/run/netns/
   ln -sf /proc/$pid/ns/net "/var/run/netns/$node"
 
-  local -r sriov_pfs=( /sys/class/net/*/device/sriov_numvfs )
+  local -r sriov_pfs=( $(find /sys/class/net/*/device/sriov_numvfs) )
+  if [ "${#sriov_pfs[@]}" -eq 0 ]; then
+    echo "FATAL: Could not find available sriov PFs"
```
EdDev, line 237: You should echo to stderr and not stdout. And in general, in bash, the convention is something like this:
`[ "${#sriov_pfs[@]}" -eq 0 ] && echo "..." .... ; exit 1`

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I would go for the first one I dont think we need to specify the role of the node
I added COUNT to indicate it is expected to be a number

oshoval: Done

oshoval:
Addressed a comment.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
EdDev: I vote for `PF_COUNT_PER_NODE`.
You could use "PFS_PER_NODE", but then that plural looks strange. Personally I do not mind, both seem to be expressive enough.

EdDev (commented):
Commit "sriov, Assign only one PF for the worker":

> Currently sriov actually uses just one of the pfs

I do not understand who you refer to by "sriov" here.
Who uses only one PF? The tests? The populated operator manifest? The scripts?

Inthe commit message, please use `PF` and not the lower case version. It appears intermixed.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -9,10 +9,16 @@ KUBECONFIG_PATH="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig"
...
 WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
+PFS_COUNT_PER_NODE=${PFS_COUNT_PER_NODE:-1}
 
 OPERATOR_GIT_HASH=8d3c30de8ec5a9a0c9eeb84ea0aa16ba2395cd68  # release-4.4
 SRIOV_OPERATOR_NAMESPACE="sriov-network-operator"
 
+if [ $PFS_COUNT_PER_NODE -le 0 ]; then
+  echo "FATAL: Wrong value of PFS_COUNT_PER_NODE, must be a positive integer, less or equal to the actual number of PFs available"
+  exit 1
+fi
```
EdDev, line 20: Please drop the `if`, there is no need for it.
The message is also too much IMO, just saying that it needs to be a positive integer should be enough.
The "the actual number of PFs available" is also redundant, if one specifies too many, then a different error should pop up.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -9,10 +9,16 @@ KUBECONFIG_PATH="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig"
 
 MASTER_NODE="${CLUSTER_NAME}-control-plane"
 WORKER_NODE_ROOT="${CLUSTER_NAME}-worker"
+PFS_COUNT_PER_NODE=${PFS_COUNT_PER_NODE:-1}
```
EdDev, line 12: PFS => PF

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -219,6 +225,7 @@ function apply_sriov_node_policy {
 
 function move_sriov_pfs_netns_to_node {
   local -r node=$1
+  local -r num_pfs_required=$2
```
EdDev, line 228: Please change the name here as well.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -241,8 +248,16 @@ function move_sriov_pfs_netns_to_node {
 
     ip link set "$pf_name" netns "$node"
     pf_array+=("$pf_name")
+    if [ "${#pf_array[@]}" -eq "$num_pfs_required" ]; then
```
EdDev, line 251: Please drop the `if`.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -241,8 +248,16 @@ function move_sriov_pfs_netns_to_node {
...
+      break
+    fi
   done
 
+  if [ "${#pf_array[@]}" -lt "$num_pfs_required" ]; then
```
EdDev, line 256: Please drop the `if`.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: You can declare `sriov_pfs` and then do the assignment
```
local sriov_pfs
sirov_pfs=( $(find /sys/class/net/*/device/sriov_numvfs) )
```
That way the error is not masked

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: About the 2nd part 
I think linter says otherwise (https://github.com/kubevirt/kubevirtci/pull/511), but np i will change now and once linter is in we discuss it again in case needed.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -241,8 +248,16 @@ function move_sriov_pfs_netns_to_node {
...
+    echo "FATAL: Could not allocate enough PFs, please check PF_BLACKLIST, PFS_COUNT_PER_NODE, and how many PF actually available or used already"
```
EdDev, line 257: Please provide the values that caused the error so the user can see what happened.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: > I dont think we should echo and search, code should do it for us.

I am not sure what you mean, at the current state I assume you added a check because the error was masked, with what I suggested you can unmask and remove that check

EdDev:
Now that we do not pass "all", it makes sense to get rid of the "black-list", replacing it with an explicit interface as an alternative (i.e. a while list).

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: Its nice, but its prone that someone will merge the lines, and then mask the error,
it might be fine even if it happens, because the code later will catch it, but i think we better fail fast, 
and reduce noises, wdyt?
also it should be `local -r`, once we get an agreement here i will try it out accordingly

oshoval:
> Now that we do not pass "all", it makes sense to get rid of the "black-list", replacing it with an explicit interface as an alternative (i.e. a while list).

I find the current method better, because its generic, and we can say how many we need, not which we need,
so it independent on what are the names on the machine

Note
It should be a list because we might want more than 1, for debugging.
For example i need now 2 for features that i work on (playing with CNI).

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
EdDev: The point is that the error does not matter. We do not want to exit on `find` failing, we want to exit on the list being empty.

Same as if you would have this function: `get_pf_count()`. You will not expect it to fail, just to return a list/array.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: In all places that you asked please drop the if you mean to do as you suggested here i guess
https://github.com/kubevirt/kubevirtci/pull/510#discussion_r550162182

EdDev:
Do you have a scenario for the black-list?
If you do, then converting it to a white-list should be equivalent.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Well in general the code is prune to someone who is going to do something wrong, 
we hope that we or a maintainer will catch it during reviews.
I think it doesn't need to contradict with "doing things the right way", maybe I am too optimistic that also an option.
We could add a comment about dividing it to two lines, about using `-r` I think it is not a must here.

About the error, I cant think of a reason for `find` to fail other than file doesn't exists or missing permissions
in both cases the script should stop at that point because it means there are no PF's on the host default netns.
In this case there I dont think we benefit from saving `find` output to array and check if its empty, its just variables substitution. Its also requiring another error checking instead of relaying on `-e` flag.

In any case if this is going to stay consider changing `exit 1` to `return 1`

oshoval: > In any case if this is going to stay consider changing `exit 1` to `return 1`

IIRC when i debugged it, i saw that we can't, because `NODE_PFS=($(move_sriov_pfs_netns_to_node "$SRIOV_NODE" "$PF_COUNT_PER_NODE"))` 
the caller is also scoped.

ormergi:
I also prefer to move PF's by quantity  and not by explicit interface names,they could change over time.

I am not sure if we even need black list for PF's, we dont use it on CI.

oshoval:
> I also prefer to move PF's by quantity and not by explicit interface names,they could change over time.
> 
> I am not sure if we even need black list for PF's, we dont use it on CI.

From what i understood, Edy meant that it will stay as alternative.
It will just changed to white list, it won't be instead of count.

I think that the blacklist was originally introduced in order to save one PF for management.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: @EdDev Sorry I missed you comment, please see the one above
About find failing, if it fails it means that there are not PF's and the script stops, so I think it is redundant to save its output in array and check if its empty.

oshoval:
Updated according comments, will leave the open discussions not resolved.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
oshoval: > In any case if this is going to stay consider changing `exit 1` to `return 1`

Done

EdDev: You are looking at "find" failure, and I am looking at unifying results.
Lets take the "list_pfs" hypothetical example function. We could allow it to fail/raise in case something is wrong, or just return an empty list. The advantage of getting an empty list is that we can handle it the same way as a non-empty list.
This means we could iterate over it, check its length, etc...

I think this is the same case here... we are not interested in finding out that "find" failed, we are interested in checking its output. If something fails, you do not really want to know that "find" failed, you want to know that there are no PF/s without entering the details on how it was detected.

It is also worth mentioning that this is too minor to deserve such a long discussion. If it is not wrong, it is just about taste and there are strong opinions, someone needs to back-off in order to get forward. It is not that important to be blocked by either a reviewer or an author. You both should consider this for the next time.

oshoval:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/510/check-up-kind-1.17-sriov/1344282316951785474

The failure might be related to this one
https://bugzilla.redhat.com/show_bug.cgi?id=1878566

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -225,7 +225,9 @@ function move_sriov_pfs_netns_to_node {
   mkdir -p /var/run/netns/
   ln -sf /proc/$pid/ns/net "/var/run/netns/$node"
 
-  local -r sriov_pfs=( /sys/class/net/*/device/sriov_numvfs )
+  local -r sriov_pfs=( $(find /sys/class/net/*/device/sriov_numvfs) )
+  [ "${#sriov_pfs[@]}" -eq 0 ] && echo "FATAL: Could not find available sriov PFs" > /dev/stderr && return 1
```
EdDev, line 233: I do not know where you saw this `> /dev/stderr` used, I know about a simple `>&2`.
And I am not clear why `&&` was used after, `echo` usually succeeds, but a more direct approach is to use `;` in case some poor author copy/paste this pattern and places something that can fail instead of the `echo`.

Note that for printing to stderr, you can also set a helper function: `echoerr() { printf "%s\n" "$*" >&2; }` that is better than `echo` as it does not require an exec.

oshoval: `false && echo bla; echo foo`
will print foo, unless i use '( )' on the right part
&& is pretty wide used i think, and is straight forward and clear.
wdyt?

I think the echoerr is overkill, unless you insist.

EdDev: You used `&& return 1` and that should be `; return 1`.

The function is more readable and avoids adding the redirection each time. The fact it is faster is less important.
I will not insist in the function.

oshoval:
```
$ false && echo bla; return 1
bash: return: can only `return' from a function or sourced script
```
You can't it will execute it anyhow, unless i use 'false && (echo bla; return 1)`

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: My point is that in this case there is no benefit in saving `find` output in an array and check its length 
because `find` could fail only if a given path doesn't exists (this script runs as sudo on CI so file permissions is not relevant)
If `find` fails the array is empty, if it doesn't fail the array is not empty.

About the long discussion, noted and appreciated.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
EdDev: Yes, I see now. My mistake.

oshoval:
https://github.com/kubevirt/kubevirtci/pull/512
Clean master is also broken.
Its not due to this PR, its either because 4.4 images were changed today, or maybe because there isn't space left on the device.
4.8 might don't have this problem, checking via #500

oshoval:
New official operator images were pushed, and locally i can't see the bug anymore, checking on this PR and on #512 clean master several times to see we are fine

oshoval:

New official operator images were pushed, and locally i can't see the bug anymore, checking on this PR and on #512 clean master several times to see we are fine.

Will create a PR that freeze versions.

oshoval:
Changed to `>&2` instead of `/dev/stderr`

EdDev (approved):
Thanks!

oshoval:
Thanks Edy,

PR #513, opened 2020-12-31
sriov, Fix VFs allocation race condition of sriov operator 

oshoval:
The flow is as follows:

1. The undercloud CNI of prow currently sets VFs number (`sriov_numvfs`)
to max allowed (according `sriov_totalvfs`).

2. The sriov operator is deployed.
3. The sriov operator sets all unused PFs to have zero VFs,
since no policy applies to these PFs yet.
Number of VFs equal zero would eventually be reported at `SriovNetworkNodeState`.

4. Kubevirtci creates a new `SriovNetworkNodePolicy`, 
with numVfs equals `sriov_totalvfs` value.

5. The sriov operator checks if it should update number of VFs,
by comparing the desired numVfs (according to policy) to the current
numVfs configured  according `SriovNetworkNodeState` status.

In case `SriovNetworkNodeState` didn't finish updating the actual
numVfs yet in the status,
the sriov operator would not create VFs, because the desired state
equals the current state.

The result would be that the cluster would stay with zero
VFs until a new policy is created,
or until the network config daemon is restarted.

By resetting the VFs before deploying the operator we fix this race condition.
The operator assumes it starts on a clean setup, so this change is natural.

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1878566

### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -262,6 +262,10 @@ ${SRIOV_NODE_CMD} mount -o remount,rw /sys     # kind remounts it as readonly wh
 ${SRIOV_NODE_CMD} chmod 666 /dev/vfio/vfio
 _kubectl label node $SRIOV_NODE sriov=true
 
+for pf in "${NODE_PFS[@]}"; do
+  docker exec $SRIOV_NODE bash -c "echo 0 > /sys/class/net/$pf/device/sriov_numvfs"
+done
+
 deploy_multus
 wait_pods_ready
 


oshoval:
/uncc @rmohr @vladikr 
So it won't noise you, thanks

oshoval:
Align commit message to PR desc.

oshoval:
Not related to this PR, but lets rerun to see if the problem is transient.

oshoval:
/retest

centos is back

oshoval:
The failures aren't related to this PR.

EdDev:
> In case SriovNetworkNodeState didn't finish updating the actual
> numVfs yet in the status,
> the sriov operator would not create VFs, because the desired state
> equals the current state.

You mean to update the number from the max total to the zero one?

> The result would be that the cluster would stay with zero
> VFs until a new policy is created,
> or until the network config daemon is restarted.

Are you saying that the operator is not reconciling the actual with the desired?

If this is the case, why not reverse the order of the operations? Create the policy before deploying/starting the operator.

oshoval:
> > In case SriovNetworkNodeState didn't finish updating the actual
> > numVfs yet in the status,
> > the sriov operator would not create VFs, because the desired state
> > equals the current state.
> 
> You mean to update the number from the max total to the zero one?

In this paragraph you quote the sriov operator (daemon pod), compares the numVfs of the policy (lets call it N)
(desired) to the one reported as the current numVfs in the node state status.
If they are equal it won't update sriov_numvfs.
It already zeroed it before, when the operator was deployed and no policy exists yet,
but the status didn't refresh, so it was a false positive that we already have N Vfs.
   
> 
> > The result would be that the cluster would stay with zero
> > VFs until a new policy is created,
> > or until the network config daemon is restarted.
> 
> Are you saying that the operator is not reconciling the actual with the desired?

True, because there is one shot, its not a periodic even.
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/release-4.4/pkg/plugins/generic/generic_plugin.go#L125
If the status wasn't updated in the status of the node, it won't do anything, forever (until new policy is updated or daemon restart).

We see here the bug: the desired state is 64 VFs (1st line)
but it think it shouldn't do anything (3rd line)

```
I1231 07:55:34.614214    6012 generic_plugin.go:105] generic-plugin Apply(): desiredState={1718 [{0000:65:00.0 64 1500 ens2f0 [{sriov_net vfio-pci 0-63}]}]}
I1231 07:55:34.631605    6012 generic_plugin.go:115] generic-plugin Apply(): lastStat={1452 []}
I1231 07:55:34.631942    6012 utils.go:113] syncNodeState(): no need update interface 0000:65:00.0
```

> 
> If this is the case, why not reverse the order of the operations? Create the policy before deploying/starting the operator.

It might work, but since it will bypass the web hook validation, i didn't think its the right thing to do.

EdDev:
> It already zeroed it before, when the operator was deployed and no policy exists yet,
> but the status didn't refresh, so it was a false positive that we already have N Vfs.

So the problem is that we did not wait for the operation of zero-ing  the VF/s to complete (e.g. check that it actually got applied and the PCI devices disappeared).
If that is completed before deploying the operator, it will come up and see no VF/s at all.

Later, when the policy is applied, it will just execute regularly.
That sounds to me like fitting with an actual stable node state, without the need to wait on k8s objects to get updated.


>> If this is the case, why not reverse the order of the operations? Create the policy before deploying/starting the operator.

> It might work, but since it will bypass the web hook validation, i didn't think its the right thing to do.

Ohh, the webhooks are deployed by the operator. Ok, makes sense then.

oshoval:
> > It already zeroed it before, when the operator was deployed and no policy exists yet,
> > but the status didn't refresh, so it was a false positive that we already have N Vfs.
> 
> So the problem is that we did not wait for the operation of zero-ing the VF/s to complete (e.g. check that it actually got applied and the PCI devices disappeared).
> If that is completed before deploying the operator, it will come up and see no VF/s at all.
> 
> Later, when the policy is applied, it will just execute regularly.
> That sounds to me like fitting with an actual stable node state, without the need to wait on k8s objects to get updated.

If i understand you correctly you mean wait_allocatable_resource ?
if so you can't remove it, it takes time to update them, and you can't deploy pods before it appears,
it might cause a flakiness if the test / pod creation runs before it managed to make them allocate able.

> 
> > > If this is the case, why not reverse the order of the operations? Create the policy before deploying/starting the operator.
> 
> > It might work, but since it will bypass the web hook validation, i didn't think its the right thing to do.
> 
> Ohh, the webhooks are deployed by the operator. Ok, makes sense then.

EdDev:
> If i understand you correctly you mean wait_allocatable_resource ?

I mean wait for the script to finish setting the number of VF/s to zero, check no VF/s exists on the PCI device list and only then deploy the operator.

oshoval:
> > If i understand you correctly you mean wait_allocatable_resource ?
> 
> I mean wait for the script to finish setting the number of VF/s to zero, check no VF/s exists on the PCI device list and only then deploy the operator.

This what i am doing, by checking the status of the nodestate, once its Succeeded, it means it done that,
and even updated the status.
checking `ip link` isn't enough because it might still not updated in the status itself, just on the OS.

EdDev:
> > I mean wait for the script to finish setting the number of VF/s to zero, check no VF/s exists on the PCI device list and only then deploy the operator.
> 
> This what i am doing, by checking the status of the node state, once its Succeeded, it means it done that,
> and even updated the status.
> checking `ip link` isn't enough because it might still not updated in the status itself, just on the OS.

You descriptions says that you wait on the object `SriovNetworkNodeState` to get updated, but this is not what I mentioned above.
I mentioned you will set the number of VF/s to zero and wait *before* you deploy the SRIOV operator.

oshoval:
> > > I mean wait for the script to finish setting the number of VF/s to zero, check no VF/s exists on the PCI device list and only then deploy the operator.
> > 
> > 
> > This what i am doing, by checking the status of the node state, once its Succeeded, it means it done that,
> > and even updated the status.
> > checking `ip link` isn't enough because it might still not updated in the status itself, just on the OS.
> 
> You descriptions says that you wait on the object `SriovNetworkNodeState` to get updated, but this is not what I mentioned above.
> I mentioned you will set the number of VF/s to zero and wait _before_ you deploy the SRIOV operator.

I edited directly, 
see please the updated message
https://github.com/kubevirt/kubevirtci/pull/513#issuecomment-754583606

It makes kubevirtci do the job of the operator imo.
(its partially good in the meaning that this is what the operator expects, but bad because it adds logic which in the current PR state isn't needed)
This was the first approach i tried in the PR.

oshoval:
> > > I mean wait for the script to finish setting the number of VF/s to zero, check no VF/s exists on the PCI device list and only then deploy the operator.
> > 
> > 
> > This what i am doing, by checking the status of the node state, once its Succeeded, it means it done that,
> > and even updated the status.
> > checking `ip link` isn't enough because it might still not updated in the status itself, just on the OS.
> 
> You descriptions says that you wait on the object `SriovNetworkNodeState` to get updated, but this is not what I mentioned above.
Waiting for it, also means the reset to zero was done and completed.

> I mentioned you will set the number of VF/s to zero and wait _before_ you deploy the SRIOV operator.

EdDev:
Summarizing per what I understand so far with my recommendation and reasoning:
- The SRIOV operator has a bug, that if it identifies a PF with non zero VF/s, it will zero it. But then, it does not wait for the operation to express in the "status" before handling a policy. And the race you mentioned occurs.
- The suggested approach in this PR is to wait for the "status" to be updated to zero VF/s before deploying the policy.
  [I find this waiting odd, just because it is a workaround to a bug, it is not expected to occur in production so it raises questions when reading it here and is playing around with a cluster object].
- I find it more straight forward, to pass to a node a "clean" PF, with no VF/s, simulating a regular just booted node. From here on everything should play nicely with no needed hacks.

oshoval:
Updated to zero the PFs,
I think the zero is enough and nothing else needed to be checked (for example that no vf exists).
The time the zero takes is proportional to the number of VFs.

I zero just the one that actually moved to the container, no need to touch the rest.

In case we see it fails, it might fail due to time it appears in the /sys/class/net of the container,
if needed, in this case i will add a retry.

oshoval:
Thanks
/retest

oshoval:
@qinqon 
Please take a look.
This PR fixes a flakiness of kubevirtci sriov lane.

ormergi:
This change should be done at the CI sriov-passthrough-cni plugin we use to attach PF's from the CI nodes to the prow job pods.
At the current state the plugin we pass PF's and VF's as they are to the prow jobs pods.

If we expect the prow pod to start "fresh", with reseted PF's and no VF's created, on sriov lane jobs
The relevant changes should apply at the sriov-passthrough-cni plugin code, and not on the provider code.

oshoval:
> This change should be done at the CI sriov-passthrough-cni plugin we use to attach PF's from the CI nodes to the prow job pods.
> At the current state the plugin we pass PF's and VF's as they are to the prow jobs pods.
> 
> If we expect the prow pod to start "fresh", with reseted PF's and no VF's created, on sriov lane jobs
> The relevant changes should apply at the sriov-passthrough-cni plugin code, and not on the provider code.

1. The CNI doesn't have today an auto deploy mechanism.
2. If we don't do it in kubevirtci, a local user will need to reset the VF to 0, after docker kill a conatiner,
else he will left with leftovers and it will happen.
3. Another reason is that even the job that the CNI gives the PF to is privileged there is this bug
https://github.com/containerd/containerd/issues/3221
It cant be worked around by remouting, but it will be less clean.
4. Moreover soon both the multi sriov effort and the migration effort will change the code, in ways
that will eliminate the need of this change in the CNI, and move it to the kubevirtci responsibility.

I don't see any reason to complicate it, nor think its right to do this change in the CNI.

ormergi:
1. What do you mean by "auto deploy mechanism"?
  
2.  When you need to do that on a local setup? 
I never had to do that.. as you said the operator do the resting

 3. This bug doesn't affect the sriov-passthrough-cni plugin pods

4. We can simplify kubevirtci code and make the process transparent if we do this kind of changes on the CNI.
Anyway it shouldn't matter, we dont need to stop pushing changes just because there going to be other changes
in the future.

> I don't see any reason to complicate it, nor think its right to do this change in the CNI.
> Kubevirtci should be self contained, and runable both on CI and locally.

Like I said it will actually simplify things, and it wont violate our expectations from kubevirtci

About to that bug you mentioned, in theory increasing the wait time would solve it?
If so I guess this explains why cluster-up fails (the nodes wont have the resource on time)
We used to wait for taints to appear and remove, I guess that due to the overall time we waited
we didnt encountered this issue. 

In any case those kind of changes deserves a comment in the code with a link to that bug 
it workaround, please add it.

oshoval:
> About to that bug you mentioned, in theory increasing the wait time would solve it?

Increasing the time between the deployment of the operator and creating the policy, yes.
Indeed this one of the reasons that the refactor which wasn't bumped yet to kubevirt uncovered.
This is the previous method i suggested, by waiting the NodeState to be completed, which in turn, means
the sriov-operator finished zeroing the VFs.

> In any case those kind of changes deserves a comment in the code with a link to that bug
> it workaround, please add it.

I am against comments, lets hear please what the other reviewers think, if they agree for a comment i will add.

EdDev:
I must say, I failed to understand why you guys took the conversation outside the immediate scope.

The change suggests to reset the VF/s count to zero before starting the operator. This is reasonable and fits with real production scenarios where the node comes up, no one changes the PF until the operator comes up.
So starting with a value of numvfs=0 make sense, just because it mimics production reality and avoids corner cases like the one explained in this PR.

And the reason it is not enough to depend on the prow cluster CNI, is that when we run this locally, we have no one to help us zero the numvfs.

@ormergi , please directly feedback on the above points and reason on why this change is not a fit.

ormergi:
There is no thumb rule for not touching the PF's until the sriov-operator comes up,
It will configure the PF's and VF's for you according to the policy.
This is the whole idea off having it in the first place.
Ii should be emphasized that this change is workaround for another bug.

We should move this change to sriov-passthrough-cni plugin which is there exclusively 
for sriov lane needs and it can handle exactly what we are talking about with minor changes.

For local setups, if one would like to run sriov lane on local setup it will require configuring sriov 
hardware on the host including creating VF's.
We can simply add a line to the README suggestion to keep VF's 0 for sriov lane tests.
Correct me if I am wrong, kubevirtci cluster should not leave footprint on the host it runs on.

EdDev:
> There is no thumb rule for not touching the PF's until the sriov-operator comes up,

Yes, there is no such rule. But there is reality. This is what happens in production with nodes, no one touches them and the operator does not need to share management on the PF/s with anyone else.

> It will configure the PF's and VF's for you according to the policy.
> This is the whole idea off having it in the first place.

The problem is not with applying the policy, it is with the status of the PF when it comes up for the first time.
If there are N VF/s defined already, it will zero them while loading.

> Ii should be emphasized that this change is workaround for another bug.

The behavior is described in the PR and the commit message, so I do not understand your point here.
It describes why the action has been taken to zero the VF/s before starting the operator.

> We should move this change to sriov-passthrough-cni plugin which is there exclusively
> for sriov lane needs and it can handle exactly what we are talking about with minor changes.

We can move the VF zero-ing to that CNI, but then it will solve only the CI flow and not the local flow.
So even if we should update that CNI logic, it will not be enough.
In addition, that CNI was not touched from the day it was deployed, if you are interested in doing this change there, you are welcome to pursue that direction. And when that will be functional, you can removed this change if you convince the reviewers it is a good step.

> For local setups, if one would like to run sriov lane on local setup it will require configuring sriov
> hardware on the host including creating VF's.

Currently all the SR-IOV interfaces are processed automatically, I do not understand what you are saying here.
After detecting all, they (one now?) are passed to the worker.
What configuration is required from the tester?
 
> We can simply add a line to the README suggestion to keep VF's 0 for sriov lane tests.

Why would we require a manual step from the tester when we can do it for him automatically?

> Correct me if I am wrong, kubevirtci cluster should not leave footprint on the host it runs on.

We are touching the HW resources, so we change the host.
That is why after one run, we may finish everything and have leftovers (like numvfs as non-zero). If we make sure to reset it on each run as part of the preparation, we cover both CI and local run.

Said that, I have no objection to add this logic to the outer CNI as well.

oshoval:
> Said that, I have no objection to add this logic to the outer CNI as well.

This draft PR, would need just one line to fix it when it will be ready to review
https://github.com/kubevirt/project-infra/pull/801
as it already removed the setting of the VFs which wasn't needed.

ormergi:
My point is that the problem occurs only on CI.
Personally I didnt encountered it on local environment, and the chance that anyone would is very low
due to the nature of using it for development and not for CI to get test results as fast as possible.

We encounter this on CI because after the operator is ready we immediately create the policy and run tests,
for local setup you would probably run the test after a while where the sriovNodeState is already in updated.
And it will take less time for the operator to be ready because it doesn't need to reconfigure the VF's driver
(it is reseted to default on CI).
I am almost sure we didnt encountered it until the last refactor because we used to wait more between 
sriov-deployment and policy creation.

For CI the proper place for this change is on the plugin code, I will propose a PR.

oshoval:
> My point is that the problem occurs only on CI.

It happened on my setup a lot.

> Personally I didnt encountered it on local environment, and the chance that anyone would is very low
> due to the nature of using it for development and not for CI to get test results as fast as possible.
> 
> We encounter this on CI because after the operator is ready we immediately create the policy and run tests,
> for local setup you would probably run the test after a while where the sriovNodeState is already in updated.
> And it will take less time for the operator to be ready because it doesn't need to reconfigure the VF's driver
> (it is reseted to default on CI).
> I am almost sure we didnt encountered it until the last refactor because we used to wait more between
> sriov-deployment and policy creation.
> 
> For CI the proper place for this change is on the plugin code, I will propose a PR.

I already did one, please see the message above, i just need to change it a bit.

oshoval:
> > My point is that the problem occurs only on CI.
> 
> It happened on my setup a lot.
> 
> > For CI the proper place for this change is on the plugin code, I will propose a PR.
> 
> I already did one, please see the message above, i just need to change it a bit.
> Please see the PR at [#513 (comment)](https://github.com/kubevirt/kubevirtci/pull/513#issuecomment-756205319)

> I am almost sure we didnt encountered it until the last refactor because we used to wait more between
sriov-deployment and policy creation.

It happens even before the refactor on kubevirt , on CI itself.
More than once, i pasted links above.

oshoval:
> > My point is that the problem occurs only on CI.

It happened on my setup a lot.

> For CI the proper place for this change is on the plugin code, I will propose a PR.
 
I already did one, please see the message above, i just need to change it a bit.
Please see the PR at [#513 (comment)](https://github.com/kubevirt/kubevirtci/pull/513#issuecomment-756205319)

> I am almost sure we didnt encountered it until the last refactor because we used to wait more between
sriov-deployment and policy creation.

It happens even before the refactor on kubevirt , on CI itself.
More than once, i pasted links above.

ormergi:
> I already did one, please see the message above, i just need to change it a bit.
> Please see the PR at [#513 (comment)](https://github.com/kubevirt/kubevirtci/pull/513#issuecomment-756205319)
> 

It is out of the PR scope and should be done on a different PR.

> It happens even before the refactor on kubevirt , on CI itself.
> More than once, i pasted links above.

The above links are jobs from the past few days, 
the refactoring (removing taints waits) was way before.
I saw those errors form time to time only after the refactoring.
but this is not the point of this dissolution so lets leave it for now.

oshoval:
> > I already did one, please see the message above, i just need to change it a bit.
> > Please see the PR at [#513 (comment)](https://github.com/kubevirt/kubevirtci/pull/513#issuecomment-756205319)
> 
> It is out of the PR scope and should be done on a different PR.

I am showing you other PR that already handled part of it.
Please let me finish my work, the task was assigned on me officially.

> 
> > It happens even before the refactor on kubevirt , on CI itself.
> > More than once, i pasted links above.
> 
> The above links are jobs from the past few days,
> the refactoring (removing taints waits) was way before.
> I saw those errors form time to time only after the refactoring.
> but this is not the point of this dissolution so lets leave it for now.

Recheck again, it happens even today on kubevirt which isnt bumped yet.

ormergi:
> Recheck again, it happens even today on kubevirt which isn't bumped yet.

What do I need to recheck?
I said that the errors we see now on CI did not occurred before the refactoring (of removing all the waits for taints).

oshoval:
Created PR on CNI
https://github.com/kubevirt/project-infra/pull/817

oshoval:
> > Recheck again, it happens even today on kubevirt which isn't bumped yet.
> 
> What do I need to recheck?
> I said that the errors we see now on CI did not occur before the refactoring (of removing all the waits for taints).

They do happen, and the refactoring isn't on kubevirt.

ormergi:
> They do happen, and the refactoring isn't on kubevirt.

The refactoring is on kubevirt since two days ago the job you pasted is from yesterday and includes the refactoring
https://github.com/kubevirt/kubevirt/blob/4ceb5f15bf74b6060b4a2eb78f540a735052bc26/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
Until two days ago those errors did not occurr on kubevirt/kubevirt, 
only on kubevirt/kubecirtci

oshoval:
> > They do happen, and the refactoring isn't on kubevirt.
> 
> The refactoring is on kubevirt since two days ago the job you pasted is from yesterday and includes the refactoring
> https://github.com/kubevirt/kubevirt/blob/4ceb5f15bf74b6060b4a2eb78f540a735052bc26/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
> Until two days ago those errors did not occurr on kubevirt/kubevirt,
> only on kubevirt/kubecirtci

Right, didnt know it was bumped 2 days ago, last time i checked it wasn't bumped yet.

ormergi:
> > > They do happen, and the refactoring isn't on kubevirt.
> > 
> > 
> > The refactoring is on kubevirt since two days ago the job you pasted is from yesterday and includes the refactoring
> > https://github.com/kubevirt/kubevirt/blob/4ceb5f15bf74b6060b4a2eb78f540a735052bc26/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
> > Until two days ago those errors did not occurr on kubevirt/kubevirt,
> > only on kubevirt/kubecirtci
> 
> Right, didnt know it was bumped 2 days ago, last time i checked it wasn't bumped yet.

We better hurry with the sriov-passthrough-cni plugin patch before it start to block PR's.

EdDev:
> My point is that the problem occurs only on CI.
> Personally I didnt encountered it on local environment, and the chance that anyone would is very low
> due to the nature of using it for development and not for CI to get test results as fast as possible.
> 
> We encounter this on CI because after the operator is ready we immediately create the policy and run tests,
> for local setup you would probably run the test after a while where the sriovNodeState is already in updated.
> And it will take less time for the operator to be ready because it doesn't need to reconfigure the VF's driver
> (it is reseted to default on CI).
> I am almost sure we didnt encountered it until the last refactor because we used to wait more between
> sriov-deployment and policy creation.

All you say here is that it has a lower chance to happen on local run because the timing.
You are not saying it is wrong to reset the numvfs to zero and you cannot assure this is not really happening (especially when one argues it does).

I also do not understand your technical reasoning here, as the argument in the flow presented in this PR has not taken into account the test run step, only the deployment of the operator and at the next step the deployment of the policy.
Aren't they deployed together?

> For CI the proper place for this change is on the plugin code, I will propose a PR.

Good luck with that.

I am playing along in this endless discussion just to make sure my agreement to this change is not broken. But so far, I did not see any convincing argument that makes it wrong.

oshoval:
> > Right, didnt know it was bumped 2 days ago, last time i checked it wasn't bumped yet.
> 
> We better hurry with the sriov-passthrough-cni plugin patch before it start to block PR's.

We can't hurry and don't need to, this PR will does the job.

qinqon:
/approve

I think the change is safe.

oshoval:
Something is weird with the current run
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/513/check-up-kind-1.17-sriov/1348618650633900032#1:build-log.txt%3A1059
first time i see dev7, i am not sure it will pass

oshoval:
The failure is again on ConnectX-5 that fails on each run of it right now i think due to the details above.
Need to catch a run on none ConnectX-5 and see it passed (it did pass on all the times i tried before this dev7 problem happened)

PR #514, opened 2021-01-04
Revert "Move etd on k8s-1.19 to in-memory storage (#498)"

rmohr:
We have mixed results. 300 MB do not seem to be enough anymore. It initially worked but fails now afer a Z release bump in kubevirtci. Reverting #498 to unblock kubevirtci.


fgimenez (approved):
/lgtm
/approve
/hold

feel free to unhold when ready

PR #499, opened 2020-12-13
QEMU VM's based providers: run etcd on RAM memory 

ormergi:
On CI and dev environments where the cluster is ephemeral and
data shouldn't be persistent it is possible to improve cluster
performance by configuring etcd to run on RAM memory.
Hopefully it will shorten the time it takes for CI job to
finish and improve CI capacity.

This is a following to what we did for KIND based provider, 
it does shorten the time it takes for a cluster to be ready and 
improve the cluster overall performance.

PR #517, opened 2021-01-07
Don't run etcd on memory by default

rmohr:
When reviewing #499 I missed that it would run etcd by default in memory. Switching the default to false.


fgimenez (approved):
/lgtm
/approve
/hold

Feel free to unhold when ready. not sure if my approve and hold commands are taken into account here, just in case :)

ormergi (approved):

ormergi:
Are we going to enable it at the provider level?

rmohr:
> Are we going to enable it at the provider level?
> If so which provider and when? :)

We can experiment with it on a PR in kubevirt/kubevirt. My first experiment with unconditionally enabling it did not go so well. With your PR we can now directly experiment in kubevirt/kubevirt and decide then if we enable it.

oshoval:
https://github.com/kubevirt/kubevirtci/pull/513 is a fix for
the failing https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/517/check-up-kind-1.17-sriov/1347105297558147072

rmohr:
@oshoval overriding then. Thanks.

/hold cancel
/override check-up-kind-1.17-sriov

PR #524, opened 2021-01-14
Move k8s 1.18 to crio 

rhrazdil:


cluster-provision/k8s/1.18/nodes.sh
```diff
@@ -16,8 +16,16 @@ while ! hostnamectl  |grep Transient ; do
     fi
 done
 
-# Wait for docker, else network might not be ready yet
-while [[ `systemctl status docker | grep active | wc -l` -eq 0 ]]
+# https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cgroup-driver
+mkdir -p /etc/crio/crio.conf.d/
+cat <<EOF >/etc/crio/crio.conf.d/01-cgroup-manager.conf
```
rmohr, line 21: Following our offline discussion, I think that we basically override kubelet settings from the provision stage instead of merging them together in `nodes.sh`.

Can you create something like a `common.sh` script where you add common featuregates and configurations and then simply add additional featuregates for `nodes.sh`, so that they have the cpumanager and such?

rmohr:
/retest

The crio repos should be fixed now.

PR #531, opened 2021-01-21
sriov, Improve CNI downloading

oshoval:
Instead of downloading the CNI for every node,
download it once, and deploy it on all the nodes.

For local deployments it will just download it
if it doesn't exists already, which will improve
cluster-up timing more.

Download time ranges between 10 seconds to around 1 minute,
depends on the connection between the client and the server.

Additional small changes:
* Silent the curl download progress.
* Reuse CNI tar name by creating a var for it.
* Rearrange installation of cni-plugins and calico cni in functions.

### DIFF
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -92,6 +92,35 @@ function _configure_registry_on_node() {
     ${NODE_CMD} $1  sh -c "echo $(docker inspect --format '{{.NetworkSettings.IPAddress }}' $REGISTRY_NAME)'\t'registry >> /etc/hosts"
 }
 
+function _install_cnis {
+    _install_cni_plugins
+    _install_calico_cni
+}
+
+function _install_cni_plugins {
+    local CNI_VERSION="v0.8.5"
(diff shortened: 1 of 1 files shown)


alonSadan (approved):
Nice!

oshoval:
Rebased

alonSadan (approved):

oshoval:
Thanks Alon

cluster-up/cluster/kind/common.sh
```diff
@@ -168,8 +168,17 @@ function setup_kind() {
         done
     fi
 
+    local CNI_VERSION="v0.8.5"
```
EdDev, line 171: It will be nicer to move all this section to a function.

oshoval: on it

oshoval:
Addressed comments.

EdDev (approved):
I won't insist on this because these are just scripts and not production code.
However, placing the initial code in a function is much easier to review than moving it after it was added to a function.

oshoval:
Rebased (${ARCH} support)

EdDev (approved):

PR #535, opened 2021-01-24
sriov, Parameterize KIND_NODE_IMAGE

oshoval:
The change allows to easily use a custom image
for testing locally.

For example a custom image with pre pulled images
(`calico`, `multus`, `sriov-operator`),
in order to improve `cluster-up` time
(almost 20-40% faster, in case image includes also the CNI),
and without getting `docker.io` pull rate limits.

Later on, in case desired, we can make an image like this
the default, with a mechanism to provision those images.
It will also support freezing `sriov-operator` version.


oshoval:
https://github.com/kubevirt/kubevirtci/pull/540 takes it to the next step

PR #542, opened 2021-02-02
Allow more than 8 nodes cluster

qinqon:
There is a bug at vm.sh that fails if KUBEVIRT_NUM_NODES=8, this change
convert the number into decimal so we can use printf as expected.

failure at vm.sh

```
+ NODE_NUM=08
++ printf %02d 08
/vm.sh: line 36: printf: 08: invalid octal number
+ n=00
```


fgimenez (approved):
/lgtm

dhiller (approved):
/approve

PR #545, opened 2021-02-09
WIP SRIOV Provider: extract sriov-operator deployment to its own script

ormergi:
On SRIOV provider cluster-up, after the cluster is ready `config_sriov.sh` script is invoked, 
it discovers the host SRIOV PF's, moves them to the cluster nodes netns and deploys sriov-network-operator.
sriov-network-operator deployment is not straightforward as one would think,
it consists of multiple phases deploying the operator, creating a policy and verifying all components are ready.

This script is very long, complex and hard to maintain.

This PR decouples this script by extracting the code for deploying sriov-network-operator to another script.
so there will be two scripts one for configuring the cluster nodes with PF's and another script for sriov-operator deployment.

Decoupling this script requires passing the _PF names_ that discovered and moved by `config_sriov.sh`
to `deploy-sriov-operator.sh` script, which eventually patch SriovNetworkNodePolicy manifest with them.
There is no reasonable way to pass data between two independent scripts.
To workaround this, each node that has sriov PF will have an annotation with the PF's name, 
this will be done by `config_sriov.sh`.
Then, `deploy-sriov-operator.sh` script will get the PF names from annotation just before SriovNetworkNodePolicy 
manifest patch.
Having those annotations will improve the visibility of which node has what PF in its netns.

This change is needed as initial step for adding  SRIOV Live Migration support on  this provider.


cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
```diff
@@ -0,0 +1,254 @@
...
+wait_pods_ready
+
+# Patch SriovNetworkNodesPolicy
+policy_template="$MANIFESTS_DIR/network_config_policy.yaml"
+policy="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/network_config_policy.yaml"
+cp -f $policy_template $policy
+
+used_pfs_names_annotations=$(_kubectl get nodes -l $SRIOV_NODE_LABEL \
+  -o jsonpath="{.items[*].metadata.annotations.$NODE_PFS_ANNOTATION}")
+vfs_count=$(cat /sys/bus/pci/devices/*/sriov_totalvfs | head -1)
```
oshoval, line 239: This might ruin the multi sriov, please verify which PFs you see here,
in case you see them all, and not just the ones the undercloud CNI allocated to this job, 
then please dont use /sys/bus/pci, but only the ones in the cluster namespace

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
```diff
@@ -0,0 +1,254 @@
...
+  return 0
+}
+
+function apply_sriov_node_policy {
+  local -r policy_file=$1
+  local -r pfs_names_json_arrays=$2
+  local -r num_vfs=$3
+  
+  # Format to singal array ["ens1","ens2"],["ens3"] => ["ens1","ens2","ens3"]
+  local all_pfs_names_json_array=$(sed 's/\].*\[/"," /g' <<< "$pfs_names_json_arrays")
```
oshoval, line 213: Please use envsubst as we did before, i find sed here too complex and not maintainable.

oshoval (changes requested):
Please make sure that you don't corrupt the multi sriov.
Lets talk if you need.

I don't find this PR mandatory yet, and its risky in its current state for multi sriov IIUC.
If possible please do it in stages, and lets check that the stages doesn't break multi sriov , and that we choose the simplest design that is possible (especially for the manifest, see inline please).

For example

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
```diff
@@ -0,0 +1,254 @@
...
+  
+  # Patch SriovNetworkNodesPolicy with PF's names and VF's count
+  sed -i "s?pfNames:.*?pfNames: $all_pfs_names_json_array?g" $policy_file
```
oshoval, line 216: please use envsubst instead, you can remove the export usage, by doing 
VAR=VALUE command

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
ormergi: `sriov_totalvfs` file represent the VF count that configured on the card firmware which is identical to all PF's on the host
thus It should not break multi cluster functionality.

ormergi:
> Please make sure that you don't corrupt the multi sriov.
> Lets talk if you need.
> 
> I don't find this PR mandatory yet, and its risky in its current state for multi sriov IIUC.
> If possible please do it in stages, and lets check that the stages doesn't break multi sriov , and that we choose the simplest design that is possible (especially for the manifest, see inline please).
> 
> Stage 1: only split to files.
> Stage 2: The additional changes.

Hi Or, unfortunately Its not possible to do this change in phases as you suggest
due to the fact that it is necessary to pass the PF names from one script to another 
(I elaborated about it in the PR description)

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
oshoval: You can configure it per PF.
In case we leave it like this its an assumption that at least need to be asserted please.

oshoval:
> > Please make sure that you don't corrupt the multi sriov.
> > Lets talk if you need.
> > I don't find this PR mandatory yet, and its risky in its current state for multi sriov IIUC.
> > If possible please do it in stages, and lets check that the stages doesn't break multi sriov , and that we choose the simplest design that is possible (especially for the manifest, see inline please).
> > Stage 1: only split to files.
> > Stage 2: The additional changes.
> 
> Hi Or, unfortunately Its not possible to do this change in phases as you suggest
> due to the fact that it is necessary to pass the PF names from one script to another
> (I elaborated about it in the PR description)

Hi,
Since you have the PFs in the node, you can look in the container instead of annotating the node or moving it between files.
Please see that current code already does it, it looks in the node after the PF is there.

For the start you can export, as there is just one PF used for now.
This will allow you to split it to two PRs:
refactor and logic changes.

You can also maybe find out a bash struct that have the information per node and then export it.
but the best is just to look inside the node imo.

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
oshoval: You can configure it per PF.
In case we leave it like this its an assumption, and we better not have it, as there is a simpler solution:
just see how we take the VF count already, by looking in the node itself, this way it will be more robust.

EdDev (changes requested):
>  config_sriov.sh: Extract operator deployment code
>
> 
> This commit only extract sriov-operator deployment code
> to new file.

How about: `This change extracts .... to a new file.

Try to explain why shortly.

> Other changes will be preseted in following commits.

This can be dropped.

> 
> Conflicts:
> 	cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh

this can be dropped.

cluster-up/cluster/kind-k8s-sriov-1.17.0/manifests/network_config_policy.yaml
```diff
@@ -8,9 +8,8 @@ spec:
   mtu: 1500
   nodeSelector:
     sriov: "true"
-  numVfs: $NODE_PF_NUM_VFS
+  numVfs:
```
phoracek, line 11: This is not a valid yaml

cluster-up/cluster/kind-k8s-sriov-1.17.0/manifests/network_config_policy.yaml
```diff
@@ -8,9 +8,8 @@ spec:
   mtu: 1500
   nodeSelector:
     sriov: "true"
-  numVfs: $NODE_PF_NUM_VFS
+  numVfs: 
   nicSelector:
     pfNames:
```
phoracek, line 13: this

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -60,8 +63,11 @@ NODE_PFS=($(move_sriov_pfs_netns_to_node "$SRIOV_NODE" "$PF_COUNT_PER_NODE"))
...
 ${SRIOV_NODE_CMD} mount -o remount,rw /sys     # kind remounts it as readonly when it starts, we need it to be writeable
 ${SRIOV_NODE_CMD} chmod 666 /dev/vfio/vfio
-_kubectl label node $SRIOV_NODE sriov=true
+_kubectl label node $SRIOV_NODE $SRIOV_NODE_LABEL
 
 for pf in "${NODE_PFS[@]}"; do
   docker exec $SRIOV_NODE bash -c "echo 0 > /sys/class/net/$pf/device/sriov_numvfs"
 done
+
+used_pfs_yaml_format_array=$(sed 's/ /, /g' <<< "[\"${NODE_PFS[@]}\"]")
```
EdDev, line 72: I found this neat join implementation: `function join_by { local IFS="$1"; shift; echo "$*"; }` , then you call it with `join_by , [\"${foolist[*]}\"]`. (the only difference is that there is no space after the `,`.

Saw it [here](https://stackoverflow.com/questions/1527049/how-can-i-join-elements-of-an-array-in-bash) on the accepted answer (it includes a more generalize solution if you need multi-char separators).

You may probably ask why it is better than `sed`, right? :)
Well, it is using bash built-in, so it does not execute it in a sub-shell.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -9,6 +9,9 @@ PF_COUNT_PER_NODE=${PF_COUNT_PER_NODE:-1}
 
 [ $PF_COUNT_PER_NODE -le 0 ] && echo "FATAL: PF_COUNT_PER_NODE must be a positive integer" >&2 && exit 1
 
+SRIOV_NODE_LABEL=${SRIOV_NODE_LABEL:-"sriov=true"}
+NODE_PFS_ANNOTATION=${NODE_PFS_ANNOTATION:-"node_pfs"}
```
EdDev, line 13: Why do you need the key to be overwritten? I could not find the reason in the commit.

EdDev (changes requested):
Review 2nd commit: config_sriov.sh: Add used PF's names annotation 

There are some typos in the text and a conflict clause appears in the end.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I added those so those variable will have default value and for visibility, we already depend on few env vars that you need to investigate where they were came from.

cluster-up/cluster/kind-k8s-sriov-1.17.0/manifests/network_config_policy.yaml
ormergi: This yaml file propose is a template, it is copied and patched later on with 
`numVfs` and `pfNames`

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
EdDev: You should be able to access the device from the interface: `/sys/class/net/ens1f0/device/sriov_totalvfs`.

You will need to build the policy from a template one pf at a time. But I am ok with asserting that the vf-count is indeed identical on all used pf/s from the list (but not `*`). If it is not, we should fail.

EdDev (changes requested):

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
```diff
@@ -27,8 +27,11 @@ function up() {
 
     # remove the rancher.io kind default storageClass
     _kubectl delete sc standard
-
+    
+    export SRIOV_NODE_LABEL=${SRIOV_NODE_LABEL:-"sriov=true"}
```
EdDev, line 31: Seems like we have it all over, lets try to place it in a single place.

EdDev (changes requested):
Review last commit: SRIOV provider: Adapt provider.sh script

Please explain what it does. The addition if the exported vars does not make it easy as well.

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
ormergi: `sriov_*total*vfs` is read only file and should not be changed by the user, 
it should be consistent across PF's as it represents the hardware limitation 
it is mentioned at intel's documentation https://www.intel.com/content/www/us/en/embedded/products/networking/sr-iov-guide-ethernet-x710-xl710-linux-brief.html

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
```diff
@@ -203,14 +206,19 @@ function deploy_sriov_operator {
...
+  local -r pfs_names_json_arrays=$2
   local -r num_vfs=$3
-
-  # Substitute $NODE_PF and $NODE_PF_NUM_VFS and create SriovNetworkNodePolicy CR
-  local -r policy=$(NODE_PF=$node_pf NODE_PF_NUM_VFS=$num_vfs envsubst < $policy_file)
-  echo "Applying SriovNetworkNodeConfigPolicy:"
-  echo "$policy"
-  _kubectl create -f - <<< "$policy"
+  
+  # Format to singal array ["ens1","ens2"],["ens3"] => ["ens1","ens2","ens3"]
```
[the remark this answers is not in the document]
ormergi: For the case where there are more then one workers with PF attached
This PR change also supporting multiple nodes

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
```diff
@@ -203,14 +206,19 @@ function deploy_sriov_operator {
...
+  local all_pfs_names_json_array=$(sed 's/\].*\[/"," /g' <<< "$pfs_names_json_arrays")
+  
+  # Patch SriovNetworkNodesPolicy with PF's names and VF's count
+  sed -i "s?pfNames:.*?pfNames: $all_pfs_names_json_array?g" $policy_file
+  sed -i "s?numVfs:.*?numVfs: $num_vfs?g" $policy_file
+
+  echo "Create SriovNetworkNodeConfigPolicy"
+  cat $policy_file
```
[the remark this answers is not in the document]
ormergi: With this PR changes there is command that echo the policy content so it is necessary to use `cat`

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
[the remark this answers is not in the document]
EdDev: Sorry, I do not get it. Each array is representing a worker? And in each array we have the PF/s that have been assign to them?

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
ormergi: I can remove the deceleration of those variable from `config_sriov.sh` and `deploy_sriov_operator.yaml` and add a check that they are initialized instead, 
what you think?

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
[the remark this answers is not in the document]
EdDev: Aha.. it comes from `_kubectl get nodes`. I think it will be easier to read by handling one node at a time in the flow.
But if it cannot work, just use a for-loop, it would be easier to name things and read the "why".

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
ormergi: At this point the PF's moved to the nodes netns so it should be invoked from the container
I trying to avoid this, so this script could do only the deployment w/o investigation the nodes.

I dont think we need to create a policy for each PF if this what you mean,
we should create single policy that will include all PF's and the VF's count (which is identical to all PF's)
so we could have single VF's pool for all the cluster at we expect on tests.

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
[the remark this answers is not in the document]
ormergi: Yes we fetch the node_pfs annotation from all nodes with label `sriov: true`.
if we would have a cluster with multiple nodes where each node have more then one PF, the annotations would looke like:
  node1: `["pf1","pf2"]`
  node1: `["pf3","pf4"]`
Then in order to include all those in the sriov policy it should be formated to single json array:
  `["pf1","pf2","pf3","pf4"]`

Maybe it will be easier to understand if the formatting was next to the part where we read the annotations,and better naming

cluster-up/cluster/kind-k8s-sriov-1.17.0/deploy_sriov_operator.sh
oshoval: > it is mentioned at intel's documentation https://www.intel.com/content/www/us/en/embedded/products/networking/sr-iov-guide-ethernet-x710-xl710-linux-brief.html

On Mellanox you can config it per PF.
and even if it wasn't the case, one PF might be corrupted on the machine, and then we wouldn't able to use it, we would black list it, but it might be the first, and mislead us in case its MAX_VF is zero.
I would suggest to make it robust and env agnostic please.

> At this point the PF's moved to the nodes netns so it should be invoked from the container
I trying to avoid this, so this script could do only the deployment w/o investigation the nodes.

I find inspecting the nodes and keeping the code more robust and simpler, better than assuming and using the annotation,
the data is already there, no need to have it twice.
Moreover if the PF is missing for any reason after it was allocated (it might happen due to bugs, especially if a container is deleted), using the annotation is less safe than inspecting the container.

oshoval:
Please consider the design that the current code allows, before doing the refactor.
It will allow also to split the work to parts, and test easier that each of the parts works both for sriov migration and multi sriov jobs.

The design:
Just update config_sriov.sh to run on all the nodes (something like you did on the other kubevirtci migration draft PR), allocate 1 PF for each, using the existing function that already support it (just node loop is missing).
And then you have all the info in the same file and can transform it to csv and update the policy in case there is just one policy, or do whatever u need if there are couple policies in the end.

I find the annotation adding a layer that cost more than the refactor, i am less fond of this refactor and the order of the operations, all encapsulated together.
Doing it in steps, would make the processes easier and potentially more natural imho.

ormergi:
/close

Thanks for the feedback! 
I will present this decoupling in another PR

PR #550, opened 2021-02-21
sriov VM, Disable qemu guest agent at startup

oshoval:
There is a race condition at Kubevirt e2e tests,
between LoginToFedora and cloud-init which sets the password.

In order to fix this race condition and make sure
cloud-init finished, we disable quest agent when creating the VM image.
The e2e tests would enable it as the last thing
in cloud-init UserData, wait for the guest agent to finish
and only then login the the VM.

Notes:
1. The next reboot won't be idempotent, because
the guest agent would be reinstalled on the first boot of the e2e tests,
using the Cloud-init UserData.
We don't reboot sriov VMs in e2e tests so its ok.
In case reboot is needed, the same sequence can be done before the
reboot from the test.

Kubevirt counterpart:
https://github.com/kubevirt/kubevirt/pull/5072

Built using:
```
export TAG="32"
export OS_VARIANT=fedora32
```

See more info on
https://github.com/kubevirt/kubevirtci/pull/533
https://github.com/kubevirt/kubevirt/pull/4839#discussion_r561004638

### DIFF
--- a/cluster-provision/images/vm-image-builder/fedora-sriov-lane/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-sriov-lane/cloud-config
@@ -22,5 +22,8 @@ runcmd:
   - sudo dnf clean all
   - sudo hostnamectl set-hostname ""
   - sudo hostnamectl set-hostname "" --transient
-  - sudo systemctl start qemu-guest-agent
+  - sudo systemctl stop qemu-guest-agent
+  - sudo systemctl disable qemu-guest-agent
+  - sudo mv /lib/systemd/system/qemu-guest-agent.service /home/fedora/
+  - sudo systemctl daemon-reload
   - sudo shutdown


cluster-provision/images/vm-image-builder/fedora-sriov-lane/cloud-config
```diff
@@ -22,5 +22,8 @@ runcmd:
   - sudo dnf clean all
   - sudo hostnamectl set-hostname ""
   - sudo hostnamectl set-hostname "" --transient
-  - sudo systemctl start qemu-guest-agent
+  - sudo systemctl stop qemu-guest-agent
+  - sudo systemctl disable qemu-guest-agent
+  - sudo mv /lib/systemd/system/qemu-guest-agent.service /home/fedora/
```
ormergi, line 27: Would you try this with `disable --now` it think remove the service as well so we wont need to do it manually

oshoval: I think i tried in the past and it didn't work,
(the important thing is that it will survive reboot, and be kept offline)
please see more info on 
https://github.com/kubevirt/kubevirtci/pull/533
"due to a service dependency tree"
If needed i will retry.

ormergi: I dont recall that I had this kind of issues when creating this image and according to [systemctl docs](https://www.freedesktop.org/software/systemd/man/systemctl.html)
using `sudo systemctl disable  --now` should do the trick in one line and simply it.

Then on the "client" (kubevirt tests, cloud init) we would need to `sudo systemctl enable  --now`
and thats it, instead of messing with the service file

EdDev:
> There is a race condition at Kubevirt e2e tests,
> between LoginToFedora and cloud-init which sets the passwor

Can you please explain what race condition?

I find requiring the tests to start the guest-agent as a problem.
I want the tests to do nothing, not even to perform a password setting.

Can we just setup the password in the provisioning of the VM instead of requiring it in the test? Will that solve the problem?

oshoval:
> > There is a race condition at Kubevirt e2e tests,
> > between LoginToFedora and cloud-init which sets the passwor
> 
> Can you please explain what race condition?

The test does the following:
1. Wait for VM to reach status Scheduled / Running
2. LoginToFedora using fedora:fedora via console expecter

Parallel to the test:
a. the VM is running and doesn't have user fedora:fedora (the one it has in the VM builder
isn't persistent).
b. The Cloud-init is the one that set the password.

In case "2" happens before "b", it won't manage to login, it does have one retry that solve some of the cases (but not all of them [1]) but it also has a timeout, and even worse, it might get out of sync with the console [2]

[1] https://github.com/kubevirt/kubevirt/pull/4839#discussion_r561004638
Shows clearly that the login attempt occurs before the password is set.
There were flakes issues that were open on it but they are auto closed after 2 weeks (build-watcher label auto close).
[2] https://github.com/kubevirt/kubevirt/issues/3821

> 
> I find requiring the tests to start the guest-agent as a problem.
> I want the tests to do nothing, not even to perform a password setting.
> 
> Can we just setup the password in the provisioning of the VM instead of requiring it in the test? Will that solve the problem?

This was the original idea that i tried (see [1] above for more info), didn't manage yet.
But i found something after finishing the PR that might explain why setting the password via the builder didn't work,
so i will retry.
It will solve the main problem which is this race.

cluster-provision/images/vm-image-builder/fedora-sriov-lane/cloud-config
oshoval: I tried it in the past, and tried it again now,
it stop it not persistently between reboots.

[fedora@testvmi1 ~]$ sudo systemctl stop qemu-guest-agent
[fedora@testvmi1 ~]$ sudo systemctl disable --now qemu-guest-agent
Removed /etc/systemd/system/dev-virtio\x2dports-org.qemu.guest_agent.0.device.wants/qemu-guest-agent.service.

Reboot
and qemu-ga is back online

anyhow i am investigating what Edy asked on the other thread so in case it will work, it will replace this method.

ormergi:
On the current version of the image we use on the SRIOV lane there is no need to configure a user and password
its already configured.

oshoval:
> On the current version of the image we use on the SRIOV lane there is no need to configure a user and password
> its already configured.

You are welcome to try to login without setting it in Cloud-init UserData in the test and you will see it doesn't work.
The config in VM builder is not persistant.
Hence there is this race.

cluster-provision/images/vm-image-builder/fedora-sriov-lane/cloud-config
ormergi: > You are welcome to try to login without setting it in Cloud-init UserData in the test and you will see it doesn't work.
I dont know what could-init does in the backgammon that wont let you use it w/o setting a user..
When we came up with this image one think we wanted to achieve is to not use cloud-init at all.
If you will spin up a VM with that image w/o cloud-init, you would see its there.

> The config in VM builder is not persistant.
It is resistant you could check other configurations it does e.g: remove the default host name, load kernel modules for sriov drivers..

ormergi:
> You are welcome to try to login without setting it in Cloud-init UserData in the test and you will see it doesn't work.

I dont know what could-init does in the backgammon that wont let you use it w/o setting a user..
When we came up with this image one think we wanted to achieve is to not use cloud-init at all.
If you will spin up a VM with that image w/o cloud-init, you would see its there.

> The config in VM builder is not persistant.

It is resistant you could check other configurations it does e.g: remove the default host name, load kernel modules for sriov drivers..

oshoval:
> > You are welcome to try to login without setting it in Cloud-init UserData in the test and you will see it doesn't work.
> 
> I dont know what could-init does in the background that wont let you use it w/o setting a user..
> When we came up with this image one of the things we wanted to achieve is to not use cloud-init on sriov tests at all (which was escalated quickly because we need it for ping tests)
> If you will spin up a VM with that image w/o cloud-init, you would see its there.

Try please first.
Take e2e test, remove the setting of the password from the UserData and see login failed.
See here a conversion about it, and logs
https://github.com/kubevirt/kubevirt/pull/4839#discussion_r561004638

ormergi:
> Try please first.

What makes you think I didnt?

> Take e2e test, remove the setting of the password from the UserData and see login failed.
> See here a conversion about it, and logs
> [kubevirt/kubevirt#4839 (comment)](https://github.com/kubevirt/kubevirt/pull/4839#discussion_r561004638)
> 
> > It is persistent if it wasn't there was no point in having it..
> 
> The config of the VM builder did add the drivers and the guest agent which is helpful, but not the password.

As I said I dont know what cloud-init does on the background, I dont know why if you wont spesify the user and password on cloud-init you can login.
what I do know is that if you spin up a VM with this image the user and password are configured.

oshoval:
> > Try please first.
> 
> What makes you think I didnt?
> 
> > Take e2e test, remove the setting of the password from the UserData and see login failed.
> > See here a conversion about it, and logs
> > [kubevirt/kubevirt#4839 (comment)](https://github.com/kubevirt/kubevirt/pull/4839#discussion_r561004638)
> > > It is persistent if it wasn't there was no point in having it..
> > 
> > 
> > The config of the VM builder did add the drivers and the guest agent which is helpful, but not the password.
> 
> As I said I dont know what cloud-init does on the background, I dont know why if you wont specify the user and password on cloud-init you cant login.
> what I do know is that if you spin up a VM with this image the user and password are configured.

Did you remove the password setting from the Cloud-init of the VMI ?
I just verified it again locally and it fails
PR for reference 
https://github.com/kubevirt/kubevirt/pull/5101

oshoval:
Moved to https://github.com/kubevirt/kubevirtci/pull/556

PR #507, opened 2020-12-24
sriov live migration setup

ormergi:
Currently SRIOV provider we use sriov-network-operator to deploy sriov-cni and sriov-device-plugin
and configure PF's and VF's.
sriov-network-operator was meant to run on full flagged Openshift with bare-metal nodes that has 
SRIOV hardware, and not on containerized Kubernetes (KIND) setup.

We need to add support to sriov provider to enable multiple sriov capable worker nodes setup 
for Kubevirt SRIOV Live Migration feature.
At the moment we use version 4.4 where sriov-device-plugin is not compatible running on 
containerized environments, causing VM's creation to fail due to VF allocation from other nodes.

This PR simplify the sriov provider by removing sriov-network-operator,
Instead it will configure PFs and VFs statically and deploy multus sriov-cni, sriov-device-plugin.
The changes includes supporting multiple sriov capable worker nodes.

PR #551, opened 2021-02-24
sriov, Support running multi clusters side by side (locally)

oshoval:
Add support for running two sriov clusters side by side,
locally, with few known limitations.

The feature allows:
1. Context switch easier between clusters in case needed,
without the need to tear down, and recreate the cluster.
2. Testable environment for the multi sriov feature.

The clusters can be spinned at the same time, as the PF allocation
is reentrant.

This feature doesn't affect the default behavior.

Limitations:
* Each cluster should be created on its own git clone folder, i.e
/root/project/kubevirtci1
/root/project/kubevirtci2
* Locally only two clusters are supported, at the same time.

Note:
In order to not have cross talk between the clusters,
for example PF resetting each other, or adding PFs which were not requested,
the bump of sriov CNI / DP is needed, and it is done on other PR. 
see #507 and #486 (without /  with the sriov-operator) for reference.
 
### DIFF
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
@@ -4,7 +4,7 @@ Provides a pre-deployed k8s cluster with version 1.17.0 that runs using [kind](h
 The KubeVirt containers are built on the local machine and are then pushed to a registry which is exposed at
 `localhost:5000`.
 
-This version also expects to have sriov-enabed nics on the current host, and will move all the physical interfaces and virtual interfaces into the `kind`'s cluster master node so that they can be used through multus.
+This version also expects to have sriov-enabled nics on the current host, and will move physical interfaces into the `kind`'s cluster worker node(s) so that they can be used through multus.
 
 ## Bringing the cluster up
 
@@ -17,8 +17,9 @@ The cluster can be accessed as usual:
 
 ```bash
 $ cluster-up/kubectl.sh get nodes
-NAME                  STATUS    ROLES     AGE       VERSION
-sriov-control-plane   Ready     master    2m33s     v1.17.0
+NAME                  STATUS   ROLES    AGE     VERSION
+sriov-control-plane   Ready    master   6m14s   v1.17.0
+sriov-worker          Ready    worker   5m36s   v1.17.0
 ```
 
 ## Bringing the cluster down
@@ -42,3 +43,33 @@ export KUBECTL_PATH="/usr/bin/kubectl"
 ```
 This allows users to test or use custom images / different kind versions before making them official.
 See https://github.com/kubernetes-sigs/kind/releases for details about node images according to the kind version.
+
+## Running multi sriov clusters locally
+Kubevirtci sriov provider supports running two clusters side by side with few known limitations.
+
+General considerations:
+
+- A sriov PF must be available for each cluster.
+In order to achieve that, there are two options:
+1. Assign just one PF for each worker node of each cluster by using `export PF_COUNT_PER_NODE=1` (this is the default value).
+2. Optional method: `export PF_BLACKLIST=<PF names>` the non used PFs, in order to prevent them from being allocated to the current cluster.
+The user can list the PFs that should not be allocated to the current cluster, keeping in mind
+that at least one (or 2 in case of migration), should not be listed, so they would be allocated for the current cluster.
+Note: another reason to blacklist a PF, is in case its has a defect or should be kept for other operations (for example sniffing).
+- The cluster names must be different.
+This can be achieved by setting `export CLUSTER_NAME=sriov2` on the 2nd cluster.
+The default `CLUSTER_NAME` is `sriov`.
+The 2nd cluster registry would be exposed at `localhost:5001` automatically, once the `CLUSTER_NAME`
+is set to a non default value.
+- Each cluster should be created on its own git clone folder, i.e
+`/root/project/kubevirtci1`
+`/root/project/kubevirtci2`
+In order to switch between them, change dir to that folder and set the env variables `KUBECONFIG` and `KUBEVIRT_PROVIDER`.
+- In case only one PF exists, for example if running on prow which will assign only one PF per job in its own DinD,
+Kubevirtci is agnostic and nothing needs to be done, since all conditions above are met.
+- Upper limit of the number of clusters that can be run on the same time equals number of PFs / number of PFs per cluster,
+therefore, in case there is only one PF, only one cluster can be created.
+Locally the actual limit currently supported is two clusters.
+- Kubevirtci supports starting `cluster-up` simultaneously, since it is capable of handling race conditions,
+when allocating PFs.
+- In order to use `make cluster-down` please make sure the right `CLUSTER_NAME` is exported.
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
@@ -240,9 +240,15 @@ function move_sriov_pfs_netns_to_node {
       continue
     fi
 
-    ip link set "$pf_name" netns "$node"
-    pf_array+=("$pf_name")
-    [ "${#pf_array[@]}" -eq "$pf_count_per_node" ] && break
+    # In case two clusters started at the same time, they might race on the same PF.
+    # The first will manage to assign the PF to its container, and the 2nd will just skip it
+    # and try the rest of the PFs available.
+    if ip link set "$pf_name" netns "$node"; then
+      if timeout 10s bash -c "until ip netns exec $node ip link show $pf_name > /dev/null; do sleep 1; done"; then
+        pf_array+=("$pf_name")
+        [ "${#pf_array[@]}" -eq "$pf_count_per_node" ] && break
+      fi
+    fi
   done
 
   [ "${#pf_array[@]}" -lt "$pf_count_per_node" ] && \
--- a/cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
+++ b/cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
@@ -2,7 +2,16 @@
 
 set -e
 
-export CLUSTER_NAME="sriov"
+DEFAULT_CLUSTER_NAME="sriov"
+DEFAULT_HOST_PORT=5000
+ALTERNATE_HOST_PORT=5001
+export CLUSTER_NAME=${CLUSTER_NAME:-$DEFAULT_CLUSTER_NAME}
+
+if [ $CLUSTER_NAME == $DEFAULT_CLUSTER_NAME ]; then
+    export HOST_PORT=$DEFAULT_HOST_PORT
+else
+    export HOST_PORT=$ALTERNATE_HOST_PORT
+fi
 
 function set_kind_params() {
     export KIND_NODE_IMAGE="${KIND_NODE_IMAGE:-kindest/node:v1.17.0}"
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -86,7 +86,7 @@ function _run_registry() {
         docker rm $REGISTRY_NAME || true
         sleep 5
     done
-    docker run -d -p 5000:5000 --restart=always --name $REGISTRY_NAME registry:2
+    docker run -d -p $HOST_PORT:5000 --restart=always --name $REGISTRY_NAME registry:2
 }
 
 function _configure_registry_on_node() {
@@ -129,7 +129,7 @@ function prepare_config() {
 master_ip="127.0.0.1"
 kubeconfig=${BASE_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
 kubectl=${BASE_PATH}/$KUBEVIRT_PROVIDER/.kubectl
-docker_prefix=localhost:5000/kubevirt
+docker_prefix=localhost:${HOST_PORT}/kubevirt
 manifest_docker_prefix=registry:5000/kubevirt
 EOF
 }


cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
```diff
@@ -2,7 +2,16 @@
 
 set -e
 
-export CLUSTER_NAME="sriov"
+DEFAULT_CLUSTER_NAME="sriov"
+DEFAULT_HOST_PORT=5000
```
alonSadan, line 6: I don't know much about it, so I could be wrong, but is it safe enough to give a hard coded port, could it be take already?
Is there a way to allocate a random free port?

alonSadan (commented):
Thank you Or. I think it will be a great feature to run to clusters in parallel.

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
oshoval: Its totally safe, as long as there isn't other port that holds it.
(else cluster-up would simply fails, which is ok)

We already have port 5000 hardcoded, now you can have two clusters, each will have its own port, chosen according the cluster name.
`make cluster-sync` will be also automatic aware of it, since its propagated in the node as well, and taken from there.
I don't think we need to over engineer here.

Its in order to be able to run two cluster locally, switch easily, and in the future debug even problems that might occur if two clusters run on the same time on CI.
Atm CI is safer than locally, because it runs in DinD.

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
```diff
@@ -41,4 +42,26 @@ export KIND_VERSION="0.9.0"
...
-See https://github.com/kubernetes-sigs/kind/releases for details about node images according to the kind version.
\ No newline at end of file
+See https://github.com/kubernetes-sigs/kind/releases for details about node images according to the kind version.
+
+## Running multi sriov clusters locally
+Kubevirtci sriov provider supports running two clusters side by side with few known limitations.
+
+General considerations:
+
+- A sriov PF must be available for the cluster.
```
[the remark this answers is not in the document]
oshoval: Yep, i can update if needed

alonSadan (commented):
Maybe add A one line explanation on how to alternate between clusters?

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
alonSadan: Thanks for explaining.

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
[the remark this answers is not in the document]
alonSadan: Thanks.

oshoval:
Addressed comments,
Thanks

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
```diff
@@ -41,4 +42,32 @@ export KIND_VERSION="0.9.0"
...
+
+## Running multi sriov clusters locally
+Kubevirtci sriov provider supports running two clusters side by side with few known limitations.
+
+General considerations:
+
+- A sriov PF must be available for each cluster.
+In order to achieve that, there are two options:
+1. Assign just one PF for each cluster by using `export PF_COUNT_PER_NODE=1` (this is the default value).
+2. Optional method: `PF_BLACKLIST` the non used PFs, in order to prevent them from being allocated to the current cluster.
```
alonSadan, line 55: What do you think about changing  "`PF_BLACKLIST` the non..." to  `export PF_BLACKLIST=<PF name>` the non..." , like you wrote one line above? 
It will be a bit clearer I think.

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
```diff
@@ -41,4 +42,32 @@ export KIND_VERSION="0.9.0"
...
+The default `CLUSTER_NAME` is `sriov`.
+The 2nd cluster registry would be exposed at `localhost:5001` automatically, once the `CLUSTER_NAME`
+is set to a non default value.
+- Each cluster should be created on its own git clone folder, i.e
+`/root/project/kubevirtci1`
+`/root/project/kubevirtci2`
+In order to switch between them, change dir to that folder and set the env variables `KUBECONFIG` and `KUBEVIRT_PROVIDER`.
+- In case only one PF exists, for example if running on prow which will assign only one PF per job in its own DinD,
+Kubevirtci is agnostic and nothing needs to be done, since all conditions above are met.
+The number of clusters that can be run on the same time equals number of PFs / number of PFs per cluster,
```
alonSadan, line 70: nice.

alonSadan (commented):
Thanks.One more tiny comment.

oshoval:
Addressed comments

alonSadan (approved):
Thanks Or :)

oshoval:
Thanks Alon
/cc @qinqon 
can you please take a look ?

ormergi (changes requested):

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
```diff
@@ -41,4 +42,32 @@ export KIND_VERSION="0.9.0"
...
+See https://github.com/kubernetes-sigs/kind/releases for details about node images according to the kind version.
+
+## Running multi sriov clusters locally
+Kubevirtci sriov provider supports running two clusters side by side with few known limitations.
+
+General considerations:
+
+- A sriov PF must be available for each cluster.
+In order to achieve that, there are two options:
+1. Assign just one PF for each cluster by using `export PF_COUNT_PER_NODE=1` (this is the default value).
```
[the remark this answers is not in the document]
oshoval: Yes, but its just a suggestion on how to run it locally, and its the default as one already, and even with #570 you will still have it 1, but you will do it for every worker if i understand right.
Anyhow its just a suggestion (for local use only), which i can remove if needed, since its the default already.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -240,9 +240,15 @@ function move_sriov_pfs_netns_to_node {
...
     fi
 
-    ip link set "$pf_name" netns "$node"
-    pf_array+=("$pf_name")
-    [ "${#pf_array[@]}" -eq "$pf_count_per_node" ] && break
+    # In case two clusters started at the same time, they might race on the same PF.
+    # The first will manage to assign the PF to its container, and the 2nd will just skip it
+    # and try the rest of the PFs available.
+    if ip link set "$pf_name" netns "$node"; then
+      if timeout 10s bash -c "until docker exec $node ip address | grep -qw $pf_name; do sleep 1; done"; then
```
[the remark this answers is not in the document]
oshoval: Its not fragile this way, but i will update it as the cni-plugins/sriov-passthrough-cni/plugin/sriov-passthrough-cni does

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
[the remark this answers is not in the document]
oshoval: once you merge the migration, u can update it that it will do it per worker, wdyt?
so the user will need to distrbute his PFs, between the cluster
where each cluster takes W*P
where W = num of workers
P = PF per worker (PF_COUNT_PER_NODE)

wdyt?

oshoval:
Addressed comments (the resolved ones, the other is not an issue imo, see please inside)

oshoval:
Updated readme about `make cluster-down`

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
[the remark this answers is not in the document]
ormergi: This line is confusing, it made me though that PF_COUNT_PER_NODE control how many PF's each cluster will get instead of node
I think we need to change to reflect what the behavior would be:
Assign just one PF for each **node on each cluster **by using `export PF_COUNT_PER_NODE=1` (this is the default value).

As you may know eventually a SRIOV setup will require total of two PF's, one for each worker node to enable SRIOV live migration testing.
In case there are two cluster running on a CI node, then we would need total of 4 PF's 
2 for each cluster where each worker node will get one.
If this is how its going to work its fine by me.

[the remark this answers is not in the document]
oshoval: Migration is not yet supported, this is why i didn't update it when i created it (3 weeks ago at least, and even more since its a stripped mode of the one that had sriov-operator), and the variable name says its per NODE.
This is why i suggest to update it once the migration is merged and active on CI, not before that.
Currently NODE === Cluster because there is only one node per cluster that we support on sriov.
The right thing to be done is to update it only after migration updates this behavior, not before that,
as we look on the current state of the master.

Of course i know that, this is what i also tried to explain above,
and this is how it will work indeed, and still PF_COUNT_PER_NODE will be left as 1,
even with migration, because its one per node (this is what we agreed when we added this env var).

I will update it just to make the pipeline faster, but please consider next time that we should look om the current state of master, as we usually do.

oshoval:
Update according comment.

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
```diff
@@ -4,7 +4,7 @@ Provides a pre-deployed k8s cluster with version 1.17.0 that runs using [kind](h
 The KubeVirt containers are built on the local machine and are then pushed to a registry which is exposed at
 `localhost:5000`.
 
-This version also expects to have sriov-enabed nics on the current host, and will move all the physical interfaces and virtual interfaces into the `kind`'s cluster master node so that they can be used through multus.
+This version also expects to have sriov-enabled nics on the current host, and will move all the physical interfaces and virtual interfaces into the `kind`'s cluster master node so that they can be used through multus.
```
ormergi, line 7: Could you also fix this part?

 _and will move all the physical interfaces and virtual interfaces into the `kind`'s cluster master node so that they can be used through multus._

 _and will move all the physical interfaces and virtual interfaces into the `kind`'s cluster **worker** node so that they can be used through multus._

oshoval:
Addressed comments (with little more modifications in the same paragraph)

ormergi:
Following this thread https://github.com/kubevirt/kubevirtci/pull/551#discussion_r596706032 (its resolved so I couldn't comment there)

 Changing this variable name to *CLUSTER is not related to the migration part,
actually its related to this PR so if want to change its name this is the right place to do so.

Considering this PR the PF_COUNT_PER_NODE no longer has proper name that describe what it does
and it should be changed

oshoval:
> Following this thread [#551 (comment)](https://github.com/kubevirt/kubevirtci/pull/551#discussion_r596706032) (its resolved so I couldn't comment there)
> 
> Changing this variable name to *CLUSTER is not related to the migration part,
> actually its related to this PR so if want to change its name this is the right place to do so.
> 
> Considering this PR the PF_COUNT_PER_NODE no longer has proper name that describe what it does
> and it should be changed

so to change PF_COUNT_PER_NODE to PF_COUNT_PER_WORKER_NODE ? 
I don't find it crucial, unless you insist (not sure if you insist or not)

Again i think that we are violating a rule here, migration PRs are not merged, and we should look on the current master state.
Please consider that.

ormergi:
> > Following this thread [#551 (comment)](https://github.com/kubevirt/kubevirtci/pull/551#discussion_r596706032) (its resolved so I couldn't comment there)
> > Changing this variable name to *CLUSTER is not related to the migration part,
> > actually its related to this PR so if want to change its name this is the right place to do so.
> > Considering this PR the PF_COUNT_PER_NODE no longer has proper name that describe what it does
> > and it should be changed
> 
> so to change PF_COUNT_PER_NODE to PF_COUNT_PER_WORKER_NODE ?
> I don't find it crucial, unless you insist (not sure if you insist or not)
> 
> Again i think that we are violating a rule here, migration PRs are not merged, and we should look on the current master state.
> Please consider that.
> 
> EDIT
> thinking on it again, this change you suggest is not related to this PR, see please PR description,
> its to allow running multi clusters locally.

I meant to change it to PF_COUNT_PER_CLUSTER

oshoval:
> > so to change PF_COUNT_PER_NODE to PF_COUNT_PER_WORKER_NODE ?
> > I don't find it crucial, unless you insist (not sure if you insist or not)
> > Again i think that we are violating a rule here, migration PRs are not merged, and we should look on the current master state.
> > Please consider that.
> > EDIT
> > thinking on it again, this change you suggest is not related to this PR, see please PR description,
> > its to allow running multi clusters locally.
> 
> I meant to change it to PF_COUNT_PER_CLUSTER

We agreed on that name and nothing changed since:
a node is the same node from the eyes of the cluster that is currently `cluster-up`
there is no person that sits outside of the cluster and looks on it and then he can know its per cluster
I don't think we should change it in this PR,
in case you do please update it in following PRs.

ormergi (approved):

oshoval:
/cc @qinqon 
Can you please take a look ?
Thanks

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
```diff
@@ -41,4 +42,33 @@ export KIND_VERSION="0.9.0"
...
+The default `CLUSTER_NAME` is `sriov`.
+The 2nd cluster registry would be exposed at `localhost:5001` automatically, once the `CLUSTER_NAME`
+is set to a non default value.
+- Each cluster should be created on its own git clone folder, i.e
+`/root/project/kubevirtci1`
+`/root/project/kubevirtci2`
+In order to switch between them, change dir to that folder and set the env variables `KUBECONFIG` and `KUBEVIRT_PROVIDER`.
+- In case only one PF exists, for example if running on prow which will assign only one PF per job in its own DinD,
+Kubevirtci is agnostic and nothing needs to be done, since all conditions above are met.
+The number of clusters that can be run on the same time equals number of PFs / number of PFs per cluster,
```
qinqon, line 70: If this is true then we have to change this https://github.com/kubevirt/kubevirtci/pull/551/files#diff-fcaa29a18b4077ecaa46dca6e69b637432306c3d2f38d3def35ff59ecc33dbbfR7 so we can have random number of registries right ?

qinqon (changes requested):
Looks good, just a small question on registries.

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
oshoval: Yes, but we now support only 2 at the same time, and its for local use only.
It doesn't affect CI, since CI runs in DinD, it can use the same port.
Its just a general readme note, i can add that we are limited to 2 locally.

oshoval:
Updated readme, to reflect that currently for local use, 2 are supported at most.

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
oshoval: Updated

qinqon (approved):

oshoval:
/retest

not related

PR #554, opened 2021-03-01
Make gocli work with podman3

rmohr:
gocli will work now with podman `3.1+`:

```
$ podman version
Version:      3.1.0-dev
API Version:  3.0.0
Go Version:   go1.14.6
Git Commit:   bde1d3f281b4d06b87320f3f001e67c79474a6a5
Built:        Fri Feb 26 10:38:59 2021
OS/Arch:      linux/amd64
```

In order to make this work, first stop or remove docker, then start podman:

```
podman system service -t 0 unix:///var/run/docker.sock
```

then run as usual:

```
make cluster-up && make cluster-sync
```

Most notable changes:

 * Podman does not fetch ports on `ContainerList`, therefore port infos are now fetched and extracted via `ContainerInspect`.
 * Podman does not like re-declaring already declared volumes. Cases where this appened were fixed.
 * Podman does not respect filters on `ContainerList`, therefore all containers are fetched and the filtering is done manually
 * Podman does not delete depending pods if the root pod is force deleted. Instead of that it fails. To overcome that, the code now ensures that the dnsmasq container is deleted last.


rmohr:
/cc @dhiller 

Looks intimidating, but almost all changes are only vendor updates.

cluster-provision/gocli/docker/docker.go
```diff
@@ -12,30 +12,40 @@ import (
...
-	return containers, err
+	return filterByPrefix(containers, prefix), nil
 }
 
-func GetPrefixedVolumes(cli *client.Client, prefix string) ([]*types.Volume, error) {
-	args, err := filters.ParseFlag("name="+prefix, filters.NewArgs())
-	if err != nil {
-		return nil, err
+func filterByPrefix(containers []types.Container, prefix string) []types.Container {
+	prefixedConatiners := []types.Container{}
```
dhiller, line 33: nit:spelling
```suggestion
	prefixedContainers := []types.Container{}
```

dhiller, line 33: nit:spelling
```suggestion
	prefixedContainers := []types.Container{}
```

cluster-provision/gocli/cmd/rm.go
```diff
@@ -37,13 +39,29 @@ func rm(cmd *cobra.Command, _ []string) error {
...
+				continue nodnsmasq
+			}
+		}
 		err := cli.ContainerRemove(context.Background(), c.ID, types.ContainerRemoveOptions{Force: true})
 		if err != nil {
 			return err
 		}
 	}
 
+	// delete dnsmasq at the end since other containers rely on ints network namespace
```
dhiller, line 57: nit:spelling
```suggestion
	// delete dnsmasq at the end since other containers rely on its network namespace
```

dhiller (approved):
/approve

Found only two nits regarding spelling, otherwise this looks fine.

PR #556, opened 2021-03-02
sriov vm image, Remove cloud-init user update module

oshoval:
Cloud-init has a module "users-groups" that configure users/groups according
to cloud-init config file.
    
While provisioning we are using the cloud-init config file in order to set
the user / password fedora:fedora (distro default).
    
When rerunning the provisioned VM, the cloud-init default configuration,
resets the users and cleans the password would run again,
In case any UserData exists.
This enforces reconfigurating the password via UserData and
cause flakiness to VM login.
    
There is no need to reset the user/passwords in subsequent VM spinning.
UserData may still be used in order to do user management commands
in case one would still need it.
Setting the user/password using the Cloud-init config is used only once during provision.
  
This commit removes the module, so subsequent Cloud-init triggering
would not reconfigure the default fedora user.

Kubevirt counterpart https://github.com/kubevirt/kubevirt/pull/5135

### DIFF
--- a/cluster-provision/images/vm-image-builder/fedora-sriov-lane/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-sriov-lane/cloud-config
@@ -23,4 +23,5 @@ runcmd:
   - sudo hostnamectl set-hostname ""
   - sudo hostnamectl set-hostname "" --transient
   - sudo systemctl start qemu-guest-agent
+  - sudo sed -i /users-groups/d /etc/cloud/cloud.cfg
   - sudo shutdown


RamLavi (approved):
/lgtm

EdDev (approved):

qinqon:
@oshoval can you paste some cloud-init doc for "users-groups" I cannot find it

oshoval:
> @oshoval can you paste some cloud-init doc for "users-groups" I cannot find it

https://cloudinit.readthedocs.io/en/latest/topics/modules.html#users-and-groups

qinqon:
> > @oshoval can you paste some cloud-init doc for "users-groups" I cannot find it
> 
> https://cloudinit.readthedocs.io/en/latest/topics/modules.html#users-and-groups

I still don't see the "users-groups" fields there, can you paste the content of cloud.cfg ?

oshoval:
You can see here that the default cloud.cfg ships with user fedora without password.
Removing the user groups module removes the updating of the users according this config
when rerunning the cloud-init (as we do in our e2e tests)

```
[fedora@testvmi1 ~]$ cat /etc/cloud/cloud.cfg
# The top level settings are used as module
# and system configuration.

# A set of users which may be applied and/or used by various modules
# when a 'default' entry is found it will reference the 'default_user'
# from the distro configuration specified below
users:
   - default

# If this is set, 'root' will not be able to ssh in and they
# will get a message to login instead as the default $user
disable_root: true

mount_default_fields: [~, ~, 'auto', 'defaults,nofail', '0', '2']
resize_rootfs_tmp: /dev
ssh_pwauth:   0

# This will cause the set+update hostname module to not operate (if true)
preserve_hostname: false

# Example datasource config
# datasource:
#    Ec2:
#      metadata_urls: [ 'blah.com' ]
#      timeout: 5 # (defaults to 50 seconds)
#      max_wait: 10 # (defaults to 120 seconds)

# The modules that run in the 'init' stage
cloud_init_modules:
 - migrator
 - seed_random
 - bootcmd
 - write-files
 - growpart
 - resizefs
 - disk_setup
 - mounts
 - set_hostname
 - update_hostname
 - update_etc_hosts
 - ca-certs
 - rsyslog
 - users-groups
 - ssh

# The modules that run in the 'config' stage
cloud_config_modules:
 - ssh-import-id
 - locale
 - set-passwords
 - spacewalk
 - yum-add-repo
 - ntp
 - timezone
 - disable-ec2-metadata
 - runcmd

# The modules that run in the 'final' stage
cloud_final_modules:
 - package-update-upgrade-install
 - puppet
 - chef
 - mcollective
 - salt-minion
 - rightscale_userdata
 - scripts-vendor
 - scripts-per-once
 - scripts-per-boot
 - scripts-per-instance
 - scripts-user
 - ssh-authkey-fingerprints
 - keys-to-console
 - phone-home
 - final-message
 - power-state-change

# System and/or distro specific settings
# (not accessible to handlers/transforms)
system_info:
   # This will affect which distro class gets used
   distro: fedora
   # Default user name + that default users groups (if added/used)
   default_user:
     name: fedora
     lock_passwd: True
     gecos: fedora Cloud User
     groups: [wheel, adm, systemd-journal]
     sudo: ["ALL=(ALL) NOPASSWD:ALL"]
     shell: /bin/bash
   # Other config here will be given to the distro class and/or path classes
   paths:
      cloud_dir: /var/lib/cloud/
      templates_dir: /etc/cloud/templates/
   ssh_svcname: sshd
```

oshoval:
Thanks all

@fgimenez 
do you know please who can build this new image and push it to quay instead the previous image?
quay.io/kubevirtci/fedora-sriov-testing (latest)

Thanks

fgimenez:
@oshoval sure I can take care of it, let's catch up online

PR #557, opened 2021-03-02
Add Kubevirt conformance tests to check-provision jobs

ormergi:
Currently its not straightforward or automated to test kubevirtci PR that involves re-provisioning an image 
with Kubevirt test suite.

This PR adds Kubevirt conformance tests to check-provision jobs, 
this will give us initial feedback whether a change cause regressions on kuebvirt/kuebvirt.


rmohr:
@ormergi if things don't work out in this run, consider for testing to first execute the kubevir conformance tests to get faster feedback.

ormergi:
Kubevirt conformance results
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/557/check-provision-k8s-1.19/1367183356285947904

cluster-provision/k8s/check-cluster-up.sh
```diff
@@ -42,7 +45,22 @@ export KUBEVIRTCI_GOCLI_CONTAINER=quay.io/kubevirtci/gocli:latest
...
+      ${ksh} apply -f "https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/kubevirt/${LATEST}/kubevirt-operator.yaml"
+      ${ksh} apply -f "https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/kubevirt/${LATEST}/kubevirt-cr.yaml"
+
+      ${ksh} wait -n kubevirt kv kubevirt --for condition=Available --timeout 15m
+
+      export SONOBUOY_EXTRA_ARGS="--plugin https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/kubevirt/${LATEST}/conformance.yaml"
+      hack/conformance.sh $conformance_config
+    fi
+
+    export SONOBUOY_EXTRA_ARGS="--plugin systemd-logs --plugin e2e"
```
rmohr, line 63: I think it makes sense to first run the k8s conformance tests before going to kubevirt. What do you think?

cluster-provision/k8s/check-cluster-up.sh
```diff
@@ -42,7 +45,22 @@ export KUBEVIRTCI_GOCLI_CONTAINER=quay.io/kubevirtci/gocli:latest
...
+
+    if [ "$RUN_KUBEVIRT_CONFORMANCE" == "true" ]; then
+      LATEST=$(curl -L "https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/kubevirt/latest")
+      ${ksh} apply -f "https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/kubevirt/${LATEST}/kubevirt-operator.yaml"
+      ${ksh} apply -f "https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/kubevirt/${LATEST}/kubevirt-cr.yaml"
+
+      ${ksh} wait -n kubevirt kv kubevirt --for condition=Available --timeout 15m
+
+      export SONOBUOY_EXTRA_ARGS="--plugin https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/kubevirt/${LATEST}/conformance.yaml"
+      hack/conformance.sh $conformance_config
```
rmohr, line 60: Would it makes ense to only add `--plugin https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/kubevirt/${LATEST}/conformance.yaml` to `SONOBUOY_EXTRA_ARGS` and basically running below then:

```
--plugin systemd-logs --plugin e2e --plugin https://<url>/conformance.yaml
```

cluster-provision/k8s/check-cluster-up.sh
ormergi: I thought running kubevirt conformance tests first is better because it take less time 
then e2e so we could fail faster.
What you think?

rmohr: Fine by me.

cluster-provision/k8s/check-cluster-up.sh
rmohr: Let's ensure that we run sonobuoy only once, since we also check the results only once. Otherwise you don't benefit from the early exit which you want after the kubevirt tests. The rest looks very good.

hack/conformance.sh
```diff
@@ -17,13 +17,27 @@ export KUBECONFIG
...
+
+    # Get each plugin junit report rename from 'junit.xml' to 'junit.<plugin-name>.<file number>.xml',
+    # and move to artifacts directory
+    for plugin in $(./sonobuoy status --json | jq -r '[.plugins[]] | unique_by(.plugin) | .[].plugin'); do
+      tar -xvzf "$results_tarball" "plugins/$plugin/"
+      plugin_artifacts_dir="${ARTIFACTS}/$plugin"
+      mkdir -p "$plugin_artifacts_dir"
+      idx=1
+      for report in $(find "plugins/$plugin/"* -name "*.xml"); do
+        plugin_report="${report/.xml/.${plugin}.${idx}.xml}"
```
rmohr, line 36: ok, a little bit more to do: the junit files need to be palced and named liek this:  `${ARTIFACTS}/junit.${something}.xml` to be recognized. Prow will merge all junit files together which match this pattern.

cluster-provision/k8s/check-cluster-up.sh
ormergi: My intention was run kubevirt conformance on a different "sonobuoy session"
where kubevirt conformance runs first (currently it takes less time).
so a job could fail faster in case of an error and enable us to control it with `RUN_KUBEVIRT_CONFORMANCE`

rmohr: I am not sure I fully understand this. Let me try to summarize like this what I would expect:
 * It is fine if you want/can somehow influence with env vars what to run and to fail fast, just ensure that
 * by default all three plugins are run in CI
 * that a failing kubevirt conformance run really leads to exiting the script early with a non-nil exit code (I am not out of the box sure if sonobuoy returns with a non-0 return code and that we exit the script). To really fail fast, like you suggest.

If that is the behaviour which we get, all is good from my perspective.

hack/conformance.sh
ormergi: On line 36 the junit file is renamed from `junit.xml` to `junit.<plugin name>.<file number>.xml`
I verivied it could be picked by spyglass according to [this pattern](https://github.com/kubevirt/project-infra/blob/master/github/ci/prow-deploy/kustom/base/manifests/test_infra/1736ab42da959a5be7de22586b495120000bc4ef/starter-s3.yaml#L78-L80)

I just noticed there is an issue with moving it to the right directory though, 
I will fix it.

cluster-provision/k8s/check-cluster-up.sh
ormergi: It should cover what you expect: 

> * It is fine if you want/can somehow influence with env vars what to run and to fail fast, just ensure that

Control running kubevirt-conformance with `RUN_KUBEVIRT_CONFORMANCE` if its set to false,
only the standard test will run ( e2e and systemd-logs)

> * by default all three plugins are run in CI (if previous plugins don't fail, I don't mind early exit)

This PR configures all tree to run as we set `RUN_KUBEVIRT_CONFORMANCE` to be true by default on the last commit.

> * that a failing kubevirt conformance run really leads to exiting the script early with a non-nil exit code (I am not out of the box sure if sonobuoy returns with a non-0 return code and that we exit the script). To really fail fast, like you suggest.

As we mentioned on other thread the caller script, check-cluster-up.sh, will runs conformance.sh twice:
first for kubevirt-conformance tests and the second for the standard tests (e2e, systemd-logs).

On `conformance.sh `script we use trap on EXIT and calling to `teardown`.
`teardown` check the tests status with `sonobouy status`, 
in case there were failing tests  `conformance.sh ` exit with error code 1.

The error will propagate to `check-cluster-up.sh` script, which runs with `-eo pipefail` flags,
propagating the erorr forwarder.

If we will find that running it this way is not best suite for us
we can always run all the tests in one shot (by setting all plugin to the same sonobuoy run)

ormergi:
/hold

Hold this PR until:
- CI is stabilized
- Decide which tests we should run on kubevirt-conformance tests
- Resources concerns
- Increase jobs timeouts

ormergi:
/hold

Hold this PR until:
- CI is stabilized
- Decide which tests we should run on kubevirt-conformance tests
- Resources concerns
- Increase jobs timeouts

rmohr:
> * CI is stabilized
> * Decide which tests should run on kubevirt-conformance tests (currently it runs 8 tests of 500+ tests)

Let's start with these 8, as a next step let's add individual tests, based on what we expect to have configured in the cluster. Here a possible list:
 * 1 test: check ipv6 is enabled and working for vmis
 * 1 test: check PVCs work with vmis
 * 1 test: check cpu manager works on node02

> * Ack on resources consumption this change will bring

I don't think that this increases the usage much. I don't see a blocker there.

> * Increase jobs timeouts

I see that on a healthy CI system, they run well below 2 hours (right?). So a timeout of 3 hours seems good enough to me. What do you think?

rmohr (approved):
/lgtm
/approve

Keeping the hold until we have a little bit more data regarding to weather we should increase the ci timeout. Also waiting for a few more green lanes to be sure to make things not worse.

PR #565, opened 2021-03-10
Use default values for node cidr mask size

qinqon:
The node-cidr-mask-size for ipv4 and ipv6 were being enforced and
overlaping with cluster-cidr. This change remove
node-cidr-mask-size-ipv4 to use the defautl value (24) and enforce
node-cidr-mask-size-ipv6 to 116 since default value (64) is too big.

Also the ipv6 calico ippool has being removed since we can enforce ipv6
nat with an environment variable CALICO_IPV6POOL_NAT_OUTGOING at the
calico-node for calico >= v3.16.0, the default ipv6 CIDR
is calculated from kubeadm.conf podSubnet.

After this fix the podCIDR appears at both nodes

```
It show two nodes with different podCIDR

[ellorent@localhost kubevirtci]$ ./cluster-up/kubectl.sh get node -o yaml |grep -i cidr
          f:podCIDR: {}
          f:podCIDRs:
    podCIDR: 10.244.0.0/24
    podCIDRs:
          f:podCIDR: {}
          f:podCIDRs:
    podCIDR: 10.244.1.0/24
    podCIDRs:
```

We also found after this fix that calico is ignoring node.spec.podCIDR and it's a expected behaviour https://github.com/projectcalico/calico/issues/2592#issuecomment-489233638.

Should fix https://github.com/kubevirt/kubevirt/issues/4961


cluster-provision/k8s/1.20/manifests/cni.diff
```diff
@@ -46,20 +49,3 @@ index bcc4ee0c..17a045ff 100644
        containers:
          - name: calico-kube-controllers
            image: docker.io/calico/kube-controllers:v3.18.0
-@@ -3869,4 +3881,16 @@ spec:
- ---
- # Source: calico/templates/configure-canal.yaml
- 
-+---
-+apiVersion: crd.projectcalico.org/v1
```
phoracek, line 54: Does Calico provide IPv6 by default now?

ormergi:
Each node should have different pod subnet, isnt it?

cluster-provision/k8s/1.20/manifests/cni.diff
qinqon: The pool was there only because we need to configure ipv6 natting, but new calico version has a new enviromnet `CALICO_IPV6POOL_NAT_OUTGOING` variable that activates ipv6 nat without creating a pool, also we are still activating IPv6 with `"assign_ipv6": "true"`

qinqon:
> Each node should have different pod subnet, isnt it?

Nope, podSubnet is cluster wise per node you configure the size of the CIDR but kubernetes choose the cidr

qinqon:
sig-network tests are passing with this change locally, I am testing other network tets to see how it rolls.

qinqon:
Also `-focus Networking` test are passing too.

EdDev (commented):
Looks very nice, thanks!

A few comments for the commit message:
- Please specify the new ranges for ipv4 and ipv6 and mention which one is a default.
- Per your explanation `CALICO_IPV6POOL_NAT_OUTGOING` is a new option from Calico, perhaps it is worth mentioning what minimum version is required.

EdDev (approved):
Thank you! Nice one.

ormergi (approved):
Thanks!

phoracek:
/approve
/hold

Feel free to cancel any time.

qinqon:
/hold cancel

Just one test failed 

https://prow.apps.ovirt.org/log?id=1370022818195968000&job=check-kubevirt-e2e-k8s-1.18

```
16:24:21: Summarizing 1 Failure:
16:24:21: 
16:24:21: [Fail] Configurations VirtualMachineInstance definition with BIOS bootloader method and no disk [It] [test_id:5266]should boot to NIC rom if a boot order was set on a network interface 
16:24:21: tests/vmi_configuration_test.go:519
16:24:21: 
16:24:21: Ran 325 of 750 Specs in 3030.891 seconds
16:24:21: FAIL! -- 324 Passed | 1 Failed | 3 Pending | 422 Skipped
```

PR #566, opened 2021-03-11
Make cluster-up/check.sh fail fast

dhiller:
Except for providers external or local we make the check script fail
fast. If the kvm device is not there or nested virtualization is not enabled
there is no chance the cluster will start up. In that case we make the
script exit with non zero error code to fail early.

/cc @qinqon @fgimenez @oshoval

### DIFF
--- a/cluster-up/check.sh
+++ b/cluster-up/check.sh
@@ -18,34 +18,46 @@
 #
 
 set -e
+fail=''
 if [ ! -c /dev/kvm ]; then
-	echo "[ERR ] missing /dev/kvm"
+    echo "[ERR ] missing /dev/kvm"
+    fail=1
 else
-	echo "[ OK ] found /dev/kvm"
+    echo "[ OK ] found /dev/kvm"
 fi
 
 KVM_ARCH=""
 KVM_NESTED="unknown"
 if [ -f "/sys/module/kvm_intel/parameters/nested" ]; then
-	KVM_NESTED=$( cat /sys/module/kvm_intel/parameters/nested )
-	KVM_ARCH="intel"
+    KVM_NESTED=$(cat /sys/module/kvm_intel/parameters/nested)
+    KVM_ARCH="intel"
 elif [ -f "/sys/module/kvm_amd/parameters/nested" ]; then
-	KVM_NESTED=$( cat /sys/module/kvm_amd/parameters/nested )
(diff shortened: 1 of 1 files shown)


cluster-up/check.sh
```diff
@@ -18,34 +18,46 @@
...
+    echo "[ OK ] $KVM_ARCH nested virtualization enabled"
 else
-	echo "[ERR ] $KVM_ARCH nested virtualization not enabled"
+    echo "[ERR ] $KVM_ARCH nested virtualization not enabled"
+    fail=1
+fi
+
+if [[ "$KUBEVIRT_PROVIDER" =~ external|local ]]; then
+    echo "nested virtualization not required"
+elif [ -z "$fail" ]; then
```
oshoval, line 58: do we have a non external | local that use Emulation (kubevirt configmap has it IIRC ?) so it doesn't need kvm ?

cluster-up/check.sh
```diff
@@ -18,34 +18,46 @@
 #
 
 set -e
+fail=''
```
oshoval, line 21: might be nicer to set 0 instead of empty string ?
and of course change the condition accordingly if so

fgimenez (commented):
/lgtm

cluster-up/check.sh
dhiller: Sorry, I don't understand your question?

oshoval: https://github.com/kubevirt/kubevirt/blob/ed8da106d0eb177b0d468a6b8db4a9b6de38115f/docs/software-emulation.md
in case useEmulation is enabled, `/dev/kvm` is allowed to not exists even on the non `external|local` providers ?

qinqon (changes requested):
Please remove the identing changes first

dhiller:
> Please remove the identing changes first

I applied `shfmt` to the file as it had some awkward indentation (5 spaces?!). What else should I use in your opinion?

dhiller:
I'm going to put this back to draft state, it's not urgent and we should iron out when it should really fail fast. IMHO it should only fail if the /dev/kvm device is missing, and even then only if it's not the `external` or `local` provider?

Opinions?

dhiller:
/close

this really does not help, I need to think about this, sorry for the noise.

PR #567, opened 2021-03-11
Configure cgroupv2 in k8s 1.20

vasiliy-ul:
This PR is an initial attempt to enable a new lane configured with cgroup v2. It is based on the existing k8s 1.20. The required changes are applied depending on the `UNIFIED_CGROUP_HIERARCHY` env var which is propagated to the provisioning script by the `--cgroupv2` gocli arg. That way the initial cluster provision (with cgroup v1) can still be used. Calling `CGROUPV2=true ../provision.sh` in turn triggers cgroup v2 configuration.

A dedicated lane is needed to enable cgroup v2 support in Kubevirt and to test end-to-end the changes in https://github.com/kubevirt/kubevirt/pull/4907

### DIFF
--- a/cluster-provision/gocli/cmd/provision.go
+++ b/cluster-provision/gocli/cmd/provision.go
@@ -41,6 +41,7 @@ func NewProvisionCommand() *cobra.Command {
 	provision.Flags().Uint("vnc-port", 0, "port on localhost for vnc")
 	provision.Flags().Uint("ssh-port", 0, "port on localhost for ssh server")
 	provision.Flags().Bool("cgroupv2", false, "set UNIFIED_CGROUP_HIERARCHY environment variable for the provision script")
+	provision.Flags().String("container-suffix", "", "use additional suffix for the provisioned container image")
 
 	return provision
 }
@@ -58,8 +59,16 @@ func provisionCluster(cmd *cobra.Command, args []string) (retErr error) {
 	}
 	base := fmt.Sprintf("quay.io/kubevirtci/%s", strings.TrimSpace(string(baseBytes)))
 
-	prefix := fmt.Sprintf("k8s-%s-provision", filepath.Base(packagePath))
-	target := fmt.Sprintf("quay.io/kubevirtci/k8s-%s", filepath.Base(packagePath))
+	containerSuffix, err := cmd.Flags().GetString("container-suffix")
+	if err != nil {
+		return err
+	}
+	name := filepath.Base(packagePath)
+	if len(containerSuffix) > 0 {
+		name = fmt.Sprintf("%s-%s", name, containerSuffix)
+	}
+	prefix := fmt.Sprintf("k8s-%s-provision", name)
+	target := fmt.Sprintf("quay.io/kubevirtci/k8s-%s", name)
 	scripts := filepath.Join(packagePath)
 
 	memory, err := cmd.Flags().GetString("memory")
--- a/cluster-provision/gocli/cmd/provision.go
+++ b/cluster-provision/gocli/cmd/provision.go
@@ -40,6 +40,7 @@ func NewProvisionCommand() *cobra.Command {
 	provision.Flags().Bool("random-ports", false, "expose all ports on random localhost ports")
 	provision.Flags().Uint("vnc-port", 0, "port on localhost for vnc")
 	provision.Flags().Uint("ssh-port", 0, "port on localhost for ssh server")
+	provision.Flags().Bool("cgroupv2", false, "set UNIFIED_CGROUP_HIERARCHY environment variable for the provision script")
 
 	return provision
 }
@@ -220,7 +221,17 @@ func provisionCluster(cmd *cobra.Command, args []string) (retErr error) {
 		return err
 	}
 
-	err = _cmd(cli, nodeContainer(prefix, nodeName), fmt.Sprintf("ssh.sh sudo version=%s /bin/bash < /scripts/provision.sh", version), "provisioning the node")
+	cgroupv2, err := cmd.Flags().GetBool("cgroupv2")
+	if err != nil {
+		return err
+	}
+
+	envVars := fmt.Sprintf("version=%s", version)
+	if cgroupv2 {
+		envVars = fmt.Sprintf("%s UNIFIED_CGROUP_HIERARCHY=1", envVars)
+	}
+
+	err = _cmd(cli, nodeContainer(prefix, nodeName), fmt.Sprintf("ssh.sh sudo %s /bin/bash < /scripts/provision.sh", envVars), "provisioning the node")
 	if err != nil {
 		return err
 	}
--- a/cluster-provision/k8s/1.20/provision.sh
+++ b/cluster-provision/k8s/1.20/provision.sh
@@ -2,12 +2,20 @@
 
 set -ex
 
+# Configure cgroup version
+if [ "${UNIFIED_CGROUP_HIERARCHY}" == "1" ]; then
+    CMDLINE_LINUX_APPEND="${CMDLINE_LINUX_APPEND} systemd.unified_cgroup_hierarchy=1"
+    CGROUP_DRIVER="cgroupfs"
+else
+    CGROUP_DRIVER="systemd"
+fi
+
 KUBEVIRTCI_SHARED_DIR=/var/lib/kubevirtci
 mkdir -p $KUBEVIRTCI_SHARED_DIR
 cat << EOF > $KUBEVIRTCI_SHARED_DIR/shared_vars.sh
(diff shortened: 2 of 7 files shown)


cluster-provision/gocli/cmd/provision.go
```diff
@@ -40,6 +40,7 @@ func NewProvisionCommand() *cobra.Command {
 	provision.Flags().Bool("random-ports", false, "expose all ports on random localhost ports")
 	provision.Flags().Uint("vnc-port", 0, "port on localhost for vnc")
 	provision.Flags().Uint("ssh-port", 0, "port on localhost for ssh server")
+	provision.Flags().Bool("use-env", false, "use additional environment file during provisioning")
```
rmohr, line 43: Would it be possible to move the cgroup decition to `cmd/run.go`? So that it is a cluster-start decision vs. a cluster-provison decision?

rmohr: Would it make sense to add a `--unified-cgroups`  argument to the `run` command? I know that at the end we basically juggle just with env vars and copy in hacky ways script around, but it may provide a nice surface toggle for users which don't know how kubevirtci looks internally.

rmohr (commented):
Very nice. A few suggestions inline on how to make it possibly easier to consume.

cluster-provision/gocli/cmd/provision.go
vasiliy-ul: I was thinking about `cmd/run.go` but the problem is that in order to enable cgroup v2 I need to pass `systemd.unified_cgroup_hierarchy=1` to the kernel. That in turn requires regeneration of grub gonfig and machine reboot. While looking at the overall flow I got impression that `cmd/run.go` should not change persistent configuration of the node. Or have I understood it wrong?

rmohr: That is a good point. It may require us to provide a kernel and initrd separately, which definitely goes beyond this pr.

rmohr: I would just somehow prefer a way to not have to provision and ship the images two times due to kernel parameter changes.

vasiliy-ul: Valid point. However apart from the kernel parameter there are also some configs which are different in case of using cgroup v2. There is a need to change the cgroup driver for `kubelet` and `crio` from `systemd` to `cgroupfs`. Unfortunately systemd in the base centos8 image does not fully support cgroup v2 (the minimum required version is `244` while in the image its `239`). Therefore kubelet and crio should not rely on it and hence changing of the driver is needed. So I assume that even if the kernel parameter is somehow handled during `run` the configs still need to be updated (and that again requires changing of the node persistant configuration).

rmohr: I think it makes sense to debate what "persistent configuration" means. The most important bit is that all external dependencies are baked into the image, which makes setting up a cluster with kubeadm very snappy, predictable and extremely reliable.
I do not care too much if additional feature gates are added during "run", or if services get changed. As you have probably seen we actively reset the cluster after we have everything in place, to get a completely fresh one on startup.
The critical thing for me would be that the `provision` jobs would have to run the conformance tests against two configurations.

I can see that it is hard to achive for that feature here. I guess we will have to duplicate the lane then (extra provision and publishing with this new env var), since I don't see us in a position at this very moment to just go blindly for cgroupsv2 for 1.20 at this stage.

vasiliy-ul: Just in theory... shouldnt the conformance tests be executed against all the possible cgroups/k8s combinations in any way (ideally)? Previously with only cgroup v1 there was a single variable in the equation which was k8s while the second (the base system) was a constant. Now the system can be in two states (cgroup v1 or v2) therefore both should be covered, no?

rmohr: > Just in theory... shouldnt the conformance tests be executed against all the possible cgroups/k8s combinations in any way (ideally)? Previously with only cgroup v1 there was a single variable in the equation which was k8s while the second (the base system) was a constant. Now the system can be in two states (cgroup v1 or v2) therefore both should be covered, no?

Definitely. That is what I meant what conformance tests have to do. All I meant is, that if we can only decide on the variant during provision, we have to ship two huge images.

The same applies for container runtimes in principle. Still, we did right now always a container runtime switch per lane. cgroupv2 differs a little there, since kubevirt should by definition work with any container runtime, without extra work. That is not the case with cgroupv2, because it creates for kubevirt visible differences.

vasiliy-ul: I think one more option might to generate two grub menuentries (one for cgroup v1 another - for v2) and to select the appropriate one via qemu serial line (I guess grub2 supports that). For that will need to implement serial communication in gocli. The actuall boot choice can be set via gocli cmdline argument. Then some magic scripts can be enabled to run on boot in the VM to detect the cgroup version and start kubelet and crio with proper configs... Well, a bit tricky but sounds doable, just need to spend some more time on it. But that way the second image will not be needed.

rmohr: I like such solutions :)

Maybe we can still right now go for both images. Let's just do the cvgroupv2 testing in a periodic job and not on PRs. I would support this.

rmohr: So let's go like this from here: We will need a dedicated `check-provision-k8s-1.20-cgroupsv2` job in `project-infra`. It will run tests and publish the provider.

We can then use this provider in an optional presubmit and in a periodic job to test kubevirt-ci. I think @davidvossel is also fine with this. For as long as cgroupsv1 is then not broken, we can bring in your cgroupsv2 PR in kubevirt/kubevirt as "cgroupsv2" preparations (hoping that the test results do not look that bad already). Then it can be iterated on until the optional presubmit and the periodic are passing. At this point we can then anounce cgroupv2 support.

Thoughts?

vasiliy-ul: Sounds like a nice plan :) I think I will need to add the new provider to `cluster-up` within this PR. Then I can open a PR in `project-infra` with the new job. Is it important in which order they get merged? It seems that `check-provision-*` jobs are used here for testing and that makes them kind of depend on each other.

I do not expect that the cgroupv2 PR in kubevirt will enable all the features. As mentioned there at least the hotplug disks topic should be addressed separately. Only after enabling the tests we will see what else might need attention.

vasiliy-ul: hm.. also I think I need to an a cmdline argument to gocli to overwrite resulting container name... otherwise it will conflict wit k8s-1.20

vasiliy-ul: So, I've created the check provsion job https://github.com/kubevirt/project-infra/pull/1036 . Is there still smth missing here? How the PRs shall be merged (in which order)?

rmohr: I am in principle ok with what this does now. Would you mind replacing the env  variables with a commandline option on the `provision` command (not `run` like discussed bevore)? Something like `--cgroupv2`? The env var pass-through is very hard to follow.

rmohr (changes requested):

vasiliy-ul:
> I am in principle ok with what this does now. Would you mind replacing the env variables with a commandline option on the provision command (not run like discussed bevore)? Something like --cgroupv2? The env var pass-through is very hard to follow.

@rmohr , probably you are right, the approach with env file is not very straightforward. Replaced with `--cgroupv2`. Please take a look. If its fine I will update the check provision job.

rmohr:
> @rmohr , probably you are right, the approach with env file is not very straightforward. Replaced with `--cgroupv2`. Please take a look. If its fine I will update the check provision job.

The approach looks good :+1:

rmohr (commented):
One general note, the publish.sh script in the root needs to be modified to also build and push the cgroupv2 provider. It will right now still only push the 1.20 provider.

vasiliy-ul:
> One general note, the publish.sh script in the root needs to be modified to also build and push the cgroupv2 provider. It will right now still only push the 1.20 provider.

That is smth I missed. Will take a look...

vasiliy-ul:
> Could not detect slice of whitelisted controller: [devices]

So now `check-provision-k8s-1.20-cgroupsv2` job fails because it tests master which does not have cgroupv2 support. I assume we need one more presubmit job for kubevirt to be able to trigget it in that PR, right?

rmohr:
/lgtm
/approve

I saw on the conformance tests for cgroupsv2, that the k8s conformance tests were passing and (like expected) the kubevirt conformance tests failed.

PR #563, opened 2021-03-10
SRIOV Provider use bare sriov components instead of sriov-network-operator

ormergi:
Currently SRIOV provider we use sriov-network-operator to deploy sriov-cni and sriov-device-plugin
and configure PF's and VF's.
sriov-network-operator was meant to run on full flagged Openshift with bare-metal nodes that has
SRIOV hardware, and not on containerized Kubernetes (KIND) setup.

We need to add support to SRIOV provider to enable multiple SRIOV capable worker nodes setup
for Kubevirt SRIOV Live Migration feature.
At the moment we use version 4.4 where sriov-device-plugin is not compatible running on
containerized environments, causing VM's creation to fail due to VF allocation from other nodes.

This PR simplify the sriov provider by removing sriov-network-operator,
Instead it will configure PFs and VFs statically and deploy multus sriov-cni, sriov-device-plugin.

Supporting multiple SRIOV capable worker nodes will be introduces in a following PR.

PR #553, opened 2021-03-01
fix make cluster-up error when using kind-k8s-1.17

zhlhahaha:
set default value for KIND_VERSION, KIND_NODE_IMAGE and
KUBECTL_PATH in kind-k8s-1.17/provider.sh

PR #570, opened 2021-03-16
Add support for SRIOV Live Migration on SRIOV provider

ormergi:
In order to test SRIOV Live Migration feature on CI it is necessary to create a cluster with two worker nodes.

This PR enables creating such such setup with the SRIOV provider.
Each worker node will have SRIOV PF attached to its netns with all necessary configuration in place.

Also re-introduce creating fake product_uuid PodPreset for virt-launcher pods,
in order to enable Libvirt live migration on containerized clusters (KIND) following #336 example.


cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -25,79 +22,127 @@ SRIOVDP_RESOURCE_NAME="sriov_net"
...
+[ "$pf_count" -lt $total_pf_required ] && \
+  echo "FATAL: there are not enough PF's on the host, try to reduce NODES_PFS_PER_NODE
+  Worker nodes count: $nodes_count
+  PF per node count:  $NODES_PFS_PER_NODE
+  Total PF count required:  $total_pf_required" >2 && return 1
 
-# KIND mounts sysfs as readonly, this script requires R/W access to sysfs
-${SRIOV_NODE_CMD} mount -o remount,rw /sys
-${SRIOV_NODE_CMD} chmod 666 /dev/vfio/vfio
+## Move SRIOV Physical Functions to worker nodes and configure VF's
```
EdDev, line 103: Can this be a function?

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
```diff
@@ -4,6 +4,8 @@ set -e
 
 export CLUSTER_NAME="sriov"
 
+SRIOV_TESTS_NS="${SRIOV_TESTS_NS:-kubevirt-test-default1}"
```
EdDev, line 7: I think you need to explain what `ubevirt-test-default1` is and from where it got here.
There is a dependency on the test code which is bad, we need to figure out how to solve it (perhaps pass in the name to the tests).

EdDev (changes requested):
Nice stuff, thanks!

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
oshoval: We solved it in the past by creating the pod preset in kubevirt, not in e2e tests, there is a PR that does it

cluster-up/cluster/kind/manifests/product-uuid-podpreset.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: settings.k8s.io/v1alpha1
+kind: PodPreset
```
oshoval, line 2: should be on e2e tests, as we done it the past, 
see please
https://github.com/kubevirt/kubevirtci/pull/570/commits/7a925f90579a3dd9d9e13078e8a10c461f267486#r596837683

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
oshoval: https://github.com/kubevirt/kubevirt/pull/3374

EdDev:
@oshoval commenting here as your last comments referred to the same thing in multiple places:

I guess that the NS dependency exposed is to move the pod preset to the tests, however I think it is worse to have knowledge in the test about what tricks needs to be done to support a specific operation on a specific cluster.
The tests should be agnostic to the cluster they run on as much as possible. This is especially relevant when such tests may be ran on a cluster that is not kind (e.g. bare-metal).

If there other reasons to run this hacking in the tests themselves, please share it with us so we can make sure we are still good.

cluster-up/cluster/kind/manifests/product-uuid-podpreset.yaml
EdDev: I continued this [here](https://github.com/kubevirt/kubevirtci/pull/570#issuecomment-801952386).

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
EdDev: I continued this [here](https://github.com/kubevirt/kubevirtci/pull/570#issuecomment-801952386).

EdDev:
@ormergi , the solution with the preset is valid until k8s 1.19, but is deprecated in 1.20. We will need to change it to a webhook.
Perhaps we should comment that in the commit and ref to an issue. WDY?
We have 2 k8s version to make this change.

cluster-up/cluster/kind/manifests/product-uuid-podpreset.yaml
oshoval: Lets continue it here please instead of in the root,
kubevirtci should be the one that is agnostic of the e2e tests,
This is what @AlonaKaplan and myself concluded when we worked on migration
and the https://github.com/kubevirt/kubevirt/pull/3374 is a consequence of it.
I find it (mentioned PR above) much better than to have the opposite dependency (kubevirtci to know the NS of the tests),
for example if the test change the NS, or even doesn't want to create a PodPreset,
this way the kubevirtci comes without PodPreset but with support for it, and the user might create it upon need, for his NS.

EdDev: Thank for your input @oshoval.

I think the reasoning you raised is valid but solvable by keeping the tests unaware of this.
You argued in favor of the previous approach but did not feedback to its problems, which seem to be no less important than the one you are raising.

How about this: You make a specific NS migration-ready, and if the relevant tests find this NS, they will use it. Otherwise, they will use the "default" one or just skip the migration tests. If we choose to skip, then we probably need to allow passing the NS name that supports migration through a flag-argument.

oshoval: Honestly i find it over engineering, and prefer that the kubevirtci will be the agnostic product (as expressed before),
i don't see any reason that it will pass stuff in a resolution such as NS.
I less like "scanning" and even less message passing between kubevirtci and kubevirt.

I think that its purely natual that the tests will create a PodPreset in case needed, i just see it otherwise than you,
i don't find it "know what tricks are needed to be done", but its matter of taste i guess.
The tests know that they actually want migration, and for that they create the test-wise layer (PodPreset),
where the kubevirtci does the platform wise prepartion.

This is at least my opinion.

EdDev: What exactly is over-engineered here?
I provided you with an example where this will not makes sense when running the cluster against an external bare-metal node. This "trick" is done only because *you* know that the cluster is kind, but in reality it may not be kind.

The suggestion from my part is to prepare on a *kind* setup a cluster that will support unique UID that allows libvirt to perform migration. It is not dependent on the tests, it just prepares it for anyone that may want to use it.

oshoval: The mentioned PR solved this as well, by creating the PodPreset only if the provider is kind
this also solve the bare metal example i think ?

Up to you, this is my opinion, i prefer agnostic separation, and feels like tprefer a specific NS that is the only one that is migration ready, is too much resolution out of the scope of kubevirtci, and btw also a little limitation,
for example cnv-tests run each tests group in its own namespace,
lets say they would use kubevirtci, and have 2 groups ? what then ?
just an example

ormergi: Its not longer possible to create the PodPrest the way presented on https://github.com/kubevirt/kubevirt/pull/3374
kubevirt uses k8s.io/api v0.20.2 package where we no longer have PodPrest type.

I think we can leave with this for now as it will be valid until the oldest version we would need to support is k8s-1.20

ormergi:
> @ormergi , the solution with the preset is valid until k8s 1.19, but is deprecated in 1.20. We will need to change it to a webhook.
> Perhaps we should comment that in the commit and ref to an issue. WDY?

Yes totally I created an issue to track this https://github.com/kubevirt/kubevirtci/issues/571

ormergi:
/hold

Should be merged after #563

ormergi:
#563 is merged
/unhold

ormergi:
rebased

ormergi:
Removed unused variable `pfs_in_use`

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -25,79 +22,127 @@ SRIOVDP_RESOURCE_NAME="sriov_net"
...
-fi
+worker_nodes=($(_kubectl get nodes -l node-role.kubernetes.io/worker -o custom-columns=:.metadata.name --no-headers))
+nodes_count=${#worker_nodes[@]}
+[ "$nodes_count" -eq 0 ] && echo "FATAL: no worker nodes found" >&2 && exit 1
 
-NODE_PFS=($(move_sriov_pfs_netns_to_node "$SRIOV_NODE" "$PF_COUNT_PER_NODE"))
-NODE_PF=$NODE_PFS
+pfs_names=( $(get_pfs_names) )
+pf_count="${#pfs_names[@]}"
+[ "$pf_count" -le 0 ] && echo "FATAL: Could not find available sriov PF's" >&2 && exit 1
```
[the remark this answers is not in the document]
ormergi: Changed it to `-eq`

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -25,79 +22,127 @@ SRIOVDP_RESOURCE_NAME="sriov_net"
...
 
-SRIOV_NODE_CMD="docker exec ${SRIOV_NODE}"
-NODE_PF_NUM_VFS=$(${SRIOV_NODE_CMD} cat /sys/class/net/$NODE_PF/device/sriov_totalvfs)
+total_pf_required=$((nodes_count*NODES_PFS_PER_NODE))
+[ "$pf_count" -lt $total_pf_required ] && \
+  echo "FATAL: there are not enough PF's on the host, try to reduce NODES_PFS_PER_NODE
+  Worker nodes count: $nodes_count
+  PF per node count:  $NODES_PFS_PER_NODE
+  Total PF count required:  $total_pf_required" >2 && return 1
```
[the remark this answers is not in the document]
ormergi: my bad thanks for the heads up, canged it to `exit 1`

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: I tried few things and yes its possible, I will change it

ormergi:
Rebased

cluster-up/cluster/kind/manifests/product-uuid-podpreset.yaml
qinqon: We don't want to implement this with an admission webhook so we are prepared for >= 1.20 right ?

ormergi: We do, it is planned, for the meantime we want to use PodPrest workaround at lease so we could test SRIOV migration on CI

qinqon:
/lgtm
/approve

We will implement an admission webhook in the future if needed.

EdDev:
I'm in the middle of the review, holding until I finish.

/hold

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -25,85 +22,134 @@ SRIOVDP_RESOURCE_NAME="sriov_net"
...
+  echo "${pfs_names[@]}"
+}
 
-  echo "${pf_array[@]}"
+function configure_nodes_sriov_pfs_and_vfs() {
+  local -r nodes_array=($1)
+  local -r pfs_names_array=($2)
+  local -r pf_count_per_node=$3
+
+  mkdir -p /var/run/netns
```
EdDev: This is surprising to have it here. Is this intentional?

EdDev (changes requested):
Review 1st commit: SRIOV provider: Add support for multiple nodes

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
```diff
@@ -0,0 +1,96 @@
...
+  local -r node_pid=$(docker inspect -f '{{.State.Pid}}' "$node_name")
+  local -r pfs_sriov_numvfs=( $(cat /proc/$node_pid/root/sys/class/net/*/device/sriov_numvfs) )
+  local total_vfs_on_node=0
+
+  for num_vfs in "${pfs_sriov_numvfs[@]}"; do
+    total_vfs_on_node=$((total_vfs_on_node + num_vfs))
+  done
+
+  echo "$total_vfs_on_node"
+}
```
EdDev, line 96: Something is wrong here, missing newline?

EdDev (changes requested):
Review: SRIOV provider, config_sriov.sh: Export nodes configurations code 

Filenames:
- I think you can call the now `sriov-node` folder just `node`.
- I do not understand what `sriov_node.sh` has compared to what `configure_vfs.sh` has. The name is too odd now and looking at the content I do not understand what each represent.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Yes, you prefer this to be outside this function?

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
EdDev: @ormergi was this missed? I do not see an explanation from where this name came from. It is hard to understand without a small explanation.

cluster-up/cluster/kind/migration.sh
```diff
@@ -0,0 +1,38 @@
...
+    local -r namespace=$1
+
+    if ! _kubectl get ns "$namespace" &>2; then
+        _kubectl create ns "$namespace"
+    fi
+
+    _kubectl apply -f "$KIND_MANIFESTS_DIR/product-uuid-podpreset.yaml" -n "$namespace"
+}
+
+function kind_migration::enable_live_migration() {
```
EdDev, line 26: I would have called it per what it does, not per what it serves for.
Something like: "expose_unique_product_uuid_per_node".

How it is done is abstracted inside.
Why it is done, is explained at the caller that wants to use this functionality.

This also means this script name should change to "podpreset.sh".

EdDev (changes requested):
Review: SRIOV provider: Add support for live migration 

I saw this while reviewing this specific commit, but I guess it is a general question: Why we need the `kind` folder? We use it only under the sriov lane, and it is confusing to have this in a different place. If there will be another kind lane later on, we can think how to handle it then.

WDYT?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
EdDev: I do not know, I think I prefer it close to the place it is actually needed. Here I do not understand where it is needed.

oshoval:
> I saw this while reviewing this specific commit, but I guess it is a general question: Why we need the `kind` folder? We use it only under the sriov lane, and it is confusing to have this in a different place. If there will be another kind lane later on, we can think how to handle it then.

There is another kind lane `cluster-up/cluster/kind-k8s-1.17`
it doesn't have CI, people from arm use it, so please consider that.
both lanes use `./cluster-up/cluster/kind/common.sh` for example, if that is what you meant.
Its basically just the kind part of the sriov-kind lane.

EdDev:
> > I saw this while reviewing this specific commit, but I guess it is a general question: Why we need the `kind` folder? We use it only under the sriov lane, and it is confusing to have this in a different place. If there will be another kind lane later on, we can think how to handle it then.
> 
> There is another kind lane `cluster-up/cluster/kind-k8s-1.17`
> it doesn't have CI, people from ARM use it, so please consider that.
> #553
> 
> both lanes use `./cluster-up/cluster/kind/common.sh` for example, if that is what you meant.
> Its basically just the kind part of the sriov-kind lane.

But we have no job that runs it anywhere publicly so how do we know it is used?
Anyway, I am also fine with moving the parts that are sriov specific, under the specific folder... that way we know who uses it.

oshoval:
> But we have no job that runs it anywhere publicly so how do we know it is used?

According https://github.com/kubevirt/kubevirtci/pull/553 for example.
Not perfect but still worth consideration.
Just raised it up so the lane won't be removed out of the blue, unless discussed first.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: We need to make sure /var/run/netns exists so that `lplink` will will "learn" the network namespaces of each worker node(container).
There is an explanation inside `prepare_node_netns` that mentioned this directory, but yeah no explanation for this line. 

I will add a comment next to it.

EdDev: So if you need it for `prepare_node_netns`, place it before the for-loop. Now it is between variables declarations.

ormergi:
> Review: SRIOV provider, config_sriov.sh: Export nodes configurations code
> 
> Filenames:
> 
> * I think you can call the now `sriov-node` folder just `node`.
> * I do not understand what `sriov_node.sh` has compared to what `configure_vfs.sh` has. The name is too odd now and looking at the content I do not understand what each represent.

@EdDev `sriov_node.sh` used as a library that, consumed by `config_sriov_cluster.sh`,
It contains operations that we need for configuring cluster nodes with SRIOV PF's and VF's such as:
discovering the SRIOV PF,  moving PF's to cluster nodes netns.

`configure_vfs.sh` is a script that creates VF's and configure their driver, 
it used by `configure_sriov_pfs_and_vfs` function of `sriov_node.sh`.

provider.sh calls` config_sriov_cluster.sh` (sources `sriov_node.sh`)
`config_sriov_cluster.sh` invokes `configure_vfs.sh` on each worker node

ormergi:
> I saw this while reviewing this specific commit, but I guess it is a general question: Why we need the `kind` folder? We use it only under the sriov lane, and it is confusing to have this in a different place. If there will be another kind lane later on, we can think how to handle it then.
> 
> WDYT?

@EdDev 
`kind/` contains the code  and manifests for creating plain KIND cluster using kubevirtci scripts (make cluster-up).
I think we shouldn't move it to be part of specific provider, the current separation make sense to me.
And kubevirtci providers tend to change frequently, we might need to move it around again.
There is no doubt there is a place for improvement, like tearing it down to few small scripts as you suggested a while back.

Currently there are two kind provider the SRIOV one and `kind-k8s-1.17` @oshoval mentioned.
I think we still have kind-k8s-1.17 provider as an example to run Kubevirt over KIND.
It contradicts with the principal of having only providers we use on CI, 
If we want to remove it I guess we should bring this discussion to the community.

EdDev:
> I think we shouldn't move it to be part of specific provider, the current separation make sense to me.

If there is another provider that uses it, fine. The fact that we have no sanity check for it makes it unsupported imo, just because we do not know when we break it.

anyway, at the minimum I would not expect we land sriov specific content in that folder like this change does.

cluster-up/cluster/kind/migration.sh
ormergi: My intention was to call what it does w/o explaining mentioning how as we plan to replace the podpreset with a webhook, its the implementation of we are going to patch virt-launcher pods.

Looking at this again what you suggests make sense, 
I wil rename it to podPreset and move `kind_migration::enable_live_migration` to the caller (provider.sh)

ormergi:
> anyway, at the minimum I would not expect we land sriov specific content in that folder like this change does.

I tried to avoid adding more to kind/common.sh so I created another script for the pod preset part.
I prefer not to move kind/common.sh on this PR and until we do
I think its the right place for podpreset.sh script.
WDYT?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -25,85 +22,134 @@ SRIOVDP_RESOURCE_NAME="sriov_net"
...
-    # In case two clusters started at the same time, they might race on the same PF.
-    # The first will manage to assign the PF to its container, and the 2nd will just skip it
-    # and try the rest of the PFs available.
-    if ip link set "$pf_name" netns "$node"; then
-      if timeout 10s bash -c "until ip netns exec $node ip link show $pf_name > /dev/null; do sleep 1; done"; then
-        pf_array+=("$pf_name")
-        [ "${#pf_array[@]}" -eq "$pf_count_per_node" ] && break
-      fi
-    fi
+    pfs_names+=( $pf_name )
```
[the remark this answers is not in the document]
ormergi: Done

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -25,85 +22,134 @@ SRIOVDP_RESOURCE_NAME="sriov_net"
...
-  SRIOV_NODE=$MASTER_NODE
-else
-  SRIOV_NODE=$WORKER
-fi
+function validate_nodes_sriov_allocatable_resource() {
+  local -r resource_name="$SRIOVDP_RESOURCE_PREFIX/$SRIOVDP_RESOURCE_NAME"
+  local -r sriov_nodes=$(_kubectl get nodes -l $SRIOV_NODE_LABEL -o custom-columns=:.metadata.name --no-headers)
+
+  for sriov_node in $sriov_nodes; do
+    num_vfs=$(total_vfs_count_on_node "$sriov_node")
```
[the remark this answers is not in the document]
ormergi: Done

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: Done

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
ormergi: Yep, I added newline

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
```diff
@@ -0,0 +1,99 @@
...
+
+function move_pf_to_node_netns() {
+  local -r node_name=$1
+  local -r pf_name=$2
+
+  # Move PF to node network-namespace
+  ip link set "$pf_name" netns "$node_name"
+  # Ensure current PF is up
+  ip netns exec "$node_name" ip link set up dev "$pf_name"
```
oshoval, line 84: It used to be 
https://github.com/kubevirt/kubevirtci/pull/570/files#diff-499cff045d8a875e8b39f4a9fd59354eb99e8f8d91a95ebc70fbf42d2d5b5beaL51

Now it means that you won't be able to spin two clusters locally at the very same time,
you will still be able to spin them if you spin one, wait for its to done spinning and then spin the 2nd.
So either please fix it, or if we don't want parallel spinning (just parallel running) locally, please update about it in the doc.

(this comment affects only locally)

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
ormergi: @EdDev  I updated the commit message with expiation about this variable and default value

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
ormergi: Why it wont affect CI jobs?

ormergi:
Resolved code review comments, changes:
- Renamed mogration.sh -> podpreset.sh, renamed function prefix accordingly
  Moved the explanation of using PodPreset to the provider.sh script
- Added `local` keyword to variables definitions

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
ormergi: I updated the doc.

oshoval: It won't affect CI because the undercloud CNI allocates orthogonal PFs to each job,
the race that could happen is just if the job itself runs two clusters at the very same time,
this what can happen locally, because locally we don't have the 2nd layer of job separation.

Thanks

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
```diff
@@ -0,0 +1,99 @@
...
+  ip netns exec "$node_name" ip link show
+}
+
+function sriov_node::total_vfs_count() {
+  local -r node_name=$1
+  local -r node_pid=$(docker inspect -f '{{.State.Pid}}' "$node_name")
+  local -r pfs_sriov_numvfs=( $(cat /proc/$node_pid/root/sys/class/net/*/device/sriov_numvfs) )
```
oshoval, line 91: we have just one node here, so why use an array ?
(also line 94)

just need to make sure it fails once changed, if it doesn't find the sriov_numvfs,
because moving PF to the node is an async command which might take some time to fnish

cluster-up/cluster/kind/podpreset.sh
```diff
@@ -0,0 +1,34 @@
...
+
+    _kubectl apply -f "$KIND_MANIFESTS_DIR/product-uuid-podpreset.yaml" -n "$namespace"
+}
+
+function pod_preset::expose_unique_product_uuid_per_node() {
+    local -r cluster_name=$1
+    local -r namespace=$2
+
+    pod_preset::enable_feature "$cluster_name"
+    sleep 3
```
oshoval, line 31: is there something else beside sleep that can check the feature is enabled (best option) ?
maybe even retry on create_virt_launcher_fake_product_uuid_podpreset would be better?

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov_cluster.sh
```diff
@@ -0,0 +1,68 @@
...
+pfs_names=($(sriov_node::discover_host_pfs))
+pf_count="${#pfs_names[@]}"
+[ "$pf_count" -eq 0 ] && echo "FATAL: Could not find available sriov PF's" >&2 && exit 1
+
+total_pf_required=$((worker_nodes_count*PF_COUNT_PER_NODE))
+[ "$pf_count" -lt "$total_pf_required" ] && \
+  echo "FATAL: there are not enough PF's on the host, try to reduce PF_COUNT_PER_NODE
+  Worker nodes count: $worker_nodes_count
+  PF per node count:  $PF_COUNT_PER_NODE
+  Total PF count required:  $total_pf_required" >2 && exit 1
```
oshoval, line 50: I would prefer if block please (and not just here, but especially here, its too long)

EdDev:
I have two points that need discussion (except of the inline comments):
- Separation and naming of the `sriov_node.sh` and `configure_vfs.sh`.
- Having the `podpreset.sh` in the common kind folder (instead in the sriov specific one which actually requires it).

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov_cluster.sh
ormergi: The logical term is short, single variables comparison, the error message is long though
I can extract it to a variable that will be just before the condition.

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
oshoval: I read it wrong, the array is indeed because there might be theoretical couple PFs per node.

The issue i possibly see here, is since the ip link set is not blocking, and the waiting of it was removed [1]
the amount of VF that it will calculate might bot be accurate, and later when it spin and wait for this number of PF to appear in the resource, it might not be accurate.
For example if there are 2 PFs, it won't exit here with an error, but it will fail the cluster-up with error,
or even worse, it won't fail it with error, because lets say each PF and N VFs,
it will wait for N instead of 2N, and once there are N it will finish successfully, without checking that the other indeed are valid in the resource.

The solution of this can be the following:
Return the code that waits for the PFs to propagate in the node [1] 
but in case it failed the two checks there, return error, unless a flag say PARALLEL=true
and then it can continue as it was
(but in that case, it should also know what is the expected PF it passed, so it can sanity check it indeed catch all their VF in this function - which is good to have anyhow, in all cases)

It will also solve
https://github.com/kubevirt/kubevirtci/pull/570#discussion_r609437642
if we will have PARALLEL=true

[1] https://github.com/kubevirt/kubevirtci/pull/570/files#diff-499cff045d8a875e8b39f4a9fd59354eb99e8f8d91a95ebc70fbf42d2d5b5beaL51

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov_cluster.sh
oshoval: Matter of test, Quique and myself from what i know prefer if,
Edy and maybe you prefer this,
your call

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
```diff
@@ -56,6 +56,7 @@ In order to achieve that, there are two options:
 The user can list the PFs that should not be allocated to the current cluster, keeping in mind
 that at least one (or 2 in case of migration), should not be listed, so they would be allocated for the current cluster.
 Note: another reason to blacklist a PF, is in case its has a defect or should be kept for other operations (for example sniffing).
+- Clusters should be created one by another and not in parallel (to avoid races over SRIOV PF's).
```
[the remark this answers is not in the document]
oshoval: possible suggestion
https://github.com/kubevirt/kubevirtci/pull/570#discussion_r609497314

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
oshoval: last paragraph of https://github.com/kubevirt/kubevirtci/pull/570#discussion_r609365714 explained

see please the elaboration
https://github.com/kubevirt/kubevirtci/pull/570#discussion_r609497314

cluster-up/cluster/kind/podpreset.sh
ormergi: k8s does not provide proper way to check which features are enabled other then do some "forensics" to kube-apiserver
I will add a check for kube-apiserver process arguments (ps -ef | grep ..)

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
ormergi: > last paragraph of [#570 (comment)](https://github.com/kubevirt/kubevirtci/pull/570#discussion_r609365714) explained
> 
You comment on a function not on the right place we dont have context and its confusing.
Please remove it from there and write it here.
If you think you saw something that need to be change, and changing it means to touch multiple places
comment about it on the PR page instead.

oshoval: Actually its not a mistake, its this code 
https://github.com/kubevirt/kubevirtci/pull/570/files/73fa656af60f4b386531306e53d54249840854a9#diff-0ab4a7b3ef24d741b22df5cd4fc6fd14ac398ecd2e6b831aa400ba7a7bd3494eR82-R84

that possible has consequences on those lines,
it may flake now, because it takes time to move the PF to the container
All the details are here https://github.com/kubevirt/kubevirtci/pull/570/files/73fa656af60f4b386531306e53d54249840854a9#r609497314
but i will try to focus and make it straight to the pint below in this comment:

I can't delete it now and order it, because there are discussions already.

The main issue is that removing the code
https://github.com/kubevirt/kubevirtci/pull/570/files#diff-499cff045d8a875e8b39f4a9fd59354eb99e8f8d91a95ebc70fbf42d2d5b5beaL51

caused at least two things:
1. We can't spin local clusters parallell in the very same time, Edy find it important to preserve it.
2. `ip netns exec "$node_name" ip link set up dev "$pf_name"` might be flaky now, as moving the PF is async, it might continue `ip link set` even that the PF is not moved yet by the previous command, but it existed aready.

I would start focusing on those, imo it can be even be on future PRs, upon need.
More possible deeper suggestions in the reference, but we should start with those 2 i think.

ormergi:
> I have two points that need discussion (except of the inline comments):
> 
> * Separation and naming of the `sriov_node.sh` and `configure_vfs.sh`.
configure_vfs.sh introduced of #563, its create VF's and configure their driver.
it meant to use inside a cluster node.

sriov_node.sh includes operations that related to the cluster nodes,
one of them is moving PF's to nodes netns's and configure VF's.
That operation invokes configure_vfs.sh script on the cluster nodes.

> * Having the `podpreset.sh` in the common kind folder (instead in the sriov specific one which actually requires it).
podpreset.sh is under kind because its general cluster feature related and not SRIOV provider exclusive

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
ormergi: 1. Kubevirt  changes frequently as you may know, and recently the migration code also changes and we dont have SRIOV migration tested on CI just yet which is very bad. 
We already had to prepare a patch after there was a change on Kubevirt that broke SRIOV migration (converting libvirt domains to have persistent config, you can have more details on https://github.com/kubevirt/kubevirt/pull/5152 https://github.com/kubevirt/kubevirt/pull/5150)
Due to this urgency I prefer to handle this after this PR get into Kubevirt so SRIOV migration will be tested on CI.

If we will agree that this is something we can leave without it, we dont need to followup on this until there will be a demand for it.
I am saying this because AFAIU the main purpose of it is to optimize CI and we dont know of someone  that uses it for development.
  
Have that said its still possible to run two cluster on a local setup its not totally not functional, 
I added this constrain that creating two clusters should be sequential to the README.

2. Personally I never encountered problems with it, I tested it many times on CI and even more on local setup.
Lets handles this when we see it happen like you suggests.

cluster-up/cluster/kind-k8s-sriov-1.17.0/README.md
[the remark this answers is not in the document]
ormergi: I tried few things and its pretty complicated and will take more time.

 As I mentioned on https://github.com/kubevirt/kubevirtci/pull/570#discussion_r611195319
We dont have SRIOV migration tested on CI until this PR is merged,
due to this urgency, and the fact that this PR doesnt totally break running two clusters it just adds another constrain to the list (clusters shoud be created sequentially)
I prefer us to solve this after this PR will land on Kubevirt.

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
ormergi: > Regarding one PF or more, this function seems not to assume anything, it just returns a number. If that is correct or not, is a question about what is assumed at the caller. If we have two PF/s, is the caller fine to continue or not?

This function calculates how many VF's there are on a node regardless of how many PF's if has.
The [caller](https://github.com/kubevirt/kubevirtci/blob/0645fa84c24db03f09d8ff5fd33dfff41c78d78f/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh#L26)  checks each sriov capable node VF's count against the node allocatable resource that `sriov-device-plugin` updated.

This is how we check that each sriov capable node has the correct allocateable resource
with the correct amount according to sriov-device-plugin config. 
If there will be a problem with missing the PF's on the node or configuring the VF's, 
`sriov-dp` would raise an error and the allocateable resource would not apear on the node state, causing the caller to fail and stop `cluster-up`.

ormergi:
Changes:
- Renamed sriov-node.sh to node.sh (directory name already indelicate its sriov nodes related)
- On podpreset.sh, rename functions, on creating UUID PodPreset replaced the sleep with proper validation

cluster-up/cluster/kind/podpreset.sh
ormergi: Done

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
oshoval: Fine by me, but see note please, just before we decide that.

Note: Returning the code 
https://github.com/kubevirt/kubevirtci/pull/570/files#diff-499cff045d8a875e8b39f4a9fd59354eb99e8f8d91a95ebc70fbf42d2d5b5beaL51
Would fix both imo, without nothing else needed.
```
    if ip link set "$pf_name" netns "$node"; then
      if timeout 10s bash -c "until ip netns exec $node ip link show $pf_name > /dev/null; do sleep 1; done"; then
        pf_array+=("$pf_name")
        [ "${#pf_array[@]}" -eq "$pf_count_per_node" ] && break
      fi
    fi
```

It verify that the PF is moved, and allow parallel spin up.
So it solves both [1] and [2].

The only corner case which is super rare if even happen (didn't saw), is that wasn't move in 10 seconds but it was moved in the background later, and then the nodes will have a PF that wasn't meant to be there.
This why I suggested an assert on the expected PF / VF, but again, we don't need it in this PR, nor to handle the theoretic corner case IMO.

Btw, did you saw a case where you needed 
`ip netns exec "$node_name" ip link set up dev "$pf_name"`?
I didn't personally, but anyhow its a good barrier in case we don't solve [2]

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
ormergi: Done

cluster-up/cluster/kind/podpreset.sh
```diff
@@ -0,0 +1,48 @@
+ #!/usr/bin/env bash
+
+set -xe
+
+source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
+
+function pod_preset::enable_admission_plugin() {
```
EdDev, line 7: nit: it should reflect the filename, so the `_` is redundant.

EdDev (commented):
I have only one nit inline.

For some reason, the CI is not happy.

cluster-up/cluster/kind/podpreset.sh
```diff
@@ -0,0 +1,48 @@
...
+
+    _kubectl apply -f "$KIND_MANIFESTS_DIR/product-uuid-podpreset.yaml" -n "$namespace"
+}
+
+function pod_preset::expose_unique_product_uuid_per_node() {
+    local -r cluster_name=$1
+    local -r namespace=$2
+
+    pod_preset::enable_admission_plugin "$cluster_name"
+    pod_preset::validate_admission_plugins_is_enabled "10" "PodPreset"
```
oshoval, line 46: It fails because the `s` of plugins

oshoval:
We used to have a final line saying `SR-IOV Kind Cluster is Ready`, if all went good
Currently the final prints are non informative
I think it worth to add it back please

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/sriov_node.sh
ormergi: Fortunately adding warping `ip link set <pf> netns <node>` with `timeout` and `until` wont enable creating clusters in parallel.
The main problem is the fact that we pick PF's before moving them to a node netns in advance, 
in case of a failure we cant pick the next PF, like on the current implementation (which is pretty straightforward because it supports only one node).

The function `node::configure_sriov_pfs_and_vfs` iterates through the nodes and move the PF's for each node netns.
Each iteration (for each node) PF's names array is sliced according to `PF_COUNT_PER_NODE` flag.
Then each PF on that slice is moved to the current node netns.
In case moving PF to node netns is failed for any reason (`ip link set <pf> netns <node>`) 
cluster-up will stop with an error.

Considering the scenario we create two clusters in parallel, `cluster1` and `cluster2`, where `cluster1` started first:
both `cluster2` and `cluster1` picked  the same PF's to move, 
cluster-up would fail on moving PF command by `cluster2` with an error that the PF does not exists (on the default netns).
We could do best effort and not fail on this, but we would end up with non consistent setups where not all nodes has the desired amount of PF's which is bad idea.

Due to the reasons I mentioned on this thread I prefer to address the problem I described on a follow up PR in case we need it.

About warping `ip link set <pf> netns <node>` with `until` `timeout`, you mentioned that this operation is non blocking.
From what I know and saw `iplinks` operations are blocking.
Could you please elaborate or give a reference that tells that this is operation is not blocking?

oshoval: About the first discussion:
Changing the flow such that it will first move the needed PFs, with the option to `continue` and then look in the PFs it have already would solve that (instead of discover_host_pfs and then move, I already said in the past that imo its not needed to split it).
But ok don't address it on this PR.

> About warping ip link set <pf> netns <node> with until timeout, you mentioned that this operation is non blocking.
From what I know and saw iplinks operations are blocking.
Could you please elaborate or give a reference that tells that this is operation is non blocking?

Based on something Edy told me in the past, and i saw it sometimes happen, for example during developing the underlcloud CNI (unless we find out it was due to other issues, but adding this retry indeed fixed it)
I don't have in hand documentation, and since we are not handling it right now its not crucial to find one imo.

cluster-up/cluster/kind/podpreset.sh
ormergi: Done

cluster-up/cluster/kind/podpreset.sh
ormergi: Good catch, I changed it to consider the right container according to cluster name

ormergi:
> We used to have a final line saying `SR-IOV Kind Cluster is Ready`, if all went good
> Currently the final prints are non informative
> 
> ```
> + '[' get == console ']'
> + '[' get == vnc ']'
> + '[' get == start ']'
> + '[' get == stop ']'
> + '[' get == migrate ']'
> + '[' get == virt ']'
> + _kubectl get nodes
> + /root/go/src/kubevirt.io/kubevirt/_ci-configs/kind-k8s-sriov-1.17.0/.kubectl --kubeconfig=/root/go/src/kubevirt.io/kubevirt/_ci-configs/kind-k8s-sriov-1.17.0/.kubeconfig get nodes
> + [[ '' != '' ]]
> ```
> 
> I think it worth to add it back please

Done

ormergi:
Changes:
- Renamed `podprest.sh` function prefix `pod_preset::..` => `podpreset::`
- Changed the PodPreset enablement validation to check the right container according to cluster-name
- Added "cluster is ready" message
- Fixed echo to `stderr` when there are not enough PF's

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
```diff
@@ -13,33 +13,40 @@ else
...
-    ${KUBEVIRTCI_PATH}/cluster/$KUBEVIRT_PROVIDER/config_sriov.sh
+    ${KUBEVIRTCI_PATH}/cluster/$KUBEVIRT_PROVIDER/config_sriov_cluster.sh
+
+    # In order to support live migration on containerized cluster we need to workaround
+    # Libvirt uuid check for source and target nodes.
+    # To do that we create PodPreset that mounts fake random product_uuid to virt-launcher pods,
+    # and kubevirt SRIOV tests namespace for the PodPrest beforhand.
+    podpreset::expose_unique_product_uuid_per_node "$CLUSTER_NAME" "$SRIOV_TESTS_NS"
+
+    echo "$KUBEVIRT_PROVIDER cluster is ready"
```
oshoval, line 46: We might have 2 clusters locally, so worth to add the `$CLUSTER_NAME` please,
as they differ by name but not with the provider

oshoval:
I would suggest if possible to test it locally please with two clusters,
all needed basically is just use a different name and different physical folder of kubevirtci

cluster-up/cluster/kind/podpreset.sh
```diff
@@ -0,0 +1,56 @@
...
+    /etc/kubernetes/manifests/kube-apiserver.yaml'
+}
+
+function podpreset::validate_admission_plugin_is_enabled() {
+    local -r wait_time=$1
+    local -r admission_plugin=$2
+    local -r cluster_name=$3
+    local -r control_plane_container="$cluster_name-control-plane"
+
+    timeout "${wait_time}s" bash <<EOT
```
oshoval, line 22: timeout should already return non zero and fail the function in case it fails, unless 
we again caused `set -e` to be ignored

ormergi: I tested this again, it wont catch **all** errors, since we got `-e` flag up.
I tried to avoid it but I can think of another option, I will warp the `timeout until` with `if` clause,
it will catch all error w/o messing with `-e` flag.

oshoval: > I tested this again, it wont catch all errors, since we got -e flag up.
It has a contradiction i think, `set -e` means it will fail all every error, unless we mess the `set -e`, this what you mean ?

We can even see that the failure before did return 124 which is timeout event of timeout, and failed the whole script (when it ran ps -ef not on the container, and couldn't find the string)
so i am not sure i understand what you saw now, that you are saying it won't catch errors ? (unless we just recently messed up the `set -e` with the latest changes)

I will strive to keep it as simple as possible please.

ormergi:
Changes:
- Improved  `PodPreset` enablement validation error handling
- New commit for adding new env var that will hold the PF names that moved to nodes netns
  It used when `sriov-dp` config is patched to have the correct PF names.

ormergi:
> I would suggest if possible to test it locally please with two clusters,
> all needed basically is just use a different name and different physical folder of kubevirtci
> use just one worker, unless you have 4 PF total

Verified, creating two clusters works

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/node.sh
```diff
@@ -0,0 +1,108 @@
...
+    $node_exec chmod 666 /dev/vfio/vfio
+
+    # Create and configure SRIOV Virtual Functions on SRIOV node
+    docker cp "$CONFIGURE_VFS_SCRIPT_PATH" "$node:/"
+    $node_exec bash -c "DRIVER=$VFS_DRIVER DRIVER_KMODULE=$VFS_DRIVER_KMODULE ./$config_vf_script"
+
+    _kubectl label node $node $SRIOV_NODE_LABEL
+  done
+
+  export PFS_IN_USE="${pfs_in_use[@]}"
```
oshoval, line 71: Why suddenly its needed?

oshoval (commented):
Please answer https://github.com/kubevirt/kubevirtci/pull/570#discussion_r612448023
saw the change of the timeout, (adding if), just want to understand please if its because the `set -e` doesn't do its job in that function, due to the `local` expression

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/node.sh
ormergi: We need this for the caller to pass the correct PF names to `sriov_components::deploy` which patch the sriodp config (`SriovNetworkNodePolicy` equivalent)
https://github.com/kubevirt/kubevirtci/blob/2a7cdf82fcb91af08892b58720a01a2597458485/cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov_cluster.sh#L52-L62

oshoval: Ok thanks

One point, why not to just allocate the PFs to the node,
and later peek into the node and then see what PFs it has ?
as we had it before.
It will simplifies the flow and maintenance of that,
is there any reason to not do that?

I am fine to do it on next PRs if there is an agreement that it should be done.

cluster-up/cluster/kind/podpreset.sh
ormergi: I mean that because we use `-e` flag in case `timeout` finished with an error we didn't get to the line with `[ $? -ne 0 ] ...`
`-e` flag works as expected here.
I warped `timeout until` this with `if` clause to catch those errors,
In case `timeout` fail with 124 we would get informative error message.

oshoval: Cool, I believe we can simplify the function, but not needed in this PR.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -25,85 +22,139 @@ SRIOVDP_RESOURCE_NAME="sriov_net"
...
 VFS_DRIVER_KMODULE="vfio_pci"
 
-function move_sriov_pfs_netns_to_node {
-  local -r node=$1
-  local -r pf_count_per_node=$2
-  local -r pid="$(docker inspect -f '{{.State.Pid}}' $node)"
-  local pf_array=()
+function prepare_node_netns() {
+  local -r node_name=$1
+  local -r node_pid=$(docker inspect -f '{{.State.Pid}}' "$node_name")
```
qinqon: Can we wrap the "docker" commands with functions like here `inspect_container $node_name` in case we move to podman or crioctl.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -25,85 +22,139 @@ SRIOVDP_RESOURCE_NAME="sriov_net"
...
+    pfs_to_move=( "${pfs_names_array[@]:$pfs_array_offset:$pf_count_per_node}" )
+    echo "Moving '${pfs_to_move[*]}' PF's to '$node' netns"
+    for pf_name in "${pfs_to_move[@]}"; do
+      move_pf_to_node_netns "$node" "$pf_name"
+    done
+    # Increment the offset for next slice
+    pfs_array_offset=$((pfs_array_offset + pf_count_per_node))
+
+    # KIND mounts sysfs as read-only by default, remount as R/W"
+    node_exec="docker exec $node"
```
qinqon: ditto

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -25,85 +22,139 @@ SRIOVDP_RESOURCE_NAME="sriov_net"
...
+    $node_exec mount -o remount,rw /sys
+    $node_exec chmod 666 /dev/vfio/vfio
+
+    # Create and configure SRIOV Virtual Functions on SRIOV node
+    docker cp "$CONFIGURE_VFS_SCRIPT_PATH" "$node:/"
```
qinqon: ditto

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
```diff
@@ -25,85 +22,139 @@ SRIOVDP_RESOURCE_NAME="sriov_net"
...
+  done
+}
 
-# this is to remove the ending 0 in case the candidate worker is the first one
-if [[ "$SRIOV_NODE" == "${WORKER_NODE_ROOT}0" ]]; then
-  SRIOV_NODE=${WORKER_NODE_ROOT}
-fi
+function total_vfs_count_on_node() {
+  local -r node_name=$1
+  local -r node_pid=$(docker inspect -f '{{.State.Pid}}' "$node_name")
```
qinqon: ditto

cluster-up/cluster/kind-k8s-sriov-1.17.0/provider.sh
```diff
@@ -20,24 +20,20 @@ function set_kind_params() {
...
 
 function up() {
-    if [[ "$KUBEVIRT_NUM_NODES" -ne 2 ]]; then
-        echo 'SR-IOV cluster can be only started with 2 nodes'
-        exit 1
-    fi
-
     # print hardware info for easier debugging based on logs
     echo 'Available NICs'
     docker run --rm --cap-add=SYS_RAWIO quay.io/phoracek/lspci@sha256:0f3cacf7098202ef284308c64e3fc0ba441871a846022bb87d65ff130c79adb1 sh -c "lspci | egrep -i 'network|ethernet'"
```
qinqon, line 29: ditto

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/node.sh
ormergi: I am not sure I understand, with this PR changes it checks what PFs are available on the host
then it will move PF's to each worker node according to `PFS_COUNT_PER_NODE`.

We need to support a setup with multiple nodes where each node could have more then one PF
and also consider multi cluster creation.
This PR add more layers to support this not trivial setup with all its constrains.

cluster-up/cluster/kind/podpreset.sh
```diff
@@ -0,0 +1,56 @@
+ #!/usr/bin/env bash
+
+set -e
+
+source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
+
+function podpreset::enable_admission_plugin() {
+    local -r cluster_name=$1
+
+    docker exec "$cluster_name-control-plane" bash -c 'sed -i \
```
qinqon, line 10: ditto

cluster-up/cluster/kind/podpreset.sh
```diff
@@ -0,0 +1,56 @@
+ #!/usr/bin/env bash
+
+set -e
+
+source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
+
+function podpreset::enable_admission_plugin() {
```
qinqon, line 7: I thinks this has to be renamed to "enable_podpreset_admission_plugin" or parameterize same as "validate_admission_plugin_is_enabled"

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/node.sh
oshoval: Multi cluster has two variants, local and on CI, lets put them aside for a moment, since we can make them agnostic for the changes almost.

I suggest that we first move to each worker node PFs according PFS_COUNT_PER_NODE.
Then you don't need to export `PFS_IN_USE`
but just peek in the nodes, and see how many PFs it has and which, once you count the VF and config the VFIO.

The additional layer i meant that isn't needed imo, is the discover, which is decoupled now from moving the PF into the node and so on.

qinqon (changes requested):
Looks pretty good in fact is kind of impressive, just added a pair of nits but they are not a must.

cluster-up/cluster/kind/podpreset.sh
ormergi: I didnt called it `enable_podpreset_admission_plugin` because it has `podprest::` prefix,
so we have sense what plugin its going to enable, WDYT?

About parameterize it, `podpreset::enable_admission_plugin` implementation is specific for `PodPreset` feature
it patches `kube-apiserver` manifest by adding `PodPreset` to `--enable-admission-plugins` and `settings.k8s.io/v1alpha1=true` to `--runtime-config`
Other admission plugins might require other parameters to be patched.
Unlike `podpreset::validate_admission_plugin_is_enabled` that could check any admission plugin
(because it checks kube-apiserver command state)

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
ormergi: If other container run-times has the same API we can simply create `CRI` variable that will be initialized from `kind/common.sh` or `provider.sh`.
IF not we probably need to create new script (that will be used as library) with all the warping functions, 
and source its on `kind/common.sh` or `provider.sh` script.

I think this change could affect other providers as well,
Would you mind if we do that on other PR?

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/node.sh
ormergi: > I suggest that we first move to each worker node PFs according PFS_COUNT_PER_NODE.
> Then you don't need to export `PFS_IN_USE`
> but just peek in the nodes, and see how many PFs it has and which, once you count the VF and config the VFIO.
> 
> The additional layer i meant that isn't needed imo, is the discover, which is decoupled now from moving the PF into the node and so on.

Doing all (discover PFs, move to nodes, configure VF's) on the same function was too much so I decided to split it.

When moving the PF's to nodes, they are saved in variable that is used later for patching sriov-dp config
I think its better then moving them and then check the nodes to get them, we already have the names when we move them.

I cant use the variable that ` node::configure_sriov_pfs_and_vfs ` returns because it will include all the stdout content it generates.

There are few possible solutions:
1. Remove the export keyword, its could work w/o it because we source node.sh script 
    so eventually its the same context.
2. Define `PFS_IN_USE` variable from `node::configure_sriov_pfs_and_vfs` on the fly by, passing the name of the variable
    (@EdDev recommendation)
3. If adding another variable is unbearable, I can split `node::configure_sriov_pfs_and_vfs` in a way that we could
    use returned variable with the used PF's
    Something like:
    ```
    PFS_IN_USE=$(node::move_pfs $worker_nodes $pfs "$PF_COUNT_PER_NODE")
    node::configure_vfs $worker_nodes
    node::label_sriov_capable $worker_nodes
    ```

> Again, its not needed to be in this PR, unless we decide its needed, but first lets please sync on the idea.

Cool

cluster-up/cluster/kind/podpreset.sh
qinqon: then what we have to do is to hardcore PodPreset into `podpreset::validate_admission_plugin_is_enabled` this function cannot be generic if it's under "podpreset" or going to the opposite direction and move this functions and a generic version of  `validate_admission_plugin_is_enabled` out of porpreset.

cluster-up/cluster/kind-k8s-sriov-1.17.0/config_sriov.sh
qinqon: You are going to be safer with wrappers docker vs pod vs crioctl don't have full compatible interfaces, but you are right that this is not the place to address this.

qinqon:
@ormergi why are the provisioning jobs failing ?

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/node.sh
oshoval: > Doing all (discover PFs, move to nodes, configure VF's) on the same function was too much so I decided to split it.

I am looking on it otherwise:
1. You move the PF to the node, and check if it moved.
2. You create VF on the PFs that the node have.

The node has all the entropy it needs (all the information), no need additional middleman to convey the information.
Its the minimal operations that are needed anyhow.

oshoval:
I don't think this PR touched the other lanes, so not sure their failure need to be handled on this PR.

Conceptual ACK from me, thanks
I didn't ack yet because Quique and Edy (maybe) still review

cluster-up/cluster/kind/podpreset.sh
```diff
@@ -0,0 +1,56 @@
...
+
+    docker exec "$cluster_name-control-plane" bash -c 'sed -i \
+    -e "s/NodeRestriction/NodeRestriction,PodPreset/" \
+    -e "/NodeRestriction,PodPreset/ a\    - --runtime-config=settings.k8s.io/v1alpha1=true" \
+    /etc/kubernetes/manifests/kube-apiserver.yaml'
+}
+
+function podpreset::validate_admission_plugin_is_enabled() {
+    local -r wait_time=$1
+    local -r admission_plugin=$2
```
EdDev, line 18: Per @qinqon [comment](https://github.com/kubevirt/kubevirtci/pull/570/files#r613772514), this needs to be hardcoded as we are in the "podpreset" context.

ormergi: Done

cluster-up/cluster/kind/podpreset.sh
ormergi: I have changed `podpreset::validate_admission_plugin_is_enabled` to have PodPreset hardcoded, 
If we will need to mess with other admission plugins we can parameterize it

ormergi:
Changes:
- Hardcode PodPreset on `podpreset::validate_admission_plugin_is_enabled`
- Replace `export` of `PFS_IN_USE` with passing the variable name to `node::configure_sriov_pfs_and_vfs`,
   this variable is now set and initialized by this function

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/node.sh
ormergi: Yes I understand that all the info is there,
I split it to improve readability, it was already long in the first place and with this PR changes it got even longer
and hard to follow.

I picked option (2) on this PR, we can implement option (3) on a following PR

qinqon:
All good from my part

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-node/node.sh
oshoval: You dont even need PFS_IN_USE in option 3, or in what i meant, the info is in the node itself,
unless we speak on the same thing

EdDev (approved):
Thank you!

ormergi:
/hold 

Test against kubevirt master on CI
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/5238/pull-kubevirt-e2e-kind-1.17-sriov/1384051821726666752

oshoval:
Any explanation why https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/5238/pull-kubevirt-e2e-kind-1.17-sriov/1384051821726666752 flaked ?

timeouts or something ?

ormergi:
> Any explanation why https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/5238/pull-kubevirt-e2e-kind-1.17-sriov/1384051821726666752 flaked ?
> 
> timeouts or something ?

I suspect kubevirt/kubevirt#5330 is the reason, it introduced big change to migration code
I saw this error on virt-launcher pod 
```
virError(Code=67, Domain=20, Message='unsupported configuration: Target domain host device count 1 does not match source 0')
```

ormergi:
kubevirt/kubevirt#5238 includes this PR changes on kubevirt latest master
Now it also include revert commit for  kubevirt/kubevirt#5330 which probably casued a regession for SRIOV migration
As we can see the lane passed with the revert commit
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/5238/pull-kubevirt-e2e-kind-1.17-sriov/1384112596952551424

I am removing the hold as this PR does not imply any regression to kubevirtci or kubevirt
/unhold

PR #574, opened 2021-03-19
fix ceph selinux permissions

mhenriks:
Signed-off-by: Michael Henriksen <mhenriks@redhat.com>

Was only setting rook/ceph selinux labelfor node01 here:  #562

This PR makes sure labeling done on all nodes


stu-gott (commented):
/approve

PR #575, opened 2021-03-19
start centos8 nodes with qemu monitor

dankenigsberg:
This would allow taking a cluster snapshot just after `make
cluster-sync`, run a disruptive test, and then revert to the initial
cluster state.

/cc @pkliczewski 

Signed-off-by: Dan Kenigsberg <danken@redhat.com>

### DIFF
--- a/cluster-provision/centos8/scripts/vm.sh
+++ b/cluster-provision/centos8/scripts/vm.sh
@@ -109,6 +109,8 @@ if [ -n "${BLOCK_DEV}" ]; then
   block_dev_arg="-drive format=qcow2,file=${BLOCK_DEV},if=virtio,cache=unsafe"
 fi
 
+MONITOR_SOCKET=/var/run/mon.sock
+
 exec qemu-system-x86_64 -enable-kvm -drive format=qcow2,file=${next},if=virtio,cache=unsafe ${block_dev_arg} \
   -device virtio-net-pci,netdev=network0,mac=52:55:00:d1:55:${n} \
   -netdev tap,id=network0,ifname=tap${n},script=no,downscript=no \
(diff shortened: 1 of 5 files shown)


cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -47,3 +47,22 @@ function up() {
         fi
     fi
 }
+
+function _send_monitor_to_nodes() {
+    for NODE_NUM in `seq "$KUBEVIRT_NUM_NODES"`
```
dankenigsberg, line 52: this is unhelpful, since each node is run on a separate container.

cluster-provision/centos8/scripts/vm.sh
```diff
@@ -107,10 +107,14 @@ if [ -n "${BLOCK_DEV}" ]; then
...
+
 exec qemu-system-x86_64 -enable-kvm -drive format=qcow2,file=${next},if=virtio,cache=unsafe ${block_dev_arg} \
   -device virtio-net-pci,netdev=network0,mac=52:55:00:d1:55:${n} \
   -netdev tap,id=network0,ifname=tap${n},script=no,downscript=no \
   -device virtio-rng-pci \
   -vnc :${n} -cpu host -m ${MEMORY} -smp ${CPU} ${QEMU_ARGS} \
   -serial pty -M q35,accel=kvm,kernel_irqchip=split \
-  -device intel-iommu,intremap=on,caching-mode=on -soundhw hda
+  -device intel-iommu,intremap=on,caching-mode=on -soundhw hda \
+  -monitor unix:${MONITOR_SOCKET},server,nowait \
```
dankenigsberg, line 119: I should probably expose the monitor over `tcp::4444` and then forward the port from the container to the host using `gocli`, as done with vnc and ssh.

PR #577, opened 2021-03-23
fix single node rook/ceph

mhenriks:
Signed-off-by: Michael Henriksen <mhenriks@redhat.com>

By fixing multi node I broke single node.  WHen will this ever end?


stu-gott (commented):
/approve

PR #580, opened 2021-03-30
fix ineffectual assignment to err

hs0210:
Signed-off-by: Hu Shuai <hus.fnst@cn.fujitsu.com>

### DIFF
--- a/pack8s/internal/pkg/podman/podman_test.go
+++ b/pack8s/internal/pkg/podman/podman_test.go
@@ -304,6 +304,7 @@ var _ = Describe("podman", func() {
 			Expect(err).To(BeNil())
 
 			images, err := handler.ListImages()
+			Expect(err).To(BeNil())
 
 			found := false
 


pack8s/internal/pkg/podman/podman_test.go
```diff
@@ -304,6 +304,7 @@ var _ = Describe("podman", func() {
 			Expect(err).To(BeNil())
 
 			images, err := handler.ListImages()
+			Expect(err).To(BeNil())
```
oshoval, line 307: Might be better to use `Expect(err).ShouldNot(HaveOccurred())`
but i understand that it will pop out because the rest use `Expect(err).To(BeNil())`

oshoval (commented):
Thanks

oshoval:
```
+ dnf install -y NetworkManager-config-server-1.30.0
Last metadata expiration check: 0:05:13 ago on Tue 06 Apr 2021 03:24:26 PM UTC.
No match for argument: NetworkManager-config-server-1.30.0
Error: Unable to find a match: NetworkManager-config-server-1.30.0
```

lets retry and see if its happening again, not due to this PR anyhow.

/test check-provision-k8s-1.19

oshoval:
Failure of the provision is known, the team will take care of it.

pack8s/internal/pkg/podman/podman_test.go
hs0210: Thanks for your review and sorry for my late reply, I have updated the error assert. PTAL.

oshoval: Thanks, I meant that I was wondering myself which style we should use,
since all the file has already `Expect(err).To(BeNil())`
I am not sure we want one place to pop up differently, sorry for the misunderstanding

hs0210: Thanks, PTAL.

oshoval:
Thanks, please take a look here, 
https://github.com/kubevirt/kubevirtci/pull/610

pack8s would be removed according that PR
so i don't think we should continue this one?

Anyhow thank you on the PR

hs0210:
Got it, thanks.

PR #584, opened 2021-04-07
Remove NetworkManager version pinning on cluster provision

ormergi:
On QEMU based providers we use Centos Stream OS which frequently
changes causing check-provision lanes to fail due to no longer
available version of NetworkManager.
NetworkManager version is pinned for compatibility
with other projects, e.g: kubernetes-nmstate, CNAO.

check-provision-k8s-1.19 and check-provision-k8s-1.18 lanes are 
broken due to missing NetworkManager version.

This PR remove the version pinning for NetworkManager.
Consumers of kubevirtci who needs specific version can install after the cluster is up.

### DIFF
--- a/cluster-provision/k8s/1.19/provision.sh
+++ b/cluster-provision/k8s/1.19/provision.sh
@@ -151,9 +151,7 @@ sysctl --system
 echo bridge >> /etc/modules
 echo br_netfilter >> /etc/modules
 
-# Bump NetworkManager to 1.22.8
-NM_VERSION=1.30.0
-dnf install -y NetworkManager-$NM_VERSION
+dnf install -y NetworkManager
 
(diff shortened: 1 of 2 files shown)


oshoval (commented):
Can you check it with kubevirt please?

qinqon:
@oshoval this is ok with kubevirt they don't need an specific NM version, also if there is an issue we can pin NM there on the kubevirtci bump PR.

/lgtm
/approve

oshoval:
> @oshoval this is ok with kubevirt they don't need an specific NM version, also if there is an issue we can pin NM there on the kubevirtci bump PR.

Imo it should be tested, else stuff pile up, and later its harder to find out what caused which problem,
as we know we don't auto bump kubevirt.

PR #586, opened 2021-04-12
sriov, Move sriov CNI and DP to point to quay

oshoval:
In order to solve docker.io rate limits,
especially when running locally without docker proxy,
point to quay instead of docker.io for DP and CNI.


oshoval:
Thanks @fgimenez for mirroring them

oshoval:
Failure of check-provision-k8s-1.20  is not related to this PR

`time="2021-04-12T14:51:24Z" level=error msg="error attempting to run sonobuoy: waiting for run to finish: Pod entered a fatal terminal status: failed"`

oshoval:
Ready to review, thanks

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-components/sriov_components.sh
```diff
@@ -107,7 +107,7 @@ function sriov_components::wait_allocatable_resource() {
 
 function sriov_components::deploy_multus() {
   echo 'Deploying Multus'
-  _kubectl apply -f "$MULTUS_MANIFEST"
+  sed "s#nfvpe/multus#quay.io/kubevirtci/multus#" "$MULTUS_MANIFEST" | _kubectl apply -f -
```
ormergi, line 110: I prefer that we control multus image through sriov-components kustomization.yaml file, like we do for sriov-cni and sriovdp.

You could do that by specifying multus image on https://github.com/kubevirt/kubevirtci/blob/aafc09f2201d66ecb973d564674c76745acad7f5/cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-components/manifests/kustomization.yaml#L4-L15

I think its fine to treat multus as part of sriov-components as to goes together with sriov-cni and sriovdp for our use case.

oshoval: It means that we would need to deploy all the outputs of the kustomize at once,
I think its better to deploy them one by one, and not merged.
Moreover the kustomize reorder the file, it might break the logic in the future (or even now)

I would'nt merge them, for sure not for one line that we need to change.
The most I would consider imo, is to create another folder for kustomize of it or create a diff file for it.

alonSadan: I agree with @oshoval  in this. Better do deploy them separately I think. What do you think @ormergi ?

ormergi: > It means that we would need to deploy all the outputs of the kustomize at once,

This is actually big advantage because we could have "single config point" for the components we need
it will actually be easier to maintain
> 
> * I think its better to deploy them one by one, and not merged,
>   easier to debug and maintain.

Each component doesn't not relay on another I dont see how this is going to be a problem.
Its actually harder to maintain because you need to have checks for each component deployment.

> * Moreover the kustomize reorder the file, it might break the logic in the future (or even now)
> 

The ordering inside yaml manifest doesn't affect anything, Kustomize ordering shouldn't cause problems.

> I wouldn't merge them imo, for sure not for one line that we need to change.
> The most I would consider imo, is to create another folder for kustomize of it (overkill)
> or create a diff file for it (possible fine, but in this case one liner is better imo).
> 
> * Decoupling would also make the bumping process of each easier, as each would be self contained.
> * Moreover this sed is safe, in case the image won't exist it will directly fail, showing the problem.

It will make out life easier if we could have single place for all components config 
and avoid patching manifests manually (e.g using sed).
Having a diff file is like doing what Kustomize does manually, we can control the versioning and images easily and in a declarative way with it.

Currently we deploy multus and then sriov-components with `kubectl -f ..` one after another,
its equivalent to deploying them though Kustomize (single manifest).
Bumping one of the components includes bringing the new manifest,mirror the image and change the image tag (to point to our registry).
Doing the last part with Kustomize (single kustomize file or not) is easier and safe.

ormergi: For the sake of migrating to our own registry, I wont insist on having Kustomization for multus version
if we will have to do more changes we defiantly want to use Kustomize.

oshoval: You can see here that reordering can break changes for this guy
https://github.com/kubernetes-sigs/kustomize/issues/1337
I saw another discussion which i don't have the link in hand for this.

It also make life harder to see what was actually changed when comparing the input versus output.

I would even suggest a feature that will allow to deploy them step by step:
1. Re/Deploy kind
2. Re/Deploy multus
3. Re/Deploy sriov

Using flags, it helped me a lot when i developed
I could update just sriov without creating the whole cluster,
Using one kustomize it won't allow it.

I an personally against combining for the reason written above, but please feel free to suggest it on other PR / discussion.

@alonSadan @ormergi 
In case we are good to go, following the results of this discussion please rereview

Thanks

ormergi: > You can see here that reordering can break changes for this guy
> [kubernetes-sigs/kustomize#1337](https://github.com/kubernetes-sigs/kustomize/issues/1337)

What broke for him is the ordering of the deployment of his application components
In out case each component is independent of the other and could be deployed in any order.

> It also make life harder to see what was actually changed when comparing the input versus output.
> 

At the current state you dont have generated output for mutlus manifest at all.

> I would even suggest a feature that will allow to deploy them step by step:
> 
> 1. Re/Deploy kind
> 2. Re/Deploy multus
> 3. Re/Deploy sriov
> 
> Using flags, it helped me a lot when i developed
> I could update just sriov without creating the whole cluster,
> Using one kustomize it won't allow it.

> This will allow to remove the provider that only deploy kind, as we will have this feature in one provider.
> 
> I an personally against combining for the reason written above, but please feel free to suggest it on other PR / discussion.

There is no problem in having single kustomize, it  would just deploy multus again

Please consider adding kustomize for multus, its not a lot of work and it could improve maintainability

oshoval: I considered that, even before this PR, and i think its not that right thing to do.
Decoupling is precious.

I have other opinions about the different phases, but lets focus on the correct PR scope if possible.

Here you said you won't block
https://github.com/kubevirt/kubevirtci/pull/586#discussion_r618280405
Alon and myself are against it, so if we are good to go, lets continue please.

alonSadan (approved):
I think the ordering of the deployment, if it is indeed not deterministic, could cause headaches, though I am still not totally convinced. Anyway I think if the kustomize issue is that controversial, it is better to discuss it in a separate PR dedicated to deploy multus with kustomize.  For now I think this PR is good to go.
\lgtm

cluster-up/cluster/kind-k8s-sriov-1.17.0/sriov-components/sriov_components.sh
ormergi: > I considered that, even before this PR, and i think its not that right thing to do,
> for sure not in this PR.
> 
> Decoupling is precious, If something fail i want to know the step that was fail, i don't want to debug which is the one out of them, when i can otherwise.
> 
> If kustomize would be use combined, there will be generated ones which are harder to compare.
> I don't want it to deploy what not need to be deployed, I prefer fastest deployments when I work locally,
> and I need sometimes to deploy just parts, and it won't allow me.
> 

You missed the part where I agree it should be separated..
Anyway its not a lot of work to add kustomization to multus please consider doing that.

oshoval: I didn't miss it 
see last paragraph here please if thats what you mean
https://github.com/kubevirt/kubevirtci/pull/586#discussion_r618300080

oshoval:
/cc @qinqon

can you take a look please?

oshoval:
/cc @fgimenez 
mind taking a look please?

oshoval:
/retest 

failures are not related to this PR

fgimenez (approved):
/approve
/hold

Feel free to unhold when ready.

oshoval:
/hold cancel

Thanks
Its ready ;]

oshoval:
/retest

can we override it maybe ? this PR doesnt affect non sriov lanes

PR #590, opened 2021-04-20
Configure Istio deployment to use multus cni

rhrazdil:
Current configuration mistakenly uses calico directly
instead of multus.

cni configuration files are created in /etc/cni/multus/net.d/
When sidecar is injected to a pod, it's also annotated with
"k8s.v1.cni.cncf.io/networks"=istio-cni

Using pods with Istio sidecar then requires creation of a
network-attachment-definition. This net-attach-def may
have empty spec, it's only important it is called istio-cni [0]:
```
apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
  name: istio-cni
```
This commit also deploys Istio using IstioOperator instead of istioctl command for better readability
of the configuration. 

[0] https://istio.io/latest/docs/setup/platform-setup/openshift/#additional-requirements-for-the-application-namespace


qinqon:
/lgtm
/approve

Now that istio is up at check-cluster-up we can depend on CI results

qinqon:
/hold

@rhrazdil cancel this hold when you thing this is ready.

rhrazdil:
Had to force-push [empty commit](https://github.com/kubevirt/kubevirtci/compare/75dce06a99fe72da45d960fd1f834505249692f9..b2883f0ec9c5f5bbbe4c121eff4c5ce9d1c7c6ce) to re-trigger the CI which wasn't running after prow migration

PR #591, opened 2021-04-21
Include vm-dump-metrics in the fedora-with-test-tooling image

rmohr:
The tool is needed for downwardsMetrics testing.


davidvossel (approved):
/lgtm
/approve

PR #593, opened 2021-04-22
Switch to booting via kernel and initrd from outside the image

rmohr:
In order to make it easy to use different kernel and qemu arguments after provisioning, add a `--additional-kernel-args` argument for cluster-up and use a kernel and initrd directly via qemu.

Every provider has now a built-in set of default kernel arguments which can be extended on cluster-up.

To sum up:

 * during`gocli provision` one can use `--additional-persistent-kernel-arguments` to add permanent kernel arguments to a provider.
 * during `gocli up` one can use `--additional-kernel-args` to add additional kernel arguments just for this run.

Use-cases are:

 * cgroupsv2
 * fakenuma

In addition this PR fixes the `--additional-qemu-args` flag which had issues with quotes in extra passed arguments.


rmohr:
/hold

Uh I forgot that some of the boot args are already set via the provision scripts. That needs to be resolved first.

vasiliy-ul:
@rmohr , that looks great. I can try to adapt the cgroupv2 lane.

> In a follow-up, we can start exposing --append-kernel-args to gocli and fix the --additional-qemu-args flag.

Are you planning to create a new PR?

rmohr:
> > In a follow-up, we can start exposing --append-kernel-args to gocli and fix the --additional-qemu-args flag.
> 
> Are you planning to create a new PR?

Yes, I want to do a follow up.

@vasiliy-ul In this PR, I think I first have to change things a little, so that we have sane base-cmdline for each provider (probably just annother file which contains the default kernel flags) and remove the commandline argument handling in the provision scripts. What do you think as this for the scope of this PR?

vasiliy-ul:
Sounds good :+1:

rmohr:
@vasiliy-ul would be great if you could have a look.

rmohr:
```
Tests Suite: [sig-network] [Serial][rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:component]Networking VirtualMachineInstance with masquerade binding mechanism [Conformance][test_id:1780][label:masquerade_binding_connectivity]should allow regular network connection [outside_connectivity]should be able to reach the outside world [IPv6] 
```

in the 1.17 lane is failing on various PRs. Needs to be fixed but not related to this PR.

cluster-provision/k8s/1.21/provision.sh
```diff
@@ -4,7 +4,6 @@ set -ex
 
 # Configure cgroup version
 if [ "${UNIFIED_CGROUP_HIERARCHY}" == "1" ]; then
-    CMDLINE_LINUX_APPEND="${CMDLINE_LINUX_APPEND} systemd.unified_cgroup_hierarchy=1"
```
vasiliy-ul, line 7: Probably the same line can be dropped from `cluster-provision/k8s/*1.20*/provision.sh` as well

cluster-up/cluster/k8s-1.21-cgroupsv2/README.md
```diff
@@ -0,0 +1,81 @@
+# Kubernetes 1.20 in ephemeral containers (with cgroups v2)
```
vasiliy-ul, line 1: Small inconsistency in this file: I think k8s version should be `1.21`

vasiliy-ul:
> @vasiliy-ul would be great if you could have a look.

Found a couple of minor issues. For me looks good :+1:

rmohr:
> Found a couple of minor issues. For me looks good

Fixed. :+1:

rmohr:
@vasiliy-ul can you have another look at the PR and lgtm it if you think it is ready?

PR #594, opened 2021-04-23
Bump k8s 1.20.5 to 1.20.6

rhrazdil:
1.20.6 fixes issue with configMap resource quota  conformance tests, where the number of config maps allowed
is constrained to only 2. This blocks running conformance tests with Istio deployment, because istio
creates it's own configMap in new namespaces, which causes this limit to be exceeded.

https://github.com/kubernetes/kubernetes/blob/v1.20.5/test/e2e/apimachinery/resource_quota.go#L1481
->
https://github.com/kubernetes/kubernetes/blob/v1.20.6/test/e2e/apimachinery/resource_quota.go#L1481


rhrazdil:
Lets do the bump together with update of istio configuratin on #590

PR #596, opened 2021-04-27
Enable invtsc

rmohr:
In order to test high-performance VMI features `invtsc` needs to be explicitly enabled so that the qemu machine hast the `nonstop_tsc` flag.


vladikr:
Thanks :)
/lgtm
/approve

PR #597, opened 2021-04-28
Create k8s-1.21 provider

dhiller:
As kubevirt/kubevirtci#579 is the result of the kubevirtci autobump, I continue here, due to the autobump regularly removing my additional commits.

Additional changes to autobump:
- Remove deprecated feature gates BlockVolume and CSIBlockVolume 
- Add cluster-up folder for k8s-1.21 
- Bump sonobuoy

/cc @EdDev @qinqon @rmohr @fgimenez 

FYI @stu-gott


dhiller:
The conformance tests seem to be flaky somehow. While [this run against 1.21](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/579/check-provision-k8s-1.21/1387083344524611584) succeeded, [the following run, where nothing was changed at the provider, succeeded](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/579/check-provision-k8s-1.21/1387318541031575552). [Also same tests failed on 1.20](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/579/check-provision-k8s-1.20/1387122806944174080)

dhiller:
See also kubevirt/project-infra#1169

cluster-provision/k8s/1.20/conformance.json
```diff
@@ -1 +1,88 @@
...
+{
+  "Description": "DEFAULT",
+  "UUID": "c3bc7d76-6ce8-4c8a-8bcb-5c7ae5fb22a3",
+  "Version": "v0.19.0",
```
EdDev: There is no need to bump this one as well?
I guess I miss the relation between the version in this manifest and the one in the conformance.sh script.

EdDev (commented):
Thank you for pushing this forward!

Do we have any clue why there are flakes on the conformance tests?

cluster-provision/k8s/1.20/conformance.json
qinqon: This change has to be done at different PR, we want the k8s-1.21 provider.

qinqon:
@dhiller can you remove the k8s-1.20 changes or it was even failing without them ?

cluster-provision/k8s/1.20/conformance.json
dhiller: Oh, interesting catch, thanks for that! I thought I had updated this already, as I had seen this also, will take another look.

dhiller:
> @dhiller can you remove the k8s-1.20 changes or it was even failing without them ?

I saw that the conformance manifest for 1.20 had not been updated to match the 0.20 version of sonobuoy. I thought that until recently (interestingly they now have released version 0.50.0 for whatever reason) the minor versions always matched with those of kubernetes. Was that wrong?

dhiller:
@EdDev @qinqon if you insist I can do the update of the 1.20 conformance configuration in another PR.

But then please bear in mind that we will have to wait for another kubevirtci-publish and another kubevirtci bump against kubevirt. And with the seemingly flaky tests this could consume even more time.

As this is a test config change and not a functionality change, I think the benefit of doing this inside this PR far outweighs the additional work required for another PR. WDYT?

cluster-provision/k8s/1.20/conformance.json
dhiller: > There is no need to bump this one as well?
> I guess I miss the relation between the version in this manifest and the one in the conformance.sh script.

The version in the script is always the latest available at the time of the update. IIUC the version in the conformance config is the one that matters when executing the conformance tests.

EdDev (approved):
Thank you!

qinqon:
Thanks for the split
/approve

qinqon:
@dhiller the 1.20 PR merged https://github.com/kubevirt/kubevirtci/pull/598

PR #606, opened 2021-05-12
Istio, Add configuration for cluster without CNAO

rhrazdil:
Currently, Istio only works when CNAO is deployed at the same time.

This commit adds Istio operator configuration
for the case when CNAO is not deployed, improving
the usability of istio in kubevirtci.

Additionally, the wait loop is improved to wait not
just for the istio-cni-node daemonset to be present
but also for all it's pods to be running.

### DIFF
--- a/cluster-provision/k8s/1.21/istio.sh
+++ b/cluster-provision/k8s/1.21/istio.sh
@@ -8,7 +8,9 @@ export PATH=/opt/istio-$ISTIO_VERSION/bin:$PATH
 kubectl --kubeconfig /etc/kubernetes/admin.conf create ns istio-system
 istioctl --kubeconfig /etc/kubernetes/admin.conf operator init
 
-kubectl --kubeconfig /etc/kubernetes/admin.conf apply -f - <<EOF
+istio_manifests_dir=/opt/istio
+mkdir -p /opt/istio
+cat <<EOF >$istio_manifests_dir/istio-operator.tpl.yaml
 apiVersion: install.istio.io/v1alpha1
 kind: IstioOperator
 metadata:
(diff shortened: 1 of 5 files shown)


cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -46,4 +46,27 @@ function up() {
...
+            sleep 5
+            retries=$((retries + 1))
+            $kubectl -n $istio_cni_ns get daemonset $istio_cni_ds_name || continue
+            ready=$(kubectl -n $istio_cni_ns get daemonset $istio_cni_ds_name -o jsonpath="{.status.numberReady}")
+            required=$(kubectl -n $istio_cni_ns get daemonset $istio_cni_ds_name -o jsonpath="{.status.desiredNumberScheduled}")
+            if [[ $ready -eq $required ]]; then
+                break
+            fi
+        done
+    fi
```
qinqon, line 71: We should rely on IstioOperator status https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#InstallStatus

cluster-provision/k8s/1.21/istio.sh
```diff
@@ -31,17 +32,33 @@ spec:
...
+      jwtPolicy: first-party-jwt
+    cni:
+      chained: true
+      cniBinDir: /opt/cni/bin
+      cniConfDir: /etc/cni/net.d
+      excludeNamespaces:
+       - istio-system
+       - kube-system
+      logLevel: debug
+EOF
```
qinqon, line 64: What about patching the current one with 
```
sidecarInjectorWebhook:
      injectedAnnotations:
        "k8s.v1.cni.cncf.io/networks": istio-cni
```
instead of creating a new CR ?

qinqon (changes requested):

cluster-provision/k8s/1.21/istio.sh
rhrazdil: There are a few more differences
`      cniConfDir: /etc/cni/net.d` vs `      cniConfDir: /etc/cni/multus/net.d`
and 
`      chained: true` vs `      chained: false`

But yeah I thought about this. The thing is patching needs to be done in `cluster-up/cluster/k8s-provider-common.sh`, because I don't have access to `KUBEVIRT_WITH_CNAO` env var in `istio.sh`. I decided to keep the logic of `k8s-provider-common.sh` simple.

qinqon: I mean creating a pair of yamls but reduce duplication by patching and we can do the following

For `cniConfDir` and `chained` we can use envsubst

For the `sidecarInjectorWebhook` we can append at the end.

qinqon (changes requested):
We can still patch creating a pair of files.

rhrazdil:
Getting `error rendering spyglass page` when displaying the 1.20 job logs, but since I saw the provision pass, I suspect the conformance tests timed out.
/test check-provision-k8s-1.20

rhrazdil:
> Getting error `rendering spy glass page` when displaying the 1.20 job logs, but since I saw the provision pass, I suspect the conformance tests timed out.

two conformance tests failed, the retest didn't happen, trying againg

/test check-provision-k8s-1.20

PR #608, opened 2021-05-13
Enabling Prometheus operator for collecting metrics from e2e test runs

### LINKED ISSUE #1211: Design & Configuration of External Multiarch CI/CD K8s clusters
**Is your feature request related to a problem? Please describe**:
The problem is this: while preparing PRs for the s390x port of kubevirt and kubevirt CDI, there needs to be a CI/CD running on an s390x K8s cluster, available to github automation.  There is currently a provisional K8s cluster provisioned in the IBM cloud for this provisioned for kubevirt build and test and there will soon be a secure permanent platform provided.  

**Describe the solution you'd like**:
On behalf of ongoing efforts to multi-arch kubevirt and CDI which need this configured and documented, it is proposed that we have some official documentation for these external CI/CD build and test automation systems. 

**Describe alternatives you've considered**:
The alternative is to not document these systems?  Seems wrong. 

**Additional context**:

marceloamaral:
**What this PR does / why we need it:**
Added an option to gocli in kubevirtci to deploy Prometheus operator

Currently, kubvirtci does not deploy Prometheus. We need to configure the Prometheus operator and configure the service/pod monitors to scrape metrics. With Prometheus, we can enable Grafana dashboards to analyze the e2e performance tests.

**Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):**
Fixes #1211

**Special notes for your reviewer:**
The Prometheus operator is disabled by default and can be enabled via `KUBEVIRT_DEPLOY_PROMETHEUS="true"` and or, `gocli --enable-prometheus`

All the yamls here are based on the repository [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus), and I included the following components:
* [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator)
* [Prometheus](https://prometheus.io/)
* [Prometheus node-exporter](https://github.com/prometheus/node_exporter)
* [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics)
* [Alertmanager](https://github.com/prometheus/alertmanager) (optional)
* [Grafana](https://grafana.com/) (optional)

To make the deployment as light as possible, I made Grafana and  Alertmanager optional and disabled them by default, since it might be necessary only for the global Prometheus.
```
KUBEVIRT_DEPLOY_PROMETHEUS=${:-false}
KUBEVIRT_DEPLOY_GRAFANA=${:-false}
```


cluster-provision/k8s/1.17/manifests/prometheus/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+To configure KubeVirt to automatically creates a ServiceMonitor, we need to configure KubeVirt as follows:
+
+```
+apiVersion: kubevirt.io/v1alpha3
+kind: KubeVirt
+metadata:
+  name: kubevirt
+spec:
+  monitorNamespace: monitoring
```
rmohr, line 38: I wonder if we can make this discoverable. Right now kubevirt checks for `openshift-monitoring` by default (default monitoring namespaces for openshift). I think it makes a lot of sense if it also checks for the default namespace of https://github.com/prometheus-operator/kube-prometheus. I can prepare a PR for that. That would make additional configuration steps obsolete.

rmohr (commented):
Very nice and straight forward implementation :)

A few high-level points (see details inline):

 * I think it is enough if you only add the manifests to the latest kubevirt provider (1.21)· If you have a specific need for older versions that is fine too, but in general we add new features only to the latest providers.
 * Replica configuration can be removed
 * I think we can make kube-prometheus fully discoverable. Checking what we can do there.

rmohr:
> * I think we can make kube-prometheus fully discoverable. Checking what we can do there.

https://github.com/kubevirt/kubevirt/pull/5652 should make any extra configuration on the `KubeVirt` CR obsolete.

cluster-provision/gocli/cmd/run.go
```diff
@@ -79,10 +79,17 @@ func NewRunCommand() *cobra.Command {
...
 	run.Flags().Uint("k8s-port", 0, "port on localhost for the k8s cluster")
 	run.Flags().Uint("ssh-port", 0, "port on localhost for ssh server")
+	run.Flags().Uint("prometheus-port", 0, "port on localhost for prometheus server")
+	run.Flags().Uint("grafana-port", 0, "port on localhost for grafana server")
 	run.Flags().String("nfs-data", "", "path to data which should be exposed via nfs to the nodes")
 	run.Flags().String("log-to-dir", "", "enables aggregated cluster logging to the folder")
 	run.Flags().Bool("enable-ceph", false, "enables dynamic storage provisioning using Ceph")
 	run.Flags().Bool("enable-istio", false, "deploys Istio service mesh")
+	run.Flags().Bool("enable-prometheus", false, "deploys Prometheus operator")
+	run.Flags().UintP("prometheus-replicas", "x", 1, "number of prometheus replicas to start")
```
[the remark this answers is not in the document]
marceloamaral: ok, I can remove that.
But the motivation to have this option is for a large-scale scenario for scalability tests. It might be that in a scenario with 1k nodes,  a single prometheus point can become a bottleneck. I don't know if a single instance can sustain such a scenario...

cluster-provision/k8s/1.17/manifests/prometheus/README.md
marceloamaral: That sounds great. I this case, I will remove this info from the README.

cluster-provision/gocli/cmd/run.go
```diff
@@ -79,10 +79,17 @@ func NewRunCommand() *cobra.Command {
...
+	run.Flags().Bool("enable-prometheus-alertmanager", false, "deploys Prometheus alertmanager")
+	run.Flags().UintP("prometheus-alertmanager-replicas", "y", 1, "number of prometheus alertmanager replicas to start")
```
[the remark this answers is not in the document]
marceloamaral: Same as before. I can remove it, but I think it might be useful for large-scale tests...

fgimenez (commented):
Looks very good, I wonder why do we need to spin up Prometheus operator on the kubevirtci clusters? They are ephemeral clusters, the configuration is known beforehand and won't change during the lifecycle of the CI cluster. Why not using plain Prometheus with kubernetes service discovery instead? It would be much simpler and more efficient in terms of resource usage.

marceloamaral:
@fgimenez thanks for the comments!
imho, Prometheus Operator (PO) itself is very light. 
For me, the key advantage to use PO instead of plain prometheus is the ServiceMonitor, which is used by KubeVirt to configure prometheus to scrape KubeVirt-related metrics.

cluster-provision/gocli/cmd/run.go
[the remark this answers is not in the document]
rmohr: Realistically we can't create more than a handful of nodes. More than 100 or 200 is not realistic. `kubevirtci` is bound to a single underlying machine. Further I think that you would need sharding for that. As far as I understand kube-prometheus, this would only result in a more highly-available setup because all replicas would scrape the same targets.

rmohr:
> @fgimenez thanks for the comments!
> imho, Prometheus Operator (PO) itself is very light.
> For me, the key advantage to use PO instead of plain prometheus is the ServiceMonitor, which is used by KubeVirt to configure prometheus to scrape KubeVirt-related metrics.

In theory @fgimenez outlines the best approach to not disturb test workloads, but given the practical simplicity of this solution (and also considering that we run other more heavy-weight components like ceph on a need-to bases inside the cluster too), I think this prometheus-operator solution is just good enough and simple.

fgimenez:
Thanks for your response @marceloamaral and @rmohr ,

> imho, Prometheus Operator (PO) itself is very light.

It is light indeed, but we will need to run it for each CI cluster on the nodes, this would multiply the resources used.

> For me, the key advantage to use PO instead of plain prometheus is the ServiceMonitor, which is used by KubeVirt to configure prometheus to scrape KubeVirt-related metrics.

But ServiceMonitors are created only if the CRD is available (if PO is installed), right? As I see it, for this specific scenario (ephemeral clusters, no changes in the monitoring components deployed, no unexpected services deployed) kubernetes SD in plain prometheus gives you all the flexibility required.

> In theory @fgimenez outlines the best approach to not disturb test workloads, but given the practical simplicity of this solution (and also considering that we run other more heavy-weight components like ceph on a need-to bases inside the cluster too). I think this prometheus-operator solution is just good enough and simple.

Yes, in terms of how performance can be affected,maybe this operator is not the same as components like ceph. I worry also about the maintenance over time, when we need to update a high number of operator-related manifests instead of a single one/

rmohr:
> But ServiceMonitors are created only if the CRD is available (if PO is installed), right? As I see it, for this specific scenario (ephemeral clusters, no changes in the monitoring components deployed, no unexpected services deployed) kubernetes SD in plain prometheus gives you all the flexibility required.

Yes, the CRDs are part of the deployment. `kubevirt` will detect prometheus-operator and  configure itself properly.
The alternative is not entirely clear to me. We would somehow proxy a prometheus instance into each kubevirtci cluster, including DNS resolution, or am I missing something?

> Yes, in terms of how performance can be affected,maybe this operator is not the same as components like ceph. I worry also about the maintenance over time, when we need to update a high number of operator-related manifests instead of a single one/

I don't think this adds much. The general rule in kubevirtci is that we only have the last three supported k8s versions, and there we only updated manifests on latest providers, except if there is an issue which we need to fix. I would not really expect that we update it often for as long as it works.

fgimenez:
> The alternative is not entirely clear to me. We would somehow proxy a prometheus instance into each kubevirtci cluster, including DNS resolution, or am I missing something?

I think it would be easier to run just plain Prometheus in the CI clusters and configure its kubernetes service discovery to watch the services/pods/endpoints created on the kubevirt namespace. Details about this SD here https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config

> I don't think this adds much. The general rule in kubevirtci is that we only have the last three supported k8s versions, and there we only updated manifests on latest providers, except if there is an issue which we need to fix. I would not really expect that we update it often for as long as it works.

Yes, the alternative would be maintaining Prometheus-related core kubernetes objects instead of Prometheus Operator manifests.

cluster-provision/gocli/cmd/run.go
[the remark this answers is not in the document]
rmohr: Oh, just ensure that you just replace all `replicas` occurences with `1` when applying the manifests. We really don't need any redundancy and it just consumes resources.

cluster-provision/k8s/1.21/manifests/prometheus/README.md
```diff
@@ -0,0 +1,28 @@
+# Prometheus
```
rmohr, line 1: Can you add documentation on how to update it?
I would expect that this is a simply copy instruction of a subfolder from the other github repo.

pack8s/cmd/run.go
```diff
@@ -24,25 +24,32 @@ import (
 )
 
 type runOptions struct {
-	privileged     bool
```
rmohr, line 27: pack8s is not really maintained or used. `gocli` now also works with latest `podman` and it s docker-api support. We should remove `pack8s`.

rmohr:
> I think it would be easier to run just plain Prometheus in the CI clusters and configure its kubernetes service discovery to watch the services/pods/endpoints created on the kubevirt namespace. Details about this SD here https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config

Ok, so you would run it inside the cluster. Then I would be even more for `prometheus-operator`. I expect this to be a simple copy all manifests from `kube-prometheus` to a place here to update and it gives us all the components which we need also for e2e testing of our alertmanager an prometheus-operator integration in kubevirt/kubevirt.

I think that this simple integration still requires quite some manifests to get everything connected (yes much less, but we have to write it, and then we have less flexibility regarding to testing, and the operator is really not that heavy).

fgimenez:
> I expect this to be a simple copy all manifests from `kube-prometheus` to a place here to update and it gives us all the components which we need also for e2e testing of our alertmanager an prometheus-operator integration in kubevirt/kubevirt.

From a pure CI perspective, I would prefer to have e2e tests with proper separation of concerns to test specific features like the Prometheus Operator integration (or for that matter any other specific feature), but if this integration is not yet tested I agree that it is a good reason to use Prometheus Operator instead of plain Prometheus inside the CI clusters.

cluster-provision/k8s/1.21/manifests/prometheus/README.md
marceloamaral: Do you mean update the Prometheus Operator to a newer release? Or update the configuration?
If it is related to update to a newer release, I did a quick search and did not find any documentation about that. So, it might be done using `kubectl patch or replace`.

pack8s/cmd/run.go
marceloamaral: Humm, ok, I will revert all changes in pack8s

cluster-provision/k8s/1.21/manifests/prometheus/README.md
rmohr: I just mean, if someone goes in and wants to replace the manifests here with the latest one from `kube-prometheus`. Did you do some extra modifications, or simply copy-paste them?

marceloamaral:
I am also more to the `prometheus-operator` approach since KubeVirt already have implemented the logic to expose metrics. Implementing our own logic to expose and scrape the KubeVirt components might be harder to maintain in the future...

fgimenez:
> I am also more to the `prometheus-operator` approach since KubeVirt already have implemented the logic to expose metrics. Implementing our own logic to expose and scrape the KubeVirt components might be harder to maintain in the future...

In fact no logic should be implemented, just enabling kubernetes SD should be enough. But yes, given that the Prometheus Operator integration is in place I agree that it is ok to exercise it.

cluster-provision/gocli/cmd/run.go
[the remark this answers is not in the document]
kwiesmueller: @marceloamaral Prometheus does not scale with replicas it just becomes more highly available.
So adding more replicas would only protect from one instance going down which if it happens in our scenario means something else is wrong or will go wrong.
Instead we just create more strain on the pods to be scraped and use twice the resources for prometheus.

cluster-provision/gocli/cmd/run.go
```diff
@@ -177,6 +186,31 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
...
+	if err != nil {
+		return err
+	}
+
+	prometheusAlertmanagerReplicas, err := cmd.Flags().GetUint("prometheus-alertmanager-replicas")
+	if err != nil {
+		return err
+	}
+
+	grafanaEnabled, err := cmd.Flags().GetBool("enable-grafana")
```
kwiesmueller, line 209: I didn't expect a Grafana instance inside the CI environment.
Not sure if I mind, but I thought the internal setup is just meant for exporting the metrics and the Grafana we'd use is an external one.
Having the option for testing manually is probably good though, I'd just not enable it for automated tests.

cluster-provision/k8s/1.21/manifests/prometheus/grafana/grafana-dashboardDatasources.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+data:
+  datasources.yaml: ewogICAgImFwaVZlcnNpb24iOiAxLAogICAgImRhdGFzb3VyY2VzIjogWwogICAgICAgIHsKICAgICAgICAgICAgImFjY2VzcyI6ICJwcm94eSIsCiAgICAgICAgICAgICJlZGl0YWJsZSI6IGZhbHNlLAogICAgICAgICAgICAibmFtZSI6ICJwcm9tZXRoZXVzIiwKICAgICAgICAgICAgIm9yZ0lkIjogMSwKICAgICAgICAgICAgInR5cGUiOiAicHJvbWV0aGV1cyIsCiAgICAgICAgICAgICJ1cmwiOiAiaHR0cDovL3Byb21ldGhldXMtazhzLm1vbml0b3Jpbmcuc3ZjOjkwOTAiLAogICAgICAgICAgICAidmVyc2lvbiI6IDEKICAgICAgICB9CiAgICBdCn0=
```
kwiesmueller, line 3: If you created this by hand you could use `stringData` instead of `data` so the content is not encoded in base64 and allows easier tracking in git.

cluster-provision/gocli/cmd/run.go
marceloamaral: Right, because of that grafana is disabled by default

cluster-provision/k8s/1.21/manifests/prometheus/grafana/grafana-dashboardDatasources.yaml
marceloamaral: I did not change the grafana yamls, they are all from the official kube-prometheus repository

cluster-provision/k8s/1.21/manifests/prometheus/README.md
marceloamaral: I see, the only change that I did was to decrease the number of replicas from 2 to 1.
I will write that in the README

cluster-provision/gocli/cmd/run.go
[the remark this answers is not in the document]
marceloamaral: It makes sense thanks!
Actually, I have already decreased all the replicas from 2 to 1. But I will remove the option to increase it

cluster-provision/gocli/cmd/run.go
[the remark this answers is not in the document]
marceloamaral: Thanks @kwiesmueller! I will remove the option to increase it

rmohr:
@fgimenez do we have an agreement so far that we continue with prometheus-operator inside kubevirtci? If it durns out to not work as well as expected, I have no issues iterating on it later on.

dhiller (commented):
Sorry for being late to the game already. Forgive me my ignorance, but I don't get why I find all the manifest file once inside the 1.21 provider folder, and the same (?) files in the parent folder? Also I found the pre-pulling of the images at all providers, while the `prometheus.sh` was found only in the 1.21 folder?

In general I was expecting to find the `prometheus.sh` shell script and the prometheus operator manifests in every provider folder that should get the ability to have it (side note: we are going to drop 1.17 and 1.18 soon, so no need to do that there)

Also we might enable prometheus install in [check-cluster-up.sh](https://github.com/kubevirt/kubevirtci/blob/master/cluster-provision/k8s/check-cluster-up.sh#L36) and look at the pods so that we can check whether everything works.

cluster-provision/k8s/1.17/provision.sh
```diff
@@ -355,6 +355,17 @@ docker_pull_retry k8s.gcr.io/sig-storage/csi-provisioner:v2.0.0
 docker_pull_retry k8s.gcr.io/sig-storage/csi-snapshotter:v3.0.0
 docker_pull_retry k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.0.1
 
+# Pre pull images used in Prometheus operator
+docker_pull_retry quay.io/prometheus-operator/prometheus-operator:v0.47.0
+docker_pull_retry quay.io/prometheus/prometheus:v2.26.0
+docker_pull_retry quay.io/brancz/kube-rbac-proxy:v0.8.0
+docker_pull_retry quay.io/prometheus/alertmanager:v0.21.0
+docker_pull_retry quay.io/brancz/kube-rbac-proxy:v0.8.0
+docker_pull_retry quay.io/prometheus/alertmanager:v0.21.0
```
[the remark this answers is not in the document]
marceloamaral: Nice catch @dhiller!
In the beginning I added prometheus operator to all k8s versions, but then we decided to support it only in 1.21.

marceloamaral:
Nice catch @dhiller!
In the beginning I added prometheus operator to all provider folders. But, I was told that, in general, new features are only added to the latest provider. Then, I removed the manifest but forgot to remove the pulling images.
I will update the check-cluster-up.sh

fgimenez:
/lgtm

> @fgimenez do we have an agreement so far that we continue with prometheus-operator inside kubevirtci? If it durns out to not work as well as expected, I have no issues iterating on it later on.

Absolutely, sorry for not being clear in my previous comments :)

marceloamaral:
Thanks @fgimenez!

cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -77,6 +77,22 @@ function _add_common_params() {
         params=" --enable-ceph $params"
     fi
 
+    if [[ $KUBEVIRT_DEPLOY_PROMETHEUS == "true" ]] &&
```
dhiller, line 80: Not sure, maybe we should check what provider is used here also, and fail if it doesn't support prometheus, like i.e. >= 1.21?

marceloamaral: I think it makes sense!
Not sure what is better: fail or echo a warn?

rmohr (approved):
/lgtm

rmohr:
Looks like we run into the image pull limitations of docker inside the VMs. That is the reason for calicos imagepull errors.

PR #610, opened 2021-05-18
Remove pack8s and document how to use kubevirtci with podman

### LINKED ISSUE #569: drop pack8s usage
According to its author @fromanirh, pack8s was outclassed by podman itself, which now supports the docker rest API; TL;DR: he suggests to drop it and just switch to newer podman

rmohr:
kubevirtci works natively with podman 3.1+, which makes pack8s obsolete and document how to use kubevirtci with podman.

Fixes #569.


PODMAN.md
```diff
@@ -0,0 +1,10 @@
+# Use kubevirtci with podman instead of docker
+
+Install podman 3.1+, then run it in docker compatible mode:
```
kwiesmueller, line 3: Maybe add that the docker-cli still needs to be there or at least a symlink to podman?
Or is that also no longer the case?

kwiesmueller: And how is the compatibility?
Will it also work to run podman without sudo and point to `unix:///home/jdoe/docker.sock` without breaking anything?
If yes that's what I'd suggest more than the sudo podman.

rmohr: The `docker-cli` should not be needed anymore. Also addressed the `sudo` requirement and documented another required socket location as well as adding the info that `sudo` is not only needed if better storage performance is a must.

fromanirh (commented):
/hold
/lgtm
good riddance to `pack8s`. Podman 3.1+ is just better nowadays.
holding to give time to other reviewers.

cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -14,18 +14,25 @@ if [ -n "${KUBEVIRTCI_TAG}" ] && [ -n "${KUBEVIRTCI_GOCLI_CONTAINER}" ]; then
...
 fi
 
-if [ "${KUBEVIRTCI_RUNTIME}" = "podman" ]; then
-    _cli="pack8s"
+if docker ps >/dev/null; then
+    _cri_bin=docker
+    echo "selecting docker as container runtime"
+elif podman ps >/dev/null; then
+    _cri_bin=podman
+    echo "selecting podman as container runtime"
```
qinqon: Can we keep an option for users to explicitly set cri checking if KUBEVIRTCI_RUNTIME has being set ?

cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -14,18 +14,25 @@ if [ -n "${KUBEVIRTCI_TAG}" ] && [ -n "${KUBEVIRTCI_GOCLI_CONTAINER}" ]; then
...
-    if [ -d /lib/modules ]; then
-        _cli="${_cli} -v /lib/modules/:/lib/modules/"
-    fi
-    _cli="${_cli} ${_cli_container}"
+    echo "no working container runtime found. Neither docker nor podman seems to work."
+    exit 1
+fi
+
+_cli_container="${KUBEVIRTCI_GOCLI_CONTAINER:-quay.io/kubevirtci/gocli:${KUBEVIRTCI_TAG}}"
+_cli="${_cri_bin} run --privileged --net=host --rm ${USE_TTY} -v /var/run/docker.sock:/var/run/docker.sock"
```
qinqon: Don't we have to consume `unix:///${HOME}/podman.sock` here ?

cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -14,18 +14,25 @@ if [ -n "${KUBEVIRTCI_TAG}" ] && [ -n "${KUBEVIRTCI_GOCLI_CONTAINER}" ]; then
     echo "WARNING: KUBEVIRTCI_GOCLI_CONTAINER is set and will take precedence over the also set KUBEVIRTCI_TAG"
 fi
 
-if [ "${KUBEVIRTCI_RUNTIME}" = "podman" ]; then
-    _cli="pack8s"
+if docker ps >/dev/null; then
+    _cri_bin=docker
+    echo "selecting docker as container runtime"
+elif podman ps >/dev/null; then
+    _cri_bin=podman
```
qinqon, line 22: What do you think about doing some sanity checks here with the stuff documented, like podman version, rest socket checks, since it's already at the documentation is not a must.

qinqon (changes requested):
Pair of suggestion and question relatd to the socket.

cluster-up/cluster/ephemeral-provider-common.sh
rmohr: Can it be that you reviewed an old version?

cluster-up/cluster/ephemeral-provider-common.sh
rmohr: That is a good idea.

cluster-up/cluster/ephemeral-provider-common.sh
rmohr: Yes.

cluster-up/cluster/ephemeral-provider-common.sh
qinqon: Yep, you can resolve this.

cluster-up/cluster/ephemeral-provider-common.sh
rmohr: Done.

cluster-up/cluster/ephemeral-provider-common.sh
rmohr: Done.

rmohr:
@kwiesmueller @qinqon please take another look.

cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -15,17 +15,34 @@ if [ -n "${KUBEVIRTCI_TAG}" ] && [ -n "${KUBEVIRTCI_GOCLI_CONTAINER}" ]; then
...
 else
-    _cli_container="${KUBEVIRTCI_GOCLI_CONTAINER:-quay.io/kubevirtci/gocli:${KUBEVIRTCI_TAG}}"
-    _cli="docker run --privileged --net=host --rm ${USE_TTY} -v /var/run/docker.sock:/var/run/docker.sock"
-    # gocli will try to mount /lib/modules to make it accessible to dnsmasq in
-    # in case it exists
-    if [ -d /lib/modules ]; then
-        _cli="${_cli} -v /lib/modules/:/lib/modules/"
-    fi
-    _cli="${_cli} ${_cli_container}"
+	if curl --unix-socket /${HOME}/podman.sock http://d/v3.0.0/libpod/info > /dev/null 2>&1 ; then
```
kwiesmueller, line 24: The path to the sock used could be a variable we can overwrite, but I don't think it's necessary now.
/lgtm
otherwise

rmohr:
Fails on an unrelated test which fails on other prs too.

PR #615, opened 2021-05-27
Add a specific label to prometheus

marceloamaral:
Signed-off-by: Marcelo Amaral <marcelo.amaral1@ibm.com>

In order to configure Prometheus Federation, we need to add a specific label to prometheus to enable it to be found via PodMonitor.


marceloamaral:
Obsolete.

PR #620, opened 2021-06-17
Fix 1.21 provider: Move fetch-images.sh into provision folder

dhiller:
Instead of copying and removing the script we put it directly into the
folder where it belongs. Also we adjust pathes and the README.

Original setup causes this error on provisioning within `publish.sh`:
```
ERROR: fetch-images.sh missing
+ '[' '!' -f /tmp/extra-pre-pull-images ']'
+ '[' '!' -f /tmp/fetch-images.sh ']'
+ echo 'ERROR: fetch-images.sh missing'
+ exit 1
```

publish failed here: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/publish-kubevirtci/1405529938345332736

### DIFF
--- a/cluster-provision/k8s/1.21/fetch-images.sh
+++ b/cluster-provision/k8s/1.21/fetch-images.sh
@@ -2,9 +2,7 @@
 
 set -euo pipefail
 
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-
-function usage {
+function usage() {
     cat <<EOF
 Usage: $0 <k8s-cluster-dir> [source-image-list]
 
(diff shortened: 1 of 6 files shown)


dhiller:
Strange, retest didn't work. Didn't use test all as I didn't want to restart 1.21 check.

dhiller:
Ah, looking at prow job history, it actually did work, looks like the pods couldn't get scheduled, so no signal to GitHub was emitted....

dhiller:
@fgimenez @rmohr FYI: the post submit does not work any more as described in this PR description. PTAL, thanks!

fgimenez (approved):
/approve
/hold

Great, unhold when ready.

BTW, I see some references to shellcheck, very nice tool. Maybe we could add some CI jobs to execute it against our scripts.

PR #624, opened 2021-06-22
Use alternative registry on conformance tests

ormergi:
When running kubernetes e2e conformance tests,
Sonobuoy pull images from docker.io by default.
Due to that CI check-up lanes that run conformance
tests hit docker pull images rate limit and fail.

Following https://github.com/vmware-tanzu/sonobuoy/pull/1194/
this PR points Sonobuoy to pull images
from 'mirror.gcr.io/library' registry instead of docker.io.

### DIFF
--- a/hack/conformance.sh
+++ b/hack/conformance.sh
@@ -2,10 +2,16 @@
 
 set -xeuo pipefail
 
+SCRIPT_PATH=$(dirname "$(realpath "$0")")
+
 ARTIFACTS=${ARTIFACTS:-${PWD}}
+
 config_file=${1:-}
 sonobuoy_version=0.50.0
 [[ -f "$config_file" ]] && sonobuoy_version=$(jq -r '.Version' "$config_file" | grep -oE '[0-9\.]+')
+
+conformance_image_config_file="$SCRIPT_PATH/confromance-image-config.yaml"
+! [[ -f "$conformance_image_config_file" ]] && echo "FATAL: Conformance image config file does not exists" 1>&2 && exit 1
 
 if [[ -z "$KUBEVIRT_PROVIDER" ]]; then
     echo "KUBEVIRT_PROVIDER is not set" 1>&2
@@ -59,7 +65,7 @@ curl -L "https://github.com/vmware-tanzu/sonobuoy/releases/download/v${sonobuoy_
 
 trap teardown EXIT
 
-run_cmd="./sonobuoy run --wait"
+run_cmd="./sonobuoy run --wait --e2e-repo-config $conformance_image_config_file"
 
 if [ "$config_file" != "" ]; then
     run_cmd+=" --config $config_file"
--- a/hack/confromance-image-config.yaml
+++ b/hack/confromance-image-config.yaml
@@ -1,0 +1,1 @@
+dockerLibraryRegistry: mirror.gcr.io/library


hack/conformance.sh
```diff
@@ -2,11 +2,17 @@
...
+SCRIPT_PATH=$(dirname "$(realpath "$0")")
+
 ARTIFACTS=${ARTIFACTS:-${PWD}}
+
 config_file=${1:-}
 sonobuoy_version=0.50.0
 [[ -f "$config_file" ]] && sonobuoy_version=$(jq -r '.Version' "$config_file" | grep -oE '[0-9\.]+')
 
+conformance_image_config_file="$SCRIPT_PATH/conformance-config-file.yaml"
+! [[ -f "$conformance_image_config_file" ]] && echo "FATAL: Conformance image config file does not exists" &>2 && return 1
```
rmohr, line 14:
```
hack/conformance.sh: line 14: return: can only `return' from a function or sourced script
```

Can it be that you wanted to write `exit 0`?

ormergi: Yes, sorry I will fix it :see_no_evil:

hack/confromance-image-config.yaml
```diff
@@ -0,0 +1 @@
+dockerLibraryRegistry: mirror.gcr.io/library
```
dhiller, line 1: Our image proxy would also be able to act as a registry, right @fgimenez? Could we maybe use that when `"${CI}" == "true"`?

hack/conformance.sh
```diff
@@ -2,11 +2,17 @@
...
 
+SCRIPT_PATH=$(dirname "$(realpath "$0")")
+
 ARTIFACTS=${ARTIFACTS:-${PWD}}
+
 config_file=${1:-}
 sonobuoy_version=0.50.0
 [[ -f "$config_file" ]] && sonobuoy_version=$(jq -r '.Version' "$config_file" | grep -oE '[0-9\.]+')
 
+conformance_image_config_file="$SCRIPT_PATH/conformance-config-file.yaml"
```
dhiller, line 13:
```suggestion
conformance_image_config_file="$SCRIPT_PATH/confromance-image-config.yaml"
```

hack/confromance-image-config.yaml
fgimenez: hmm I don't think so, we set our proxy in the daemon's config and it acts transparently to the clients using it, they keep requesting images from dockerhub, quay, etc. For this to work we need additionally to setup the proxy with a concrete CA and add it as a trusted CA to the hosts where the damons run. All this is explained here https://github.com/rpardini/docker-registry-proxy#configuring-the-docker-clients--kubernetes-nodes--linux-clients

hack/conformance.sh
ormergi: Sorry my bad
Done

ormergi:
I verified on a local setup that Kubernetes conformance test actually pull the images from `mirror.gcr.io/library`
by checking the image tag of one of the pods that the test creates:
```
 kubectl get po -A -o yaml | grep "image.*gcr"
    - image: mirror.gcr.io/library/httpd:2.4.38-alpine
      image: mirror.gcr.io/library/httpd:2.4.38-alpine
...
```

hack/confromance-image-config.yaml
ormergi: I think it should be the default because we can hit docker pull rate limit easily also on a local setup

rmohr: It would be great if we would add a way to inject the proxy settings in CI. Let's do that in a follow-up.

rmohr (approved):
/lgtm

let's see what CI thinks.

ormergi:
I think it failed due to docker pull rate limits when pulling calico image, we have no informative error message though

ormergi:
> /approve

Thank you :slightly_smiling_face:

PR #629, opened 2021-06-25
Configure cgroup v2 settings on cluster-up

vasiliy-ul:
This is a follow-up of the https://github.com/kubevirt/kubevirtci/pull/593. Now it is possible to pass additional kernel arguments to qemu and that allows moving the configuration of cgroup v2 from the proviosion step to cluster-up. That way cgroup v2 testing lanes can reuse the existing "pure" 1.20 and 1.21 images.

Th PR does the following:
* adds `systemd.unified_cgroup_hierarchy=1` to qemu kernel args
* configures cgroup settings during VM boot in `node01.sh/nodes.sh`
* removes cgroup configuration during provision step

### DIFF
--- a/cluster-provision/gocli/cmd/provision.go
+++ b/cluster-provision/gocli/cmd/provision.go
@@ -41,7 +41,6 @@ func NewProvisionCommand() *cobra.Command {
 	provision.Flags().Bool("random-ports", false, "expose all ports on random localhost ports")
 	provision.Flags().Uint("vnc-port", 0, "port on localhost for vnc")
 	provision.Flags().Uint("ssh-port", 0, "port on localhost for ssh server")
-	provision.Flags().Bool("cgroupv2", false, "set UNIFIED_CGROUP_HIERARCHY environment variable for the provision script")
 	provision.Flags().String("container-suffix", "", "use additional suffix for the provisioned container image")
 	provision.Flags().StringArray("additional-persistent-kernel-arguments", []string{}, "additional persistent kernel arguments applied after provision")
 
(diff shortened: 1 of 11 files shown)


rmohr:
/lgtm
/approve

Looks great.

rmohr:
@vasiliy-ul I think the developer build from kubevirt was broken. It should be fixed with https://github.com/kubevirt/kubevirt/pull/5951. Let's see if the kubevirt intallation works again now.

vasiliy-ul:
> @vasiliy-ul I think the developer build from kubevirt was broken. It should be fixed with [kubevirt/kubevirt#5951](https://github.com/kubevirt/kubevirt/pull/5951). Let's see if the kubevirt intallation works again now.

Ah, okay. I though it was a random failure. Locally the provision check works fine for me.

vasiliy-ul:
hm... does not look like a random failure now:
```
No match for argument: kernel-modules-4.18.0-315.el8.x86_64
Error: Unable to find a match: kernel-modules-4.18.0-315.el8.x86_64
```
locally I was not able to reproduce it...

rmohr:
> hm... does not look like a random failure now:
> 
> ```
> No match for argument: kernel-modules-4.18.0-315.el8.x86_64
> Error: Unable to find a match: kernel-modules-4.18.0-315.el8.x86_64
> ```
> 
> locally I was not able to reproduce it...

That is indeed interesting. It seems to exist: https://centos.pkgs.org/8-stream/centos-baseos-x86_64/kernel-modules-4.18.0-315.el8.x86_64.rpm.html

vasiliy-ul:
yeah, I also found it in the repo. Now running only 1.19 and looks like that point has passed already...

oshoval:
@vasiliy-ul 
due to the discussion on https://github.com/kubevirt/project-infra/pull/1397
it might be worth to /hold this PR so it won't merged? (as this PR should be changed)

vasiliy-ul:
@oshoval , I was thinking about handling it in a separate PR. Now the cgroupv2 lanes are optional and all the mandatory ones are checked anyway. But if you think its better to address it here I dont have objections.

oshoval:
Up to you, we just usually try to combine the feature + tests in the same PR
Thanks

PR #630, opened 2021-06-25
Switch the default to k8s-1.20

maya-r:
All kubevirt projects are attempting to handle k8s 1.22 deprecations,
and some of the newer versions are not yet available in k8s 1.18,
like networking.k8s.io/v1 (available since 1.19), causing weird
failures to people using the default values.


dhiller (approved):
/approve

Thanks, nice catch!

PR #631, opened 2021-06-25
Calico interface regex

rhrazdil:
Currently, kubevirtci uses Calico's default IP auto-
detection method: first-found.

The disadvantage of this method is that it makes only simplified
guess and on top of that, when new interfaces are added (bridges, veths,
etc.) calico may decide to change the NODE IP, which breaks BGP updates.

This method is not recommended by the documentation either [0].

It's quite simple to reproduce this, just creating a bridge on a node
and assigning an IP address to it makes calico-node pod to pick up the
assigned address and start using that address for BGP.

Since this is certainly not a desired behaviour, this commit adds
IP_AUTODETECTION_METHOD=interface=eth.*
IP6_AUTODETECTION_METHOD=interface=eth.*
setting to only consider interfaces called eth0, eth1, etc. for IP
address autodetection.

[0] https://docs.projectcalico.org/reference/node/configuration#ip-setting

### DIFF
--- a/cluster-provision/k8s/1.20/manifests/cni.diff
+++ b/cluster-provision/k8s/1.20/manifests/cni.diff
@@ -1,5 +1,5 @@
---- a/cluster-provision/k8s/1.19/manifests/cni.do-not-change.yaml
-+++ b/cluster-provision/k8s/1.19/manifests/cni.do-not-change.yaml
+--- a/cluster-provision/k8s/1.20/manifests/cni.do-not-change.yaml
++++ b/cluster-provision/k8s/1.20/manifests/cni.do-not-change.yaml
 @@ -32,7 +32,12 @@ data:
            "nodename": "__KUBERNETES_NODE_NAME__",
            "mtu": __CNI_MTU__,
--- a/cluster-provision/k8s/1.20/manifests/cni.diff
+++ b/cluster-provision/k8s/1.20/manifests/cni.diff
@@ -14,6 +14,18 @@
            },
            "policy": {
                "type": "k8s"
+@@ -3642,6 +3642,11 @@ spec:
+             # Auto-detect the BGP IP address.
+             - name: IP
+               value: "autodetect"
++            # Use IP/IPv6 addresses on ifaces eth0, eth1, eth2, etc.
++            - name: IP_AUTODETECTION_METHOD
++              value: "interface=eth.*"
++            - name: IP6_AUTODETECTION_METHOD
++              value: "interface=eth.*"
+             # Enable IPIP
+             - name: CALICO_IPV4POOL_IPIP
+               value: "Always"
 @@ -3671,6 +3676,8 @@ spec:
              # no effect. This should fall within `--cluster-cidr`.
              # - name: CALICO_IPV4POOL_CIDR
--- a/cluster-provision/k8s/1.21/manifests/cni.diff
+++ b/cluster-provision/k8s/1.21/manifests/cni.diff
@@ -1,5 +1,5 @@
---- a/cluster-provision/k8s/1.19/manifests/cni.do-not-change.yaml
-+++ b/cluster-provision/k8s/1.19/manifests/cni.do-not-change.yaml
+--- a/cluster-provision/k8s/1.21/manifests/cni.do-not-change.yaml
++++ b/cluster-provision/k8s/1.21/manifests/cni.do-not-change.yaml
 @@ -32,7 +32,12 @@ data:
            "nodename": "__KUBERNETES_NODE_NAME__",
            "mtu": __CNI_MTU__,
--- a/cluster-provision/k8s/1.21/manifests/cni.diff
+++ b/cluster-provision/k8s/1.21/manifests/cni.diff
@@ -14,6 +14,18 @@
            },
            "policy": {
                "type": "k8s"
+@@ -3642,6 +3642,11 @@ spec:
+             # Auto-detect the BGP IP address.
+             - name: IP
+               value: "autodetect"
++            # Use IP/IPv6 addresses on ifaces eth0, eth1, eth2, etc.
++            - name: IP_AUTODETECTION_METHOD
++              value: "interface=eth.*"
++            - name: IP6_AUTODETECTION_METHOD
++              value: "interface=eth.*"
+             # Enable IPIP
+             - name: CALICO_IPV4POOL_IPIP
+               value: "Always"
 @@ -3671,6 +3676,8 @@ spec:
              # no effect. This should fall within `--cluster-cidr`.
              # - name: CALICO_IPV4POOL_CIDR


cluster-provision/k8s/1.20/manifests/cni.diff
```diff
@@ -14,6 +14,18 @@
            },
            "policy": {
                "type": "k8s"
+@@ -3642,6 +3642,11 @@ spec:
+             # Auto-detect the BGP IP address.
+             - name: IP
+               value: "autodetect"
++            # Use IP/IPv6 addresses on ifaces eth0, eth1, eth2, etc.
++            - name: IP_AUTODETECTION_METHOD
++              value: "interface=eth.*"
```
maiqueb, line 23: I wonder if we want to limit this more - i.e. `[0-9]*` . I don't know.

cluster-provision/k8s/1.20/manifests/cni.diff
```diff
@@ -14,6 +14,18 @@
...
++            - name: IP6_AUTODETECTION_METHOD
++              value: "interface=eth.*"
+             # Enable IPIP
+             - name: CALICO_IPV4POOL_IPIP
+               value: "Always"
```
maiqueb, line 28: And this one? Why are you updating it ?

cluster-provision/k8s/1.21/manifests/cni.diff
```diff
@@ -14,6 +14,18 @@
...
+             - name: IP
+               value: "autodetect"
++            # Use IP/IPv6 addresses on ifaces eth0, eth1, eth2, etc.
++            - name: IP_AUTODETECTION_METHOD
++              value: "interface=eth.*"
++            - name: IP6_AUTODETECTION_METHOD
++              value: "interface=eth.*"
+             # Enable IPIP
+             - name: CALICO_IPV4POOL_IPIP
+               value: "Always"
```
maiqueb, line 28: ditto

cluster-provision/k8s/1.20/manifests/cni.diff
```diff
@@ -1,5 +1,5 @@
---- a/cluster-provision/k8s/1.19/manifests/cni.do-not-change.yaml
```
maiqueb, line 1: I honestly don't understand this change

rhrazdil: It seems like this file was copied from older k8s provision without changing the path.
You can see that this `cni.diff` file is in `cluster-provision/k8s/1.20/manifests`, but it's applied to older
`a/cluster-provision/k8s/ => 1.19 <= /manifests/cni.do-not-change.yaml`.

`cni.do-not-change.yaml` hasn't changed between versions, but still, the patch should be applied to the right manifest

cluster-provision/k8s/1.20/manifests/cni.diff
rhrazdil: I thought that I messed up, but it's just so confusing to see diff in a diff.

The lines I'm adding here are:
```
@@ -3642,6 +3642,11 @@ spec:
             # Auto-detect the BGP IP address.
             - name: IP
               value: "autodetect"
+            # Use IP/IPv6 addresses on ifaces eth0, eth1, eth2, etc.
+            - name: IP_AUTODETECTION_METHOD
+              value: "interface=eth.*"
+            - name: IP6_AUTODETECTION_METHOD
+              value: "interface=eth.*"
             # Enable IPIP
             - name: CALICO_IPV4POOL_IPIP
               value: "Always"
```

So in Github diff, only the lines with two ++ are being added, the lines with one + are just part of the diff context

cluster-provision/k8s/1.20/manifests/cni.diff
maiqueb: i see 4 lines changes, instead of 2.

I would have understood something like: 
```
--- a/cluster-provision/k8s/1.19/manifests/cni.do-not-change.yaml
+++ b/cluster-provision/k8s/1.20/manifests/cni.do-not-change.yaml
```

I see instead: 
```
--- a/cluster-provision/k8s/1.19/manifests/cni.do-not-change.yaml
+++ b/cluster-provision/k8s/1.19/manifests/cni.do-not-change.yaml
--- a/cluster-provision/k8s/1.20/manifests/cni.do-not-change.yaml
+++ b/cluster-provision/k8s/1.20/manifests/cni.do-not-change.yaml
```

Am I missing something ?

cluster-provision/k8s/1.20/manifests/cni.diff
maiqueb: Inception 0_o .

Thanks for explaining.

cluster-provision/k8s/1.20/manifests/cni.diff
maiqueb: diff of a diff once again.

maiqueb (approved):

cluster-provision/k8s/1.20/manifests/cni.diff
maiqueb: one thing that just occurred to me: instead of pointing to the "correct" copy (all files have the same contents) you could just delete them, and refer to the 1.19 configuration. Until there's a need for a different config, moment on which we'd have to copy the correct config.

ormergi:
@rhrazdil Hey could you please give more context for this change? do you need it for a new feature?
I am asking because we should avoid relaying on certain CNI feature/behavior.
How things are working on other CNI plugins when adding a bridge with fixed IP address?

As you may already know calico is very complex and has a lot of capabilities,
we want to avoid using its magical capabilities :grimacing:

rhrazdil:
> @rhrazdil Hey could you please give more context for this change? do you need it for a new feature?
I am asking because we should avoid relaying on certain CNI feature/behavior.
How things are working on other CNI plugins when adding a bridge with fixed IP address?

Hey, of course. This is **not** a requirement for a new feature.
I have hit an issue with calico's pods becoming `NotReady` and other weird networking issues after creating a NetworkAttachmentDefinition specifying a bridge with  `"isDefaultGateway": true,`.

With this setting, the bridge is assigned an IP address and when this happens, calico picks up the change and starts using this IP address for exchanging BGP updates.

See the relevant part of calico-node pod log:
```
2021-06-22 13:38:05.235 [INFO][57] monitor-addresses/startup.go 768: Using autodetected IPv4 address on interface eth0: 192.168.66.101/24   => BEFORE ASSIGNING NEW BRIDGE AN IP
2021-06-22 13:38:05.236 [INFO][57] monitor-addresses/startup.go 768: Using autodetected IPv6 address on interface eth0: fd00::101/128
2021-06-22 13:38:31.303 [INFO][55] felix/iface_monitor.go 187: Netlink address update. addr="10.10.10.1" exists=true ifIndex=28      => HERE it notices that an IP was added to an interface
2021-06-22 13:38:31.303 [INFO][55] felix/int_dataplane.go 1036: Linux interface addrs changed. addrs=set.mapSet{"10.10.10.1":set.empty{}, "fe80::f096:71ff:fe62:95fe":set.empty{}} ifaceName="newbridge"
-- omitted --
2021-06-22 13:38:31.303 [INFO][55] felix/int_dataplane.go 1463: Received interface addresses update msg=&intdataplane.ifaceAddrsUpdate{Name:"newbridge", 2021-06-22 13:38:31.305 [INFO][55] felix/ipsets.go 760: Doing full IP set rewrite family="inet" numMembersInPendingReplace=5 setID="this-host"
2021-06-22 13:39:05.238 [INFO][57] monitor-addresses/startup.go 768: Using autodetected IPv4 address on interface newbridge: 10.10.10.1/24  => AND HERE IT'S STARTED TO USE THAT IP FOR BGP 
```

> As you may already know calico is very complex and has a lot of capabilities,
we want to avoid using its magical capabilities grimacing

Indeed, I understand that.
However, this change doesn't introduce any new calico capability, it just configures the already
used IP autodetection mechanism in a way that's recommened by calico's [documentation](https://docs.projectcalico.org/reference/node/configuration#ip-setting)

maiqueb:
@rhrazdil what should we do about this PR ? Do you have plans to keep working it, or can it rot away ? 

Does the issue still persist ?

PR #638, opened 2021-07-01
Restore provision checking for cgroupsv2 lanes

vasiliy-ul:
The PR restores provision checking for cgroupsv2 lanes (leveraging the existing jobs `check-provision-k8s-1.20-cgroupsv2` and `check-provision-k8s-1.21-cgroupsv2`). This is a follow-up of the https://github.com/kubevirt/kubevirtci/pull/629.

**Note**: the existing cgroup v2 lanes are optional and now they actually use 'pure' `k8s-1.2{0,1}` images. There is a possibility to include `cluster-up` check for cgroupsv2 into the `check-provision-k8s-1.2{0,1}` jobs (just by running cluster-up twice: with and without cgroupsv2). It will spare one provision step per lane but the jobs will take more time to complete (also it will make cgroupsv2 checking mandatory). For now I just decided to reuse the existing jobs. But if it makes sense I can implement it as described.


vasiliy-ul:
Hm... cgroupsv2 checking fails on:
```
Step 1/15 : FROM fedora@sha256:8fa60b88e2a7eac8460b9c0104b877f1aa0cea7fbc03c701b7e545dacccfb433
Get https://registry-1.docker.io/v2/: x509: certificate signed by unknown authority
```
I checked in `project-infra` and seems that the cgroupsv2 jobs use a different image. E.g
```yaml
- image: gcr.io/k8s-testimages/kubekins-e2e@sha256:8c80c98d035a0f285ad49b964e5d0b62004844f3340407367841618e80e1503c
```

https://github.com/kubevirt/project-infra/blob/e15936dbc22251312454ffaf7156e8eee8e68f8c/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml#L181

while all other jobs have:

```yaml
- image: quay.io/kubevirtci/golang:v20210316-d295087
```

https://github.com/kubevirt/project-infra/blob/e15936dbc22251312454ffaf7156e8eee8e68f8c/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml#L154


Might it be the reason? @rmohr  maybe you have some hints?

rmohr:
@vasiliy-ul yes that can be the reason. In our won CI images we inject some certificates for our docker mirror.

vasiliy-ul:
It looks like my initial idea of having dedicated providers `k8s-1.2{0,1}-cgroupsv2` for cgroupv2 clusters does not fit well. Initially I wanted just to substitute `KUBEVIRT_PROVIDER` (e.g. from `k8s-1.20-cgroupsv2` to `k8s-1.20`) while doing cluter-up but it turns out that it causes issues in several other places and fixing those seems more like a hack. So eventually I decided to make it simpler by leveraging `KUBEVIRT_CGROUPSV2` env var:
```bash
export KUBEVIRT_PROVIDER=k8s-1.20
export KUBEVIRT_NUM_NODES=2 # master + one node
export KUBEVIRT_CGROUPV2=true
make cluster-up
```
That approach seems to work almost out of the box. If its fine I will also need to change the periodic and presubmit jobs in project-infra and adapt to new changes.

rmohr:
> That approach seems to work almost out of the box. If its fine I will also need to change the periodic and presubmit jobs in project-infra and adapt to new changes.

Sure, no issue at all.

vasiliy-ul:
@rmohr , could you maybe take one more look here? I just realized that I messed up the cgroupsv2 periodics and they are currently broken :( This PR and then the https://github.com/kubevirt/project-infra/pull/1414 should fix them.

Here basically I just removed the dedicated cgroupsv2 providers (1.20 and 1.21) and now cgroup settings within the test cluster are configured depending on the env var `KUBEVIRT_CGROUPV2` (set to `false` by default).

PR #644, opened 2021-07-18
SRIOV provider: Deploy default CNI

ormergi:
While back then we decided to deploy Calico CNI on KinD providers
in order to create dual-stack cluster for DEV and CI env's.

As for today we managed to enable dual-stack clusters using our
VM based providers, and eventually removed the IPV6 KinD based provider.
Though we still deploy old version of Calico on the SRIOV lane provider.

We should use the default KinD CNI 'kindnet' instead,
it will reduce complexity of the setup and maintenance.


ormergi:
SRIOV lane with kindnet CNI plugin pass 
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1462/rehearsal-pull-kubevirt-e2e-kind-1.19-sriov/1417481998527631360

ormergi:
@phoracek @qinqon @maiqueb @EdDev SRIOV lane with kindnet CNI plugin pass 
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/1462/rehearsal-pull-kubevirt-e2e-kind-1.19-sriov/1417481998527631360

cluster-up/cluster/kind/common.sh
```diff
@@ -102,7 +102,6 @@ function _configure_registry_on_node() {
 
 function _install_cnis {
     _install_cni_plugins
-    _install_calico_cni
```
EdDev, line 105: Can the function be deleted now?

EdDev (commented):
Very nice!

cluster-up/cluster/kind/common.sh
ormergi: Yes we can remove all related code, if we would want to use Calico again we have git history

ormergi: Done

ormergi:
@EdDev I addressed you review comment could you please take another look?

EdDev (commented):
Does this change mean we do not deploy dual-stack anymore on the SR-IOV lane? I could not understand that part from the commit message.

And if not, does `kindnet` support it (so we can add it)?

ormergi:
> Does this change mean we do not deploy dual-stack anymore on the SR-IOV lane? I could not understand that part from the commit message.

Yes its single stack now,
The test that checks connectivity between VM's though SRIOV interface with IPv6 is green.

> And if not, does `kindnet` support it (so we can add it)?

In general kindnet supports dual-stack

EdDev:
> In general kindnet supports dual-stack

If it is easily addable, then can we do it now? It is not a must, but it will be nicer so we will not need to think about it in the future.

ormergi:
> > In general kindnet supports dual-stack
> 
> If it is easily addable, then can we do it now? It is not a must, but it will be nicer so we will not need to think about it in the future.

With the new Kind API its pretty straight forward https://kind.sigs.k8s.io/docs/user/configuration/#networking
the problem is that it supported on k8s-1.20+ clusters.
Due to the fact we still use `PodPreset` (deprecated on k8s-1.20) we cant upgrade the provider now.
As soon as https://github.com/kubernetes-sigs/kind/pull/2321 will get in we will be able to upgrade the provider 
stop using `PodPreset` and eventually turn on dual-stack.

EdDev (approved):
Thanks!

EdDev (changes requested):
ohh, what about `./cluster-up/cluster/kind/manifests/kube-calico.yaml.in`?

EDIT: Also `./cluster-up/cluster/kind/manifests/kube-calico.diff.in`

ormergi:
> ohh, what about `./cluster-up/cluster/kind/manifests/kube-calico.yaml.in`?
> 
> EDIT: Also `./cluster-up/cluster/kind/manifests/kube-calico.diff.in`

@EdDev Done

EdDev (approved):
Thanks!

enp0s3 (approved):
@ormergi Thank you!
/approve

PR #645, opened 2021-07-18
Add check-cluster-up script for KinD providers

ormergi:
Currently we dont run Kubevirt conformance tests on SR-IOV check up lane
This PR adds check-cluster-up script for KinD based providers, similar to what we have on check-provision lanes.

Basically it will check that the cluster is healthy, dump the nodes network configurations
and run conformance tests.

This script will be used on SR-IOV prow job command in order to prevent bad changes to SR-IOV provider.


oshoval:
> This PR check-cluster-up script for KinD based providers, similar to what we have on check-provision lanes.

Missing the word `adds`

cluster-up/cluster/kind/check-cluster-up.sh
```diff
@@ -0,0 +1,76 @@
...
+  cd $kubevirtci_path
+  kubectl="./cluster-up/kubectl.sh"
+  echo "Wait for pods to be ready.."
+  timeout 5m bash -c "until ${kubectl} wait --for=condition=Ready pod --timeout=30s --all  -A; do sleep 1; done"
+  timeout 5m bash -c "until ${kubectl} wait --for=condition=Ready pod --timeout=30s -n kube-system --all; do sleep 1; done"
+  ${kubectl} get nodes
+  ${kubectl} get pods -A
+  echo ""
+  
+  nodes=$(${kubectl}  get no --no-headers | awk '{print $1}')
```
oshoval, line 38: please use `get nodes` here,
and also there is double space before

cluster-up/cluster/kind/check-cluster-up.sh
```diff
@@ -0,0 +1,76 @@
...
+    ${node_exec} cat /etc/hosts
+    echo ""
+    echo "[$node] resolve config:"
+    ${node_exec} cat /etc/resolv.conf
+    echo ""
+  done
+
+  if [ "$RUN_KUBEVIRT_CONFORMANCE" == "true" ]; then
+    nightly_build_base_url="https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/kubevirt"
+    latest=$(curl -L "${nightly_build_base_url}/latest")
```
oshoval, line 57: please add `-s` (all over)

cluster-up/cluster/kind/check-cluster-up.sh
```diff
@@ -0,0 +1,76 @@
...
+
+    echo "Deploy latest nighly build Kubevirt"
+    if [ $(kubectl get kubevirts -n kubevirt kubevirt -ojsonpath='{.status.phase}') != "Deployed"; then
```
oshoval, line 60: maybe worth to assert that kubevirt is not deployed instead ?
i do understand it gives freedom to install specific kubevirt, so up to you

btw usually on kubevirtci the sonobuoy conformance are without kubevirt, right?

cluster-up/cluster/kind/check-cluster-up.sh
```diff
@@ -0,0 +1,76 @@
...
+      ${kubectl} apply -f "${nightly_build_base_url}/${latest}/kubevirt-operator.yaml"
+      ${kubectl} apply -f "${nightly_build_base_url}/${latest}/kubevirt-cr.yaml"
+    fi
+    ${kubectl} wait -n kubevirt kv kubevirt --for condition=Available --timeout 15m
```
oshoval, line 64: maybe should be part of the if, else it means that the kubevirt is deployed but here you are checking that its also Available, again up to you

cluster-up/cluster/kind/check-cluster-up.sh
```diff
@@ -0,0 +1,76 @@
...
+
+    echo "Run latest nighly build Kubevirt conformance tests"
+    kubevirt_plugin="--plugin ${nightly_build_base_url}/${latest}/conformance.yaml"
+    SONOBUOY_EXTRA_ARGS="${kubevirt_plugin} ${SONOBUOY_EXTRA_ARGS}"
```
oshoval, line 68: `="$SONOBUOY_EXTRA_ARGS ..."` first i think (also in the 2nd place)

cluster-up/cluster/kind/check-cluster-up.sh
```diff
@@ -0,0 +1,76 @@
...
+
+    commit=$(curl -L "${nightly_build_base_url}/${latest}/commit")
+    container_tag="--plugin-env kubevirt-conformance.CONTAINER_TAG=${latest}_${commit:0:9}"
```
oshoval, line 71: maybe the `0:9` should be in the previous line ?

oshoval (commented):
Thanks

cluster-up/cluster/kind/check-cluster-up.sh
ormergi: > maybe worth to assert that kubevirt is not deployed instead ?
> i do understand it gives freedom to install specific kubevirt, so up to you

In case its not deployed it wont get into the if block

> btw usually on kubevirtci the sonobuoy conformance are without kubevirt, right?

Currently we run Kubevirt conformance as well on all check-provision lanes

cluster-up/cluster/kind/check-cluster-up.sh
ormergi: This check is not part of the if in order to prevent running tests in case not all kubevirt components are ready.

cluster-up/cluster/kind/check-cluster-up.sh
ormergi: In general the order doesn't matter, I placed `SONOBUOY_EXTRA_ARGS` at the end so it will be easier to understand what we just added.

cluster-up/cluster/kind/check-cluster-up.sh
ormergi: you mean adding another line?
`commit=${commit:0:9}`

cluster-up/cluster/kind/check-cluster-up.sh
ormergi: Done

cluster-up/cluster/kind/check-cluster-up.sh
ormergi: Done

cluster-up/cluster/kind/check-cluster-up.sh
oshoval: of course it doesnt matter, but most of the places all over our code are appending, imo it should be the same here

cluster-up/cluster/kind/check-cluster-up.sh
oshoval: ye, just so it will be less hidden

ormergi: Done

cluster-up/cluster/kind/check-cluster-up.sh
ormergi: Done

cluster-up/cluster/kind/check-cluster-up.sh
oshoval: maybe 15 minutes is too much ?
how much other repos use

oshoval (approved):
Thanks

cluster-up/cluster/kind/check-cluster-up.sh
```diff
@@ -0,0 +1,77 @@
...
+  kubectl="./cluster-up/kubectl.sh"
+  echo "Wait for pods to be ready.."
+  timeout 5m bash -c "until ${kubectl} wait --for=condition=Ready pod --timeout=30s --all  -A; do sleep 1; done"
+  timeout 5m bash -c "until ${kubectl} wait --for=condition=Ready pod --timeout=30s -n kube-system --all; do sleep 1; done"
+  ${kubectl} get nodes
+  ${kubectl} get pods -A
+  echo ""
+  
+  nodes=$(${kubectl} get nodes --no-headers | awk '{print $1}')
+  for node in $nodes; do
```
oshoval, line 39: some thing nice is to create an array of command strings and then iterate on it , executing it,
wdyt?

ormergi:
Added missing quotes to kubevirt deployed if condition

oshoval:
I think that we need a way to easily run the provision without checking kubevirt as well
maybe a 2nd lane which is optional and does all the same, beside the check-up?

else it will make testing where there are iterations on CI, slower
and will need hacking in order to skip temporarily the check

cluster-up/cluster/kind/check-cluster-up.sh
ormergi: I agree its nicer, I didnt do it because I wanted to keep the log messages before each command.
Putting it in bash equivalent for map doesn't make things prettier in this case.

cluster-up/cluster/kind/check-cluster-up.sh
ormergi: on hack/cluster-deploy.sh it check every 5 minutes for 25 minutes
https://github.com/kubevirt/kubevirt/blob/371958b07a340d0ba9e9d1e9a2936411e395a5df/hack/cluster-deploy.sh#L154
From what I saw sometime it takes 2-3 rounds to be ready, rarely more then 10 minutes , so I picked 15 minutes timeout.

ormergi:
> I think that we need a way to easily run the provision without checking kubevirt as well
> maybe a 2nd lane which is optional and does all the same, beside the check-up?
> 
> else it will make testing where there are iterations on CI, slower
> and will need hacking in order to skip temporarily the check

I understand it could slow down iterations on CI, but think about the time it will save to revert a bad change 
after you discovered it only when its land at kubevirt/kubevirt, 
the time you would spend on debugging it.

Instead we will get initial feedback of how a change will affect kubevirt/kubevirt, 
Also it will save the hassle to create secondary PR and test the changes on kubevirt/kubevirt.

For iterating on CI Its pretty straight forward to skip Kubevirt conformance,
all you need to do is override RUN_KUBEVIRT_CONFORMANCE env var (on provider.sh)
I think is simple then having another secondary job that does only check-up (w/o conformance)

cluster-up/cluster/kind/check-cluster-up.sh
oshoval: You can still keep log messages or whatever you desire in the loop isnt it ?
something like this wouldn't be nicer?
```
declare -a arr=("element1" "element2" "element3")
for i in "${arr[@]}"
do
   echo "$i" # exec here instead
done
```

ormergi: > You can still keep log messages or whatever you desire in the loop isnt it ?

It will print the command we executes and not its descriptions, for example
for `ip a` command, it will print 
`[sriov-worker] ip r` 
instead of:
`network interfaces status:`

It makes the log easier to follow, cosmetics.

oshoval:
the last push is pure rebase right ?

ormergi:
> the last push is pure rebase right ?

Yes

oshoval (approved):
Thanks

dhiller (approved):
/approve

Sorry for the delay, this slipped through :-/

ormergi:
/hold 

Wait until https://github.com/kubevirt/kubevirt/pull/6199 is merged as we might need to do some changes to this PR

PR #648, opened 2021-07-23
Consolidate test test tool images into a single one

rmohr:
As a build- and CI improvement only have one fedora instead of two.

Signed-off-by: Roman Mohr <rmohr@redhat.com>


rmohr:
/hold

still testing this.

cluster-provision/images/vm-image-builder/fedora-sriov-lane/cloud-config
```diff
@@ -1,27 +0,0 @@
...
-    - path: /etc/modules-load.d/igb.conf
-      content: |
-        igb
-runcmd:
-  - sudo dnf install -y kernel-modules-$(uname -r) qemu-guest-agent stress
-  - sudo dnf clean all
-  - sudo hostnamectl set-hostname ""
-  - sudo hostnamectl set-hostname "" --transient
-  - sudo systemctl start qemu-guest-agent
-  - sudo sed -i /users-groups/d /etc/cloud/cloud.cfg
```
oshoval, line 26: We chosen this method in order to eliminate the race condition between cloud-init setting the password and the fedora login from the tests,
now it means that the tests will set the password
see 
https://github.com/kubevirt/kubevirt/pull/6108/files#diff-da14ca97e77db6bb275427ac69187a539c58fc297e64519628e5385f9f411c5fR682

I am less fond of this, please consider
@EdDev FYI

rmohr: I am not sure if that solves anything. Doesn't that mean that you potentially can already log in before cloud-init is done and as such you work on an OS which is not done with the setup?

Note that the consolidated image only starts the guest agent now as the last step in cloud-init (like the other image already did) which allows a race-free login. You basically wait for the guest agent first and then start your test..

oshoval: > I am not sure if that solves anything. Doesn't that mean that you potentially can already log in before cloud-init is done and as such you work on an OS which is not done with the setup?

Both methods did solve the races we encountered (David and myself), where both don't solve waiting for all cloud-init modules to be finished.
So far so good, there are no flakes anymore about the login.
If we want to wait for all modules to be finished, we can wait for cloud-init done event and for guest-agent connected event  (which we are already waiting for right?)
The races were due to the credentials recreation.
I don't think creating this binding is the right thing to do, but to create independent methods to wait for the needed modules where they are actually needed.

> 
> Note that the consolidated image only starts the guest agent now as the last step in cloud-init (like the other image already did) which allows a race-free login. You basically wait for the guest agent first and then start your test..

The guest-agent connected signal which the tests are waiting to doesn't mean the whole cloud-init is done, it has several modules.

rmohr: > The guest-agent connected signal which the tests are waiting to doesn't mean the whole cloud-init is done, it has several modules.

That can be., but I would argue that waiting for the guest-agent this way is more accurate.

rmohr: This consolidates it to doing it one way, and it can be moved later on  to another way.

rmohr: > The guest-agent connected signal which the tests are waiting to doesn't mean the whole cloud-init is done, it has several modules.

According to the  docs it is at least very very late in the process: https://cloudinit.readthedocs.io/en/latest/topics/boot.html#final: "This stage runs as late in boot as possible".

oshoval: Adding `- sudo sed -i /users-groups/d /etc/cloud/cloud.cfg`
means that we are not forced to add cloud-init
nor need to set again passwords,
nor create binding between modules,
and that we can wait for guest-agent anyway because it already runs (if we don't disable it),
cloud-init has a file that signals the finishing of it, i think we should do the other way around when consolidating,
it will just means that we need to add wait for cloud-init done where actually needed / used.

rmohr: > Adding `- sudo sed -i /users-groups/d /etc/cloud/cloud.cfg`
> means that we are not forced to add cloud-init
> nor need to set again passwords,
> nor create binding between modules,
> and that we can wait for guest-agent anyway because it already runs (if we don't disable it),
> cloud-init has a file that signals the finishing of it, i think we should do the other way around when consolidating,
> it will just means that we need to add wait for cloud-init done where actually needed / used.

You may not use user-data in all of your tests, but as far as I see you use other parts of cloud-init a lot (like network setup). You know for sure that this is done this way.

oshoval: right, but only where we do need it, and without creating any binding between modules,
all what we need to wait for is explicit.

rmohr: > right, but only where we do need it, and without creating any binding between modules,
> all what we need to wait for is explicit.

I don't understand this. It just makes everything more complicated. The mental model this way is very simple: just wait for the guest agent and most likely things are fine. I don't have to look at different contexts so much, reducing the maintenance burden. I am not sure what exactly you mean with "without creating any binding between modules". If guest-agent is ready. Things are most likely fine.

oshoval: Imho creating binding between guest agent and cloud-init, and enforcing the user to always run cloud-init in order to set password is mental burden,
keeping tests as simple and explicit / controlled is a plus.

Things are most likely fine anyhow with the approach of the sed line.
Simply adding boot finish event of cloud-init will detach the connection between the cloud-init and guest-agent
and then tests will have the option to call

1. wait for guest-agent done event (we already have in code of course)
2. wait for cloud-init done event (only where needed)

without enforcing tests to run both cloud-init and guest-agent in order to set password where only part of those 3 are needed.

oshoval: btw `sudo sed -i /users-groups/d /etc/cloud/cloud.cfg` keeping this line, means that the VM can be logged out of the box, without setting cloud-init which is a plus i think.
I don't see any reason to remove this line, whichever the solution is chosen right?
It will work both ways (unless i am missing something).

moreover it means that disabling the guest-agent means one must use cloud-init or run it manually when running the VM standalone,
I think that it is not a good practice, it would be nicer to have a standalone machine without proprietary parts, but more straightforward.

rmohr: Readded it.

rmohr: But note that I now have to remove your guest-agent check, since the guest agent will not be started anymore by default.

ormergi: Hi, why do we want to disable qemu-guest-agent by default?

rmohr: Because all other tests, except a few networking tests use the guest-agent as the readiness check.

oshoval: > But note that I now have to remove your guest-agent check, since the guest agent will not be started anymore by default.

Its orthogonal for readding sed line isnt it ?

I don't think its that the right thing to do :/ (disabling it in first place) :/

If you do so, need to check please each place where it is removed if its allowed to be removed, or that we need to start it (which is a mental burden i think?)

PS. all those fixes from David and myself started only because the login problem.

ormergi: Hey, back then I enabled qemu-guest agent by default for SRIOV test suite because there were flakes
when it was disabled and we needed to enabled it for the tests, 
Also to untangle provisioning the VM from tests suite.
I am pretty lost here on this thread.. why do we want to disable qmeu-guest agent by default again?
What does it mean for SRIOV tests? we would need to enabled it from the tests again?

rmohr: > Its orthogonal for readding sed line isnt it ?
> 
> I don't think its that the right thing to do :/ (disabling it in first place) :/
> 
> If you do so, need to check please each place where it is removed if its allowed to be removed, or that we need to start it (which is a mental burden i think?)
> 
> PS. all those fixes from David and myself started only because the login problem.

Yes. I am just consolidating it. I understand why they are there. This is just the simplest way forward until something better. As you can see right now we have complexity in multiple places:

 * having multiple images (maintenance and build download is an issue at scale and during local development)
 * having one image with multiple flows (adds complexity at provisioning and individual tests because one has to maintain the flows and has to know which test chooses wich)

Therefore I chose it. I can leave the line in but I would want to insist on the guest agent.

rmohr: > What does it mean for SRIOV tests? we would need to enabled it from the tests again?

@ormergi  I took care of that in https://github.com/kubevirt/kubevirt/pull/6108/files#diff-da14ca97e77db6bb275427ac69187a539c58fc297e64519628e5385f9f411c5fR682. The guest agent will be started as always, but just very late when cloud-init is almost done (the same like for the other tests).

rmohr: @oshoval @ormergi  btw. don't get me wrong. I would not mind if you have your own images with your own flows, but we create big images of a few hundred megabytes which are pulled back and forth without much benefit at this stage. If needed the images can also be forked later on again.

rmohr: Please let me know if you can live with the guest-agent and cloud-init solution. If not, then I will keep the separate image for now and close the PR. That is fine too.

rmohr:
@ormergi @oshoval I changed the guest-agent startup. It is now delayed until cloud-init reports that it is done. PTAL.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -2,13 +2,45 @@
...
+      owner: root:root
+      permissions: '0755'
+      content: |
+          [Unit]
+          Description=Wait for cloud-init to complete
+          Before=qemu-guest-agent.service
+
+          [Service]
+          Type=oneshot
+          ExecStartPre=bash -c "while [ ! -f /var/lib/cloud/data/result.json ] && [ ! -f /etc/cloud/cloud-init.disabled ]; do sleep 1; done"
```
oshoval, line 30: very nice
maybe worth to wait for /var/lib/cloud/instance/boot-finished instead result.json ?
as this is the really last thing that occur i think ?

oshoval: hmm, but this service will run every boot right ?
it means that the first thing that need to be done is to clean the previous cloud-init indications (for resets)
or at least clean it on the first boot before the shutdown (in this file)

oshoval: Question - sorry for triple posts, just didnt want to EDIT, so it will jump
it means that one must config cloud-init or at least have the cloud-init disabled on the machine right (the 2nd is not possible via tests, chicken and egg maybe) ?
it means that it isn't perfect yet, and there is dependency, and burden to run manually stuff that otherwise we are not must to

rmohr: I have a better way already. Just wait a little. :)

rmohr: Ok, updated. I am now waiting until `cloud-final.service` is done. Then I do a blocking wait with `cloud-init status --wait` in a pre-check. Once the check passes, the `wait-for-cloud-init.service` is done. Since it is marked as something which has to run `Before` qemu-guest-agent the agent will be started afterwards.

This works in both cases, if cloud init is enabled or disabled:

 * If cloud-init gets enabled by the systemd generator: `wait-for-cloud-init` will be enabled and run, and qemu-guest-agent will wait for it
 * If cloud-init is disabled by  the systemd generator: Nothing will want `wait-for-cloud-init`. Since `wait-for-cloud-init` is not `wanted` by `qemu-guest-agent` it will be ignored in this case and `qemu-guest-agent` starts immediately.

oshoval: Nice !

oshoval: Nice !
once there will be the updated kubevirt PR we can review it please together

Just making sure please
`cloud-init status --wait` will work also after the 2nd reboot ? (first one after provision) ?
i mean the indication files are still there from the first boot, unless it doesn't use them

rmohr: Yes it always works.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -2,13 +2,46 @@
...
+  - sudo systemctl enable wait-for-cloud-init
+  - sudo systemctl enable qemu-guest-agent.service
   - sudo hostnamectl set-hostname ""
   - sudo hostnamectl set-hostname "" --transient
-  - sudo systemctl stop qemu-guest-agent
-  - sudo systemctl disable qemu-guest-agent
-  - sudo mv /lib/systemd/system/qemu-guest-agent.service /home/fedora/
-  - sudo systemctl daemon-reload
+  - sudo sed -i /users-groups/d /etc/cloud/cloud.cfg
+  - sudo sed -i 's/SELINUX=.*/SELINUX=permissive/' /etc/selinux/config
```
oshoval, line 46: it will also change the remark line FWIW
`# SELINUX= can take one of these three values:`
to
`# SELINUX=permissive`
maybe add ^ ?

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
ormergi: Looks good!

I have a question, will qemu-guest-agent start if we create a VM w/o cloud-init?

rmohr: Yes: If cloud-init is disabled by the systemd generator: Nothing will want wait-for-cloud-init. Since wait-for-cloud-init is not wanted by qemu-guest-agent it will be ignored in this case and qemu-guest-agent starts immediately. 

To rephrase it: For as long as no unit is enabled which `wants` `wait-for-cloud-init` it will not be pulled in, and therefore `qemu-guest-agent` will not wait for it. `qemu-guest-agent` is not even aware of it.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
rmohr: Done.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
ormergi: FWIW I verified that `qemu-guest-agent` starts on a VM w/o cloud-config
``` 
[fedora@vmi-masquerade ~]$ sudo systemctl status wait-for-cloud-init.service
● wait-for-cloud-init.service - Wait for cloud-init to complete
     Loaded: loaded (/etc/systemd/system/wait-for-cloud-init.service; enabled; >
     Active: inactive (dead)
```
And for a VM **with** cloud-config that it does starts after cloud-init finished (using kubevirt example/vmi-masqurade.yaml)
```
...
[  168.941298] cloud-init[733]: ci-info: no authorized ssh keys fingerprints found for user fedora.
[  168.962179] cloud-init[733]: Cloud-init v. 19.4 finished at Mon, 26 Jul 2021 14:19:16 +0000. Datasource DataSourceNoCloud [seed=/dev/vdb][dsmode=net].  Up 168.85 seconds
[  OK  ] Finished Execute cloud user/final scripts.
[  OK  ] Reached target Cloud-init target.
         Starting Wait for cloud-init to complete...
[  OK  ] Finished Wait for cloud-init to complete.
[  OK  ] Started QEMU Guest Agent.
[  OK  ] Finished man-db-cache-update.service.
...
```
```
[fedora@vmi-masquerade2 ~]$ sudo systemctl status  wait-for-cloud-init.service
● wait-for-cloud-init.service - Wait for cloud-init to complete
     Loaded: loaded (/etc/systemd/system/wait-for-cloud-init.service; enabled; vendor preset: disabled)
     Active: active (exited) since Mon 2021-07-26 14:19:18 UTC; 8min ago
    Process: 3132 ExecStartPre=/usr/bin/cloud-init status --wait (code=exited, status=0/SUCCESS)
    Process: 4186 ExecStart=/usr/bin/true (code=exited, status=0/SUCCESS)
   Main PID: 4186 (code=exited, status=0/SUCCESS)
        CPU: 770ms

Jul 26 14:19:17 vmi-masquerade2 systemd[1]: Starting Wait for cloud-init to complete...
Jul 26 14:19:18 vmi-masquerade2 cloud-init[3132]: status: done
Jul 26 14:19:18 vmi-masquerade2 systemd[1]: Finished Wait for cloud-init to complete.
```

ormergi (approved):

oshoval (approved):

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -2,13 +2,46 @@
...
+      permissions: '0755'
+      content: |
+          [Unit]
+          Description=Wait for cloud-init to complete
+          Before=qemu-guest-agent.service
+          After=cloud-final.service
+
+          [Service]
+          Type=oneshot
+          ExecStartPre=cloud-init status --wait
```
oshoval, line 31: btw this means the cloud-init is finished, but maybe with errors right ?

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
AlonaKaplan: Does it mean that now all the tests with cloud-init will have to wait for "agent connected condition" (even if they don't need/care about the ga)? Otherwise the test will be flaky since it will be able to connect to the vm before the cloud-init is ready.

rmohr: @AlonaKaplan yes, that was never different.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
rmohr: According to my experience `cloud-final` will not complete then, so it will not even reach this command. Here an example of failing tests. The guest agent here never starts because the user-data is bad and `cloud-final` never succeeds.

This may not be the most convenient behaviour with my unit, but at least we see right now clearly a failing test.

rmohr: Forgot to add the link: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/6108/pull-kubevirt-e2e-k8s-1.19-sig-network/1419925101213650944

oshoval: maybe worth to update kubevirt "Retried login attempt after two minutes failed" to say something like (did cloud-init finish ?)

rmohr:
/unhold

CI tests work with the new image. Unholding and publishing the image.

/approve

due to two lgtm

rmohr:
Error is unrelated:

```
time="2021-08-03T13:25:17Z" level=error msg="error attempting to run sonobuoy: waiting for run to finish: Pod entered a fatal terminal status: failed"
```

/override check-provision-k8s-1.20

PR #651, opened 2021-07-30
Autostart the CNI DHCP daemon on CNAO lanes

jean-edouard:
That will allow admins to define network-attachment-definitions that use the DHCP IPAM plugin.

### DIFF
--- a/cluster-provision/k8s/1.19/provision.sh
+++ b/cluster-provision/k8s/1.19/provision.sh
@@ -316,6 +316,39 @@ echo "tmpfs /var/provision/kubevirt.io/tests tmpfs rw,context=system_u:object_r:
 
 dnf install -y NetworkManager-config-server
 
+# Add CNI DHCP daemon service and socket, disabled by default
+# TODO: remove after switching from kubernetes-cni.rpm to containernetworking-plugins,
+#   as containernetworking-plugins provides these 2 files
+cat > /etc/systemd/system/cni-dhcp.service <<EOF
+[Unit]
+Description=CNI DHCP service
+Documentation=https://github.com/containernetworking/plugins/tree/master/plugins/ipam/dhcp
+After=network.target cni-dhcp.socket
+Requires=cni-dhcp.socket
+
+[Service]
+ExecStart=/opt/cni/bin/dhcp daemon
+
+[Install]
+WantedBy=multi-user.target
+EOF
+cat > /etc/systemd/system/cni-dhcp.socket <<EOF
+[Unit]
+Description=CNI DHCP service socket
+Documentation=https://github.com/containernetworking/plugins/tree/master/plugins/ipam/dhcp
+PartOf=cni-dhcp.service
+
+[Socket]
+ListenStream=/run/cni/dhcp.sock
+SocketMode=0660
+SocketUser=root
+SocketGroup=root
+RemoveOnStop=true
+
+[Install]
+WantedBy=sockets.target
+EOF
+
 # Cleanup the existing NetworkManager profiles so the VM instances will come
 # up with the default profiles. (Base VM image includes non default settings)
 rm -f /etc/sysconfig/network-scripts/ifcfg-*
--- a/cluster-up/cluster/k8s-provider-common.sh
+++ b/cluster-up/cluster/k8s-provider-common.sh
@@ -50,6 +50,11 @@ function up() {
             $kubectl create -f /opt/cnao/network-addons-config-example.cr.yaml
             $kubectl wait networkaddonsconfig cluster --for condition=Available --timeout=200s
         fi
+
+        # When CNAO is enabled, start the CNI dhcp daemon on every node by default
+        for node in $( seq -w 1 99 | head -n $KUBEVIRT_NUM_NODES ); do
+            ${_cli} --prefix $provider_prefix ssh node$node -- sudo systemctl enable --now cni-dhcp
+        done
     fi
 
     if [ "$KUBEVIRT_DEPLOY_ISTIO" == "true" ] && [[ $KUBEVIRT_PROVIDER =~ k8s-1\.1.* ]]; then
--- a/cluster-provision/k8s/1.20/provision.sh
+++ b/cluster-provision/k8s/1.20/provision.sh
@@ -325,6 +325,39 @@ echo "tmpfs /var/provision/kubevirt.io/tests tmpfs rw,context=system_u:object_r:
 
 dnf install -y NetworkManager-config-server
 
+# Add CNI DHCP daemon service and socket, disabled by default
+# TODO: remove after switching from kubernetes-cni.rpm to containernetworking-plugins,
+#   as containernetworking-plugins provides these 2 files
+cat > /etc/systemd/system/cni-dhcp.service <<EOF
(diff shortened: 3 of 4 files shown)


cluster-provision/k8s/1.19/provision.sh
```diff
@@ -316,6 +316,24 @@ echo "tmpfs /var/provision/kubevirt.io/tests tmpfs rw,context=system_u:object_r:
...
+Description=CNI DHCP daemon
+After=network.target
+StartLimitIntervalSec=0
+
+[Service]
+Type=simple
+Restart=always
+RestartSec=1
+ExecStartPre=rm -f /run/cni/dhcp.sock
+ExecStart=/opt/cni/bin/dhcp daemon
```
phoracek, line 331: Interesting, is this deployed on kubevirtci nodes by default? My understanding was that this is installed only with CNAO - so in case user does not install CNAO, there would be a failing service.

jean-edouard: The service is actually disabled by default, and only enabled on cluster-up when CNAO is enabled

jean-edouard: @phoracek does the above answer your concern?

phoracek: It does. Sorry I haven't got back to this yet... I will ask somebody from the team to assist with this and https://github.com/kubevirt/kubevirtci/pull/651

cluster-provision/k8s/1.19/provision.sh
```diff
@@ -316,6 +316,24 @@ echo "tmpfs /var/provision/kubevirt.io/tests tmpfs rw,context=system_u:object_r:
 
 dnf install -y NetworkManager-config-server
 
+# Add CNI DHCP daemon service, disabled by default
```
ormergi, line 319: Why is it necessary to create this service, cant we just enable it?

jean-edouard: I'm not sure I understand your question.
This is creating a systemd service for `/opt/cni/bin/dhcp daemon`, since I could not find any existing one.
Since it is just in `/etc/systemd/system`, it gets added to the list of systemd available services but it is not enabled by default.
The idea is to enable it only when we need it, i.e. when creating CNAO lanes that potentially have multiple NICs.
Without that daemon, CNI can't do DHCP on pods using secondary interfaces.

cluster-provision/k8s/1.19/provision.sh
EdDev: There is something I am missing here.
The service is suppose to be there already, supplied by `kubernetes-cni.rpm` (which is actuallu suppose to be replaced by `containernetworking-plugins` nowdays).
So why you create/override it here?

I find this strange also because I do not really expect an operator to have to deploy this manually like this.

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -50,6 +50,11 @@ function up() {
             $kubectl create -f /opt/cnao/network-addons-config-example.cr.yaml
             $kubectl wait networkaddonsconfig cluster --for condition=Available --timeout=200s
         fi
+
+        # When CNAO is enabled, start the CNI dhcp daemon on every node by default
+        for node in $( seq -w 1 99 | head -n $KUBEVIRT_NUM_NODES ); do
+            ${_cli} --prefix $provider_prefix ssh node$node -- sudo systemctl enable --now cni-dhcp
+        done
```
EdDev, line 57: Is there an option to do this from the tests? It would help support running the U/S tests on any D/S (not only kubevirtci). Conformance tests could also take advantage of this.

cluster-provision/k8s/1.19/provision.sh
jean-edouard: `kubernetes-cni` does not seem to deploy a systemd service (`# repoquery --list kubernetes-cni`).
Grepping around on a CNAO-enabled node, no systemd service seems to have the ability to start the CNI DHCP daemon.
Am I the one missing something? :)
Thanks

EdDev: I think I was too optimistic :)
It appears in [containernetworking-plugins-1.0.0](https://rpmfind.net/linux/RPM/fedora/updates/testing/34/x86_64/Packages/c/containernetworking-plugins-1.0.0-1.fc34.x86_64.html).

I think it did not landed yet on centos-stream, which is a bit odd, because it usually is a bit ahead of fedora.
Can you just make sure it follows those so we can just switch when it lands on centos-stream?

jean-edouard: Ah, yes, got it!
I will update the PR to use those.
Thanks!

jean-edouard: Done!

EdDev: Thanks!

EdDev (commented):
This looks good to me, I would just appreciate if you share your opinion on the [activation part](https://github.com/kubevirt/kubevirtci/pull/651/commits/542fe987504789cceb168147c7ec1c35a90e4f6d#r691260618).
I guess the tests that will use it, will have to be condition per running under kubevirtci or only if the service is available.

cluster-provision/k8s/1.19/provision.sh
ormergi: Thanks for explaining I understand this better now,
I meant if there is already an existing service on systemd cant we just enable and start it when needed? (e.g CNAO tests suite) what did I miss?

EdDev: There is no such service, it was discussed in the other thread.

cluster-up/cluster/k8s-provider-common.sh
jean-edouard: Sorry I missed that one!
Starting and stopping services on every node from the test suite feels a bit cumbersome to me.
I guess my question would be: is there ever a reason to *not* be running this daemon? We (and everybody else) could just enable it by default and be done with this.

EdDev: Yes, I agree, it could be enabled without condition. The problem is, that it may not be started in practice on other clusters. I guess it will be enough to make sure the test is not ran on clusters where the service is not running.

I think the main point of all this is that we introduce a new dependency on the cluster, for running the tests. If we could avoid it, it would have been simpler for all.
(e.g. using static IP)

cluster-provision/k8s/1.19/provision.sh
ormergi: It seems that containernetworking-plugins v1 is now available for centos steams [link](https://rpmfind.net/linux/RPM/centos/8-stream/appstream/x86_64/Packages/containernetworking-plugins-1.0.0-1.module_el8.5.0+890+6b136101.x86_64.html)
We can get it by updating it as part of provision script, though it doesn't include cni-dhcp service and socket files.
I think its still better to use v1 version.

jean-edouard:
Closing.
We'll use `whereabouts` for testing dedicated migration networks, so this is not needed anymore.

PR #656, opened 2021-08-08
Correct error caused by #654

akalenyu:
https://github.com/kubevirt/kubevirtci/pull/654 breaks kubevirtci:
https://github.com/kubevirt/kubevirt/pull/6188

We cannot just forward everything from port->port because the containers
that get created by kubevirtci (dnsmasq, registry, VM carriers) share the same network NS.

Signed-off-by: Alex Kalenyuk <akalenyu@redhat.com>


enp0s3:
/lgtm
/approve
/hold
waiting for CI pass

PR #664, opened 2021-08-17
cnao, Bump CNAO to 0.58.0

RamLavi:
This PR bumps CNAO to 0.58.0 on all currently provisioned k8s providers

### DIFF
--- a/cluster-provision/k8s/1.19/manifests/cnao/namespace.yaml
+++ b/cluster-provision/k8s/1.19/manifests/cnao/namespace.yaml
@@ -5,3 +5,4 @@ metadata:
   name: cluster-network-addons
   labels:
     name: cluster-network-addons
+    openshift.io/cluster-monitoring: "true"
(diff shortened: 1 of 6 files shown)


RamLavi:
@EdDev do we want to bump it on k8s-1.20?

EdDev:
> @EdDev do we want to bump it on k8s-1.20?

 I would prefer to bump on all active ones so we do not need to filter the future test. Do you see a problem with this?
 The active ones are 1.19, 1.20 and 1.21.  1.22 is in the making I think.

RamLavi:
> > @EdDev do we want to bump it on k8s-1.20?
> 
> I would prefer to bump on all active ones so we do not need to filter the future test. Do you see a problem with this?
> The active ones are 1.19, 1.20 and 1.21. 1.22 is in the making I think.

shouldn't be a problem and if we do, it's a bug.. will update them all

RamLavi:
/retest
`check-provision-k8s-1.20-cgroupsv2` failed on `should allow regular network connection IPv6 with a specific port number [IPv6]` but I don't see how it could be relevant to the bump.

RamLavi:
Hey @rmohr, where can I find conformance tests code?
do you think these tests failing is related to this change?

rmohr:
/retest

@RamLavi there was an issue with not available test images which should be resolved.

EdDev (approved):
Thanks!

enp0s3 (approved):
Thank you!
/approve

PR #669, opened 2021-08-31
Kind infra, Increase PodPreset enabled validation timeout

ormergi:
Since we upgraded kind node image to k8s-1.19, 
it seems that it takes more time for changes on `kube-apiserver.yaml` to propagate (e.g: enabling `PodPreset`).

Unfortunately there is no API to enable and verify that the `PodPreset` feature-gate is enabled.
As for today we check `kube-apiserver` process command on control-plane node to validate that the `PodPeset` feature is enabled.

This workaround is temporary, once  https://github.com/kubernetes-sigs/kind/issues/2318 will land at kubevirtci we can stop using `PodPreset`.

### DIFF
--- a/cluster-up/cluster/kind/podpreset.sh
+++ b/cluster-up/cluster/kind/podpreset.sh
@@ -4,6 +4,8 @@ set -e
 
 source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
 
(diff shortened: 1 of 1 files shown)


EdDev (approved):
Thanks!

fgimenez (approved):
/approve

ormergi:
Thank you all :slightly_smiling_face:

PR #670, opened 2021-09-01
add get_kubectl to external provider

zhlhahaha:
As kubectl binary is not always provided in the host machine,
add get_kubectl function, which can facilitate e2e test on
external provider.

Signed-off-by: Howard Zhang <howard.zhang@arm.com>

### DIFF
--- a/cluster-up/cluster/external/README.md
+++ b/cluster-up/cluster/external/README.md
@@ -11,6 +11,7 @@ The build machine should be a client of the cluster.
 export KUBEVIRT_PROVIDER=external
 export DOCKER_PREFIX=myregistry:5000/kubevirt
 export KUBECONFIG=mycluster.conf
+export KUBEVERSION=v1.19.7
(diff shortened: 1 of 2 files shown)


cluster-up/cluster/external/README.md
```diff
@@ -11,6 +11,7 @@ The build machine should be a client of the cluster.
 export KUBEVIRT_PROVIDER=external
 export DOCKER_PREFIX=myregistry:5000/kubevirt
 export KUBECONFIG=mycluster.conf
+export KUBEVERSION=v1.19.7
```
rmohr, line 14: Hm, the assumption is that we don't care about the version of the external cluster. With this we break this assumption.

If you could find a way to probe the cluster for the version and then always download the right kubectl it may be nice, but if that is not possible, I would prefer that we keep the task of getting the right kubectl outside of the provider.

PR #672, opened 2021-09-02
Add setup to emulate NVMe device

alicefr:
The setup allows to create and emulate in the provisioned nodes NVMe
device. This can be particularly useful for testing scenarios with
NVMe like NVMe passthrough.

How to enable it during the cluster creation:
```bash
$ export KUBEVIRT_PROVIDER_EXTRA_ARGS="--nvme 1G --nvme 500M"
$ make cluster-up
[...]
$ cluster-up/ssh.sh node01
WARNING: KUBEVIRTCI_GOCLI_CONTAINER is set and will take precedence over the also set KUBEVIRTCI_TAG
2021/09/02 10:34:58 Waiting for host: 192.168.66.101:22
2021/09/02 10:34:58 Connected to tcp://192.168.66.101:22
Last login: Thu Sep  2 10:24:50 2021 from 192.168.66.2
[vagrant@node01 ~]$ sudo lspci -nnD | grep NVM
0000:00:05.0 Non-Volatile memory controller [0108]: Intel Corporation QEMU NVM Express Controller [8086:5845] (rev 02)
[vagrant@node01 ~]$ lsblk
NAME    MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
sr0      11:0    1 1024M  0 rom  
vda     252:0    0   35G  0 disk 
└─vda1  252:1    0   35G  0 part /
nvme0n1 259:0    0    1G  0 disk 
nvme1n1 259:1    0  500M  0 disk 

````

### DIFF
--- a/cluster-provision/centos8/scripts/vm.sh
+++ b/cluster-provision/centos8/scripts/vm.sh
@@ -20,6 +20,7 @@ while true; do
     -n | --next-disk ) NEXT_DISK="$2"; shift 2 ;;
     -b | --block-device ) BLOCK_DEV="$2"; shift 2 ;;
     -s | --block-device-size ) BLOCK_DEV_SIZE="$2"; shift 2 ;;
+    -n | --nvme-device-size ) NVME_DISK_SIZES+="$2 "; shift 2 ;;
     -- ) shift; break ;;
     * ) break ;;
   esac
@@ -107,6 +108,14 @@ if [ -n "${BLOCK_DEV}" ]; then
   qemu-img create -f qcow2 ${BLOCK_DEV} ${block_device_size}
   block_dev_arg="-drive format=qcow2,file=${BLOCK_DEV},if=virtio,cache=unsafe"
 fi
+
+disk_num=0
+for size in ${NVME_DISK_SIZES[@]}; do
+  echo "Creating disk "$size" for NVMe disk emulation"
+  disk="/nvme-"${disk_num}".img"
+  qemu-img create -f raw $disk $size
+  let "disk_num+=1"
+done
 
 exec qemu-system-x86_64 -enable-kvm -drive format=qcow2,file=${next},if=virtio,cache=unsafe ${block_dev_arg} \
   -device virtio-net-pci,netdev=network0,mac=52:55:00:d1:55:${n} \
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -49,8 +49,10 @@ systemctl restart docker
 EOF
 `
 const etcdDataDir = "/var/lib/etcd"
+const nvmeDiskImagePrefix = "/nvme"
 
 var cli *client.Client
+var nvmeDisks []string
 
 type dockerSetting struct {
 	Proxy string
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: cluster-provision/centos8/scripts/vm.sh [hunk 1]
```diff
@@ -119,4 +129,4 @@ exec qemu-system-x86_64 -enable-kvm -drive format=qcow2,file=${next},if=virtio,c
   -serial pty -M q35,accel=kvm,kernel_irqchip=split \
   -device intel-iommu,intremap=on,caching-mode=on -soundhw hda \
   -uuid $(cat /proc/sys/kernel/random/uuid) \
-  ${QEMU_ARGS}
+  ${QEMU_ARGS} ${NVME_QEMU_ARG}
```
qinqon, line 132: Why do we need to touch vm.sh ? The NVMe qemu arguments must be after the other arguments ? If not this can be implemented just at gocli passing the NVMe arguments to QEMU_ARGS.

qinqon (changes requested):

cluster-provision/centos8/scripts/vm.sh
alicefr: @qinqon we need to create the image before calling qemu. For the rest, you are right I can pass the arguments as extra qemu args. I can also move the image creation in gocli

alicefr:
@qinqon @vladikr I added the setup to create multiple NVMe devices and I move the qemu argument in gocli.

cluster-provision/gocli/cmd/run.go
```diff
@@ -482,6 +485,16 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 			nodeQemuArgs = fmt.Sprintf("%s -device vfio-pci,host=%s", nodeQemuArgs, gpuAddress)
 		}
 
+		var vmArgsNvmeDisks []string
+		if len(nvmeDisks) > 0 {
```
qinqon, line 489: Would be possible to execute `qemu-img create` command at gocli so we don't touch vm.sh ? maybe adding additional steps at the vmContainerConfig Cmd:.

We want to move as much bash code to gocli as possible.

qinqon (changes requested):

cluster-provision/gocli/cmd/run.go
alicefr: no because the gocli is executed in a separate container and it only sets the command to run in the node container. If we want to remove the script then we need some kind of program that executes the commands inside the node container. The current etup is not different than the way how the script creates the image for block devices (https://github.com/kubevirt/kubevirtci/blob/main/cluster-provision/centos8/scripts/vm.sh#L85)

alicefr: `qemu-img` needs to be executed in the container that provisions the node

cluster-provision/centos8/scripts/vm.sh
rmohr: @alicefr I think there is a pretty simple way to still do this at `cluster-up` in contrast to `cluster-provision` while keeping most of your changes like they are. The `vm.sh` script is executed in both phases. You could move the `gocli` flag from `provision` to `up` and the disks would then only be created when `vm.sh` is executed for `cluster-up`. Similar like you can sets different memory for  both phases.

alicefr: sorry @rmohr I don't quite understand the comment. Do you mean moving the `--nvme` flag from `gocli run` to `gocli provision`?

rmohr: No. Forget my comment. Everything is where it should be. I hod for some reason the impression that you provision the images during the provision phase. :+1:

PR #674, opened 2021-09-02
Add force flag, auto detect latest provision dir if not given

dhiller:
This is a preparation to use the script to update CDI whenever a new provider is created via the job.

It does two things:
 - adds force flag `-f` to be able to replace cdi manifests with newer version
 - auto detects the latest provider directory in case no argument is given

This way we can call the script after a new provider is created to update the CDI manifest to the latest version.

Signed-off-by: Daniel Hiller <dhiller@redhat.com>

/cc @oshoval @enp0s3

### DIFF
--- a/cluster-provision/k8s/fetch-latest-cdi.sh
+++ b/cluster-provision/k8s/fetch-latest-cdi.sh
@@ -2,16 +2,41 @@
 
 set -euo pipefail
 
-provision_dir="$1"
+force=
+while getopts ":f" opt; do
+    case "${opt}" in
+    f)
+        force=true
+        shift
+        ;;
+    \?)
+        usage
+        exit 1
+        ;;
+    esac
+done
+
+if [ "$#" -gt 0 ]; then
+    provision_dir="$1"
+else
+    provision_dir=$(find "$(readlink --canonicalize "$(dirname "$0")")" -mindepth 1 -maxdepth 1 -type d -regex '^.*[0-9]\.[0-9]+$' -regextype 'posix-extended' | sort -rV | head -1)
+fi
+
 [ ! -d $provision_dir ] && echo "directory $provision_dir does not exist!" && exit 1
 [ ! -d $provision_dir/manifests ] && echo "directory $provision_dir/manifests does not exist!" && exit 1
 
 tag_name=$(curl -s -L -f https://api.github.com/repos/kubevirt/containerized-data-importer/releases/latest | jq -r '.tag_name')
 
-if [ $(find "$provision_dir/manifests" -name "cdi-*.yaml" | wc -l) -gt 0 ]; then
-    echo "$provision_dir/manifests/cdi-*.yaml already exists!"
+if [ "$(find "$provision_dir/manifests" -name "cdi-*.yaml" | wc -l)" -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


oshoval (commented):
can you add please a flag that force specific version as well (line 44) ?
this way we can force update the the same version that kubevirt uses
kubevirt doesn't use latest always, for example now it uses 1.36.0 
and the latest is 37+

once we have this, we can add code in the provision that will check what kubevirt uses and fetch that version,
kubevirtci uses the CDI yaml only for prefetching the images, so it will make kubevirtci always aligned to the actual kubevirt it uses

dhiller:
> can you add please a flag that force specific version as well (line 44) ?
> this way we can force update the the same version that kubevirt uses
> kubevirt doesn't use latest always, for example now it uses 1.36.0
> and the latest is 37+
> 
> once we have this, we can add code in the provision that will check what kubevirt uses and verify that version is used using the script (else fail if there is a diff), and use the same method in order to actual bump when needed,
> kubevirtci uses the CDI yaml only for prefetching the images, so it will make kubevirtci always aligned to the actual kubevirt it uses
> 
> Its not mandatory, as we can achieve this purpose also without the script, it will just improve the script i think.

IMHO the long term goal should be that we add the CDI spinup here as we do with most of the other features (CNAO, Prometheus, etc) and kubevirt just consumes what is baked in here, as it does with the other things. In other words, I'd rather remove all the CDI spinup from KubeVirt and add a flag, like KUBEVIRT_DEPLOY_CDI or the like here: https://github.com/kubevirt/kubevirtci/blob/1ad0cbd6f8621e0d8583c921581960a46155098b/cluster-up/cluster/ephemeral-provider-common.sh#L93

What do others think?

/cc @rmohr @aglitke @awels

dhiller:
/close

Will create another one where I'll incorporate this change to update cdi manifest while also adding an install for cdi

oshoval:
> > can you add please a flag that force specific version as well (line 44) ?
> > this way we can force update the the same version that kubevirt uses
> > kubevirt doesn't use latest always, for example now it uses 1.36.0
> > and the latest is 37+
> > once we have this, we can add code in the provision that will check what kubevirt uses and verify that version is used using the script (else fail if there is a diff), and use the same method in order to actual bump when needed,
> > kubevirtci uses the CDI yaml only for prefetching the images, so it will make kubevirtci always aligned to the actual kubevirt it uses
> > Its not mandatory, as we can achieve this purpose also without the script, it will just improve the script i think.
> 
> IMHO the long term goal should be that we add the CDI spinup here as we do with most of the other features (CNAO, Prometheus, etc) and kubevirt just consumes what is baked in here, as it does with the other things. In other words, I'd rather remove all the CDI spinup from KubeVirt and add a flag, like KUBEVIRT_DEPLOY_CDI or the like here:
> 
> https://github.com/kubevirt/kubevirtci/blob/1ad0cbd6f8621e0d8583c921581960a46155098b/cluster-up/cluster/ephemeral-provider-common.sh#L93

https://github.com/kubevirt/kubevirt/pull/6339
I am already working on it

Note that i think we still need to align CDI images between kubevirt and kubevirtci, but it can be done on other PR

> 
> What do others think?
> 
> /cc @rmohr @aglitke @awels

PR #676, opened 2021-09-08
Add cdi deployment to kubevirtci clusters

dhiller:
# Main change

Add support for environment variable `KUBEVIRT_DEPLOY_CDI`  that triggers
a deployment of bundled cdi if set to `true`. This then uses the bundled cdi manifests
and installs them after the k8s cluster is running.

The aim here is to just `export KUBEVIRT_DEPLOY_CDI=true` before `cluster-up` and then
have cdi deployed when cluster-up is finished. With this change we can get rid of the
kubevirt/kubevirt manifest for cdi which we are using currently during cluster-deploy.

# Further changes

Split the cdi manifests into two files for operator and custom resource
again so we can install cdi in the usual way.

Change `cluster-provision/k8s/fetch-latest-cdi.sh` so that it creates two files again.
Also adds a `-f` flag that removes existing cdi manifests and replaces them
with the newer ones instead.
Also if the provision directory is omitted
the latest provision directory is chosen as default. Together we can now
use it to automate cdi manifest updates.

### DIFF
--- a/cluster-provision/k8s/1.22/manifests/cdi-1.37.1-operator.yaml
+++ b/cluster-provision/k8s/1.22/manifests/cdi-1.37.1-operator.yaml
@@ -1,20 +1,3 @@
----
-apiVersion: cdi.kubevirt.io/v1beta1
-kind: CDI
-metadata:
-  name: cdi
-spec:
-  imagePullPolicy: IfNotPresent
-  infra:
-    nodeSelector:
-      kubernetes.io/os: linux
-    tolerations:
-    - key: CriticalAddonsOnly
-      operator: Exists
-  workload:
-    nodeSelector:
-      kubernetes.io/os: linux
----
 apiVersion: v1
 kind: Namespace
 metadata:
@@ -25,12 +8,10 @@ metadata:
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
-  labels:
-    operator.cdi.kubevirt.io: ""
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.3.0
   name: cdis.cdi.kubevirt.io
 spec:
-  conversion:
-    strategy: None
   group: cdi.kubevirt.io
   names:
     kind: CDI
@@ -2005,6 +1986,13 @@ rules:
 - apiGroups:
   - cdi.kubevirt.io
   resources:
+  - datasources
+  verbs:
+  - list
+  - get
+- apiGroups:
+  - cdi.kubevirt.io
+  resources:
   - cdis
   verbs:
   - get
@@ -2231,24 +2219,24 @@ spec:
         - name: DEPLOY_CLUSTER_RESOURCES
           value: "true"
         - name: OPERATOR_VERSION
-          value: v1.36.0
+          value: v1.37.1
         - name: CONTROLLER_IMAGE
-          value: quay.io/kubevirt/cdi-controller:v1.36.0
+          value: quay.io/kubevirt/cdi-controller:v1.37.1
         - name: IMPORTER_IMAGE
-          value: quay.io/kubevirt/cdi-importer:v1.36.0
+          value: quay.io/kubevirt/cdi-importer:v1.37.1
         - name: CLONER_IMAGE
-          value: quay.io/kubevirt/cdi-cloner:v1.36.0
+          value: quay.io/kubevirt/cdi-cloner:v1.37.1
         - name: APISERVER_IMAGE
-          value: quay.io/kubevirt/cdi-apiserver:v1.36.0
+          value: quay.io/kubevirt/cdi-apiserver:v1.37.1
         - name: UPLOAD_SERVER_IMAGE
-          value: quay.io/kubevirt/cdi-uploadserver:v1.36.0
+          value: quay.io/kubevirt/cdi-uploadserver:v1.37.1
         - name: UPLOAD_PROXY_IMAGE
-          value: quay.io/kubevirt/cdi-uploadproxy:v1.36.0
+          value: quay.io/kubevirt/cdi-uploadproxy:v1.37.1
         - name: VERBOSITY
           value: "1"
         - name: PULL_POLICY
           value: IfNotPresent
-        image: quay.io/kubevirt/cdi-operator:v1.36.0
+        image: quay.io/kubevirt/cdi-operator:v1.37.1
         imagePullPolicy: IfNotPresent
         name: cdi-operator
         ports:
--- a/cluster-provision/k8s/fetch-latest-cdi.sh
+++ b/cluster-provision/k8s/fetch-latest-cdi.sh
@@ -2,6 +2,20 @@
 
 set -euo pipefail
 
+function usage() {
+    cat <<EOF
+usage: $0 [-f] [<provision_dir>]
+
+        Downloads the latest containerized-data-importer manifests from the github release page and stores them into the manifests directory of the provision_dir.
+
+        If the provision_dir is omitted, it is assumed that the latest provision_dir should be the target
+
+options:
+    -f
+        forces the download of the manifests, existing manifests are deleted beforehand
+EOF
+}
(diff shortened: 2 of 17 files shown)


oshoval:
FYI already worked on https://github.com/kubevirt/kubevirt/pull/6339 as presented in the demo

dhiller:
@aglitke @awels @mhenriks  I noticed that we have the cdi namespace parameterized at kubevirt/kubevirt cdi manifest, but I didn't see any usage of the `cdi_namespace` anywhere? Do we need this here so that we can change it during tests?

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -80,4 +80,14 @@ function up() {
...
     fi
+
+    if [ "$KUBEVIRT_DEPLOY_CDI" == "true" ]; then
+        $kubectl create -f /opt/cdi-*-operator.yaml
+        $kubectl create -f /opt/cdi-*-cr.yaml
+        while [ $($kubectl get pods --namespace cdi | grep 'cdi-' | wc -l) -lt 4 ]; do
+            $kubectl get pods --namespace cdi
+            sleep 10
+        done
+        until $kubectl wait --for=condition=Ready pod --timeout=60s --all --namespace cdi; do sleep 10; done
```
oshoval, line 91: maybe we want to make sure this doesnt drift forever in case of an error ? (according lane max time)
not a must

oshoval (changes requested):
Thanks

cluster-provision/k8s/fetch-latest-cdi.sh
```diff
@@ -2,24 +2,60 @@
 
 set -euo pipefail
 
-provision_dir="$1"
-[ ! -d $provision_dir ] && echo "directory $provision_dir does not exist!" && exit 1
-[ ! -d $provision_dir/manifests ] && echo "directory $provision_dir/manifests does not exist!" && exit 1
+force=
```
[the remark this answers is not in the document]
dhiller: IIRC kubevirt should run with whatever latest version is available currently. IMHO it makes more sense to just integrate the latest version and not bother in kubevirt itself what version is used.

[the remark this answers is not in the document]
oshoval: so far kubevirt uses 1.36.0 even that for some time 1.37 exists

I personally prefer to separate logic like the two here, wdyt about asking Storage team ?
they need to know about it anyway i think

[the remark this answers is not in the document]
dhiller: like said before, our main goal is to relieve anyone on explicitly having to install a certain manifest and keep versions in sync. By using what is bundled with the kubevirtci providers users need not care about the version, while they still can opt out of the install and install any other version they want.

Makes sense?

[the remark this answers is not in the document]
rmohr: @oshoval I think this was now explained enough. We will proceed with the bundling. Please ping me offline if you really feel that you have to continue discussing this.

cluster-provision/k8s/1.22/manifests/cdi-1.37.1-operator.yaml
```diff
@@ -25,12 +8,10 @@ metadata:
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
-  labels:
-    operator.cdi.kubevirt.io: ""
+  annotations:
```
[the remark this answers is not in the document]
dhiller: yes

dhiller:
/cc @jean-edouard 

Hey, would be great if you could take a look at this, please! Thanks!

cluster-up/hack/common.sh
```diff
@@ -26,6 +26,7 @@ KUBEVIRT_DEPLOY_PROMETHEUS=${KUBEVIRT_DEPLOY_PROMETHEUS:-false}
 KUBEVIRT_DEPLOY_PROMETHEUS_ALERTMANAGER=${KUBEVIRT_DEPLOY_PROMETHEUS_ALERTMANAGER-false}
 KUBEVIRT_DEPLOY_GRAFANA=${KUBEVIRT_DEPLOY_GRAFANA:-false}
 KUBEVIRT_CGROUPV2=${KUBEVIRT_CGROUPV2:-false}
+KUBEVIRT_DEPLOY_CDI=${KUBEVIRT_DEPLOY_CDI:-false}
```
oshoval, line 29: You might want to have default as true

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -83,4 +83,14 @@ function up() {
             exit 1
         fi
     fi
+
+    if [ "$KUBEVIRT_DEPLOY_CDI" == "true" ]; then
```
oshoval, line 87: since its a boolean, no need quotes here

cluster-up/hack/common.sh
dhiller: No, because in general we want to opt in, as not only kubevirt/kubevirt is using this.

cluster-up/cluster/k8s-provider-common.sh
dhiller: IIRC bash doesn't have the concept of boolean itself when declaring variables, no?

cluster-provision/k8s/fetch-latest-cdi.sh
```diff
@@ -2,24 +2,60 @@
...
+while getopts ":f" opt; do
+    case "${opt}" in
+    f)
+        force=true
+        shift
+        ;;
+    \?)
+        usage
```
enp0s3, line 13: @dhiller I think the implementation of `usage` is missing

enp0s3 (changes requested):
Thank you! a small comment below, otherwise looks good

cluster-provision/k8s/fetch-latest-cdi.sh
dhiller: Good catch, thanks!

dhiller:
@enp0s3 addressed review comments, PTAL, thanks!

xpivarc (approved):

enp0s3:
@dhiller From I see in the logs it looks like an unpredictable behavior of the `ip-reconciler`

cluster-up/cluster/k8s-provider-common.sh
enp0s3: @oshoval I think the timeout is already defined in the command in the flag `--timeout=60s`

enp0s3 (approved):
/approve
CI failures are related to the flakiness of `ip-reconciler`, not related to PR changes.

dhiller:
> /approve
> CI failures are related to the flakiness of `ip-reconciler`, not related to PR changes.

Going to check that, maybe we can improve there.

dhiller:
Ha, this one is it:
```
default                  ip-reconciler-27206505--1-v5khn                       0/1     Completed   0          61s  
```

It's created by a cron job, will try to find out for how to exclude this pod from the check

dhiller:
@enp0s3 I changed the check in check-cluster-up.sh to avoid including the ip-reconciler pods. Unfortunately there was no other option IMHO.

cluster-up/cluster/k8s-provider-common.sh
oshoval: it is until forever loop in case the command fails, because the do part is conditionless
it will just sleep and retry
the timeout is per try, which act as the condition

cluster-up/hack/common.sh
oshoval: understood from @rmohr otherwise, in case i understood correct

cluster-up/cluster/k8s-provider-common.sh
oshoval: not sure, as it will work also on other strings indeed

PR #677, opened 2021-09-13
Adjust KCI for k8s-1.22 provider

enp0s3:
Adjust the configuration files and the scripts to meet the k8s 1.22 requirements

Signed-off-by: Igor Bezukh <ibezukh@redhat.com>


oshoval (commented):
Nice, 
I remember something about EndpointSliceProxying that Quique worked with, might worth to run sig-network (unless already done), or verify with Quique that we are safe ?

enp0s3:
@oshoval Hello! Thank you for the review. I would like to test the lanes on the CI infra rather than doing it locally. My plan is to get this code merged, then bump KCI in KV, and then I will create a temp. PR and will start testing the 1.22 lanes. Currently the 1.22 lanes are optional in our prow configuration files. 

Regarding the `EndpointSliceProxying` I've found the PR of @qinqon where he disabled it and IIUC the reason was that it's not stable. @qinqon please keep me honest here.

oshoval:
> @oshoval Hello! Thank you for the review. I would like to test the lanes on the CI infra rather than doing it locally. My plan is to get this code merged, then bump KCI in KV, and then I will create a temp. PR and will start testing the 1.22 lanes. Currently the 1.22 lanes are optional in our prow configuration files.

I meant on CI, but np, since its a new lane, and not mandatory yet, it is fine to do in steps.

dhiller (approved):
/approve

Thanks for the fixes!

PR #680, opened 2021-09-15
Add support for realtime tuning kernel arguments

jordigilh:
This PR allows to define the amount of hugepages to be passed on to the kernel. The current implementation has the value hardcoded to 64 pages, which translate into 128Mi per node. This change is required to successfully boot up the fedora container disk used in the realtime functional tests in kubevirt, since the kernel is unable to boot with the 128Mi available in any node.

This PR adds 3 new flags to the `gocli run` cmd:
*  `hugepages-count`: Determines the number of hugepages to create. Defaults to 64. The value of this flag is captured via the environment variable `KUBEVIRT_HUGEPAGES_COUNT`
* `hugepages-2m`: When enabled, it uses `2M` page sizes. Defaults to true.
* `enable-realtime-scheduler`: When enabled, it configures each node to allow processes to be scheduled in realtime without time constrains by running the command: `sysctl kernel.sched_rt_runtime_us=-1`. Without this change, it is not possible to change the scheduling of the VM to use type FIFO and set the process priority to 1. The value of this flag is captured via the environment variable `KUBEVIRT_REALTIME_SCHEDULER`

/cc @vladikr @rmohr @davidvossel @pkliczewski

### DIFF
--- a/cluster-provision/centos8/scripts/kernel.args
+++ b/cluster-provision/centos8/scripts/kernel.args
@@ -1,1 +1,1 @@
-root=/dev/vda1 ro no_timer_check console=tty0 console=ttyS0,115200n8 net.ifnames=0 biosdevname=0 elevator=noop spectre_v2=off nopti hugepagesz=2M hugepages=64 intel_iommu=on modprobe.blacklist=nouveau
+root=/dev/vda1 ro no_timer_check console=tty0 console=ttyS0,115200n8 net.ifnames=0 biosdevname=0 elevator=noop spectre_v2=off nopti intel_iommu=on modprobe.blacklist=nouveau
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
@@ -33,8 +34,9 @@ import (
 	"github.com/alessio/shellescape"
 )
 
-const soundcardPCIID = "8086:2668"
-const proxySettings = `
+const (
+	soundcardPCIID = "8086:2668"
+	proxySettings  = `
 mkdir -p /etc/systemd/system/docker.service.d/
 
 cat <<EOT  >/etc/systemd/system/docker.service.d/proxy.conf
@@ -48,8 +50,11 @@ systemctl daemon-reload
 systemctl restart docker
 EOF
 `
-const etcdDataDir = "/var/lib/etcd"
-const nvmeDiskImagePrefix = "/nvme"
+	etcdDataDir         = "/var/lib/etcd"
+	nvmeDiskImagePrefix = "/nvme"
+
+	hugepagesSize2M = "2M"
+)
 
 var cli *client.Client
 var nvmeDisks []string
@@ -101,7 +106,9 @@ func NewRunCommand() *cobra.Command {
 	run.Flags().StringArrayVar(&nvmeDisks, "nvme", []string{}, "size of the emulate NVMe disk to pass to the node")
 	run.Flags().Bool("run-etcd-on-memory", false, "configure etcd to run on RAM memory, etcd data will not be persistent")
 	run.Flags().String("etcd-capacity", "512M", "set etcd data mount size.\nthis flag takes affect only when 'run-etcd-on-memory' is specified")
-
+	run.Flags().Uint("hugepages-count", 64, "number of hugepages to allocate")
+	run.Flags().Bool("hugepages-2m", true, "use 2M size for hugepages")
+	run.Flags().Bool("enable-realtime-scheduler", false, "configures the kernel to allow unlimited runtime for processes that require realtime scheduling")
 	return run
 }
 
@@ -245,6 +252,20 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 		return err
 	}
 	resource.MustParse(etcdDataMountSize)
+
+	hugepagesCount, err := cmd.Flags().GetUint("hugepages-count")
+	if err != nil {
+		return err
+	}
+
+	hugepages2M, err := cmd.Flags().GetBool("hugepages-2m")
+	if err != nil {
+		return err
+	}
+	realtimeSchedulingEnabled, err := cmd.Flags().GetBool("enable-realtime-scheduler")
+	if err != nil {
+		return err
+	}
 
 	cli, err = client.NewEnvClient()
 	if err != nil {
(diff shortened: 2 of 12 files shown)


cluster-provision/centos8/scripts/kernel.args
```diff
@@ -1 +1 @@
-root=/dev/vda1 ro no_timer_check console=tty0 console=ttyS0,115200n8 net.ifnames=0 biosdevname=0 elevator=noop spectre_v2=off nopti hugepagesz=2M hugepages=64 intel_iommu=on modprobe.blacklist=nouveau
\ No newline at end of file
+root=/dev/vda1 ro no_timer_check console=tty0 console=ttyS0,115200n8 net.ifnames=0 biosdevname=0 elevator=noop spectre_v2=off nopti hugepagesz=2M intel_iommu=on modprobe.blacklist=nouveau
```
rmohr, line 1: Since we already take out hugepages, let's also move out the `hugepagesz=2M` setting from here and move it up to `gocli`. We have thenall parts which one would potentially want to modify removed from here.

cluster-provision/k8s/1.20/nodes.sh
```diff
@@ -5,7 +5,10 @@ set -ex
 source /var/lib/kubevirtci/shared_vars.sh
 
 # Ensure that hugepages are there
-cat /proc/meminfo | sed -e "s/ //g" | grep "HugePages_Total:64"
+if [ -n $HUGEPAGES_COUNT ]; then
```
rmohr, line 8: Let's just remove this check. We know that we pass it to the kernel. That should be good enough. If we want to test that, I think we should do it with our conformance tests.

rmohr (changes requested):
Looks very good. Two suggestions.

cluster-provision/centos8/scripts/kernel.args
jordigilh: Sounds good. I'll parametrize it as well. Since I don't know which kind of environments it will run (x86, arm), I will not check the value of the page size to avoid locking it to certain platforms (x86). What do you think?

rmohr: > Sounds good. I'll parametrize it as well. Since I don't know which kind of environments it will run (x86, arm), I will not check the value of the page size to avoid locking it to certain platforms (x86). What do you think?

How about adding the following flag: `--hugepages-2m` instead of an extra parameter?

jordigilh: That would be easier indeed. I'll just add the flag and set the default value as true for now.

This means we're skipping support for `1G` and ppc64 arch page sizes, right?  [(ppc64 uses 4k and 16M)](https://www.kernel.org/doc/html/latest/admin-guide/mm/hugetlbpage.html).

jordigilh:
@rmohr finished implementing all changes. I've asked @vladikr for a review :smile:

vladikr:
Thanks @jordigilh 
Roman's comments have been addressed. It looks good to me.
/approve

cluster-provision/gocli/cmd/run.go
```diff
@@ -499,8 +519,15 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 		if len(nodeQemuArgs) > 0 {
 			additionalArgs = append(additionalArgs, "--qemu-args", shellescape.Quote(nodeQemuArgs))
 		}
-		if len(kernelArgs) > 0 {
-			additionalArgs = append(additionalArgs, "--additional-kernel-args", shellescape.Quote(kernelArgs))
+
+		var hugepagesAmount string
```
jean-edouard, line 523: Maybe I missed something but what's the point of this variable?
Why not just use `hugepagesSize2M` directly?

cluster-provision/gocli/cmd/run.go
```diff
@@ -706,6 +733,20 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 		}
 	}
 
+	if realtimeSchedulingEnabled {
+		nodeName := nodeNameFromIndex(1)
+		success, err := docker.Exec(cli, nodeContainer(prefix, nodeName), []string{
+			"/bin/bash",
+			"-c",
+			"ssh.sh sudo sysctl kernel.sched_rt_runtime_us=-1",
```
jean-edouard, line 741: That won't store the value into sysctl.conf, right?
Maybe we should also do that, in case the node ever reboots?

cluster-provision/gocli/cmd/run.go
```diff
@@ -792,7 +833,6 @@ func prepareDeviceForAssignment(cli *client.Client, nodeContainer, pciID, pciAdd
 func prepareEtcdDataMount(node string, etcdDataDir string, mountSize string) error {
 	var err error
 	var success bool
-
```
jean-edouard, line 795: Nit: unnecessary whitespace removal

cluster-provision/k8s/1.19/node01.sh
```diff
@@ -2,9 +2,6 @@
 
 set -ex
 
-# Ensure that hugepages are there
-cat /proc/meminfo | sed -e "s/ //g" | grep "HugePages_Total:64"
```
jean-edouard, line 6: Pedantic: `tr -d ' '`? :)

jean-edouard (commented):
Looks good, just have a few questions in-line, mostly nits.

cluster-provision/k8s/1.19/node01.sh
jordigilh: Yep, would have done the trick. Unfortunately these 2 lines are no longer there... they were casualties to this PR :smile: (or should be if I missed them in any of the other scripts :smile:  )

cluster-provision/gocli/cmd/run.go
jordigilh: Sure, I'll put it back :smile:

cluster-provision/gocli/cmd/run.go
jordigilh: I wanted to leave the structure there in case it was required to support other sizes... (I initially planned to do so). I can definitely trim the code.

cluster-provision/k8s/1.19/node01.sh
jean-edouard: Facepalm, ignore me, time for the week-end! :)

cluster-provision/gocli/cmd/run.go
jordigilh: I don't think that the node will reboot during a run for realtime tests, but I can add the setting to `/etc/sysctl.d/`.

jean-edouard (approved):

cluster-provision/gocli/cmd/run.go
jordigilh: Ok done: The value is now stored in `/etc/sysctl.d/realtime.conf` and loaded with `sysctl --system`. This change only applies  to k8s-1.22 since the functional tests will only run with this version onwards.

jordigilh:
@jean-edouard can I get another `lgtm`? Thanks! :smile:

cluster-provision/centos8/scripts/kernel.args
rmohr: I meant converging all `2m` settings into one single parameter and not adding a boolean. So that `--hugepages-2m=64` would add both kernel parameters. But happy if it works for you as is.

jordigilh: Makes sense and it looks cleaner (1 flag vs 2 flags). I'll raise a new PR to adjust to this suggestion. Should be painless :smile:

vladikr: We will still need another parameter if we will need to support 1G pages, won't we?

jordigilh: That's true, but I wonder if that case is coming anytime soon (disclaimer: I don't know). If there is no plans in the horizon to support other than 2MB sizes for functional testing, it might be a case of over engineering.
Otherwise, I'm happy to leave the structure in the code to support other sizes :smile:

PR #683, opened 2021-09-19
Improve PodPreset enablement validation

ormergi:
Kubevirt SRIOV and GPU lanes fails more often during cluster-up due to 
PodPreset enablement failures [link](https://search.ci.kubevirt.io/?search=FATAL%3A+failed+to+enable+PodPreset+admission+plugin&maxAge=336h&context=1&type=all&name=.*sriov%7C.*vgpu&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job)

This PR intent is to improve how we validate that PodPreset is enabled in order
to prevent such failures:
- Increase validation timeout
  It seems that when kube-apiserver is under pressure it takes 
  more than 60 seconds for new configurations to propagate.

- Wait for kube-system pods to be ready
   PodPreset is enabled changing by changing kube-apiserver pod
   manifest, which when changed triggers kube-system pods to restart
   along with the kube-apiserver pod.

### DIFF
--- a/cluster-up/cluster/kind/podpreset.sh
+++ b/cluster-up/cluster/kind/podpreset.sh
@@ -5,6 +5,7 @@ set -ex
 source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
 
 VALIDATE_PODPRESET_TIMEOUT="3m"
+SYSTEM_CONTAINERS_READY_TIMEOUT="8m"
 
 function podpreset::enable_admission_plugin() {
     local -r cluster_name=$1
@@ -37,6 +38,11 @@ EOT
     fi
 }
 
+function podpreset::wait_for_kube_system_components_ready() {
+    local -r timeout=$1
+    until _kubectl wait --for=condition=Ready pod --all -n kube-system --timeout "${timeout}"; do sleep 1 ;done
+}
+
 function podpreset::create_virt_launcher_fake_product_uuid_podpreset() {
     local -r namespace=$1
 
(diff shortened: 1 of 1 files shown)


fgimenez (commented):
/lgtm

Thanks!

cluster-up/cluster/kind/podpreset.sh
```diff
@@ -37,6 +38,11 @@ EOT
     fi
 }
 
+function podpreset::wait_for_kube_system_components_ready() {
+    local -r timeout=$1
```
dhiller, line 42: What does the `-r` flag do?

cluster-up/cluster/kind/podpreset.sh
```diff
@@ -20,7 +21,7 @@ function podpreset::validate_admission_plugin_is_enabled() {
     local -r wait_time=$2
     local -r control_plane_container="$cluster_name-control-plane"
 
-    if ! timeout "${wait_time}s" bash <<EOT
+    if ! timeout "${wait_time}" bash <<EOT
```
dhiller, line 24: I find this "closure pattern" a bit awkward. Is that a common pattern? Can't we express this with a normal function declaration and a call to it?

cluster-up/cluster/kind/podpreset.sh
ormergi: Its an option of `local`, it sets the variable to be readonly

cluster-up/cluster/kind/podpreset.sh
ormergi: Unfortunatly `timeout` won't work with a function name, its expects a binary to execute.
We workaround it by passing `bash <command>`
this way it will run the code we want in place.

dhiller: OK, thanks!

cluster-up/cluster/kind/podpreset.sh
dhiller: Thanks!

dhiller (approved):
/approve

PR #686, opened 2021-09-24
Delete ip reconciler job pods before and during the check

dhiller:
As there still is some flakiness (i.e ip-reconciler pods that are Completed), we delete the pods before AND during the check loop.

/cc @enp0s3 @jean-edouard

### DIFF
--- a/cluster-provision/k8s/check-cluster-up.sh
+++ b/cluster-provision/k8s/check-cluster-up.sh
@@ -6,85 +6,100 @@ set -exuo pipefail
 make -C ../gocli container
 
 CI=${CI:-"false"}
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 RUN_KUBEVIRT_CONFORMANCE=${RUN_KUBEVIRT_CONFORMANCE:-"true"}
 
 provision_dir="$1"
 provider="${provision_dir}"
 
-function cleanup {
-  cd "$DIR" && cd ../..
-  make cluster-down
+function cleanup() {
+    cd "$DIR" && cd ../..
+    make cluster-down
+}
+
+function delete_ip_reconciler_pods() {
+    set +e
+    pods="$(${ksh} get pods -n default | grep 'ip-reconciler' | awk '{ print $1; }' | tr "\n" " ")"
+    set -e
+    [ -z "${pods}" ] && return
+    set +e
+    ${ksh} delete --ignore-not-found=true pods ${pods}
+    set -e
 }
 
 export KUBEVIRTCI_GOCLI_CONTAINER=quay.io/kubevirtci/gocli:latest
 # check cluster-up
 (
-  ksh="./cluster-up/kubectl.sh"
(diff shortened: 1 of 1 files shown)


cluster-provision/k8s/check-cluster-up.sh
```diff
@@ -6,85 +6,98 @@ set -exuo pipefail
...
+    # whereabouts ip-reconciler cleanup: ip-reconciler cronjob pods must get removed directly after creation
+    ${ksh} patch cronjob ip-reconciler --type merge -p '{ "spec": { "jobTemplate": { "spec": { "ttlSecondsAfterFinished": 0 } } } }'
 
-    export SONOBUOY_EXTRA_ARGS="--plugin systemd-logs --plugin e2e"
-    hack/conformance.sh $conformance_config
-  fi
+    # whereabouts ip-reconciler cleanup: delete eventually lingering pods before and during the loop
+    delete_ip_reconciler_pods
+    until ${ksh} wait --for=condition=Ready pod --timeout=30s --all; do
+        delete_ip_reconciler_pods
```
enp0s3, line 62: @dhiller Hi. I see that you've decided to abandon your previous approach of using `timeout 210s bash -c "until ${ksh} wait --for=condition=Ready pod --timeout=30s --all; do delete_ip_reconciler_pods; sleep 1; done"` , wondering if it's because it may take more than 210 secs for the ip-reconciler pods to get cleaned?

dhiller: I did this as the inner bash scope would not know of the declared function. I thought about doing some kind of retry logic, but then thought that the test pod would get killed anyway. WDYT?

dhiller:
Meh, seems to still not work... https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/686/check-provision-k8s-1.21/1443550395556171776

/hold

cluster-provision/k8s/check-cluster-up.sh
enp0s3: @dhiller I see what you mean about the inner bash scope. I have another concern, are we sure that at this stage the ip-reconciler pods will always be in `Completed` state? Is the assumption that if they were still executing the actual job we won't get to here?

ormergi: @dhiller FYI you can export the function so it could be used from another subshell `export -f delete_ip_reconciler_pods`

cluster-provision/k8s/check-cluster-up.sh
```diff
@@ -6,85 +6,100 @@ set -exuo pipefail
...
 
-function cleanup {
-  cd "$DIR" && cd ../..
-  make cluster-down
+function cleanup() {
+    cd "$DIR" && cd ../..
+    make cluster-down
+}
+
+function delete_ip_reconciler_pods() {
```
rmohr, line 21: It may make sense to either fetch all unready pods and filter out pods which we don't care about, or to just patch the yaml in our repo by script when updating manifests.

A delete would be very specific to whereabouts and I have the feeling that other solutions would work in a more general fashion by similar complexity.

dhiller:
Abandoned for https://github.com/kubevirt/kubevirtci/pull/702

PR #688, opened 2021-09-26
rename env var for hugepages count

jordigilh:
Renames the environment variable to define the number of hugepages to allocate to be more meaningful with its purpose:
`KUBEVIRT_HUGEPAGES_2M` -> `KUBEVIRT_HUGEPAGES2M_COUNT`

@rmohr @vladikr can you take a look? I forgot to rename the variable in my last PR.


rmohr:
@jordigilh the current env name sounds good to me. Do we have to change it?

jordigilh:
> @jordigilh the current env name sounds good to me. Do we have to change it?

No need to change it if it's all good. I will discard this PR :smile:

PR #690, opened 2021-10-04
Fix container disk create script to run in a prow job

jordigilh:
This PR adds a waiting condition right after the `virt-install` command is triggered before it continues with the execution in the `customize-image.sh` script. 

The wait condition is needed when running this script as part of a prow job because the `virt-install` [command fails to attach to the VM's console](https://storage.googleapis.com/kubevirt-prow/logs/publish-fedora-realtime-container-image/1443601489757999104/build-log.txt) and reverts to running in the background. 

```
+ virt-install --memory 2048 --vcpus 2 --name provision-vm --disk fedora-realtime_build/copy-source-image.qcow2,device=disk --disk cloudinit.iso,device=cdrom --os-type Linux --os-variant fedora32 --virt-type kvm --graphics none --network default --import
error: Cannot run interactive console without a controlling TTY
```

Since the script was designed to expect the command to hang in the install process while executing, the script resumes the execution considering the installation command as completed, long before it is done.

You can test the behaviour in your local environment by using this additional flag option in `virt-install`:
```
  --autoconsole none
```

Using this flag, it will render an output similar to [this](https://storage.googleapis.com/kubevirt-prow/logs/publish-fedora-realtime-container-image/1446485605511860224/build-log.txt):
```
Customize image by booting a VM with
 the image and cloud-init disk
 press ctrl+] to exit
+ virt-install --memory 2048 --vcpus 2 --name provision-vm --disk fedora-realtime_build/copy-source-image.qcow2,device=disk --disk cloudinit.iso,device=cdrom --os-type Linux --os-variant fedora32 --virt-type kvm --graphics none --network default --import
error: Cannot run interactive console without a controlling TTY


Starting install...
Domain creation completed.
+ wait_for_install_to_complete
+ count=0
+ retries=100
++ virsh list --state-running --name
++ grep provision-vm
+ [[ provision-vm == '' ]]
+ [[ 0 -gt 100 ]]
+ sleep 5
...
...
++ virsh list --state-running --name
++ grep provision-vm
+ [[ provision-vm == '' ]]
+ [[ 35 -gt 100 ]]
+ sleep 5
+ count=36
++ virsh list --state-running --name
++ grep provision-vm
+ [[ '' == '' ]]
+ [[ 36 -gt 100 ]]
+ echo done
done
+ virsh destroy provision-vm
error: Failed to destroy domain provision-vm
error: Requested operation is not valid: domain is not running

+ true
+ virt-sysprep -d provision-vm --operations machine-id,bash-history,logfiles,tmp-files,net-hostname,net-hwaddr
[   0.0] Examining the guest ...
...
...
+ virsh undefine provision-vm
Domain 'provision-vm' has been undefined

+ qemu-img convert -c -O qcow2 fedora-realtime_build/copy-source-image.qcow2 fedora-realtime_build/provisioned-image.qcow2
+ cleanup
...
...
+ FULL_IMAGE_NAME=quay.io/kubevirt/fedora-realtime-container-disk:v20211008-22109a3
+ docker tag fedora-realtime:devel quay.io/kubevirt/fedora-realtime-container-disk:v20211008-22109a3
+ docker push quay.io/kubevirt/fedora-realtime-container-disk:v20211008-22109a3
```

@dhiller @fgimenez can you take a look? If you're OK with the changes, can you trigger the job to generate the fedora-realtime container disk? I'll test the generated image to see if it fixed the issue.

### DIFF
--- a/cluster-provision/images/vm-image-builder/customize-image.sh
+++ b/cluster-provision/images/vm-image-builder/customize-image.sh
@@ -11,6 +11,27 @@ function cleanup() {
   virsh undefine "${DOMAIN_NAME}" || true
 }
 
+VIRT_INSTALL_WAIT_INTERVAL=${VIRT_INSTALL_WAIT_INTERVAL:-5}
+VIRT_INSTALL_TIMEOUT=${VIRT_INSTALL_TIMEOUT:-"$((100*$VIRT_INSTALL_WAIT_INTERVAL))"}
+
+function wait_for_install_to_complete() {
+  local -r timeout_seconds=$1
+  local -r interval_seconds=$2
+  local count=0
+  local -r retries=$((timeout_seconds/interval_seconds))
+  until [[ $(virsh list --state-running --name | grep $DOMAIN_NAME) == "" || $count -gt $retries ]]; do
+    sleep $interval_seconds
+    count=$((count + 1))
+  done
+  if [[ $count -gt $retries ]]; then
+    echo "VM '$DOMAIN_NAME' is still in "$(virsh list --all |grep $DOMAIN_NAME |awk '{print $3}')" state after waiting for "$(expr $count \* $retries) "seconds"
+    return 1
+  else
+    echo "done"
+  fi
+}
+
+
 SOURCE_IMAGE_PATH=$1
 OS_VARIANT=$2
 CUSTOMIZE_IMAGE_PATH=$3
@@ -39,6 +60,8 @@ virt-install \
   --graphics none \
   --network default \
   --import
+
+wait_for_install_to_complete $VIRT_INSTALL_TIMEOUT $VIRT_INSTALL_WAIT_INTERVAL
 
 # Stop VM
 virsh destroy $DOMAIN_NAME || true
--- a/cluster-provision/images/vm-image-builder/fedora-realtime/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-realtime/cloud-config
@@ -35,17 +35,18 @@ write_files:
           [Install]
           WantedBy=cloud-init.service
 runcmd:
-  - sudo systemctl daemon-reload
+  - sudo -i
+  - systemctl daemon-reload
   - export FEDORA_VERSION=$(cat /etc/redhat-release |awk '{print $3}')
-  - sudo -E dnf install -y http://ccrma.stanford.edu/planetccrma/mirror/fedora/linux/planetccrma/$FEDORA_VERSION/x86_64/planetccrma-repo-1.1-3.fc$FEDORA_VERSION.ccrma.noarch.rpm
-  - sudo dnf install -y kernel-rt tuned-profiles-realtime realtime-tests
-  - sudo grubby --set-default=$(ls /boot/vmlinuz-*rt*)
-  - sudo echo -e "isolated_cores=0,1\nisolate_managed_irq=Y" >/etc/tuned/realtime-variables.conf
-  - sudo dnf clean all
-  - sudo systemctl enable wait-for-cloud-init
-  - sudo systemctl enable qemu-guest-agent.service
-  - sudo hostnamectl set-hostname ""
-  - sudo hostnamectl set-hostname "" --transient
-  - sudo sed -i /users-groups/d /etc/cloud/cloud.cfg
-  - sudo sed -i 's/^SELINUX=.*/SELINUX=permissive/' /etc/selinux/config
-  - sudo shutdown
+  - dnf install -y http://ccrma.stanford.edu/planetccrma/mirror/fedora/linux/planetccrma/$FEDORA_VERSION/x86_64/planetccrma-repo-1.1-3.fc$FEDORA_VERSION.ccrma.noarch.rpm
+  - dnf install -y kernel-rt tuned-profiles-realtime realtime-tests
+  - grubby --set-default=$(ls /boot/vmlinuz-*rt*)
+  - grub2-mkconfig -o /boot/grub2/grub.cfg
+  - echo -e "isolated_cores=0,1\nisolate_managed_irq=Y" >/etc/tuned/realtime-variables.conf
+  - dnf clean all
+  - systemctl enable wait-for-cloud-init
+  - hostnamectl set-hostname ""
+  - hostnamectl set-hostname "" --transient
+  - sed -i /users-groups/d /etc/cloud/cloud.cfg
+  - sed -i 's/^SELINUX=.*/SELINUX=permissive/' /etc/selinux/config
+  - shutdown


cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -40,6 +55,8 @@ virt-install \
   --network default \
   --import
 
+wait_for_vm_state "shutoff"
```
ormergi, line 58: I think this should be optional in order to support local environments in case one would want to have live console to the provisioned VM.

ormergi (changes requested):
Hi Jordi,
Nice initiative! 
We did plan to run it on CI we didnt do it until now due to capacity.
Note that the vm-image-builder was meant to run on local environments.
In order to be able to run it on CI we need to containerize it; 
add dependencies and configure QEMU, Libvirt and virt-install to run inside a container
similar to what we do on virt-launcher.
Last time I tried it I had troubles with running `virt-install`.

cluster-provision/images/vm-image-builder/customize-image.sh
jordigilh: It works well in local environments, meaning that the only impact is that the script is be executed just once after the `virt-install` command completes but it does not modify flow in any other way. Initially I planned to hide the console output, but I thought that it added value for local runs/debugging, so I left it intact.
When running in a CI, this function will make sure that the execution does not continue until `virt-install` has completed executing.

jordigilh:
> Hi Jordi, Nice initiative! We did plan to run it on CI we didnt do it until now due to capacity. Note that the vm-image-builder was meant to run on local environments. In order to be able to run it on CI we need to containerize it; add dependencies and configure QEMU, Libvirt and virt-install to run inside a container similar to what we do on virt-launcher. Last time I tried it I had troubles with running `virt-install`.

Thanks!
The prow job that triggers the build [starts](https://github.com/kubevirt/project-infra/pull/1515/files#diff-3d91c222953e9c709dcf668b322dc1ee73f679af939354a5752d3cb78d3c98beR81-R82) the `libvirtd` and `virtlogd` daemons before starting the scripts. We did a [run](https://storage.googleapis.com/kubevirt-prow/logs/publish-fedora-realtime-container-image/1443601489757999104/build-log.txt) of the job and it worked fine in that scope, however the problem was that the image did not contain any of the required packages in the `cloud-config`. And that's the motivation for this PR :smile:

cluster-provision/images/vm-image-builder/customize-image.sh
ormergi: Since we use this function only here, I think its better to not accept arguments and put `shutoff` arg as part of the function and rename it to something like `wait_for_vm_shutoff`

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -11,6 +11,21 @@ function cleanup() {
   virsh undefine "${DOMAIN_NAME}" || true
 }
 
+function wait_for_vm_state() {
+  count=0
+  retries=100
+  until [[ $(virsh list --state-$1 --name |grep . ) =~ $DOMAIN_NAME || $count -gt $retries ]]; do
```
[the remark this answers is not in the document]
jordigilh: To remove the extra line that is generated by the `virsh list --state-shutoff --name`. Here is an example of the same command using the `--state-running`.
```
[root@localhost ~]# virsh list --name --state-running
provision-vm

[root@localhost ~]#
```

With `grep .`, it removes the extra line:
```
[root@localhost ~]# virsh list --name --state-running |grep .
provision-vm
[root@localhost ~]#
```

If you prefer a different way of removing the empty spaces, I'm all ears :smile:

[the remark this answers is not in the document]
ormergi: I think there is no need to use `[[` + `=~` which is basicly will perform string matching, 
we know that the domain name would be `DOMAIN_NAME` so `==` will do.
Also if you are going to use single brackets `[` you can specify `-o` instead of `||`.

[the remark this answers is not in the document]
jordigilh: > > if you prefer a different way of removing the empty spaces, I'm all ears ￼
> 
> From what I saw It will work w/o removing the extra line

Removing the extra line won't work with `==`. it works with the `=~` since it's a substring, but won't work with `==`. I'd much prefer to be concise and compare against the matching name instead of using a weak comparison with `=~`, And I can't get the other suggestion with `-o` to work.

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -11,6 +11,22 @@ function cleanup() {
   virsh undefine "${DOMAIN_NAME}" || true
 }
 
+function wait_for_install_to_complete() {
+  count=0
+  retries=100
+  until [[ $(virsh list --all --name|grep $DOMAIN_NAME) == "" || $(virsh list --state-shutoff --name|grep $DOMAIN_NAME) =~ "" || $count -gt $retries ]]; do
```
dhiller, line 17: Why do you check against an empty regular expression in the second condition?

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -11,6 +11,22 @@ function cleanup() {
...
+    sleep 5
+    count=$((count + 1))
+  done
+  if [[ $count -gt $retries ]]; then
+    echo "VM '$DOMAIN_NAME' is still in "$(virsh list --all |grep $DOMAIN_NAME |awk '{print $3}')" state after waiting for "`expr $count \* $retries` "seconds"
```
dhiller, line 22: Why do you use backticks here?

cluster-provision/images/vm-image-builder/customize-image.sh
jordigilh: Good catch. I'll compare against equals.

jordigilh: Done.

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -11,6 +11,21 @@ function cleanup() {
   virsh undefine "${DOMAIN_NAME}" || true
 }
 
+function wait_for_install_to_complete() {
+  count=0
+  retries=100
```
ormergi, line 16: I would change this function to get the total timeout and waiting interval  as arguments 
in a way we could set the timeout outside this function.

```
VIRT_INSTALL_TIMEOUT="$((60*8))"
...
function wait_for_install_to_complete() {
    local -r timeout_seconds=$1
    local -r interval_seconds=$2
    local -r count=0
    local -r retries=$((timeout_seconds\interval_seconds))
...
}
...
wait_for_install_to_complete "${VIRT_INSTALL_TIMEOUT}" "5"

It will also make it easier to tune it on CI directly from the prow-job yaml.

File: cluster-provision/images/vm-image-builder/publish-containerdisk.sh [hunk 6]
```diff
@@ -6,9 +6,20 @@ if [ "$#" -ne 2 ]; then
...
+}
+
 export IMAGE_NAME=$1
 export TAG=devel
 export FULL_IMAGE_NAME=$2
 
+if [[ ! "$FULL_IMAGE_NAME" =~ ":" ]]; then
+  FULL_IMAGE_NAME=$FULL_IMAGE_NAME":"$(get_image_tag)
+fi
+
```
ormergi, line 23: I think its better to keep the tag of the final image as part of part of `FULL_IMAGE_NAME`
And generate the commit tag outside this script, on CI use case as part of the prow-job command
So the usage would be: 
```
# vm-image-builder prow-job
...
commit="$(git rev-parse HEAD)"
date="$(date +%Y%m%d)"
tag="v${today}-${current_commit:0:7}"
...
./cluster-provision/images/vm-image-builder/publish-containerdisk.sh "fedora" "quay.io/kubevirtci/fedora:${tag}
```

ormergi (changes requested):

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -40,6 +55,8 @@ virt-install \
   --network default \
   --import
 
+wait_for_install_to_complete
+
```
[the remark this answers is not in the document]
jordigilh: I fail to see any impact with running in a local environment where the TTY is available: the script is only run once and it has no impact on the execution time because the loop will exit right away: You will only see the condition being checked once in the log. I'm not convinced this flag will add any improvement, but if you are keen to have a cleaner output I will add it there :smile:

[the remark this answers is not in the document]
jordigilh: no worries :smile:

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -11,6 +11,21 @@ function cleanup() {
   virsh undefine "${DOMAIN_NAME}" || true
 }
 
+function wait_for_install_to_complete() {
+  count=0
+  retries=100
+  until [[ $(virsh list --state-running --name|grep $DOMAIN_NAME) == "" || $count -gt $retries ]]; do
```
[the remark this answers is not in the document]
jordigilh: Done

ormergi:
> @ormergi if you don't have any more comments, can this PR be merged?

Thanks!
It looks like these changes are not reflected 
https://github.com/kubevirt/kubevirtci/pull/690#discussion_r725609705
https://github.com/kubevirt/kubevirtci/pull/690#discussion_r725611174

PR #692, opened 2021-10-13
Do not double mount /lib/modules for kind

fgimenez:
Kind by default mounts `/lib/modules`, here for docker https://github.com/kubernetes-sigs/kind/blob/a3076d6118bb01a65bc06839b4f7b98ee92ed139/pkg/cluster/internal/providers/docker/provision.go#L236 and here for podman https://github.com/kubernetes-sigs/kind/blob/a3076d6118bb01a65bc06839b4f7b98ee92ed139/pkg/cluster/internal/providers/podman/provision.go#L199. When running kind we were instructing the CRI to mount /lib/modules, this was ok for docker but podman complains about double mounts like here https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.19-sriov/1447596047848181760#1:build-log.txt%3A443

These changes remove the unnecessary configuration to mount `/lib/modules` (it is already done by default anyway).

/cc @ormergi @oshoval @dhiller  


oshoval (approved):
Nice thanks
It will still work locally as well because its mounted by kind right ?

PR #696, opened 2021-10-26
Add bump-cdi shell script

dhiller:
Adds the shell script that will bump the cdi manifests on all k8s
provider directories that are present.

Required for the periodic job that will automate this.

### DIFF
--- a/hack/bump-cdi.sh
+++ b/hack/bump-cdi.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+#
+# This file is part of the KubeVirt project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


dhiller:
FYI @ShellyKa13

This is the script I talked about regarding automation of cdi bumping. Related job that uses this script is here: https://github.com/kubevirt/project-infra/pull/1695

brybacki:
this is great, 
/lgtm

brybacki:
@dhiller @mhenriks do we think a bump to a specific version might be needed? Is yes, is there a a way to do it?

dhiller:
> @dhiller @mhenriks do we think a bump to a specific version might be needed? Is yes, is there a a way to do it?

No, there currently is no way to specify a version (see [here](https://github.com/kubevirt/kubevirtci/blob/main/cluster-provision/k8s/fetch-latest-cdi.sh#L42)).

Although this might not be hard to do (in the called script) I'd say that this might be out of scope for this PR.

Maybe we should create a separate PR for this if need be?

brybacki:
> Maybe we should create a separate PR for this if need be?
Agree, if there are good arguments for it, a new PR might add the option

enp0s3 (approved):
Thank you! I've verified the script locally
/approve

PR #697, opened 2021-10-26
Do not declare disk mount for node container

fgimenez:
This change will prevent errors during `make cluster-up` when using podman as the container runtime because of /var/run/disk being mounted as readonly in the node: 
```
XCreating disk "/var/run/disk/disk.qcow2 backed by /disk01.qcow2 with size 37580963840".
a+ echo 'Creating disk "/var/run/disk/disk.qcow2 backed by /disk01.qcow2 with size 37580963840".'
^+ qemu-img create -f qcow2 -o backing_file=/disk01.qcow2 /var/run/disk/disk.qcow2 37580963840
Qqemu-img: /var/run/disk/disk.qcow2: Could not create file: Read-only file system
�Formatting '/var/run/disk/disk.qcow2', fmt=qcow2 size=37580963840 backing_file=/disk01.qcow2 cluster_size=65536 lazy_refcounts=off refcount_bits=16
volume: k8s-1.20-node01
could not establish a connection to the node after a generous timeout: waiting for node to come up failed: Error response from daemon: can only create exec sessions on running containers: container state improper
```
and indeed when inspecting the container:
```
"Mounts": [
            {
                "Type": "volume",
                "Name": "k8s-1.20-node01",
                "Source": "/var/lib/containers/storage/volumes/k8s-1.20-node01/_data",
                "Destination": "/var/run/disk",
                "Driver": "local",
                "Mode": "",
                "Options": [
                    "nosuid",
                    "nodev",
                    "rbind"
                ],
                "RW": false,
                "Propagation": "rprivate"
            }
        ],
```

During provision we are creating a volume in the node at `/var/run/disk` here https://github.com/kubevirt/kubevirtci/blob/main/cluster-provision/gocli/cmd/provision.go#L162-L164. If we try to mount another volume in the `run` command we end up with a read only volume when using podman (docker seems to deal with this differently).  

Just running a simple container from `quay.io/kubevirtci/k8s-1.20:2110210919-bf9b729` shows that the volume is created and mounted even when we don't specify it when creating the container, both for podman and docker:
```
$ podman run --name test -ti --entrypoint /bin/bash quay.io/kubevirtci/k8s-1.20:2110210919-bf9b729
# in other shell
$ podman inspect test
.........................
"Mounts": [
            {
                "Type": "volume",
                "Name": "5b8cedcea483b197d61ad72b3fe5fa85de6d7b0e94c84c727f8697185fff3147",
                "Source": "/var/lib/containers/storage/volumes/5b8cedcea483b197d61ad72b3fe5fa85de6d7b0e94c84c727f8697185fff3147/_data",
                "Destination": "/var/run/disk",
                "Driver": "local",
                "Mode": "",
                "Options": [
                    "nodev",
                    "exec",
                    "nosuid",
                    "rbind"
                ],
                "RW": true,
                "Propagation": "rprivate"
            }
        ],
......................
``` 

/cc @rmohr 

### DIFF
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -268,7 +268,7 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 	ctx, cancel := context.WithCancel(b)
 
 	stop := make(chan error, 10)
-	containers, volumes, done := docker.NewCleanupHandler(cli, stop, cmd.OutOrStderr(), false)
+	containers, _, done := docker.NewCleanupHandler(cli, stop, cmd.OutOrStderr(), false)
 
 	defer func() {
 		stop <- retErr
@@ -467,14 +467,6 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 			nodeNum = fmt.Sprintf("%02d", (int(nodes) - x))
 		}
 
-		vol, err := cli.VolumeCreate(ctx, volume.VolumeCreateBody{
-			Name: fmt.Sprintf("%s-%s", prefix, nodeName),
-		})
-		if err != nil {
-			return err
-		}
-		volumes <- vol.Name
-
 		// assign a GPU to one node
 		var deviceMappings []container.DeviceMapping
 		if gpuAddress != "" && x == int(nodes)-1 {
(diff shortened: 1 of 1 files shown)


rmohr (approved):
/lgtm
/approve

It is already added in the provision phase.

PR #700, opened 2021-11-08
ks8-1.22, provision:add podSecurity feature

alonSadan:
podSecurity feature allows to enorce pod security standarts[1]
as an admission controller. In order to properly test this with
a kubevirtCi cluster this has to be enabled.

This commit adds this feature to the 1.22 lane.(the feature is only
available since 1.22)

This feature takes affect on the namesapce level, on namespaces that
are labeled with: `pod-security.kubernetes.io/<MODE>: <LEVEL>`.
So this change should have no side effects on tests that don't explicitly
ask for it.

Signed-off-by: alonsadan <asadan@redhat.com>


alonSadan:
The need for this feature has been reduced as the PR who needed it [was closed](https://github.com/kubevirt/cluster-network-addons-operator/pull/1074#issuecomment-964173953) So this can be added later when explicitly needed.

PR #701, opened 2021-11-10
Use configured kubectl instead of one in PATH

dhiller:
/cc @rhrazdil @Barakmor1

### DIFF
--- a/cluster-provision/centos8/scripts/vm.sh
+++ b/cluster-provision/centos8/scripts/vm.sh
@@ -46,7 +46,7 @@ n="$(printf "%02d" $(( 10#${NODE_NUM} )))"
 cat >/usr/local/bin/ssh.sh <<EOL
 #!/bin/bash
 set -e
-dockerize -wait tcp://192.168.66.1${n}:22 -timeout 300s
+dockerize -wait tcp://192.168.66.1${n}:22 -timeout 300s &>/dev/null
 ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no vagrant@192.168.66.1${n} -i vagrant.key -p 22 -q \$@
 EOL
 chmod u+x /usr/local/bin/ssh.sh
--- a/cluster-up/cluster/k8s-provider-common.sh
+++ b/cluster-up/cluster/k8s-provider-common.sh
@@ -51,7 +51,7 @@ function wait_for_istio_ready() {
         while [[ $retries -lt $max_retries ]]; do
             echo "waiting for istio-operator to be healthy"
             sleep 5
-            health=$(kubectl -n $istio_operator_ns get istiooperator istio-operator -o jsonpath="{.status.status}")
+            health=$($kubectl -n $istio_operator_ns get istiooperator istio-operator -o jsonpath="{.status.status}")
             if [[ $health == "HEALTHY" ]]; then
                 break
             fi


dhiller:
@Barakmor1 FYI

iholder-redhat:
Thanks @dhiller!
/lgtm

maiqueb (approved):
FWIW, I think this is fine.

fgimenez (approved):
/approve
/retest

Thanks!

dhiller:
@rhrazdil i see istio is failing on us on this fix:  https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/701/check-provision-k8s-1.20-cgroupsv2/1458403116712988672#1:build-log.txt%3A4347
do you have an idea why now the istio stuff is failing

maiqueb:
Couldn't help noticing this on the logs:

```bash
+ health='2021/11/10 12:17:00 Waiting for host: 192.168.66.101:22
2021/11/10 12:17:00 Connected to tcp://192.168.66.101:22
HEALTHY'
+ [[ 2021/11/10 12:17:00 Waiting for host: 192.168.66.101:22
2021/11/10 12:17:00 Connected to tcp://192.168.66.101:22
HEALTHY == \H\E\A\L\T\H\Y ]]
```

Why do we ssh into the node and `kubectl` from it ?

rhrazdil:
@dhiller Hey
From the logs, it seems that calling `kubectl` produces more output than we would like
see, this is how kubectl is called in `cluster-up/cluster/k8s-provider-common.sh`
`health=$(kubectl -n $istio_operator_ns get istiooperator istio-operator -o jsonpath="{.status.status}")`

And from the logs, the health variable contains:
```
health='2021/11/10 12:15:59 Waiting for host: 192.168.66.101:22
2021/11/10 12:15:59 Connected to tcp://192.168.66.101:22
HEALTHY'
```

So the Istio oprator is healthy, there is somehow extra stuff coming from `kubectl`

dhiller:
/hold

will first try to fix the health check problem, then remove the hold

dhiller:
> @dhiller Hey From the logs, it seems that calling `kubectl` produces more output than we would like see, this is how kubectl is called in `cluster-up/cluster/k8s-provider-common.sh` `health=$(kubectl -n $istio_operator_ns get istiooperator istio-operator -o jsonpath="{.status.status}")`
> 
> And from the logs, the health variable contains:
> 
> ```
> health='2021/11/10 12:15:59 Waiting for host: 192.168.66.101:22
> 2021/11/10 12:15:59 Connected to tcp://192.168.66.101:22
> HEALTHY'
> ```
> 
> So the Istio oprator is healthy, there is somehow extra stuff coming from `kubectl`

I am thinking this is the result of the ssh command itself, will try to make it silent. If that does not work, I'll go with the `tail -1` approach. WDYT?

maiqueb:
I'll ask again, since apparently https://github.com/kubevirt/kubevirtci/pull/701#issuecomment-965137012 went unnoticed: Why do we ssh into the node and kubectl from it ? @dhiller / @rhrazdil

dhiller:
> I'll ask again, since apparently [#701 (comment)](https://github.com/kubevirt/kubevirtci/pull/701#issuecomment-965137012) went unnoticed: Why do we ssh into the node and kubectl from it ? @dhiller / @rhrazdil

@maiqueb that's a really good question. At that point we should have configured everything to just use the outer kubectl. Will look into that.

dhiller:
> > I'll ask again, since apparently [#701 (comment)](https://github.com/kubevirt/kubevirtci/pull/701#issuecomment-965137012) went unnoticed: Why do we ssh into the node and kubectl from it ? @dhiller / @rhrazdil
> 
> @maiqueb that's a really good question. At that point we should have configured everything to just use the outer kubectl. Will look into that.

Now I remember: because the manifests that we need to deploy are inside the cluster. I don't know of a fast way besides changing a lot of the logic to make this work without that trick.

maiqueb:
> > > I'll ask again, since apparently [#701 (comment)](https://github.com/kubevirt/kubevirtci/pull/701#issuecomment-965137012) went unnoticed: Why do we ssh into the node and kubectl from it ? @dhiller / @rhrazdil
> > 
> > 
> > @maiqueb that's a really good question. At that point we should have configured everything to just use the outer kubectl. Will look into that.
> 
> Now I remember: because the manifests that we need to deploy are inside the cluster. I don't know of a fast way besides changing a lot of the logic to make this work without that trick.

I see. So the manifests that deploy the cluster are shipped **in** the virtualized nodes that will run the cluster. That's not a good dependency, as exhibited by the particular issue you're facing. 

I also can't see a simple way out of that.

dhiller:
> > > > I'll ask again, since apparently [#701 (comment)](https://github.com/kubevirt/kubevirtci/pull/701#issuecomment-965137012) went unnoticed: Why do we ssh into the node and kubectl from it ? @dhiller / @rhrazdil
> > > 
> > > 
> > > @maiqueb that's a really good question. At that point we should have configured everything to just use the outer kubectl. Will look into that.
> > 
> > 
> > Now I remember: because the manifests that we need to deploy are inside the cluster. I don't know of a fast way besides changing a lot of the logic to make this work without that trick.
> 
> I see. So the manifests that deploy the cluster are shipped **in** the virtualized nodes that will run the cluster. That's not a good dependency, as exhibited by the particular issue you're facing.
> 
> I also can't see a simple way out of that.

The upside of this is that kubevirtci is pretty self contained here, i.e. it does not need to fetch any manifests (think of versioning and such) but has it all contained. We maybe could extract those beforehand as we do with kubectl but I am not sure if that's much better...

rmohr:
> @dhiller Hey From the logs, it seems that calling `kubectl` produces more output than we would like see, this is how kubectl is called in `cluster-up/cluster/k8s-provider-common.sh` `health=$(kubectl -n $istio_operator_ns get istiooperator istio-operator -o jsonpath="{.status.status}")`
> 
> And from the logs, the health variable contains:
> 
> ```
> health='2021/11/10 12:15:59 Waiting for host: 192.168.66.101:22
> 2021/11/10 12:15:59 Connected to tcp://192.168.66.101:22
> HEALTHY'
> ```
> 
> So the Istio oprator is healthy, there is somehow extra stuff coming from `kubectl`

We were not too strict in the past in bash scripts, but this looks like there are just some messages which should go to stderr instead of stdout.

rmohr:
> > I see. So the manifests that deploy the cluster are shipped **in** the virtualized nodes that will run the cluster. That's not a good dependency, as exhibited by the particular issue you're facing.
> > I also can't see a simple way out of that.
> 
> The upside of this is that kubevirtci is pretty self contained here, i.e. it does not need to fetch any manifests (think of versioning and such) but has it all contained. We maybe could extract those beforehand as we do with kubectl but I am not sure if that's much better...

Yes that is the big benefit. The manifests are (potentially) verified in CI and/or pre-pulled. Since they are baked into the very specific kubevirtci container build it is best to keep it all together.

As I see it a progress message should just have been gone to stderr.

dhiller:
> As I see it a progress message should just have been gone to stderr.

Thanks! Will look into that.

dhiller:
> > As I see it a progress message should just have been gone to stderr.
> 
> Thanks! Will look into that.

TBH I don't have a grip on where the message comes from, regarding to the three levels of inception we have here (gocli/docker client/ssh into node).

Does it come from the call to the docker client in the cli? Or where else could it come from?

dhiller:
@rmohr @maiqueb I've redirected the stderr from dockerize to avoid this problem. PTAL, thanks!

maiqueb (approved):
Thanks.

PR #702, opened 2021-11-10
Use kustomize to patch ip reconciler job, applying a label for filtering

dhiller:
This is a cleaner approach compared to https://github.com/kubevirt/kubevirtci/pull/686, which tried to tackle this problem in a different way.

It essentially tackles problems when ip-reconciler pods originating from whereabouts are present, which hinder the ready check. Examples: https://search.ci.kubevirt.io/?search=waiting+for+the+condition+on+pods%2Fip-reconciler&maxAge=48h&context=1&type=build-log&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

We change the way of bumping whereabouts such as that we add kustomization and put the result into the manifests directory instead of copying the raw yamls.


jean-edouard (commented):
/approve
Nice to see there's a new release, allowing us to move away from master!
This looks good to me, just one question, what prompted the big re-indentation of `check-cluster-up.sh`? It's fine, but it makes the PR hard to read, especially since it's not in its own commit.

dhiller:
> /approve Nice to see there's a new release, allowing us to move away from master! This looks good to me, just one question, what prompted the big re-indentation of `check-cluster-up.sh`? It's fine, but it makes the PR hard to read, especially since it's not in its own commit.

This was me aligning to the default indentation that we use in other places. I think the other places already got adjusted, this one was missing.

maiqueb:
So we now plan on tracking whereabouts releases ? 

Shouldn't we push to have https://github.com/k8snetworkplumbingwg/whereabouts/pull/150 merged instead ? (IIUC, having that merged would remove the need for this PR). @dhiller

rmohr:
> So we now plan on tracking whereabouts releases ?
> 
> Shouldn't we push to have [k8snetworkplumbingwg/whereabouts#150](https://github.com/k8snetworkplumbingwg/whereabouts/pull/150) merged instead ? (IIUC, having that merged would remove the need for this PR). @dhiller

I don't see that as mutual exclusive. It is common that we have and had to patch things until they are fixed, or because we want different settings.

maiqueb:
> > So we now plan on tracking whereabouts releases ?
> > Shouldn't we push to have [k8snetworkplumbingwg/whereabouts#150](https://github.com/k8snetworkplumbingwg/whereabouts/pull/150) merged instead ? (IIUC, having that merged would remove the need for this PR). @dhiller
> 
> I don't see that as mutual exclusive. It is common that we have and had to patch things until they are fixed, or because we want different settings.

Correct me if I'm wrong, but here, we don't want different settings, do we ? We just want to patch things until they are fixed. Which imo means that this PR is redundant to https://github.com/k8snetworkplumbingwg/whereabouts/pull/150 . Or do we want to kustomize the whereabouts deployment further ?

rmohr:
> Correct me if I'm wrong, but here, we don't want different settings, do we ? We just want to patch things until they are fixed. Which imo means that this PR is redundant to [k8snetworkplumbingwg/whereabouts#150](https://github.com/k8snetworkplumbingwg/whereabouts/pull/150) . Or do we want to kustomize the whereabouts deployment further ?

It is just a workaround. If the PR gets in and a release gets out with it, we can just drop the changes, but This temporary fixing is a general pattern which we can use for workarounds like this. I am not sure I understand the problem.

maiqueb:
> > Correct me if I'm wrong, but here, we don't want different settings, do we ? We just want to patch things until they are fixed. Which imo means that this PR is redundant to [k8snetworkplumbingwg/whereabouts#150](https://github.com/k8snetworkplumbingwg/whereabouts/pull/150) . Or do we want to kustomize the whereabouts deployment further ?
> 
> It is just a workaround. If the PR gets in and a release gets out with it, we can just drop the changes, but This temporary fixing is a general pattern which we can use for workarounds like this. I am not sure I understand the problem.

There is no problem, since we both agree this is a temporal work-around.

dhiller:
@maiqueb @rmohr so I understand that we agree to include this temporary fix until the other PR gets merged and has been released. While that is not done, this PR would drastically reduce the flakyness of the check-provision lanes. So are there still problems with this PR that have not been mentioned here?

maiqueb (approved):
> @maiqueb @rmohr so I understand that we agree to include this temporary fix until the other PR gets merged and has been released. While that is not done, this PR would drastically reduce the flakyness of the check-provision lanes. So are there still problems with this PR that have not been mentioned here?

Not from my end. Let's proceed as you suggest above. I'll try to push the whereabouts PR through as well (the tricky thing will be getting a release out ...) . 

/hold

Feel free to uplift the hold (just not sure if @rmohr 's OK with this... seems so).

rmohr:
> Feel free to uplift the hold (just not sure if @rmohr 's OK with this... seems so).

I am :)

/unhold

maiqueb:
The [whereabouts PR](https://github.com/k8snetworkplumbingwg/whereabouts/pull/150) was merged.

We should get rid of this only after that code is released, right ?

PR #703, opened 2021-11-11
Add registry volume during provision

fgimenez:
Will prevent errors like https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-e2e-k8s-1.21-sig-network/1458811162673025024#1:build-log.txt%3A789, using podman as the container runtime makes the registry volume being mounted as readonly, this is the result of an inspect on the container created with the current version of gocli:
```
"Mounts": [
            {
                "Type": "volume",
                "Name": "k8s-1.20-registry",
                "Source": "/var/lib/containers/storage/volumes/k8s-1.20-registry/_data",
                "Destination": "/var/lib/registry",
                "Driver": "local",
                "Mode": "",
                "Options": [
                    "nosuid",
                    "nodev",
                    "rbind"
                ],
                "RW": false,
                "Propagation": "rprivate"
            }
        ],
``` 
Using a bind mount instead gives:
```
"Mounts": [
            {
                "Type": "bind",
                "Source": "/tmp/kubevirt_registry",
                "Destination": "/var/lib/registry",
                "Driver": "",
                "Mode": "",
                "Options": [
                    "rbind"
                ],
                "RW": true,
                "Propagation": "rprivate"
            }
        ],
``` 

/cc @rmohr @dhiller 

### DIFF
--- a/cluster-up/cluster/ephemeral-provider-common.sh
+++ b/cluster-up/cluster/ephemeral-provider-common.sh
@@ -72,7 +72,7 @@ function _registry_volume() {
 
 function _add_common_params() {
     # shellcheck disable=SC2155
-    local params="--nodes ${KUBEVIRT_NUM_NODES} --memory ${KUBEVIRT_MEMORY_SIZE} --cpu 6 --secondary-nics ${KUBEVIRT_NUM_SECONDARY_NICS} --random-ports --background --prefix $provider_prefix --registry-volume $(_registry_volume) ${KUBEVIRT_PROVIDER} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
+    local params="--nodes ${KUBEVIRT_NUM_NODES} --memory ${KUBEVIRT_MEMORY_SIZE} --cpu 6 --secondary-nics ${KUBEVIRT_NUM_SECONDARY_NICS} --random-ports --background --prefix $provider_prefix ${KUBEVIRT_PROVIDER} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
     if [[ $TARGET =~ windows_sysprep.* ]] && [ -n "$WINDOWS_SYSPREP_NFS_DIR" ]; then
         params=" --nfs-data $WINDOWS_SYSPREP_NFS_DIR $params"
     elif [[ $TARGET =~ windows.* ]] && [ -n "$WINDOWS_NFS_DIR" ]; then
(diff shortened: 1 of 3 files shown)


cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -66,8 +66,11 @@ EOF
 }
 
 function _registry_volume() {
-    # shellcheck disable=SC2154
-    echo ${job_prefix}_registry
+    local -r registry_dir=/tmp/${job_prefix}_registry
```
fgimenez, line 69: Can be /tmp mounted as tmpfs in some cases? Maybe better use `${HOME}`?

rmohr:
@fgimenez I know we talked about this already but I am still confused. I think one key point is that I only want this to be a volume to not write into the container overlay. Why do I now have to care about the specific location? maybe it just needs the rw mode extra passed?

fgimenez:
@rmohr In fact we have a `ReadOnly` bool field in the mount struct https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/api/types/mount/mount.go#L30, which should be set to false by default. I've tried to explicitly set it to false without any luck. 

I agree that having to specify the path of the bind mount is a pain, will try to create the volume during provision (same as we do with `/var/run/disk`)

fgimenez:
@rmohr changes pushed to create the registry volume during provision as we do with `/var/run/disk`, tested with quay.io/kubevirtci/gocli:podman-test19 and quay.io/kubevirtci/k8s-1.20:podman-test19 and works fine PTAL

rmohr:
> @rmohr changes pushed to create the registry volume during provision as we do with `/var/run/disk`, tested with quay.io/kubevirtci/gocli:podman-test19 and quay.io/kubevirtci/k8s-1.20:podman-test19 and works fine PTAL

Awesome. I guess all this confusion comes from the fact that we add the volumes and then snapshot the container during provision.

rmohr (approved):
> @rmohr changes pushed to create the registry volume during provision as we do with /var/run/disk, tested with quay.io/kubevirtci/gocli:podman-test19 and quay.io/kubevirtci/k8s-1.20:podman-test19 and works fine PTAL

/lgtm
/approve

/hold

Could you do one final check with `docker` if the volumes are also there created as expected? If that looks fine to, feel free to unhold.

fgimenez:
/unhold
/retest

thanks @rmohr tested successfully with docker, log here https://gist.github.com/fgimenez/87dfa36106d7a1b4b841bcae27188dc0

PR #704, opened 2021-11-12
network, multus: remove multus delegate config

maiqueb:
The delegate config provided via `ConfigMap` is not being used,
since the entrypoint script is being provided
the `--multus-conf-file=auto` option - this means the delegate
configuration is computed from the contents of the CNI config dir.

As such we can safely remove it, thus simplifying the
configuration.

Signed-off-by: Miguel Duarte Barroso <mdbarroso@redhat.com>


maiqueb:
/retest

Some port was already assigned on https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/704/check-provision-k8s-1.21/1460565084072841216

PR #707, opened 2021-11-16
(vfio) If at first we don't succeed, sleep 10 and try again

### LINKED ISSUE #457: KubeVirtCi failing on cluster-up with 'binding device to vfio driver failed'
Excerpt from the job log:
```
06:18:23:  + export QEMU_AUDIO_DRV=none
06:18:23:  + QEMU_AUDIO_DRV=none
06:18:23: � + exec qemu-system-x86_64 -enable-kvm -drive format=qcow2,file=/var/run/disk/disk.qcow2,if=virtio,cache=unsafe -device virtio-net-pci,netdev=network0,mac=52:55:00:d1:55:02 -netdev tap,id=network0,ifname=tap02,script=no,downscript=no -device virtio-rng-pci -vnc :02 -cpu host -m 10240M -smp 6 -serial pty -M q35,accel=kvm,kernel_irqchip=split -device intel-iommu,intremap=on,caching-mode=on -soundhw hda
06:18:23: 6 char device redirected to /dev/pts/0 (label serial0)
06:18:23: volume: pull-kubevirt-e2e-windows2016-node01
06:18:23: volume: pull-kubevirt-e2e-windows2016-node02
06:18:23: binding device to vfio driver failed
Makefile:113: recipe for target 'cluster-up' failed
make: *** [cluster-up] Error 1
```

Seen on two presubmits at least:

* https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/batch/pull-kubevirt-e2e-windows2016/1318796169635172352
* https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/4392/pull-kubevirt-e2e-k8s-1.18/1318654239735549952/build-log.txt

/cc @vladikr @jean-edouard

maya-r:
Attempt to fix #457 -- perhaps it's just slow to bind the device.

### DIFF
--- a/cluster-provision/k8s/1.19/bind_device_to_vfio.sh
+++ b/cluster-provision/k8s/1.19/bind_device_to_vfio.sh
@@ -50,6 +50,10 @@ fi
 # The device should now be using the vfio-pci driver
 new_driver=$(get_device_driver)
 if [[ $new_driver != "vfio-pci" ]]; then
-    echo "Error: Failed to bind to vfio-pci driver" 1>&2
-    exit 1
+    sleep 10
+    new_driver=$(get_device_driver)
+    if [[ $new_driver != "vfio-pci" ]]; then
+        echo "Error: Failed to bind to vfio-pci driver" 1>&2
+        exit 1
+    fi
 fi
--- a/cluster-provision/k8s/1.20/bind_device_to_vfio.sh
+++ b/cluster-provision/k8s/1.20/bind_device_to_vfio.sh
@@ -50,6 +50,10 @@ fi
 # The device should now be using the vfio-pci driver
 new_driver=$(get_device_driver)
 if [[ $new_driver != "vfio-pci" ]]; then
-    echo "Error: Failed to bind to vfio-pci driver" 1>&2
-    exit 1
+    sleep 10
+    new_driver=$(get_device_driver)
+    if [[ $new_driver != "vfio-pci" ]]; then
+        echo "Error: Failed to bind to vfio-pci driver" 1>&2
+        exit 1
+    fi
 fi
--- a/cluster-provision/k8s/1.21/bind_device_to_vfio.sh
+++ b/cluster-provision/k8s/1.21/bind_device_to_vfio.sh
@@ -50,6 +50,10 @@ fi
 # The device should now be using the vfio-pci driver
 new_driver=$(get_device_driver)
 if [[ $new_driver != "vfio-pci" ]]; then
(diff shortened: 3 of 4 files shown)


cluster-provision/k8s/1.19/bind_device_to_vfio.sh
```diff
@@ -50,6 +50,10 @@ fi
...
 if [[ $new_driver != "vfio-pci" ]]; then
-    echo "Error: Failed to bind to vfio-pci driver" 1>&2
-    exit 1
+    sleep 10
+    new_driver=$(get_device_driver)
+    if [[ $new_driver != "vfio-pci" ]]; then
+        echo "Error: Failed to bind to vfio-pci driver" 1>&2
+        exit 1
+    fi
 fi
```
oshoval, line 59: it might work, (similar solutions were done in the past)
maybe worth to use timeout until ... sleep 1 instead ? 
see cluster-up/cluster/kind/check-cluster-up.sh for an example please

maya-r: I think it's going to look worse because instead of blocking until it's true, we fail if it isn't.
so it will look like
timeout bash -c 'until [[ long-conditions ]]; do sleep 1; done || echo "failed" && exit
or something like that. it looks like a really long complicated line :)

oshoval: you dont need the exit,
the timeout will give an error, so scripts with -e will fail directly (in case script here is with -e)
`timeout -s10 bash -c 'until [[ long-conditions ]]; do sleep 1; done`
u need just this (added the -s10 or whatever the syntax is)

maya-r: so I tried this:
`timeout -v 10s bash -c "until [ x$(get_device_driver) = "xvfio-pci" ]; do sleep 1; done"`
and I'm not super sure about order of evaluating variables in bash because the output looks like this:
```
+ timeout -v 10s bash -c 'until [ x = xvfio-pci ]; do sleep 1; done'
timeout: sending signal TERM to command ‘bash’
```
It might only run this function once, and not have the desired effect of re-testing.

(I am not sure it's worth pursuing it further)

oshoval: $ timeout 5s bash -c "until false; do echo bla; sleep 1; done"
bla
bla
bla
bla
bla


it does repeat this way

maya-r:
doesn't seem to happen much now

PR #708, opened 2021-11-18
vgpu, Remove podpreset

oshoval:
Live migration is not supported yet on vgpu.
and the lane assumes there is only one node.

Podpreset is flaky a bit, so better to remove it
since it is not needed.

Once it will be needed we can use k8s-1.22 with new kind
version that supports built-in unique uuid.

Note:
`cluster-up/cluster/kind/podpreset.sh` can't be removed since other providers
still use it.

### DIFF
--- a/cluster-up/cluster/kind-1.19-vgpu/provider.sh
+++ b/cluster-up/cluster/kind-1.19-vgpu/provider.sh
@@ -13,10 +13,6 @@ else
     export HOST_PORT=$ALTERNATE_HOST_PORT
 fi
 
-#'kubevirt-test-default1' is the default namespace of
-# Kubevirt VGPU tests where the VGPU VM's will be created.
-VGPU_TESTS_NS="${VGPU_TESTS_NS:-kubevirt-test-default1}"
-
 function set_kind_params() {
     export KIND_VERSION="${KIND_VERSION:-0.11.1}"
     export KIND_NODE_IMAGE="${KIND_NODE_IMAGE:-quay.io/kubevirtci/kindest_node:v1.19.11@sha256:cbecc517bfad65e368cd7975d1e8a4f558d91160c051d0b1d10ff81488f5fb06}"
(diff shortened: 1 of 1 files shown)


oshoval:
/cc @vladikr 

btw it seems we don't have vgpu kubevirtci lane ?

vladikr:
```
/home/prow/go/src/github.com/kubevirt/kubevirt/cluster-up//cluster/kind-1.19-vgpu/provider.sh: line 47: /home/prow/go/src/github.com/kubevirt/kubevirt/cluster-up//cluster/kind/podpreset.sh: No such file
```

oshoval:
yep removed

oshoval:
job passed on kubevirt

vladikr:
Thanks @oshoval 
/lgtm

fgimenez (commented):
Looks very good to me, but shouldn't we remove it from `kind-1.19-sriov` provider too? https://github.com/kubevirt/kubevirtci/blob/main/cluster-up/cluster/kind-1.19-sriov/provider.sh#L66

fgimenez:
we no longer have a check-provision job for 1.19-sriov, but there are presubmit and periodics based on it. If we don't plan to use it maybe better remove the provider?

oshoval:
> Looks very good to me, but shouldn't we remove it from `kind-1.19-sriov` provider too? https://github.com/kubevirt/kubevirtci/blob/main/cluster-up/cluster/kind-1.19-sriov/provider.sh#L66

We still use sriov-1.19 on kubevirt, the fact that there aren't kubevirtci is just because flakes were found in the middle of removing it, please we actually still need sriov-1.19, until sriov-1.22 is not flaky
therefore we can't remove sriov-1.19 lane / can't remove the podpreset from there, since it need it for migration

sriov-1.22 won't need anymore podpreset because the new kind already have a unique uuid per node

oshoval:
> ok makes sense, then we can't remove `cluster-up/cluster/kind/podpreset.sh` in this PR right? it is being sourced from the 1.19-sriov provider.

Right, thanks !
returning it, it shows that we do need back the sriov-1.19 indeed

oshoval:
fixed, thanks

fgimenez (approved):
/approve

thanks!

oshoval:
Thanks Federico
@vladikr can you take please another look ?

PR #709, opened 2021-11-19
[wip] Remove non inclusive terms

fossedihelm:
Remove some non-inclusive terms from the codebase replacing it with `main` or `control-plane` depending on what it refers to:
- control-plane when it refers to nodes.
- main where possible if it refers to a branch.

Replace the usages of `node-role.kubernetes.io/master` with `node-role.kubernetes.io/control-plane`


dhiller (commented):
Looks great already! Thanks!

One note though: please be aware that the files in or below path `cluster-provision/k8s/1.x/manifests` may be overwritten whenever a bump of those files is required. So maybe we should rather not touch them but wait for bumping instead. What you could do though is ask the teams or authors that added those file for when the required changes are about to be made in the manifests, so that we know when that will happen.

cluster-up/cluster/kind/common.sh
```diff
@@ -150,12 +150,12 @@ function _get_pods() {
...
     # Due to inconsistent labels and taints state in multi-nodes clusters,
     # it is nessecery to remove taint NoSchedule and set role labels manualy:
-    #   Master nodes might lack 'scheduable=true' label and have NoScheduable taint.
+    #   Control-plane nodes might lack 'scheduable=true' label and have NoScheduable taint.
     #   Worker nodes might lack worker role label.
     master_nodes=$(_get_nodes | grep -i $MASTER_NODES_PATTERN | awk '{print $1}')
     for node in ${master_nodes[@]}; do
         # removing NoSchedule taint if is there
-        if _kubectl taint nodes $node node-role.kubernetes.io/master:NoSchedule-; then
+        if _kubectl taint nodes $node node-role.kubernetes.io/control-plane:NoSchedule-; then
```
oshoval, line 158: Thanks for the PR

About `node-role.kubernetes.io/master` (appears few times in the PR),
IIUC it is set by k8s `kubeadm init` itself when creating the cluster, so we can't replace reference to `node-role.kubernetes.io/control-plane`, unless we also set k8s itself to use it,
or at least rename this taint one time in the very beginning after the cluster is created ?

worth to add ok-to-test to let tests run

acardace: It should not be an issue, per https://v1-20.docs.kubernetes.io/docs/setup/release/notes/#changelog-since-v1-19-0 the new label is added by default so no changes to scripts creating the cluster should be required.

acardace (approved):

cluster-up/cluster/kind/common.sh
oshoval: Please see the jobs, they failed to find this taint

oshoval:
@qinqon 
please see here https://github.com/kubevirt/kubevirtci/pull/9/files#diff-2119e7bf277e9db8e8750506ac71a3f6d7e19f94acc48eadf70666461a30cf4e

It seems we created this file in the very beginning (or at least didnt change it since),
so it should be adapted to the image is uses, as it export some env vars with the words that need to be changed

cluster-up/cluster/kind/common.sh
fossedihelm: Thanks. I'll do some trying

acardace: Had a second read of the docs and actually the new label is applied but the new taint has not been added yet, I guess we'd have to keep the old taint reference for the time being.

fossedihelm: Thanks @oshoval @acardace I force pushed the changes without `node-role.kubernetes.io/control-plane:NoSchedule`

acardace (approved):

cluster-up/cluster/kind/common.sh
```diff
@@ -39,7 +39,13 @@ ETCD_IN_MEMORY_DATA_DIR="/tmp/kind-cluster-etcd"
 
 function _wait_kind_up {
     echo "Waiting for kind to be ready ..."
-    while [ -z "$(docker exec --privileged ${CLUSTER_NAME}-control-plane kubectl --kubeconfig=/etc/kubernetes/admin.conf get nodes --selector=node-role.kubernetes.io/master -o=jsonpath='{.items..status.conditions[-1:].status}' | grep True)" ]; do
+    # For providers with version > 1.19 we use control-plane node role, master otherwise
```
stu-gott, line 42: Does this check need to exist in perpetuity, or will we be able to retire it at some future point?

oshoval: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/709/check-up-kind-1.19-sriov/1469329208550363136

kind job indeed failed, changes might be related ?

fossedihelm: @stu-gott We have to keep this check as long as we use a k8s 1.19 based provider. After that it can be deleted.

fossedihelm: @oshoval the changes was made to make kind 1.19 job works again.But for some reason it seems does not work. I'm investigating on it. Thanks

fossedihelm: Found it. I pushed the changes. The check was wrong

cluster-up/cluster/kind/common.sh
```diff
@@ -39,7 +39,13 @@ ETCD_IN_MEMORY_DATA_DIR="/tmp/kind-cluster-etcd"
...
+    if [[ $KUBEVIRT_PROVIDER =~ k8s-1\.1.* ]] || [[ $KUBEVIRT_PROVIDER =~ kind-1\.1.* ]]; then
```
oshoval, line 43: this would work `if [[ $KUBEVIRT_PROVIDER =~ (k8s|kind)-1\.1.* ]]; then` i think as well
but on the other hand, i think we should not limit kind to specific versions, so maybe lets
change please `kind-1\.1.*` to simply `kind-.*` ?
so once we bump kind it won't stop working, because what matters here is that it is kind, not the version for now

fossedihelm: @oshoval Thanks so much for revision. Maybe I'm missing something, but I think that version is relevant. Only kind with version <1.20 doesn't have `control-plane` label. Indeed, in future k8s versions (and kind too) `master` will not be available anymore. In fact kind with version 1.22 works with `control-plane` selector. Anyway you are right about `if [[ $KUBEVIRT_PROVIDER =~ (k8s|kind)-1\.1.* ]]; then`. If you agree I will change the if with the latter, otherwise can you explain why version is not important? :slightly_smiling_face:  Thanks

oshoval: ah ok you are right then, thanks

oshoval: if so, you are right, but
i am just confused with one thing, we already have kind-1.22-sriov
and it still using master

another note, this function _wait_kind_up is only for kind, so you can remove the k8s regex i think ?

cluster-up/cluster/kind/common.sh
oshoval, line 42: i think we don't need please the comment, the code would be enough, (and soon to be outdated)

cluster-up/cluster/kind/common.sh
fossedihelm: `control-plane` label is added in parallel of `master` by kubeadm itself. In this co-living period it can be used master and control-plane indiscriminately. So maybe for the next providers I think that is better to use always `control-plane`. About the regex I left the k8s part because I see that there is a provider named `kind-k8s-1.19`, but we are not using it, so maybe I was too conservative. I will remove it. Thanks

oshoval: people from `arm` do use kind-k8s-1.19 indeed, but having `kind-.*` will include that as well
thanks

fossedihelm: If you think that something like `kind-k8s-1.21.1` should never happen ok, otherwise we can switch to `kind-(k8s-)?1\.1.*`

oshoval: we don't specify the Z stream as part of the providers
(since we bump the same provider with Z stream, but the major and minor are pinned)

fossedihelm: Perfect, I will use `kind-.*1\.1.*`. Thanks for information and support :slightly_smiling_face:

oshoval:
Thanks
/lgtm

There will be follow-up as mentioned here
https://github.com/kubevirt/kubevirtci/pull/709#pullrequestreview-812371647
CNAO
logging.yaml (we first need to make sure the current feature works, and then to update it)
Calico CNI (which we can't change as we depend on external repo)

@qinqon are you fine with it ?

dhiller (commented):
/approve

PR #714, opened 2021-11-28
k8s-1.2x, provision: mv original cni manifest

oshoval:
Since fetch-images.sh runs on /tmp folder,
it will try to fetch images that we mirrored to quay
from their original image repository docker.io.

Once we move the original manifest before patching it
instead of copying it, it will not exist anymore in the
/tmp folder and won't cause for unneeded pull request
that might fail due to rate limit.


enp0s3 (approved):
Thank you!
/lgtm
/approve

PR #724, opened 2021-12-16
KinD providers: Mount audit logs as read-only

ormergi:
Following #716 

On Kind clusters auditing at the node level is not supported.
We mount the host audit logs dir in order to enable Kubevirt test
suite reporter to dump the logs at the time a test fails.

The mount is misconfigured and should be mounted as read-only
to prevent logs corruption.

### DIFF
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -228,6 +228,7 @@ function _add_worker_extra_mounts() {
   extraMounts:
   - containerPath: /var/log/audit
     hostPath: /var/log/audit
+    readOnly: true
 EOF
 
     if [[ "$KUBEVIRT_PROVIDER" =~ sriov.* || "$KUBEVIRT_PROVIDER" =~ vgpu.* ]]; then


cluster-up/cluster/kind/common.sh
```diff
@@ -228,6 +228,7 @@ function _add_worker_extra_mounts() {
   extraMounts:
   - containerPath: /var/log/audit
     hostPath: /var/log/audit
+    readOnly: false
```
rmohr, line 231: `true` :)

rmohr (changes requested):

cluster-up/cluster/kind/common.sh
ormergi: Sorry my brain is locked I guess :facepalm:, fixed

ormergi:
Thanks :slightly_smiling_face:

oshoval:
you might want to close https://github.com/kubevirt/kubevirt/pull/6961 then ?
so this PR will be included in the bump ?

rmohr:
> you might want to close [kubevirt/kubevirt#6961](https://github.com/kubevirt/kubevirt/pull/6961) then ?
> so this PR will be included in the bump ?

The bot will do the bump eventually (and update the PR if already open). No need to close. It will also not propoagate it faster.

oshoval:
> The bot will do the bump eventually (and update the PR if already open). No need to close. It will also not propoagate it faster.

unless it is lgtmed or approved and then it doesnt update it

rmohr:
> unless it is lgtmed or approved and then it doesnt update it

Possible (I may have missed an update there then), but that would not be the normal flow which we have in general.

oshoval:
> > unless it is lgtmed or approved and then it doesnt update it
> 
> Possible (I may have missed an update there then), but that would not be the normal flow which we have in general. And if so I would wonder if we should revert that behaviour.

https://github.com/kubevirt/project-infra/pull/1528
imo it is needed, even empty rebase was causing it bump again and again and stall the merge for around a week

oshoval:
btw something is weird,
if you try to `make bump-kubevirtci` now that a new publish job has finished
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/publish-kubevirtci/1471457550510592000
(triggered as post submit of this PR IIUC)

it doesn't take this change, but it does take the previous #716 PR

PR #723, opened 2021-12-12
CI-cluster swap support

Barakmor1:
**_This PR enables swap on the nodes. This can be helpful for a number of use-cases, such as:_**
- Swap available for node-level performance tuning and stability/reducing noisy neighbour issues.
- Prevents failure on memory over commitment cases.
- Can be Useful for live migration of VMs https://github.com/kubernetes/kubernetes/issues/53533#issuecomment-754878431
- Require for further investigation of kubernetes swap https://kubernetes.io/blog/2021/08/09/run-nodes-with-swap-alpha/
(might reduce tests flakiness?)

**_description:_**
- Enable the NodeMemorySwap feature flag on the kubelet
- Set --fail-on-swap flag to false
- Swap is disabled by default 
- Can only work with kubevirtci version > 1.21

**_environment variables:_**
- KUBEVIRT_SWAP_ON-enable swap
- KUBEVIRT_SWAP_SIZE_IN_GB- Setting swap size
- KUBEVIRT_SWAPPINESS- Setting swappiness
- KUBEVIRT_UNLIMITEDSWAP- if set to 'true' Kubernetes workloads can use as much swap memory as they request, up to the system limit otherwise Kubernetes workloads can use as much swap memory as they request, up to the system limit

**_to verify that it works:_**
set the environment variables:
export KUBEVIRT_SWAP_ON=true
export KUBEVIRT_SWAP_SIZE_IN_GB=5

deploy the cluster

label node02:
kubectl label nodes node02 node02=true

create the following manifest : 

```
---
apiVersion: kubevirt.io/v1
kind: VirtualMachineInstance
metadata:
  labels:
    special: vmi-masquerade
  name: vmi-masquerade
spec:
  nodeSelector:
    node02: "true"
  domain:
    devices:
      disks:
      - disk:
          bus: virtio
        name: containerdisk
      - disk:
          bus: virtio
        name: cloudinitdisk
      interfaces:
      - masquerade: {}
        name: testmasquerade
        ports:
        - name: http
          port: 80
          protocol: TCP
      rng: {}
    resources:
      overcommitGuestOverhead: true
      requests:
        memory: 3Gi
    memory:
      guest: 6Gi
  networks:
  - name: testmasquerade
    pod: {}
  terminationGracePeriodSeconds: 0
  volumes:
  - containerDisk:
      image: registry:5000/kubevirt/fedora-with-test-tooling-container-disk:devel
    name: containerdisk
  - cloudInitNoCloud:
      networkData: |
        version: 2
[...]

PR #725, opened 2021-12-16
Configure the nodes and kubeadm to avoid failure when swap is enabled with version 1.23

Barakmor1:
Signed-off-by: Barak Mordehai <bmordeha@redhat.com>

The 1.23 provider landed before the following PR:
https://github.com/kubevirt/kubevirtci/pull/723
, so we should copy those changes to the 1.23 provider as well.

I removed the --ignore-preflight-errors=SWAP flags because according to:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md#:~:text=Lifecycle%20and%20Windows%5D-,Kubeadm%3A%20switch%20the%20preflight%20check%20(called%20%27Swap%27)%20that%20verifies%20if,via%20--ignore-preflight-errors%20or%20the%20kubeadm%20config.%20(%23104854%2C%20%40pacoxu),-Kubelet%20did%20not
it is no longer needed on versions >= 1.23.

### DIFF
--- a/cluster-provision/k8s/1.23/node01.sh
+++ b/cluster-provision/k8s/1.23/node01.sh
@@ -41,6 +41,9 @@ do
     sleep 2
 done
 
+# Disable swap
+sudo swapoff -a
+
 # 1.23 has deprecated --experimental-patches /provision/kubeadm-patches/, we now mention the patch directory in kubeadm.conf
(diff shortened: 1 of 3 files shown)


dhiller (approved):
/approve

Thanks for this!

PR #741, opened 2022-02-01
cluster-up: Use podman.socket's default path

0xFelix:
These changes set the path of the podman's socket to
'${XDG_RUNTIME_DIR}/podman/podman.socket' so podman can be started with
'systemctl start --user podman.socket' which is the standard way of
starting podman as user on distributions using systemd.

Also silence stderr output of `docker ps` command so behavior is uniform.

### DIFF
--- a/PODMAN.md
+++ b/PODMAN.md
@@ -3,7 +3,7 @@
 Install podman 3.1+, then run it in docker compatible mode:
 
 ```
-podman system service -t 0 unix:///${HOME}/podman.sock
+systemctl start --user podman.socket
 ```
 
 After this command, you can use kubevirtci with the typical `make cluster-*`
@@ -13,5 +13,11 @@ This will use `fuse-overlayfs` as storage layer. If the performance is not
 satisfactory, consider running podman as root to use plain `overlayfs2`:
 
 ```
-sudo podman system service -t 0 unix:///${HOME}/podman.sock
+mkdir -p $XDG_RUNTIME_DIR/podman
+sudo podman system service -t 0 unix:///$XDG_RUNTIME_DIR/podman/podman.sock
+sudo chown $USER $XDG_RUNTIME_DIR/podman/podman.sock
 ```
+
+Note that `podman system service` will keep running in the foreground so the
+current terminal must be kept open and the last command must be executed in a
+new terminal.
--- a/cluster-up/cluster/ephemeral-provider-common.sh
+++ b/cluster-up/cluster/ephemeral-provider-common.sh
@@ -16,16 +16,16 @@ fi
 
 if [ "${KUBEVIRTCI_RUNTIME}" = "podman" ]; then
     _cri_bin=podman
-    _docker_socket="${HOME}/podman.sock"
+    _docker_socket="${XDG_RUNTIME_DIR}/podman.sock"
 elif [ "${KUBEVIRTCI_RUNTIME}" = "docker" ]; then
     _cri_bin=docker
     _docker_socket="/var/run/docker.sock"
 else
-    if curl --unix-socket /${HOME}/podman.sock http://d/v3.0.0/libpod/info >/dev/null 2>&1; then
+    if curl --unix-socket "${XDG_RUNTIME_DIR}/podman/podman.sock" http://d/v3.0.0/libpod/info >/dev/null 2>&1; then
         _cri_bin=podman
-        _docker_socket="${HOME}/podman.sock"
+        _docker_socket="${XDG_RUNTIME_DIR}/podman/podman.sock"
         >&2 echo "selecting podman as container runtime"
-    elif docker ps >/dev/null; then
+    elif docker ps >/dev/null 2>&1; then
         _cri_bin=docker
         _docker_socket="/var/run/docker.sock"
         >&2 echo "selecting docker as container runtime"


0xFelix:
See also https://github.com/kubevirt/kubevirt/pull/7160

PODMAN.md
```diff
@@ -3,7 +3,7 @@
 Install podman 3.1+, then run it in docker compatible mode:
```
lyarwood, line 3: I don't think this makes any sense anymore right?

PODMAN.md
```diff
@@ -3,7 +3,7 @@
...
 
 ```
-podman system service -t 0 unix:///${HOME}/podman.sock
+systemctl start --user podman.socket
```
lyarwood, line 6: I've yet to see cluster-up work with rootless podman FWIW.

PODMAN.md
```diff
@@ -13,5 +13,11 @@ This will use `fuse-overlayfs` as storage layer. If the performance is not
 satisfactory, consider running podman as root to use plain `overlayfs2`:
 
 ```
-sudo podman system service -t 0 unix:///${HOME}/podman.sock
+mkdir -p $XDG_RUNTIME_DIR/podman
```
lyarwood, line 16: This shouldn't be required AFAICT with rootless podman.

PODMAN.md
```diff
@@ -13,5 +13,11 @@ This will use `fuse-overlayfs` as storage layer. If the performance is not
...
+sudo podman system service -t 0 unix:///$XDG_RUNTIME_DIR/podman/podman.sock
+sudo chown $USER $XDG_RUNTIME_DIR/podman/podman.sock
```
lyarwood, line 18: Ditto the above shouldn't be required.

lyarwood (commented):
The actual change LGTM but I wouldn't document anything about rootless podman until it's actually working again somewhere. I'd also drop references to manually creating `$XDG_RUNTIME_DIR/podman` directories or starting the service outside of systemd.

PODMAN.md
0xFelix: Are you speaking about the version or the docker compatible mode?

PODMAN.md
0xFelix: It was required for me when running rootful podman manually.

PODMAN.md
0xFelix: When running rootful podman the socket intially belongs to `root:root`, which the user could not access then.

I documented it this way so you can run rootful podman without make any other changes to the repository files (when #741 and https://github.com/kubevirt/kubevirt/pull/7160 are merged).

PODMAN.md
lyarwood: I was about to say both but 3.1 is barely a year old so that should likely stay.

The docker compatible mode doesn't make any sense to me tbh.

PODMAN.md
lyarwood: Right but what you end up with there isn't rootless, it's rootful podman that's accessiable to your user via that socket.

I still think we should just document the working setup at the moment of using rootful podman until the pure rootless usecase works:

```
$ sudo systemctl start podman
$ sudo ln -sf /var/run/podman/podman.sock /root/podman.sock # Workaround pull/7160
$ sudo make cluster-up
```

0xFelix: It was my intention to run a rootful podman accessible to the user via that socket. You can then run `make cluster-up` without `sudo` so that at least not everything is run as root. That's similar to how it's usually done with docker. [1](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user) In the future it would also be compatible with the default rootless setup (no path changing).

Another possible way would be

```
sudo systemctl start podman
mkdir -p $XDG_RUNTIME_DIR/podman
sudo ln -sf /var/run/podman/podman.sock $XDG_RUNTIME_DIR/podman/podman.sock
sudo chown $USER /var/run/podman/podman.sock
```

Or maybe introduce a detection whether `make cluster-up` is run as root / non-root and select the appropriate socket automatically?

PODMAN.md
0xFelix: From what I've heard, it probably worked with 3.3 at some point. On 3.4 it is not working for me.

I think running in docker compatible mode means letting podman listen on a socket, which it doesn't do by default. Mentioning this probably still makes sense.

lyarwood: I'm pretty sure it does by default now on Fedora >=35,

0xFelix: You don't need to start podman.service or podman.socket to use podman. In contrast to docker it works daemonless by default.

lyarwood: Yeah apologies that's a fair point, I think my reliance on systemd units to do this isn't the main approach people will take here.

PODMAN.md
lyarwood: ACK apologies that's entirely valid and better than just running everything under sudo as you say.

lyarwood (approved):
LGTM, apologies for all of the noise around the podman documentation!

PODMAN.md
0xFelix: Nothing to apologize for :) I think systemd units are actually the way to easily run podman in docker compatibility mode.

PR #742, opened 2022-02-01
provision: Install ovs

qinqon:
Some projects at networking need ovs at kubevirtci and they install as
part of cluster-up, provisioning it make things more stable and reduce
CI time.

### DIFF
--- a/cluster-provision/k8s/1.20/provision.sh
+++ b/cluster-provision/k8s/1.20/provision.sh
@@ -191,6 +191,9 @@ nmcli connection modify "System eth0" \
    ipv6.addr-gen-mode eui64
 nmcli connection up "System eth0"
 
+dnf install -y centos-release-nfv-openvswitch
+dnf install -y openvswitch2.16
+
 mkdir -p $kubeadmn_patches_path
 
 cat >$kubeadmn_patches_path/kube-apiserver.yaml <<EOF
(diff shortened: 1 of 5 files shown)


rhrazdil (approved):
/lgtm

cluster-provision/k8s/1.20/provision.sh
```diff
@@ -191,6 +191,11 @@ nmcli connection modify "System eth0" \
    ipv6.addr-gen-mode eui64
 nmcli connection up "System eth0"
 
+dnf install -y centos-release-nfv-openvswitch
+dnf install -y openvswitch2.16
+systemctl enable openvswitch
+systemctl restart openvswitch
```
EdDev, line 197: I do not really like the idea of forcing every project to consume OVS when they do not care about it.
But if we must, at least do not let it run there by default. Can you place the service start as part of cluster-up?

We also do not need this for 1.20 anymore.

cluster-provision/k8s/1.22-ipv6/provision.sh
```diff
@@ -208,6 +208,11 @@ nmcli connection modify "System eth0" \
    ipv6.addr-gen-mode eui64
 nmcli connection up "System eth0"
 
+dnf install -y centos-release-nfv-openvswitch
+dnf install -y openvswitch2.16
+systemctl enable openvswitch
+systemctl restart openvswitch
```
EdDev, line 214: These have been left intentionally?

qinqon: Nope, removing...

EdDev (approved):
Could you add a few words on the PR description about centos-release-nfv-openvswitch ? I am not familiar with it.

dhiller (approved):
/approve

PR #748, opened 2022-02-11
vm-image-builder: small fixes

jean-edouard:
- Add a comment with the necessary packages
- Prevent backticks in `usage` from running commands
- Exit at the end of `usage`
- Add `--nvram` to `virsh undefine` which is needed on aarch64

### DIFF
--- a/cluster-provision/images/vm-image-builder/create-containerdisk.sh
+++ b/cluster-provision/images/vm-image-builder/create-containerdisk.sh
@@ -1,10 +1,12 @@
 #!/usr/bin/env bash
 set -exuo pipefail
 
+# dnf -y install bridge-utils libvirt virt-install qemu-kvm cloud-utils guestfs-tools
 
 if [ "$#" -ne 1 ]; then
-    echo "Usage: create-containerdisk.sh image-directory"
-    echo "Run `create-continerdisk.sh example` to build the `example` image in the `example` folder"
+    echo 'Usage: create-containerdisk.sh image-directory'
+    echo 'Run `create-continerdisk.sh example` to build the `example` image in the `example` folder'
+    exit 1
 fi
 
 SCRIPT_PATH=$(dirname "$(realpath "$0")")
--- a/cluster-provision/images/vm-image-builder/customize-image.sh
+++ b/cluster-provision/images/vm-image-builder/customize-image.sh
@@ -8,7 +8,7 @@ function cleanup() {
 
   rm -rf "${CLOUD_INIT_ISO}"
   virsh destroy "${DOMAIN_NAME}" || true
-  virsh undefine "${DOMAIN_NAME}" || true
+  virsh undefine "${DOMAIN_NAME}" --nvram --remove-all-storage || true
 }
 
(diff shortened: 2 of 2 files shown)


cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -8,7 +8,7 @@ function cleanup() {
 
   rm -rf "${CLOUD_INIT_ISO}"
   virsh destroy "${DOMAIN_NAME}" || true
-  virsh undefine "${DOMAIN_NAME}" || true
+  virsh undefine "${DOMAIN_NAME}" --nvram --remove-all-storage || true
```
ormergi, line 11: Why is it necessary to pass `--nvram`? isnt it enough to use `virsh undefine` as is?
Specifying `--remove-all-storage` will remove the provisioned image as well, isnt it?
Note that it is already being deleted at lines 5-7

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -47,7 +47,7 @@ virsh destroy $DOMAIN_NAME || true
 virt-sysprep -d $DOMAIN_NAME --operations machine-id,bash-history,logfiles,tmp-files,net-hostname,net-hwaddr
 
 # Remove VM
-virsh undefine $DOMAIN_NAME
+virsh undefine $DOMAIN_NAME --nvram
```
ormergi, line 50: Why is it necessary to pass --nvram? isnt it enough to use `virsh undefine` as is?

ormergi (changes requested):
Hi Jed, thanks for addressing these.
Could you please add more to the commit message such as why are these necessary and more info about `virsh` command flags that need to be added?

cluster-provision/images/vm-image-builder/create-containerdisk.sh
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/env bash
 set -exuo pipefail
 
+# dnf -y install bridge-utils libvirt virt-install qemu-kvm cloud-utils guestfs-tools
```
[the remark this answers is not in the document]
jean-edouard: I find it really convenient to have here, in 1 command instead of bullet points, especially when having to run the script on multiple throw-away installs, but I understand it effectively duplicates the docs. Will remove.

cluster-provision/images/vm-image-builder/customize-image.sh
jean-edouard: `--nvram` is necessary on aarch64, otherwise it fails, like there: https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1371
`--remove-all-storage` can't hurt and will ensure all volumes are gone. I can remove it if it causes any issue.

cluster-provision/images/vm-image-builder/customize-image.sh
jean-edouard: `virsh undefine` fails without it on aarch64

ormergi (approved):
Thanks

PR #749, opened 2022-02-15
Use skopeo to push image as workaround for "unknown blob"

dhiller:
As `docker push` is failing nearly everytime we use skopeo to push images into the registry.

Local run succeeded in pushing images with tag `2202141654-55f845f3`

### DIFF
--- a/publish.sh
+++ b/publish.sh
@@ -2,7 +2,8 @@
 
 set -ex
 
-export KUBEVIRTCI_TAG=$(date +"%y%m%d%H%M")-$(git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


publish.sh
```diff
@@ -25,7 +26,11 @@ done
 IMAGES="${CLUSTERS}"
 docker push ${TARGET_REPO}/gocli:${KUBEVIRTCI_TAG}
```
brianmcarey, line 27: @dhiller is a skopeo copy required here too?

dhiller: Good point :blush:

dhiller: @brianmcarey fixed, PTAL, thanks!

enp0s3:
/approve
/hold
feel free to unhold upon CI pass

PR #756, opened 2022-03-03
vm.sh: Add iptables DNAT rules for rootless podman

0xFelix:
This adds iptables DNAT rules for rootless podman. Because traffic is
originating from the loopback adapter when using rootles podman /
slirp4netns, the existing iptables rules did not apply and
'make cluster-up' was failing on rootless podman. This patch checks
whether a rootless container is used and adds the corresponding
iptables rules.

### DIFF
--- a/cluster-provision/centos8/scripts/vm.sh
+++ b/cluster-provision/centos8/scripts/vm.sh
@@ -58,15 +58,30 @@ until ip link show tap${n}; do
   sleep 0.1
 done
 
+ROOTLESS=0
+if [ -f /run/.containerenv ]; then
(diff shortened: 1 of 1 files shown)


EdDev:
@0xFelix , thank you for this change.

Could we check at this stage if we are in a rootful or rootless container? If we could, then we could adjust each with its own proper focused configuration.

I think this addition will not harm the rootful run, but we better check nevertheless (that we did not increased the run time).

EdDev (approved):
Nice! Thank you!!

enp0s3 (commented):
/approve

PR #757, opened 2022-03-04
cluster-up: Unify usage of rootless or rootful podman

0xFelix:
This adds the '--remote' and '--url' argument to every podman call to
unify the usage of rootless or rootful podman. By using the same socket
everywhere it is guaranteed that rootless and rootful podman are not mixed.
Prior providing a socket to a rootful podman would still leave some
instances where a rootless podman (without socket) was used.


EdDev (approved):
Thanks!

enp0s3 (approved):
/approve
@0xFelix Thank you!

PR #766, opened 2022-03-22
provision: Clean CNI folder before cluster creation

oshoval:
Crio and podman creates their own files under the CNI folder.
Having a file in this folder will make k8s cluster use it
without waiting for the CNI to be installed (calico for example).
For now we see errors without cleaning the folder only for ipv6,
but a good practice is just to have this folder clean,
so calico will be the one creating the config.

It is also safer in case some installation would update the folder,
and cause unknown effect on the cluster.


oshoval:
sriov not related to this PR

PR #768, opened 2022-03-23
Add rook-cephfs storage class for filesystem PVCs

rmohr:
If rook-ceph gets enabled, it can now used for filesystem PVCs.

```
$ kubectl.sh get storageclasses
NAME                        PROVISIONER                     RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
local                       kubernetes.io/no-provisioner    Delete          WaitForFirstConsumer   false                  33m
rook-ceph-block (default)   rook-ceph.rbd.csi.ceph.com      Delete          Immediate              true                   31m
rook-cephfs                 rook-ceph.cephfs.csi.ceph.com   Delete          Immediate              false                  31m
```

Also bumping the centos-stream base version, since for the initrd of the older kenerl we don't have kernel modules anymore.


akalenyu:
Awesome!

The only suggestion I have is that we add a ready check for cephfs similarly to https://github.com/kubevirt/kubevirtci/blob/ab5a0f27c874b85a32e99f2be43fd378b73a5e59/cluster-provision/k8s/1.23/rook-ceph.sh#L23

cluster-provision/centos8/version
```diff
@@ -1 +1 @@
-20210210.0
+20220125.1
```
xpivarc, line 1: Just from curiosity why is it needed?

rmohr:
> The only suggestion I have is that we add a ready check for cephfs similarly to

Will add it asap :+1:

PR #769, opened 2022-03-28
Add cirros container disk

ormergi:


ormergi:
No need to invest in Cirros image now that we have Alpine image.
/close

PR #732, opened 2022-01-10
enable vm-image-builder to build Arm64 images

zhlhahaha:
Currently, the images, fedora-with-test-tooling and fedora-realtime, are used in kubevirt e2e tests.
We need these images support multiarch, thus we can verify more e2e tests on Arm64.
~~This PR is used to support build multiarch images in vm-image-builder expect fedora-realtime, because the repository
http://ccrma.stanford.edu/planetccrma/mirror/fedora does not have package for aarch64~~

The package planetccrma-repo-1.1-3.fc32.ccrma.noarch.rpm is a noarch rpm package which can be install in Arm64 VM.

This PR does not support to build multiarch alpine-cloud-init, because the build tool https://raw.githubusercontent.com/alpinelinux/alpine-make-vm-image/master/alpine-make-vm-image does not support to build Arm64 version image, please refer to https://github.com/alpinelinux/alpine-make-vm-image/issues/10

Signed-off-by: Howard Zhang <howard.zhang@arm.com>

PR #770, opened 2022-03-30
alpine: Create cloud-init aware alpine

qinqon:
Create the alpine image using the official [tool](https://github.com/alpinelinux/alpine-make-vm-image) it does not set networking since it depends on kubevirt tests and it can be set with cloud-init, the idea is to set as minimal stuff as possible and then customize with cloud-init at kubevirt tests.

This has to work with https://github.com/kubevirt/kubevirtci/pull/732


rmohr:
/lgtm
/approve

works so far. Initial push: quay.io/kubevirtci/alpine-cloud-init:v3.15-2203311632

PR #773, opened 2022-04-04
Avoid `make cluster-down` failures

jean-edouard:
I keep seeing test lanes failing on `make cluster-down`, even if all the tests succeeded.
That leads to a lot of unnecessary retests and therefore a lot of pressure on the CI infrastructure.
Example: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/7103/pull-kubevirt-e2e-kind-1.23-vgpu-nonroot/1511039474501423104

### DIFF
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -303,7 +303,8 @@ function down() {
     if [ -z "$($KIND get clusters | grep ${CLUSTER_NAME})" ]; then
         return
     fi
-    $KIND delete cluster --name=${CLUSTER_NAME}
+    # On CI, avoid failing an entire test run just because of a deletion error
+    $KIND delete cluster --name=${CLUSTER_NAME} || [ "$CI" = "true" ]
     docker rm -f $REGISTRY_NAME >> /dev/null
     rm -f ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
 }


cluster-up/down.sh
```diff
@@ -9,4 +9,4 @@ fi
 
 source ${KUBEVIRTCI_PATH}hack/common.sh
 source ${KUBEVIRTCI_CLUSTER_PATH}/$KUBEVIRT_PROVIDER/provider.sh
-down
+down || true
```
xpivarc, line 12: Would you be inclined to move the fix closer to root cause? (`cluster-up/cluster/kind/common.sh`

xpivarc (changes requested):
I have one suggestion and one question. Did you have time to look at the root cause of this issue?

cluster-up/down.sh
jean-edouard: I do think that `make cluster-down` should never cause entire test lanes to fail, but at the same time you're right, let's "fix" only what's broken for now! Pushed.

cluster-up/cluster/kind/common.sh
```diff
@@ -304,6 +304,6 @@ function down() {
         return
     fi
     $KIND delete cluster --name=${CLUSTER_NAME}
-    docker rm -f $REGISTRY_NAME >> /dev/null
+    docker rm -f $REGISTRY_NAME >> /dev/null || true
```
oshoval, line 307: I remember vaguely, that overriding a value in a function will eliminate the `-e` in it totally,
so if for example the previous commands fail they won't fail the test as well

maybe we are find with it in this case
if we are not, we can check if this is the behavior (for example add `false` line above it)
and consider trying something like  `! docker rm -f $REGISTRY_NAME >> /dev/null`
that will achieve same effect but might still respect the `-e`

oshoval (commented):
Thanks
maybe worth as an offline effort to try and understand why it happens ?

oshoval:
note that the failing command is docker rm -f -v, maybe this is not the failing command that was changed?

cluster-up/cluster/kind/common.sh
jean-edouard: Thanks for the review @oshoval !
- I don't think your concern about eliminating `-e` is a problem here, as the `||` only applies to the one command and not an entire function
- `! <cmd>` would indeed succeed on failure, but also fail on success, so that's a problem :)
- Completely agree on `>` vs `>>`, but it works so I was going to leave it alone
- I think originally you mentioned that I was modifying the wrong line, and you're absolutely right, it must be the `kind` command above that calls `docker rm`, I should probably do both

oshoval: Thanks

> Thanks for the review @oshoval !
> 
> * I don't think your concern about eliminating `-e` is a problem here, as the `||` only applies to the one command and not an entire function

We saw at the past, that masking one line affect the whole function, even that it isn't logical in the first place, it is a gotcha of bash

> * `! <cmd>` would indeed succeed on failure, but also fail on success, so that's a problem :)

It won't fail on success, nothing checks the return code, because you add `!` only, 
it is syntactic sugar for `|| true`

cluster-up/cluster/kind/common.sh
```diff
@@ -303,7 +303,7 @@ function down() {
     if [ -z "$($KIND get clusters | grep ${CLUSTER_NAME})" ]; then
         return
     fi
-    $KIND delete cluster --name=${CLUSTER_NAME}
-    docker rm -f $REGISTRY_NAME >> /dev/null
+    $KIND delete cluster --name=${CLUSTER_NAME} || true
+    docker rm -f $REGISTRY_NAME >> /dev/null || true
```
oshoval, line 307: do we need also `docker rm -f $REGISTRY_NAME >> /dev/null || true` ?
we didn't see it fail so far
i think we better remove the change here?

cluster-up/cluster/kind/common.sh
jean-edouard: > Thanks
> 
> > Thanks for the review @oshoval !
> > 
> > * I don't think your concern about eliminating `-e` is a problem here, as the `||` only applies to the one command and not an entire function
> 
> We saw at the past, that masking one line affect the whole function, even that it isn't logical in the first place, it is a gotcha of bash (not sure it is exactly the case here, but generally speaking)
> 
> > * `! <cmd>` would indeed succeed on failure, but also fail on success, so that's a problem :)
> 
> It won't fail on success, nothing checks the return code, because you add `!` only, it is syntactic sugar for `|| true` (but anyhow it is needed only if the above is relevant here)

Indeed, my bad! I had to write a little test to convince myself:
``` 
$ cat test.sh 
#!/bin/bash -e

test() {
	echo 1
	false || true
	echo 2
	! true
	echo 3
	! false
	echo 4
	false
	echo 5
}

test
$ ./test.sh 
1
2
3
4
$ 
```

oshoval (approved):
Thanks

EdDev:
> I keep seeing test lanes failing on `make cluster-down`, even if all the tests succeeded. That leads to a lot of unnecessary retests and therefore a lot of pressure on the CI infrastructure. Example: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/7103/pull-kubevirt-e2e-kind-1.23-vgpu-nonroot/1511039474501423104

While I mostly agree, I think there needs to be an explanations why this is ok to do.
And at the same time, add it to the commit as well.

It is not trivial to understand that on CI, the environment, including the docker daemon is lives in a prow pod which is removed eventually.
On the other hand, when run locally (on a proper machine that has the needed HW), we may end up with leftovers.

There is also a question about if any leftovers are left on CI, I want to hope we do not. And I tend to take the risk and say that if we do leave them, we have a bug. And this bug will probably also exist today even without this change.

So I am left with asking if we can limit this to CI runs only and not include local runs.

jean-edouard:
> > I keep seeing test lanes failing on `make cluster-down`, even if all the tests succeeded. That leads to a lot of unnecessary retests and therefore a lot of pressure on the CI infrastructure. Example: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/7103/pull-kubevirt-e2e-kind-1.23-vgpu-nonroot/1511039474501423104
> 
> While I mostly agree, I think there needs to be an explanations why this is ok to do. And at the same time, add it to the commit as well.
> 
> It is not trivial to understand that on CI, the environment, including the docker daemon is lives in a prow pod which is removed eventually. On the other hand, when run locally (on a proper machine that has the needed HW), we may end up with leftovers.
> 
> There is also a question about if any leftovers are left on CI, I want to hope we do not. And I tend to take the risk and say that if we do leave them, we have a bug. And this bug will probably also exist today even without this change.
> 
> So I am left with asking if we can limit this to CI runs only and not include local runs.

Right, having leftovers or not is unrelated to returning failure or success.
It looks like the root cause of the issue is being investigated there: https://github.com/kubevirt/kubevirt/pull/7527
All I'm trying to do is give CI a bit of much needed immediate relief, instead of running long successful test suites over and over. 
I'm fine with limiting the change to CI, but I don't know how, please advise.
Thank you

EdDev:
> I'm fine with limiting the change to CI, but I don't know how, please advise.

Maybe `[ "$CI" != "true" ]`?

EdDev (approved):

PR #776, opened 2022-04-05
alpine: Reduce size and startup time

qinqon:
The alpine qcow2 image can be reduced by reserving less space and removing packages also the startup time can be reduced by removing guest agent

### DIFF
--- a/cluster-provision/images/vm-image-builder/alpine-cloud-init/configure.sh
+++ b/cluster-provision/images/vm-image-builder/alpine-cloud-init/configure.sh
@@ -6,15 +6,6 @@ step() {
 	printf '\n\033[1;36m%d) %s\033[0m\n' $_step_counter "$@" >&2  # bold cyan
 }
 
-step 'Set up qemu-guest-agent'
-cat > /etc/conf.d/qemu-guest-agent <<-EOF
-GA_METHOD="virtio-serial"
-GA_PATH="/dev/vport1p1"
-EOF
-
-step 'Set up timezone'
-setup-timezone -z Europe/Prague
-
 step 'Adjust rc.conf'
 sed -Ei \
 	-e 's/^[# ](rc_depend_strict)=.*/\1=NO/' \
@@ -23,9 +14,4 @@ sed -Ei \
 	/etc/rc.conf
 
 step 'Enable services'
-rc-update add acpid default
-rc-update add chronyd default
-rc-update add crond default
-rc-update add termencoding boot
-rc-update add qemu-guest-agent default
 rc-update add cloud-init default
--- a/cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
+++ b/cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
@@ -12,10 +12,10 @@ if [ ! -f alpine-make-vm-image ]; then
     chmod 755 alpine-make-vm-image
 fi
 
-docker run --platform=$PLATFORM --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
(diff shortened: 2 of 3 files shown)


cluster-provision/images/vm-image-builder/alpine-cloud-init/configure.sh
```diff
@@ -6,15 +6,6 @@ step() {
 	printf '\n\033[1;36m%d) %s\033[0m\n' $_step_counter "$@" >&2  # bold cyan
 }
 
-step 'Set up qemu-guest-agent'
```
xpivarc, line 9: Can you comment on this change?

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
```diff
@@ -12,10 +12,10 @@ if [ ! -f alpine-make-vm-image ]; then
     chmod 755 alpine-make-vm-image
 fi
 
-docker run --platform=$PLATFORM --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
+docker run --rm --platform=$PLATFORM -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
```
ormergi, line 15: nit: you can pass `--workdir $(pwd)` instead of passing it to `sh -c `

Just to understand, why is it necessary to mount /lib/modules and /dev? is it safe? 
would it work if we mount it as read only?

cluster-provision/images/vm-image-builder/alpine-cloud-init/packages
```diff
@@ -1,10 +1,3 @@
-chrony
-less
-logrotate
-openssh
-ssmtp
-sudo
 cloud-init
-qemu-guest-agent
```
ormergi, line 8: Is it necessary for kubevirt tests?

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
qinqon: --privileged is supose to be mounting those already don't know why I added them, I will remove and use --workdir.

cluster-provision/images/vm-image-builder/alpine-cloud-init/packages
qinqon: Not yet, let's go step by step, so we have the fastest alpline cloud-init image possible.

EdDev (commented):
> The alpine qcow2 image can be reduced by reseving less space for the
disk.

- Typo with `reseving`.
- The removal of the guest-agent is reducing the usage potential significantly.
   Can we control this through cloud-init user-data? i.e. tell it to run only if we explicitly ask for it there.
   I agree that by default it should not be running.

dhiller (approved):
/approve

Have verified that the image works. Holding for remaining reviewers to settle. Please remove hold when it's ready and ping me again in case you need another lgtm.

/hold

qinqon:
/hold cancel

Tested by @AlonaKaplan at https://github.com/kubevirt/kubevirt/pull/7520#issuecomment-1096072028

qinqon:
> /approve
> 
> Have verified that the image works. Holding for remaining reviewers to settle. Please remove hold when it's ready and ping me again in case you need another lgtm.
> 
> /hold

@dhiller can you push an image to kubevirtci with this similar to one one we did already (it would be better to have some automation for this though).

qinqon:
/override check-provision-k8s-1.23 check-provision-k8s-1.22-ipv6

PR #778, opened 2022-04-06
podman, docs: clarify rootless podman does not work w/ cluster-sync

maiqueb:
Currently, kubevirtci only works successfully with rootful podman;
`make cluster-sync` fails pulling the images from quay.

### DIFF
--- a/PODMAN.md
+++ b/PODMAN.md
@@ -2,15 +2,26 @@
 
 Install podman 3.1+, then run it in docker compatible mode:
 
+## Rootless podman
+
 ```
 systemctl start --user podman.socket
 ```
 
-After this command, you can use kubevirtci with the typical `make cluster-*`
-commands.
+Currently rootless podman is **not** working with the `make cluster-sync`
+command, essentially because incoming traffic is coming from the loopback device
+instead of eth0.
+
+The current rules - [ssh](https://github.com/kubevirt/kubevirtci/blob/962d90cead28fc2aadcc07388b18d2479b2b6714/cluster-provision/centos8/scripts/vm.sh#L73), [restricted ports](https://github.com/kubevirt/kubevirtci/blob/962d90cead28fc2aadcc07388b18d2479b2b6714/cluster-provision/centos8/scripts/vm.sh#L83) - allow `make cluster-up` to run successfully, but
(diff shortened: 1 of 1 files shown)


PODMAN.md
```diff
@@ -2,22 +2,39 @@
...
 systemctl start --user podman.socket
 ```
 
-After this command, you can use kubevirtci with the typical `make cluster-*`
-commands.
+Currently rootless podman is **not** working with the `make cluster-sync`
+command, essentially because incoming traffic is coming from the loopback device
+instead of eth0.
+
+The current rules - [0], [1] - allow `make cluster-up` to run successfully, but
```
0xFelix, line 15: Make [0] and [1] links instead of adding them below?

PODMAN.md
```diff
@@ -2,22 +2,39 @@
...
+unfortunately they break the cluster's network connectivity in a subtle way:
+image pulling fails because outgoing traffic to ports 22 6443 8443 80 443 30007
+30008 31001 is redirected to the VM in the node01 container (i.e. itself)
```
0xFelix, line 18: Maybe it is more accurate to say "redirected to the VM in the respective node container (i.e. itself)".

PODMAN.md
```diff
@@ -2,22 +2,39 @@
...
+instead of going to the specified host (e.g. quay.io).
 
 This will use `fuse-overlayfs` as storage layer. If the performance is not
 satisfactory, consider running podman as root to use plain `overlayfs2`:
 
+## Rootfull podman
```
0xFelix, line 24: I think it's rootful?

0xFelix (changes requested):

PODMAN.md
maiqueb: Can I add 2 links to same word ? That's the reason I put this as I did :)

PODMAN.md
maiqueb: Done. Good point. It read weird before.

PODMAN.md
maiqueb: Done.

PODMAN.md
0xFelix: Looks good now :)

PR #779, opened 2022-04-07
Bump kube-state-metrics to enable usage of some advanced metrics

akalenyu:
Allows usage of advanced kube-state-metrics in kubevirt which are enabled by default on OpenShift like
kube_persistentvolumeclaim_labels/kube_pod_labels.

The second commit adds an allow list, without this these object label metrics are denied by default.


marceloamaral:
Looks good!
/lgtm

Maybe It might be useful to write a comment where this metric is being used that it requires setting the `-metric-labels-allowlist` in prometheus.

PR #780, opened 2022-04-12
alpine: Add guest agent to the image

qinqon:


qinqon:
@AlonaKaplan has test this too and it does not introduce latencies, so we can merge it.

dhiller (approved):
/approve

PR #755, opened 2022-03-03
kind: Use quay registry image

oshoval:
Overcome docker.io rate limits

Signed-off-by: Or Shoval <oshoval@redhat.com>

PR #782, opened 2022-04-12
Revert "kind: Use quay registry image"

oshoval:
This reverts commit 42b307938d0bb771119d1bab8255dc4f072e6fc7.

The image doesn't support arm64.

Follow effort will be to fix https://github.com/kubevirt/project-infra/pull/1762
by using docker pull and push instead of skopeo copy,
which might cause the problem (we checked and the source image that we mirrored
does support arm64 where the target doesn't, tried source `library/registry:2.7.1`). 
See discussion on https://github.com/kubevirt/kubevirtci/pull/755

### DIFF
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -94,7 +94,7 @@ function _run_registry() {
         docker rm $REGISTRY_NAME || true
         sleep 5
     done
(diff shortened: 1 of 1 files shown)


zhlhahaha:
Thanks!

enp0s3 (approved):
/approve

oshoval:
Thanks Igor

/retest

not related, only kind was changed

PR #784, opened 2022-04-13
cluster-up, kind: add kind-1.23

zhlhahaha:
This is necessary for testing kubeVirt on Arm64 server, as, currently, Arm64 server does not support nested virtualization.

I update the default `KIND_VERSION` to 0.12.0, because kind 0.12.0 support multi-arch kindest/node:v1.23.4 image. https://github.com/kubernetes-sigs/kind/releases
~~The default `KIND_NODE_IMAGE`, quay.io/kubevirtci/kindest_node:v1.23.3, does not support Arm64~~
Update the default KIND_NODE_IMAGE to `quay.io/kubevirtci/kindest_node:v1.23.4` which support both x86_64 and Arm64
Signed-off-by: Howard Zhang <howard.zhang@arm.com>

### DIFF
--- a/cluster-up/cluster/kind-1.23/provider.sh
+++ b/cluster-up/cluster/kind-1.23/provider.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -e
+
+DEFAULT_CLUSTER_NAME="kind-1.23"
+DEFAULT_HOST_PORT=5000
+ALTERNATE_HOST_PORT=5001
+export CLUSTER_NAME=${CLUSTER_NAME:-$DEFAULT_CLUSTER_NAME}
+
+if [ $CLUSTER_NAME == $DEFAULT_CLUSTER_NAME ]; then
+    export HOST_PORT=$DEFAULT_HOST_PORT
+else
+    export HOST_PORT=$ALTERNATE_HOST_PORT
+fi
+
+function set_kind_params() {
+    export KIND_VERSION="${KIND_VERSION:-0.12.0}"
+    export KIND_NODE_IMAGE="${KIND_NODE_IMAGE:-quay.io/kubevirtci/kindest-node:v1.23.4}"
+    export KUBECTL_PATH="${KUBECTL_PATH:-/bin/kubectl}"
+}
+
+function configure_registry_proxy() {
+    [ "$CI" != "true" ] && return
+
+    echo "Configuring cluster nodes to work with CI mirror-proxy..."
+
+    local -r ci_proxy_hostname="docker-mirror-proxy.kubevirt-prow.svc"
+    local -r kind_binary_path="${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kind"
+    local -r configure_registry_proxy_script="${KUBEVIRTCI_PATH}/cluster/kind/configure-registry-proxy.sh"
+
+    KIND_BIN="$kind_binary_path" PROXY_HOSTNAME="$ci_proxy_hostname" $configure_registry_proxy_script
+}
+
+function up() {
+    cp $KIND_MANIFESTS_DIR/kind.yaml ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
+    _add_worker_kubeadm_config_patch
+    _add_worker_extra_mounts
+    kind_up
+
+    configure_registry_proxy
+
+    # remove the rancher.io kind default storageClass
+    _kubectl delete sc standard
+
+    echo "$KUBEVIRT_PROVIDER cluster '$CLUSTER_NAME' is ready"
+}
+
+set_kind_params
+
+source ${KUBEVIRTCI_PATH}/cluster/kind/common.sh
--- a/cluster-up/cluster/kind-1.23/README.md
+++ b/cluster-up/cluster/kind-1.23/README.md
@@ -1,0 +1,45 @@
+# K8S 1.23.3 with mdev support in a Kind cluster
+
+Provides a pre-deployed k8s cluster with version 1.23.3 that runs using [kind](https://github.com/kubernetes-sigs/kind) The cluster is completely ephemeral and is recreated on every cluster restart.
+The KubeVirt containers are built on the local machine and are then pushed to a registry which is exposed at
+`localhost:5000`.
+
+## Bringing the cluster up
+
+The following needs to be executed as root.
+
+```bash
+export KUBEVIRT_PROVIDER=kind-1.23
+make cluster-up
+```
+
+The cluster can be accessed as usual:
+
+```bash
+$ cluster-up/kubectl.sh get nodes
+NAME                  STATUS   ROLES    AGE     VERSION
+kind-1.23-control-plane   Ready    master   6m14s   v1.23.3
+```
+
+## Bringing the cluster down
+
+```bash
+make cluster-down
+```
+
+This destroys the whole cluster.
+
+## Setting a custom kind version
+
+In order to use a custom kind image / kind version,
+export KIND_NODE_IMAGE, KIND_VERSION, KUBECTL_PATH before running cluster-up.
+For example in order to use kind 0.9.0 (which is based on k8s-1.19.1) use:
+```bash
+export KIND_NODE_IMAGE="kindest/node:v1.19.1@sha256:98cf5288864662e37115e362b23e4369c8c4a408f99cbc06e58ac30ddc721600"
+export KIND_VERSION="0.9.0"
+export KUBECTL_PATH="/usr/bin/kubectl"
+```
+This allows users to test or use custom images / different kind versions before making them official.
+See https://github.com/kubernetes-sigs/kind/releases for details about node images according to the kind version.
+
+- In order to use `make cluster-down` please make sure the right `CLUSTER_NAME` is exported.


jean-edouard (commented):
Thank you @zhlhahaha for this, looks good to me!
/cc @dhiller 
Can we add a `check-provision` test lane for this?

cluster-up/cluster/kind-k8s-1.23/README.md
```diff
@@ -0,0 +1,45 @@
...
+export KUBEVIRT_PROVIDER=kind-k8s-1.23
+make cluster-up
+```
+
+The cluster can be accessed as usual:
+
+```bash
+$ cluster-up/kubectl.sh get nodes
+NAME                  STATUS   ROLES    AGE     VERSION
+kind-1.23-control-plane   Ready    master   6m14s   v1.23.3
```
jean-edouard, line 21: If there's only 1 node, shouldn't it also be a worker?

zhlhahaha: Yes, it should be both master and worker node.

zhlhahaha: But I tried to create one node cluster, and use `kubectl get nodes`. It do not show worker role.

jean-edouard: I assume it can't be used for much then... We probably need something like that:
https://github.com/kubevirt/kubevirtci/blob/main/cluster-up/cluster/k8s-provider-common.sh#L153
Alternatively, I guess we could enforce a minimum of 2 nodes

zhlhahaha: I am not quite sure about it. As the taint of master node is none and master node is schedulable, I can deploy kubevirt and vmi successfully on this one node cluster. Do I still need to change the label of this node?

jean-edouard: Ah, got it, sounds like we're good then!

dhiller:
@zhlhahaha just to make sure I understand the use case: can we then use kind to run the tests as we do on the existing prow-workloads cluster?

/hold

Other than that, changes look good to me.

/approve

zhlhahaha:
> Can we then use kind to run the tests on the arm cluster as we do on the existing prow-workloads cluster? Like in: create arm presubmit jobs that test things in parallel?

For this PR, my purpose is using kind to quick deploy a development environment on Arm64, not for CI test. And I understand your idea. I think `running parallel testing in kind cluster` is a feasible way, but may need more works to do. I can open an issue for it.

> Or can we even run the arm tests on the prow-workloads cluster?

I am not sure what prow-workloads cluster is. Currently the problems for CI testing on Arm64 are:
1. Arm64 server does not nested virtualization, so the test scenario for x86_64 can not be used on Arm64.
2. Prow bootstrap image does not support Arm64, we are unable to use prow to create a kind cluster on Arm64.

dhiller (approved):

cluster-up/cluster/kind-k8s-1.23/provider.sh
```diff
@@ -0,0 +1,50 @@
...
+
+if [ $CLUSTER_NAME == $DEFAULT_CLUSTER_NAME ]; then
+    export HOST_PORT=$DEFAULT_HOST_PORT
+else
+    export HOST_PORT=$ALTERNATE_HOST_PORT
+fi
+
+function set_kind_params() {
+    export KIND_VERSION="${KIND_VERSION:-0.12.0}"
+    export KIND_NODE_IMAGE="${KIND_NODE_IMAGE:-quay.io/kubevirtci/kindest_node:v1.23.3@sha256:0df8215895129c0d3221cda19847d1296c4f29ec93487339149333bd9d899e5a}"
```
enp0s3, line 18: @zhlhahaha Following the PR description `v1.23.3` doesn't support ARM, so shouldn't we set `kindest/node:v1.23.4@sha256:0e34f0d0fd448aa2f2819cfd74e99fe5793a6e4938b328f657c8e3f81ee0dfb9` for the `KIND_NODE_IMAGE`

enp0s3 (commented):
@zhlhahaha Hi! thank you for the PR. Please see my question below

cluster-up/cluster/kind-k8s-1.23/provider.sh
zhlhahaha: As the kind node image can be set via environmental variable `KIND_NODE_IMAGE`, so I keep it sync with the node image in kind-1.23-vgpu. I am total fine with both image. What is your suggestion?

oshoval:
Thanks

Maybe we should jiust bump sriov one to 1.23 and have the sriov installment there as opt-out ?
this way we will have one provider less to maintain
(unless there are more features you activated that should be opt-in)

/cc @EdDev

oshoval:
maybe worth to simply call it `kind-1.23` instead of `kind-k8s-1.23` ?
this way it is named like the sriov one without the sriov suffix
also we know that it refer to k8s so it is redundant

zhlhahaha:
> maybe worth to simply call it `kind-1.23` instead of `kind-k8s-1.23` ? this way it is named like the sriov one without the sriov suffix also we know that it refer to k8s so it is redundant

Updated

oshoval (approved):
Thanks

Just change please PR title as well

cluster-up/cluster/kind-1.23/provider.sh
```diff
@@ -0,0 +1,50 @@
...
+
+if [ $CLUSTER_NAME == $DEFAULT_CLUSTER_NAME ]; then
+    export HOST_PORT=$DEFAULT_HOST_PORT
+else
+    export HOST_PORT=$ALTERNATE_HOST_PORT
+fi
+
+function set_kind_params() {
+    export KIND_VERSION="${KIND_VERSION:-0.12.0}"
+    export KIND_NODE_IMAGE="${KIND_NODE_IMAGE:-quay.io/kubevirtci/kindest_node:v1.23.3@sha256:0df8215895129c0d3221cda19847d1296c4f29ec93487339149333bd9d899e5a}"
```
oshoval, line 18: Aint this version contradicts what you wrote on the PR desc?
btw if you want a multi arch image, we can mirror it to quay, now when we mirror it supports multi arch

zhlhahaha: > Aint this version contradicts what you wrote on the PR desc? 

Do you mean that kind-1.23 contradicts with KIND_VERSION which is 0.12.0? 1.23 refer to kubernetes version rather than kind version, and here it is related to the KIND_NODE_IMAGE tag which is v1.23.3.

> btw if you want a multi arch image, we can mirror it to quay, now when we mirror it supports multi arch you need to add it here if you want https://github.com/kubevirt/project-infra/blob/main/hack/images_to_mirror.csv

Is it means that I only need to put the image name (like following line) in the csv document, then multi-arch image would automatically be pull?
```
docker.io,kindest/node:v1.23.4,quay.io/kubevirtci
```

oshoval: > > Aint this version contradicts what you wrote on the PR desc?
> 
> Do you mean that kind-1.23 contradicts with KIND_VERSION which is 0.12.0? 1.23 refer to kubernetes version rather than kind version, and here it is related to the KIND_NODE_IMAGE tag which is v1.23.3.

"(The default KIND_NODE_IMAGE, quay.io/kubevirtci/kindest_node:v1.23.3, does not support Arm64)"
I mean that this line of the PR desc, is the actual line that you use, don't you want an image which is multi arch ?

> 
> > btw if you want a multi arch image, we can mirror it to quay, now when we mirror it supports multi arch you need to add it here if you want https://github.com/kubevirt/project-infra/blob/main/hack/images_to_mirror.csv
> 
> Is it means that I only need to put the image name (like following line) in the csv document, then multi-arch image would automatically be pull?
> 
> ```
> docker.io,kindest/node:v1.23.4,quay.io/kubevirtci
> ```

A post submit job would push it, at the first time a repo is created under quay, one of the maintainer might need to make it public, but once those two are done you will be able to use it, yes, and it will be multi arch

zhlhahaha: I see, thanks. I will submit a PR for it.

zhlhahaha:
As the multi-arch kind-node image have mirrored, I update the default kind node image to `quay.io/kubevirtci/kindest-node:v1.23.4`. @oshoval

oshoval (approved):
Thanks

Can you please update PR desc now that we do have a multi arch image?

oshoval:
Are we good to remove the hold ?
https://github.com/kubevirt/kubevirtci/pull/784#issuecomment-1108466803
@dhiller @zhlhahaha

oshoval:
Once https://github.com/kubevirt/kubevirtci/pull/826 is merged (if merged before this PR),
please consider to do the same change for the new provider

dhiller:
> Once #826 is merged (if merged before this PR), please consider to do the same change for the new provider

The PR looks good to me. I think that we can add this auto detection in another PR, as this is a new provider, it will not hurt IMHO.

xpivarc:
@dhiller `check-provision-k8s-1.23` seems to fail a lot on this PR and it is the only job holding this PR. Do you think it is worth overriding?

zhlhahaha:
Hi @dhiller 
It seems the test `check-provision-k8s-1.23` failure is not caused by my PR.

oshoval:
It seems that https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/784/check-provision-k8s-1.23/1548619678551642112 is broken

oshoval:
/hold

CI just test it again and again because the lanes have problems
lets either override them because they are not related to the PR, or keep hold please until they are fixed

PR #788, opened 2022-04-25
alpine: Use virt-install to support multiarch

qinqon:
It's not possible to construct non native alpine images with current
method since there is no syslinux package for amd64 and using nbd kernel
possible is impossible either. This change go back to virt-install +
expect with golang.

TODO:
- Download image
- Use multiarch manifest


cluster-provision/images/vm-image-builder/create-containerdisk.sh
```diff
@@ -54,7 +56,11 @@ if [ -f "${SCRIPT_PATH}/${IMAGE_NAME}/create-image.sh" ]; then
       ./create-image.sh "${build_directory}/${VM_IMAGE}" 
 
       echo "Creating the containerdisk ..."
-      docker build . -t ${IMAGE_NAME}:${TAG} -f - <<END
+      PLATFORM=linux/$ARCHITECTURE
+      if [ "$ARCHITECTURE" = "aarch64" ]; then
+          PLATFORM=linux/arm64
+      fi
+      docker buildx build --load --platform $PLATFORM . -t ${IMAGE_NAME}:${TAG} -f - <<END
```
oshoval, line 63: Maybe worth to support please podman (using buildah, qemu-static-user is needed in case of a cross compile) as well
example: hack/build-push-container.podman.sh at https://github.com/nmstate/kubernetes-nmstate/pull/1085

cluster-provision/images/vm-image-builder/alpine-cloud-init/setup/setup.sh
```diff
@@ -1,11 +1,23 @@
-#!/bin/sh
+#!/bin/sh -xe
```
oshoval, line 1: the -x is a leftover?

oshoval (commented):
Thanks

zhlhahaha:
hi @qinqon why close this PR?

PR #798, opened 2022-05-23
alpine: Use vport2p1 ga path

qinqon:
At kubevirt the GA path is different. This adapt the alpine image to use
the kubevirt one.


EdDev (commented):
> At kubevirt the GA path is different.

I think this part is no longer true, right? In the commit and in the PR description.

What I do not understand is, why these are not defined by default upon installation of the package. I think this is the case with the rpm on Fedora, no?

ormergi:
> > At kubevirt the GA path is different.
> 
> I think this part is no longer true, right? In the commit and in the PR description.
> 
> What I do not understand is, why these are not defined by default upon installation of the package. I think this is the case with the rpm on Fedora, no?

It seems `/dev/virtio-ports/org.qemu.guest_agent.0` is the default path https://qemu.readthedocs.io/en/latest/interop/qemu-ga.html

ormergi (approved):
Thanks for addressing this!

PR #734, opened 2022-01-14
Add second sound card

alicefr:
An additional sound cards can cover more advanced testing of the KubeVirt device plugin.

Signed-off-by: Alice Frosi <afrosi@redhat.com>

PR #799, opened 2022-05-23
Update cluster provisioning image to fedora 36

brianmcarey:
Fedora 31 is no longer supported so we should update to use Fedora 36.
This change also includes a bump to the dockerize version to the latest
release.

### DIFF
--- a/cluster-provision/centos8/Dockerfile
+++ b/cluster-provision/centos8/Dockerfile
@@ -1,10 +1,9 @@
 
-#FROM fedora:31
-FROM fedora@sha256:8fa60b88e2a7eac8460b9c0104b877f1aa0cea7fbc03c701b7e545dacccfb433
+FROM quay.io/fedora/fedora@sha256:38813cf0913241b7f13c7057e122f7c3cfa2e7c427dca3194f933d94612e280b
 
 ARG centos_version
 
-RUN dnf -y update nettle && dnf -y install jq iptables iproute dnsmasq qemu openssh-clients screen && dnf clean all
+RUN dnf -y install jq iptables iproute dnsmasq qemu openssh-clients screen && dnf clean all
 
 WORKDIR /
 
@@ -12,7 +11,7 @@ COPY vagrant.key /vagrant.key
 
 RUN chmod 700 vagrant.key
 
-ENV DOCKERIZE_VERSION v0.3.0
+ENV DOCKERIZE_VERSION v0.6.1
 
 RUN curl -LO https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
   && tar -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
--- a/cluster-provision/centos8/scripts/vm.sh
+++ b/cluster-provision/centos8/scripts/vm.sh
@@ -99,7 +99,7 @@ if [ $disk_size -lt $default_disk_size ]; then
 fi
 
 echo "Creating disk \"${next} backed by ${last} with size ${disk_size}\"."
-qemu-img create -f qcow2 -o backing_file=${last} ${next} ${disk_size}
+qemu-img create -f qcow2 -o backing_file=${last} -F qcow2 ${next} ${disk_size}
 
 echo ""
 echo "SSH will be available on container port 22${n}."
--- a/cluster-provision/gocli/cmd/provision.go
+++ b/cluster-provision/gocli/cmd/provision.go
@@ -234,6 +234,11 @@ func provisionCluster(cmd *cobra.Command, args []string) (retErr error) {
 		return err
 	}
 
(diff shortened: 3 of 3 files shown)


cluster-provision/centos8/scripts/vm.sh
```diff
@@ -150,6 +150,6 @@ exec qemu-system-x86_64 -enable-kvm -drive format=qcow2,file=${next},if=virtio,c
   -append "$(cat /kernel.args) $(cat /additional.kernel.args) ${KERNEL_ARGS}" \
   -vnc :${n} -cpu host,migratable=no,+invtsc -m ${MEMORY} -smp ${CPU} \
   -serial pty -M q35,accel=kvm,kernel_irqchip=split \
-  -device intel-iommu,intremap=on,caching-mode=on -soundhw hda -soundhw ac97 \
+  -device intel-iommu,intremap=on,caching-mode=on -device intel-hda -device hda-duplex -device AC97 \
```
dankenigsberg, line 153: now that `-shoundhw` is gone, `-device AC97` is more opaque; but why do we need audio on kubernetes nodes? Can it be dropped (on another PR?)

brianmcarey: Sorry I removed these too soon - soundhw is only removed qemu 7.1[1]. Fedora 36 has qemu 6.2. I have added the soundhw flags back in as they are more descriptive of what the device is. These soundcards are used for testing scenarios with the Kubevirt device plugin - https://github.com/kubevirt/kubevirtci/pull/734

[1] https://github.com/qemu/qemu/blob/master/docs/about/removed-features.rst#creating-sound-card-devices-using--soundhw-removed-in-71

dhiller:
@brianmcarey looks like we at beyond capacity again, thus the job doesn't get started...

cluster-provision/centos8/Dockerfile
```diff
@@ -1,18 +1,17 @@
 
-#FROM fedora:31
-FROM fedora@sha256:8fa60b88e2a7eac8460b9c0104b877f1aa0cea7fbc03c701b7e545dacccfb433
+FROM quay.io/fedora/fedora@sha256:38813cf0913241b7f13c7057e122f7c3cfa2e7c427dca3194f933d94612e280b
 
 ARG centos_version
 
-RUN dnf -y update nettle && dnf -y install jq iptables iproute dnsmasq qemu openssh-clients screen && dnf clean all
+RUN dnf -y install jq iptables iproute dnsmasq qemu openssh-clients screen && dnf clean all
```
oshoval, line 6: @AlonaKaplan can we deprecate iptables here ?

dhiller: IIUC this will be done in another step, right?

brianmcarey: @oshoval  - it is used by a couple of the scripts for provisioning the VMs[1] - I think it would be better in a follow on PR

[1] https://github.com/kubevirt/kubevirtci/blob/main/cluster-provision/centos8/scripts/vm.sh

oshoval, line 6: Can you explain why nettle is not required ? 
(in PR desc / commit preferably if possible please)

brianmcarey: Yes - when I was running this - the dnf update was not actually updating anything - nothing to update.

brianmcarey:
@oshoval  - this change will help with issues like this - https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/824/check-provision-k8s-1.24/1542137453714870272#1:build-log.txt%3A33 - can you take a look?

cluster-provision/centos8/scripts/vm.sh
oshoval: Just note please that we do have this warning
```
qemu-system-x86_64: warning: '-soundhw ac97' is deprecated, please use '-device AC97' instead
qemu-system-x86_64: warning: '-soundhw hda' is deprecated, please use '-device intel-hda -device hda-duplex' instead
```

jean-edouard:
```
$ docker pull quay.io/fedora/fedora@sha256:38813cf0913241b7f13c7057e122f7c3cfa2e7c427dca3194f933d94612e280b
Error response from daemon: manifest for quay.io/fedora/fedora@sha256:38813cf0913241b7f13c7057e122f7c3cfa2e7c427dca3194f933d94612e280b not found: manifest unknown: manifest unknown
```
Am I doing something wrong?

brianmcarey:
> Am I doing something wrong? Also, any reason not to use `quay.io/fedora/fedora:36`?

It looks like old fedora images get cleaned up over time from quay. I would be in favour of just using  `quay.io/fedora/fedora:36` - it would prevent us from falling too far behind. 

I see that it has been updated to use the kubevirtci mirror - https://github.com/kubevirt/kubevirtci/pull/849 - but there could still be some value in pulling the latest f36 from quay.

oshoval:
> > Am I doing something wrong? Also, any reason not to use `quay.io/fedora/fedora:36`?
> 
> It looks like old fedora images get cleaned up over time from quay. I would be in favour of just using `quay.io/fedora/fedora:36` - it would prevent us from falling too far behind.

It might break with unknown errors from time to time,
especially when we are using something very new
we might want to have some cron update job that runs one a month ?
so it will be deterministic
see https://github.com/kubevirt/kubevirtci/pull/847#discussion_r933931344

> 
> I see that it has been updated to use the kubevirtci mirror - #849 - but there could still be some value in pulling the latest f36 from quay.

PR #801, opened 2022-05-26
alpine: Install iputils

qinqon:
The vanilla alpine image is using light ip utils version from busybox,
this change add the full feature tools to it.

### DIFF
--- a/cluster-provision/images/vm-image-builder/alpine-cloud-init/packages
+++ b/cluster-provision/images/vm-image-builder/alpine-cloud-init/packages
@@ -2,3 +2,4 @@ qemu-guest-agent
 cloud-init
 e2fsprogs-extra
 util-linux
+iputils


oshoval (approved):

EdDev:
> The vanilla alpine image is using light ip utils version from busybox, this change add the full feature tools to it.

Can you please explain the reason this is needed?

dhiller (approved):
/approve

qinqon:
> > The vanilla alpine image is using light ip utils version from busybox, this change add the full feature tools to it.
> 
> Can you please explain the reason this is needed?

The kubevirt e2e tests use some arguments from the proper IP command.

EdDev:
> The kubevirt e2e tests use some arguments from the proper IP command.

It will be nice to see this reasoning with an example of what is not supported in the commit message, so future us will be able to understand why it was done.

qinqon:
> > The kubevirt e2e tests use some arguments from the proper IP command.
> 
> It will be nice to see this reasoning with an example of what is not supported in the commit message, so future us will be able to understand why it was done.

This is the issue:

```
ping 10.244.140.85 -c 5 -w 10
ping: -c: Name does not resolve
```

But moving the destination as the last argument fix it, so we don't need to install iptools, busybox ping is enough if destination passed correctly (it's also compatible with iptools ping).

PR #804, opened 2022-05-31
fixed crash for ipv6 disabled machines

ConnorOgorchock:
prevent script from crashing if ipv6 directory does not exist


oshoval:
FYI @EdDev 

At the past we concluded that ipv6 should be enabled, there was an issue about it as well
please wait for the sig-network team to review it

/hold

cluster-provision/centos8/scripts/dnsmasq.sh
```diff
@@ -6,8 +6,8 @@ NUM_NODES=${NUM_NODES-1}
 NUM_SECONDARY_NICS=${NUM_SECONDARY_NICS:-0}
 
 ip link add br0 type bridge
-echo 0 > /proc/sys/net/ipv6/conf/br0/disable_ipv6
-echo 1 > /proc/sys/net/ipv6/conf/all/forwarding
+echo 0 > /proc/sys/net/ipv6/conf/br0/disable_ipv6 || true
+echo 1 > /proc/sys/net/ipv6/conf/all/forwarding || true
```
oshoval, line 10: Did you check it with the latest providers ?
The nodes wait for ipv6 address by dnsmasq, so this fix might not be enough

(and also we concluded that we enforce even ipv6 but this is other story, see please the other comment about it)

EdDev (changes requested):
kubevirtci is an opinionated test infrastructure for testing kubevirt.
We expect to have IPv6 enabled in order to test dual-stack and single stack IPv6 clusters.
This should be the default.

If we fail to define it like that, we have a problem and I would expect everything to blow up.

If some need to test this without IPv6, I would expect it to be an explicit configuration exception (e.g. an env variable that does not attempt to set IPv6) that is requested during the buid/run.

I would also expect to see a good reasoning on why we need to support it (as at the end, these options have maintenance cost).

jean-edouard:
> If some need to test this without IPv6, I would expect it to be an explicit configuration exception (e.g. an env variable that does not attempt to set IPv6) that is requested during the buid/run.

I think adding an environment variable would be a nice compromise here. Perhaps with an added warning in the script when entering that case, saying it's an unsupported/experimental configuration option.

> I would also expect to see a good reasoning on why we need to support it (as at the end, these options have maintenance cost).

I know of at least one company that provides their engineers with machines running a pre-installed Linux distro that disables IPv6 on the kernel command line.
It would be nice if they could somehow still deploy kubevirt without making core changes to the kernel configuration.

jean-edouard:
/approve
@oshoval @EdDev PTAL!

cluster-provision/centos8/scripts/dnsmasq.sh
```diff
@@ -6,8 +6,17 @@ NUM_NODES=${NUM_NODES-1}
...
+  if [ -z $ALLOW_MISSING_IPV6 ]; then
+    echo "Error: IPV6 is disabled on this machine and ALLOW_MISSING_IPV6 is not set." >&2
+    exit 1
+  else
+    echo "Warning: proceeding without IPV6, this configuration is not supported" >&2
+  fi
+else
+  echo 0 > /proc/sys/net/ipv6/conf/br0/disable_ipv6
+  echo 1 > /proc/sys/net/ipv6/conf/all/forwarding
+fi
```
oshoval, line 19: lets have it in a function please ?

oshoval (commented):
Thanks

Did you try to provision the latest providers without ipv6 ?

I wonder if it will stuck or work, they are waiting for ipv6 from dnsmasq
do we need to change dnsmasq as well ? (see in dnsmasq.sh at the end)

jean-edouard:
> Did you try to provision the latest providers without ipv6 ?
> 
> I wonder if it will stuck or work, they are waiting for ipv6 from dnsmasq do we need to change dnsmasq as well ? (see in dnsmasq.sh at the end)

That's a good point, let's make sure it all works fine before merging this.
/approve cancel

oshoval:
> > Did you try to provision the latest providers without ipv6 ?
> > I wonder if it will stuck or work, they are waiting for ipv6 from dnsmasq do we need to change dnsmasq as well ? (see in dnsmasq.sh at the end)
> 
> That's a good point, let's make sure it all works fine before merging this. /approve cancel

Btw note that even the updated non ipv6 providers do wait for ipv6 by dnsmasq on the nodes

ConnorOgorchock:
Thanks for the feedback. Closing for now, may revisit later

oshoval:
Thanks,
Note for future, I am not sure how it was planned to use, as `ALLOW_MISSING_IPV6` should be declared in the node which spins the dnsmasq container, it can't simply be exported on the bare metal

PR #808, opened 2022-06-06
k8s-1.24: Support SingleStack IPv6 provisioning

oshoval:
In order to ease creation of SingleStack IPv6 providers,
adapt k8s-1.24 code to support provisioning of both DualStack and SingleStack IPv6.

Duplicate k8s-1.24 as k8s-1.24-ipv6
Provisioning `k8s-1.24-ipv6` auto detects the ipv6 suffix and configures the cluster as
SingleStack IPv6.

Deprecated `IPv6DualStack` was removed all over.

Note:
k8s-1.23 doesn't support SingleStack, only k8s-1.22, and by this PR k8s-1.24 as well.
In case we will need to have k8s-1.23-ipv6 we will adapt it.
Adapting it without having lane for it isn't good practice.


oshoval:
Testing k8s-1.24-ipv6 here https://github.com/kubevirt/project-infra/pull/2145

oshoval:
/cc @brianmcarey 

k8s-1.24-ipv6 is tested here https://github.com/kubevirt/project-infra/pull/2145 with CRI-O 1.22 (since it hanged on CI with CRI-O 1.24)

cluster-provision/k8s/1.24/k8s_provision.sh
```diff
@@ -127,9 +132,12 @@ rules:
 EOF
 
 kubeadm_manifest="/etc/kubernetes/kubeadm.conf"
-envsubst < /tmp/kubeadm.conf > $kubeadm_manifest
+envsubst < $kubeadm_raw > $kubeadm_manifest
+
+until ip address show dev eth0 | grep global | grep inet6; do sleep 1; done
```
brianmcarey, line 137: Is there a case here that this will sleep continuously if the interface fails to get an IPv6 address? Should it fail at some point?

oshoval: well if the dnsmasq has a bug it theoretically might, but we didn't see it never yet (we are already using this logic for few months now)
In case we see we can add a protection, just so it will break after few minutes, imo there is no need now unless you think otherwise

oshoval:
Rebased to use CRI-O 1.22 fix

oshoval:
/test check-provision-k8s-1.24 

it doesnt started, retriggering

brianmcarey:
> it doesnt started, retriggering

the job was recently made required here (https://github.com/kubevirt/project-infra/pull/2133) - thats why it appears but didn't get triggered.

brianmcarey (commented):
/lgtm

oshoval:
Thanks Brian
@dhiller / @qinqon can you take a look please?

PR #812, opened 2022-06-15
SR-IOV Provider: Patch Multus manifest using Kustomize

ormergi:
On Kubevirtci various components manifests are kept as they are and if a change is required the manifest is being patched in place.

Following 'sriov-cni' and 'sriov-device-plugin' example where their manifests are patched using Kustomize,
this PR applies the same pattern for Multus manifest.


cluster-up/cluster/kind-1.22-sriov/sriov-components/manifests/multus/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- multus.yaml
+images:
+- name: nfvpe/multus:v3.4
+  newName: quay.io/kubevirtci/multus
+  newTag: v3.4
```
qinqon, line 8: Can we keep the incoming tag so we just replace the name ? 
```suggestion
- name: nfvpe/multus
  newName: quay.io/kubevirtci/multus
```

qinqon (changes requested):
question

cluster-up/cluster/kind-1.22-sriov/sriov-components/manifests/multus/kustomization.yaml
ormergi: Its possible, I specified the tag for visibility.
You prefer to remove it anyway?

qinqon: Let's remove it, If we bump multus we will not need to update this file.

ormergi: Done

cluster-up/cluster/kind-1.22-sriov/sriov-components/sriov_components.sh
```diff
@@ -106,8 +108,12 @@ function sriov_components::wait_allocatable_resource() {
 }
 
 function sriov_components::deploy_multus() {
-  echo 'Deploying Multus'
-  sed "s#nfvpe/multus#quay.io/kubevirtci/multus#" "$MULTUS_MANIFEST" | _kubectl apply -f -
+  _kubectl kustomize "$KUSTOMIZE_MULTUS_DIR" > "$MULTUS_MANIFEST"
+
+  echo "Deploying Multus:"
+  cat "$MULTUS_MANIFEST"
```
oshoval, line 114: I don't think we need it, it makes the log / deploy have too much info

oshoval: I still think it is not needed, once it works it works, this code isn't breathing but making the log yet more heavy forever.

The fact it is heavy already doesn't mean we can add to it, we just need to make it lighter and to show only the important stuff.

qinqon: Let's have all the bits at the SRIOV lanes we are talking about something at it is show at startup once just before hours of tests, that's not much noise. Real noise is when we end up printing a line every 5 mintues or so .

PR #813, opened 2022-06-16
SR-IOV provider: Reduce the amount of VFs to create

ormergi:
As part of SR-IOV provider cluster-up, each cluster node is provisioned with SR-IOV PF and VFs.
Currently node SR-IOV PF will be configured with maximal possible amount of VFs (i.e: according to the SR-IOV PF 'sriov_totalvfs').

There is no need to create so many VFs since only a few are actually being used.

With this PR changes, the amount of VFs that will be created is reduced to fixed number of 8, slightly more then enough.
It will reduce some resource consumption by redundant VFs.


EdDev (approved):
Thank you!

cluster-up/cluster/kind-1.22-sriov/sriov-node/configure_vfs.sh
```diff
@@ -79,6 +79,9 @@ function ensure_driver_is_loaded() {
 
 DRIVER="${DRIVER:-vfio-pci}"
 DRIVER_KMODULE="${DRIVER_KMODULE:-vfio_pci}"
+VFS_COUNT=${VFS_COUNT:-0}
```
qinqon, line 82: We should default to the at least the previous behaviour but not to zero, defaults should be something that works.

qinqon (changes requested):
Use sane defaults

cluster-up/cluster/kind-1.22-sriov/sriov-node/configure_vfs.sh
ormergi: For what its worth, if VFS_COUNT is not overrides the script wont do any thing (also not mess anything).
I can set the default to 1, or 6 as we were going to use, WDYT?

ormergi:
[Changes](https://github.com/kubevirt/kubevirtci/compare/267b4679c3c768868e4ae9d7e63650d33b51b9cd..4673cea719dae3ca67440a34e2139a163c33fa06): Set default value for VFS_COUNT, also refined commit messages.

EdDev (approved):
Thank you!

EdDev:
> /hold
> 
> Please make sure we "reset" the `numvfs` (set to 0 and back to 8) so all the config will get removed (hopefully).

Done, removing the hold.

/unhold

brianmcarey:
/override check-provision-alpine-with-test-tooling

Changes are just related to the kind provider for SRIOV

PR #816, opened 2022-06-20
dnsmasq: Don't use ipv6 local address a dns server

qinqon:
At some envs[1] having a resolv.conf with an entry referencing the
interface name is not supported, this happend when a IPv6 link local
address is used as DNS server. This change instruct dnsmasq to only use
global addresses.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2095207

Signed-off-by: Enrique Llorente <ellorent@redhat.com>


enp0s3 (approved):
/approve

PR #819, opened 2022-06-22
alpine: Revert boot timeout removal

qinqon:
At alpine with tooling containerdisk if timeout is set to 0 it waits
human intervention, this change go back to default. Follow up PR will
investigate how to remove menu so it directly boot without wait.

Depends on:
- https://github.com/kubevirt/project-infra/pull/2171

Signed-off-by: Enrique Llorente <ellorent@redhat.com>


enp0s3 (approved):
/approve

qinqon:
@brianmcarey Is there any issue after upgrade on prow nodes ? looks like nbd cannot be mounted anymore.

PR #823, opened 2022-06-28
Sync ceph bumps to 1.24 as well

akalenyu:
The ceph bumps must have gone in after the 1.24 addition PR,
so the bumps never reached 1.24.
```bash
[akalenyu@localhost kubevirtci]$ cat cluster-provision/k8s/1.23/manifests/ceph/operator.yaml | grep quay.io/cephcsi
  # ROOK_CSI_CEPH_IMAGE: "quay.io/cephcsi/cephcsi:v3.5.1"
[akalenyu@localhost kubevirtci]$ cat cluster-provision/k8s/1.24/manifests/ceph/operator.yaml | grep quay.io/cephcsi
  # ROOK_CSI_CEPH_IMAGE: "quay.io/cephcsi/cephcsi:v3.4.0"
```
This results in a parallel PVC creation issue on 1.24 periodic (same issue that made us bump ceph in the first place):
https://testgrid.k8s.io/kubevirt-periodics#periodic-kubevirt-e2e-k8s-1.24-sig-storage&width=20

Signed-off-by: Alex Kalenyuk <akalenyu@redhat.com>

### DIFF
--- a/cluster-provision/k8s/1.24-ipv6/manifests/ceph/cluster-test.yaml
+++ b/cluster-provision/k8s/1.24-ipv6/manifests/ceph/cluster-test.yaml
@@ -26,7 +26,7 @@ metadata:
 spec:
   dataDirHostPath: /var/lib/rook
   cephVersion:
-    image: ceph/ceph:v16.2.4
+    image: quay.io/ceph/ceph:v16.2.7
     allowUnsupported: true
   mon:
     count: 1
--- a/cluster-provision/k8s/1.24-ipv6/manifests/ceph/common.yaml
+++ b/cluster-provision/k8s/1.24-ipv6/manifests/ceph/common.yaml
@@ -1,65 +1,199 @@
-###################################################################################################################
+####################################################################################################
 # Create the common resources that are necessary to start the operator and the ceph cluster.
 # These resources *must* be created before the operator.yaml and cluster.yaml or their variants.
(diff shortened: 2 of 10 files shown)


dhiller:
@akalenyu how hard would it be to automate the bumping? In a similar fashion as we do with cdi?

dhiller (commented):
/approve

akalenyu:
> > > @akalenyu how hard would it be to automate the bumping? In a similar fashion as we do with cdi?
> > > https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-periodics.yaml#L100
> > > EDIT: added link
> > 
> > 
> > I am not sure about the difficulty, what I do is diff the non-trivial yamls (the ones with env vars) and try to select the correct values. Also, we do diverge in some places from the example yamls so that can present some challenges
> 
> OK, thanks for the clarification. However, if the update is crucial for ceph to work with CDI, we should at least define some process around this, i.e. when that MUST happen, what is to do. This is to make sure we don't miss this again. WDYT?

Definitely agree, would this be a doc in here?

For visibility I wanted to mention the sequence of events:
We had noticed rook had API calls that included some deprecated ones, so that was a good reason to bump.
After making this bump, we started hitting this parallel PVC creation issue:
https://github.com/ceph/ceph-csi/issues/2521

PR #824, opened 2022-06-29
podman: Support non root user via root socket

oshoval:
In order to use roootful podman by a non root user,
podman supports a socket (like docker does).

Adapt code to support it.
Changes:
1. Support the default root socket path `/run/podman/podman.socket`
2. Download the kubectl according the cluster version instead
copying it from the node.
We used to copy it from the node as a workaround of podman issue (affect only 1.8M+ files),
but it is copied with root only permissions, and then the non root
user can't access it.
Downloading it instead allows non root user to access it.
Note that kubectl needs to be presented.
It will be used for checking the cluster version and then the kubectl with the same
version as the cluster will be downloaded and used (as before).
(potential improvement - we can download only if not present already, in order to save few seconds).

Kind podman support will be on a follow PR.

Note: even as root we would need to start `podman.socket`
because kubevirtci and kubevirt are trying to use socket for podman at all cases.

### DIFF
--- a/PODMAN.md
+++ b/PODMAN.md
@@ -23,16 +23,33 @@ satisfactory, consider running podman as root to use plain `overlayfs2`:
 
 ## Rootful podman
 
+In order to use rootful podman by a non root user, we will need to bind podman
+to a socket, accessible by the user (as docker does).
+
+Assuming the user is in `wheel` group please do the following (one time):
+
+As root, create a Drop-In file `/etc/systemd/system/podman.socket.d/10-socketgroup.conf`
+with the following content:
 ```
-mkdir -p $XDG_RUNTIME_DIR/podman
-sudo podman system service -t 0 unix:///$XDG_RUNTIME_DIR/podman/podman.sock
-sudo chown $USER $XDG_RUNTIME_DIR/podman/podman.sock
+[Socket]
+SocketGroup=wheel
+ExecStartPost=/usr/bin/chmod 755 /run/podman
 ```
 
-After this command, you can use kubevirtci with the typical `make cluster-*`
-commands.
+The 1st line is needed in order to create the socket accessible by the `wheel` group.
+2nd line because systemd-tmpfiles recreates the folder as root:root without group reading rights.
 
-Note that `podman system service` will keep running in the foreground so the
-current terminal must be kept open and the last command must be executed in a
-new terminal.
+Stop `podman.socket` if it is running,
+reload the daemon `systemctl daemon-reload` since we changed the systemd settings
+and restart it again `systemctl enable --now podman.socket`
 
+As the user add the following to your ~/.bashrc
+```
+alias podman="podman --remote"
+export CONTAINER_HOST=unix:///run/podman/podman.sock
+```
+
+Validate it by running `podman run hello-world` as the non root user
+and see that as root `podman ps -a` shows the same exited container (or vice versa).
+
+Tested on fedora 35.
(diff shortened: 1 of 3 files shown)


oshoval:
/cc @0xFelix 
Thanks for the help here

PODMAN.md
```diff
@@ -23,16 +23,33 @@ satisfactory, consider running podman as root to use plain `overlayfs2`:
...
+
+As root, create a Drop-In file `/etc/systemd/system/podman.socket.d/10-socketgroup.conf`
+with the following content:
 ```
-mkdir -p $XDG_RUNTIME_DIR/podman
-sudo podman system service -t 0 unix:///$XDG_RUNTIME_DIR/podman/podman.sock
-sudo chown $USER $XDG_RUNTIME_DIR/podman/podman.sock
+[Socket]
+SocketGroup=wheel
+ExecStartPre=/usr/bin/chmod -f 755 /run/podman
```
phoracek, line 36: This should be ExecStartPost. Otherwise it may fail on a fresh install due to a chicken egg issue - this systemd file creates the directory, but we are trying to run a command on the (non-existent) directory.

phoracek (changes requested):
Thanks for this improvement!

PODMAN.md
oshoval: Thanks, the `-f` is for this, for some reason it didn't happen here even if I deleted the folder,
anyhow will change as you said because it is a good fix

cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -14,16 +14,24 @@ if [ -n "${KUBEVIRTCI_TAG}" ] && [ -n "${KUBEVIRTCI_GOCLI_CONTAINER}" ]; then
...
+        echo "/run/podman/podman.sock"
+    elif curl --unix-socket "${XDG_RUNTIME_DIR}/podman/podman.sock" http://d/v3.0.0/libpod/info >/dev/null 2>&1; then
+        echo "${XDG_RUNTIME_DIR}/podman/podman.sock"
+    fi
+}
+
 if [ "${KUBEVIRTCI_RUNTIME}" = "podman" ]; then
-    _cri_bin="podman --remote --url=unix://${XDG_RUNTIME_DIR}/podman/podman.sock"
-    _docker_socket="${XDG_RUNTIME_DIR}/podman/podman.sock"
+    _docker_socket=$(detect_podman_socket)
```
brianmcarey, line 26: Should we call this something like ``_cri_socket`` ?

oshoval: We can (note that it is not a regression), will add as a prev commit

brianmcarey (commented):
/lgtm

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -121,23 +121,23 @@ function up() {
     fi
     eval ${_cli:?} run $params
 
-    # Copy k8s config and kubectl
+    ${_cli} scp --prefix $provider_prefix /etc/kubernetes/admin.conf - >${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
+
+    # Set server and disable tls check
+    export KUBECONFIG=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
+    kubectl config set-cluster kubernetes --server="https://$(_main_ip):$(_port k8s)"
+    kubectl config set-cluster kubernetes --insecure-skip-tls-verify=true
```
qinqon, line 129: We should use the copied kubectl so people don't need to have installed.

qinqon (changes requested):

cluster-up/cluster/k8s-provider-common.sh
oshoval: We can't, because it is copied with root permissions and then a non root user can't use podman via the socket
See please comment on PR desc
It is fine to ask to have kubectl installed, it will be used for checking the cluster version

PR #832, opened 2022-07-12
single stack: Remove k8s-1.22-ipv6

oshoval:
k8s-1.24-ipv6 is used instead


oshoval:
/retest

not related

oshoval:
/test check-provision-k8s-1.22 

not related
basically we can just override all and save time, best if we had run_if_changed instead

PR #833, opened 2022-07-17
podman: Present socket env var

oshoval:
In order to make the code easier to maintain
and agnostic of custom socket path,
preset an env var `KUBEVIRTCI_PODMAN_SOCKET`.
The default value will be `/run/podman/podman.sock`
which is podman.socket root user default value.

The user can override this value in case he uses
a custom path such as one depends on `XDG_RUNTIME_DIR`.

### DIFF
--- a/PODMAN.md
+++ b/PODMAN.md
@@ -52,4 +52,8 @@ export CONTAINER_HOST=unix:///run/podman/podman.sock
 Validate it by running `podman run hello-world` as the non root user
 and see that as root `podman ps -a` shows the same exited container (or vice versa).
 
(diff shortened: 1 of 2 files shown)


oshoval:
Thanks
Failures are because kubevirtci is broken atm (not due the PR), retest won't help

oshoval:
@brianmcarey 
can you review please ?

brianmcarey (commented):
/lgtm

oshoval:
/test check-provision-k8s-1.24


failures here are due to
https://github.com/kubevirt/kubevirtci/issues/843

PR #841, opened 2022-07-20
vm based providers: Pin NetworkManager version 

oshoval:
NetworkManager 1.39.7-x causes the node to get stuck on
various stages.
For example during CNI deploy or during kubeadm reset.
The provision / publish fails due to that without informative data.
For example
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/831/check-provision-k8s-1.24/1549349853673295872
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/publish-kubevirtci/1549206667130834944

It happens even on idle VM after modifying the interface
and running the following command
"nmcli connection up" without creating a cluster.

In talk with NM team offline in order to find the root cause.
Meanwhile pinning the NM version 2 versions back, where it doesn't happen.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2109285

### DIFF
--- a/cluster-provision/k8s/1.22/k8s_provision.sh
+++ b/cluster-provision/k8s/1.22/k8s_provision.sh
@@ -47,7 +47,9 @@ systemctl daemon-reload
 systemctl enable crio && systemctl start crio
 systemctl enable kubelet && systemctl start kubelet
 
-dnf install -y NetworkManager NetworkManager-ovs
+# Version 1.39.7-x cause the VM to hang on random places
+NETWORK_MANAGER_VERSION=1.39.5-1
+dnf install -y NetworkManager-${NETWORK_MANAGER_VERSION}.el8.x86_64 NetworkManager-ovs-${NETWORK_MANAGER_VERSION}.el8.x86_64
 
 # configure additional settings for cni plugin
 cat <<EOF >/etc/NetworkManager/conf.d/001-calico.conf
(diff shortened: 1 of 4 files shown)


oshoval:
/cc @brianmcarey @qinqon 
FYI
I think this is the root of the problems we see on CI / locally during provision
sometimes it seems CNI hangs. sometimes kubeadm reset hangs
but as a matter of fact the machine or at least its network hangs
(socat / screen doesn't work even on healthy machibe so couldn't console it)

oshoval:
New problem not related to the PR
first time I see, we might need to pin it as well if returns 
```
+ dnf install -y kernel-modules-4.18.0-408.el8.x86_64
CentOS Stream 8 - AppStream                      13 MB/s |  24 MB     00:01    
CentOS Stream 8 - BaseOS                         14 MB/s |  23 MB     00:01    
CentOS Stream 8 - Extras                         31 kB/s |  18 kB     00:00    
No match for argument: kernel-modules-4.18.0-408.el8.x86_64
Error: Unable to find a match: kernel-modules-4.18.0-408.el8.x86_64
```

oshoval:
Once the BZ https://bugzilla.redhat.com/show_bug.cgi?id=2109285 is fixed, do we want to unpin NM ?

PR #845, opened 2022-07-24
kind: Adapt check-cluster-up to auto detect CRI

oshoval:
`cluster-up/cluster/kind/check-cluster-up.sh` is the CI entry point of
kubevirtci kind presubmits.
Support using a parameterized CRI instead hard coding `docker`.

### DIFF
--- a/cluster-up/cluster/kind/check-cluster-up.sh
+++ b/cluster-up/cluster/kind/check-cluster-up.sh
@@ -25,6 +25,12 @@ PROVIDER_PATH="${kubevirtci_path}/cluster-up/cluster/${KUBEVIRT_PROVIDER}"
 
 RUN_KUBEVIRT_CONFORMANCE=${RUN_KUBEVIRT_CONFORMANCE:-"false"}
 
+function detect_cri() {
+    if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi
+}
+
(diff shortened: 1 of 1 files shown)


oshoval:
errors not related, retest might solve them, some network hiccup

cluster-up/cluster/kind/check-cluster-up.sh
```diff
@@ -25,6 +25,12 @@ PROVIDER_PATH="${kubevirtci_path}/cluster-up/cluster/${KUBEVIRT_PROVIDER}"
 
 RUN_KUBEVIRT_CONFORMANCE=${RUN_KUBEVIRT_CONFORMANCE:-"false"}
 
+function detect_cri() {
+    if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi
```
brianmcarey, line 29: with kind - should docker be the default if both are installed? I thought that podman support was still experimental in kind?

oshoval: We don't have both on the same bootstrap
That PR with both was reverted to the best of my knowledge
so we just keep podman first because once we are done podman will have priority
This what we did everywhere, and it totally fine as long as we dont have both
one can still override CRI if needed

brianmcarey: ok sounds good to me.

brianmcarey (commented):
/lgtm

PR #846, opened 2022-07-24
vm based: Add validation that single stack is indeed as expected

oshoval:
Update `check-cluster-up` script to validate single stack cluster is indeed
IPv6 single stack by validating it has only one IP, and this IP is IPv6.

Additional change:
Print the pods with `-owide` to see the IPs for each pod.
It would help to see if the cluster IPs are as expected.
Both in case of failure and generally.


enp0s3 (approved):
/approve

PR #847, opened 2022-07-29
whereabouts: bump to v0.5.4

jean-edouard:
We're currently using v0.5, which appears to be unreliable, probably due to the race condition issues they've been working on upstream.
v0.5.4 should contain all the necessary fixes.


jean-edouard:
Added a commit to switch to Fedora 36 for building the Centos 8 image (oO).
The static hash there doesn't seem to work anymore.
Probably not the right thing to do but otherwise check-provision fails right away and I want to test this PR.

oshoval (commented):
Thanks, nice
Can you please update PR desc about the reason as appears here?
https://github.com/kubevirt/kubevirt/pull/7168#issuecomment-1199468059

Also if you can update PR desc about pinning to fedora:36 because the hash doesn't work

cluster-provision/centos8/Dockerfile
```diff
@@ -1,5 +1,5 @@
 
-FROM quay.io/fedora/fedora@sha256:38813cf0913241b7f13c7057e122f7c3cfa2e7c427dca3194f933d94612e280b
+FROM quay.io/fedora/fedora:36
```
EdDev, line 2: Isn't this risky? Fedora is known to change frequently with little assurance of stability.

hack/kustomization/whereabouts/kustomization.yaml
```diff
@@ -6,6 +6,3 @@ resources:
 - ./daemonset-install.yaml
 - ./whereabouts.cni.cncf.io_ippools.yaml
 - ./whereabouts.cni.cncf.io_overlappingrangeipreservations.yaml
-
-transformers:
```
oshoval, line 10: I think that the files appear above are deleted as well ?
(they were merged into the current file)
also once there isn't a transformers we basically not need this file and the references of it

On the other hand, we will need on a follow PR to use kustomize again in order to pin the version
until https://github.com/k8snetworkplumbingwg/whereabouts/issues/252 is fixed

cluster-provision/centos8/Dockerfile
oshoval: We can mirror it if needed to our quay
note that atm it blocks kubevirtci PR

xpivarc: Let's split this part to separate PR and pin this down to a new hash. Keeping a floating tag can cause an issue that will be discovered later than missing sha.

jean-edouard:
Removed the fedora fix commit, as it was addressed by another (already merged) PR.
Added the appropriate kustomize bits to address the image issues upstream.
PTAL!

hack/kustomization/whereabouts/kustomization.yaml
```diff
@@ -7,5 +7,16 @@ resources:
 - ./whereabouts.cni.cncf.io_ippools.yaml
 - ./whereabouts.cni.cncf.io_overlappingrangeipreservations.yaml
 
-transformers:
-  - patch-ip-reconciler-job.yaml
+patches:
```
oshoval, line 10: Thanks for the changes
Can we do it please with this kustomization style ?
https://github.com/kubevirt/kubevirtci/blob/main/cluster-up/cluster/kind-1.22-sriov/sriov-components/manifests/kustomization.yaml#L9

and then the sed in the script is not needed

oshoval:
Please also update the manifest of k8s-1.24-ipv6

oshoval:
Hi @maiqueb
ain't 0.5.4 should pin the version as 0.5.3 does ?
https://github.com/k8snetworkplumbingwg/whereabouts/pull/253

Then we won't need to kustomize it ourself

hack/kustomization/whereabouts/kustomization.yaml
oshoval: `newTag` i think ?

something like this
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- grafana-deployment.yaml
images:
  - name: grafana/grafana
    newName: quay.io/kubevirtci/grafana-grafana
    newTag: <TAG>
```

jean-edouard: And.. how do you set `<TAG>`? :)

oshoval: envsubst / sed it according "whereabouts_version", wdyt?

jean-edouard: Done, PTAL!

hack/kustomization/whereabouts/kustomization.yaml
```diff
@@ -7,5 +7,7 @@ resources:
 - ./whereabouts.cni.cncf.io_ippools.yaml
 - ./whereabouts.cni.cncf.io_overlappingrangeipreservations.yaml
 
-transformers:
-  - patch-ip-reconciler-job.yaml
+images:
+  - name: ghcr.io/k8snetworkplumbingwg/whereabouts
+    newName: ghcr.io/k8snetworkplumbingwg/whereabouts
```
oshoval, line 12: Can we remove newName or it is a must to keep it even that it is the same ?
beside that lgtm, thanks

maiqueb:
> Hi @maiqueb ain't 0.5.4 should pin the version as 0.5.3 does ? [k8snetworkplumbingwg/whereabouts#253](https://github.com/k8snetworkplumbingwg/whereabouts/pull/253) (maybe this PR was reverted ?)
> 
> Then we won't need to kustomize it ourself

v0.5.4 **is** using a [pinned version](https://github.com/k8snetworkplumbingwg/whereabouts/commit/d979fa57d1e06d09b50b91d755ee828c96587cd4).

... might be a bit confusing to use the generated version for v0.5.3, but, it **is** pinned.

Going forward, we'll try to make sure we don't screw up again.

maiqueb:
> > > Hi @maiqueb ain't 0.5.4 should pin the version as 0.5.3 does ? [k8snetworkplumbingwg/whereabouts#253](https://github.com/k8snetworkplumbingwg/whereabouts/pull/253) (maybe this PR was reverted ?)
> > > Then we won't need to kustomize it ourself
> > 
> > 
> > v0.5.4 **is** using a [pinned version](https://github.com/k8snetworkplumbingwg/whereabouts/commit/d979fa57d1e06d09b50b91d755ee828c96587cd4).
> > ... might be a bit confusing to use the generated version for v0.5.3, but, it **is** pinned.
> > Going forward, we'll try to make sure we don't screw up again.
> 
> I am lost a bit tbh here Jed had to pin to 0.5.4

I don't follow what is it you don't understand; you asked me to pin whereabouts v0.5.4, and I've replied: it **is** already pinned - check the commit to which the [whereabouts repo v0.5.4 release is pointing to](https://github.com/k8snetworkplumbingwg/whereabouts/releases/tag/v0.5.4).

> 
> https://github.com/kubevirt/kubevirtci/pull/847/files#diff-a6d30590f227704f4eb26e7638a13dcece7afba0a44baa1f15942ed5a47ddd50R218

oshoval (approved):
Thanks

oshoval:
> I don't follow what is it you don't understand; you asked me to pin whereabouts v0.5.4, and I've replied: it **is** already pinned - check the commit to which the [whereabouts repo v0.5.4 release is pointing to](https://github.com/k8snetworkplumbingwg/whereabouts/releases/tag/v0.5.4).

I expect that users of the repo should not have to pin a version like this PR is doing
so if someone uses manifest of 0.5.4 he will have version 0.5.4 already in the image field
we can continue this conv offline if needed

hack/kustomization/whereabouts/kustomization.yaml
```diff
@@ -7,5 +7,6 @@ resources:
 - ./whereabouts.cni.cncf.io_ippools.yaml
 - ./whereabouts.cni.cncf.io_overlappingrangeipreservations.yaml
 
-transformers:
-  - patch-ip-reconciler-job.yaml
+images:
```
oshoval, line 10: The old resources can be deleted please if possible (line 4)

xpivarc (commented):
/lgtm

oshoval:
/hold

need to cordon PR to be merged first, so lanes will pass

oshoval:
I think we are good 
for example this passed today
the PR was merged and it fixed it

we can remove the hold imo

PR #848, opened 2022-08-01
registry proxy: Fix crio support

oshoval:
Fix registry proxy support for crio (the CRI that we use).
It is an opt-in optional feature.

It will allow to to overcome `docker.io` rate limits,
and to cache container images locally in order to save time.

Tested with the following commands (works also with podman):
```
docker volume create registry-ca
docker volume create registry-mirror-cache

docker run -d --restart=always --name docker_registry_proxy -it -p 0.0.0.0:3128:3128 -e ENABLE_MANIFEST_CACHE=true -e ALLOW_PUSH=true -e PROXY_CONNECT_TIMEOUT=600s -e PROXY_CONNECT_READ_TIMEOUT=600s -e PROXY_CONNECT_CONNECT_TIMEOUT=600s -e PROXY_CONNECT_SEND_TIMEOUT=600s -e SEND_TIMEOUT=600s -e CLIENT_BODY_TIMEOUT=600s -e CLIENT_HEADER_TIMEOUT=600s -e PROXY_READ_TIMEOUT=600s -e PROXY_CONNECT_TIMEOUT=600s -e PROXY_SEND_TIMEOUT=600s --mount type=volume,source=registry-mirror-cache,target=/docker_mirror_cache --mount type=volume,source=registry-ca,target=/ca quay.io/kubevirtci/rpardini-docker-registry-proxy:0.6.2
```

Get proxy ip by
`docker inspect docker_registry_proxy -f '{{.NetworkSettings.Networks.bridge.IPAddress}}'`

Tested sig-network tests, kubevirt conformance, and k8s conformance with both
single stack IPv6 and dual stack with the proxy locally.

Ref: https://github.com/kubevirt/kubevirtci/pull/70 (which did the same for the previous CRI)

### DIFF
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -34,17 +34,19 @@ import (
 
 const (
 	proxySettings = `
-mkdir -p /etc/systemd/system/docker.service.d/
+curl {{.Proxy}}/ca.crt > /etc/pki/ca-trust/source/anchors/docker_registry_proxy.crt
+update-ca-trust
 
-cat <<EOT  >/etc/systemd/system/docker.service.d/proxy.conf
+mkdir -p /etc/systemd/system/crio.service.d
+cat <<EOT >/etc/systemd/system/crio.service.d/override.conf
 [Service]
 Environment="HTTP_PROXY={{.Proxy}}"
 Environment="HTTPS_PROXY={{.Proxy}}"
-Environment="NO_PROXY=localhost,127.0.0.1"
+Environment="NO_PROXY=localhost,127.0.0.1,registry,10.96.0.0/12,10.244.0.0/16,192.168.0.0/16,fd00:10:96::/112,fd00:10:244::/112,fd00::/64"
 EOT
 
 systemctl daemon-reload
-systemctl restart docker
+systemctl restart crio.service
 EOF
 `
 	etcdDataDir         = "/var/lib/etcd"
--- a/K8S.md
+++ b/K8S.md
@@ -37,7 +37,7 @@ SSH into a node
 cluster-up/ssh.sh node01                                                   
 ```                                                                        
                                                                            
-Attach to node console with screen and pty
+## Attach to node console with screen and pty
 ```                                                  
 # Attach to node01 console                           
 docker exec -it ${KUBEVIRT_PROVIDER}-node01 screen /dev/pts/0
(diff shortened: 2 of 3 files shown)


oshoval:
/cc @EdDev @qinqon 

This is a step to help us potentially use uncached kubevirtci providers
It is needed because of 2 reasons:
1. Overcome docker.io rate limits locally
2. Faster cluster-up, else each cluster-up will need to refetch the uncached images again and again

oshoval:
/cc @rmohr 
Mind taking a look please ?

rmohr:
/uncc @rmohr 
no time at the moment.

cluster-provision/gocli/cmd/run.go
```diff
@@ -34,17 +34,19 @@ import (
...
+update-ca-trust
 
-cat <<EOT  >/etc/systemd/system/docker.service.d/proxy.conf
+mkdir -p /etc/systemd/system/crio.service.d
+cat <<EOT >/etc/systemd/system/crio.service.d/override.conf
 [Service]
 Environment="HTTP_PROXY={{.Proxy}}"
 Environment="HTTPS_PROXY={{.Proxy}}"
-Environment="NO_PROXY=localhost,127.0.0.1"
+Environment="NO_PROXY=localhost,127.0.0.1,registry,10.96.0.0/12,10.244.0.0/16,192.168.0.0/16,8.8.8.8,fd00:10:96::/112,fd00:10:244::/112,fd00::/64,2001:db8:1::1"
```
enp0s3, line 45: From what I understand, the proxy configuration should only affect the traffic that is being originated from a host node, when CRIO tries to pull an image. Its not clear to me how does this proxy configuration affects pod-to-pod traffic, pod-to-service traffic, pod-to-internet traffic. In which way does the `iptables` configuration is being affected by these vars, that you had to add `8.8.8.8` to the NO_PROXY list?

enp0s3 (commented):
Hi, I have a question below

cluster-provision/gocli/cmd/run.go
oshoval: It configures a proxy, all the traffic of CRIO is going there, not only traffic from host node when trying to pull an image.
The fact that the proxy we are using is an image cache is just a special use case.
You can run it locally, and see that if you don't put for example the `registry` it will fail to load from it, because it will contact the proxy instead of the `registry`

enp0s3:
@brianmcarey Hi, can you PTAL?

cluster-provision/gocli/cmd/run.go
oshoval: Rechecked several times, it seems that when running the tests not part of the conformance,
they do pass even if 8.8.8.8, but flaky sometimes, the proxy adds some delay or so so the ping timeouts.
I am removing `8.8.8.8` and `2001:db8:1::1`, as It is controversial, works basically, and not crucial when running with proxy locally.

oshoval:
Updated NO_PROXY field

enp0s3 (approved):
/approve
@oshoval Thank you!

PR #850, opened 2022-08-02
Retire use of rook ceph toolbox

maya-r:
Was intended for a single kubevirt test that was retired in
kubevirt/kubevirt#6942


oshoval (approved):
Thanks

dhiller (approved):
/approve

Thanks!

PR #852, opened 2022-08-07
vm based providers: Create uncached version

oshoval:
In order to save time during development,
create a slim flavor of the providers by skip the caching part
of the optional images.

Using slim mode will also allow to provision providers
faster during development when the uncaching is not needed.


oshoval:
The problem that affected the nightly seems to affect all providers here as well 
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirt-push-nightly-conformance/1556105387738927104

not due to this PR

KUBEVIRTCI_LOCAL_TESTING.md
```diff
@@ -99,6 +99,12 @@ by the `linux` phase.
 Once you are done, either check the cluster manually, or use:  
 `export PHASES=k8s; export CHECK_CLUSTER=true; (cd cluster-provision/k8s/1.21; ../provision.sh)`
 
+### provision without caching images
```
rhrazdil, line 102: I'm not sure if the caching is immediately clear - we've called it `pre pulling` in kubevirtci provision.sh.
Maybe
```suggestion
### provision without pre-pulling images
```
Would be more self-descriptive?

KUBEVIRTCI_LOCAL_TESTING.md
```diff
@@ -99,6 +99,12 @@ by the `linux` phase.
...
+
+In order to develop faster, you can skip caching the optional images,
```
rhrazdil, line 104: ditto

K8S.md
```diff
@@ -36,7 +36,13 @@ SSH into a node
 ```                                                                        
 cluster-up/ssh.sh node01                                                   
 ```                                                                        
-                                                                           
+
+Use slim provider (without cached images)
```
rhrazdil, line 40: ditto

K8S.md
```diff
@@ -36,7 +36,13 @@ SSH into a node
 ```                                                                        
 cluster-up/ssh.sh node01                                                   
 ```                                                                        
-                                                                           
+
+Use slim provider (without cached images)  
+In order to save time downloading the image, set the following before cluster-up
```
rhrazdil, line 41: Maybe worth mentioning, in which use-cases this is beneficial - IUUC typically when extra components like CNAO,CNI,Istio are note required.

Otherwise pulling these separately after cluster-up wouldn't really save much, compared to pulling them as part of kubevirtci image.

rhrazdil (changes requested):

oshoval:
Thanks, addressed comments

K8S.md
oshoval: updated to use short explanation
bit more info is on KUBEVIRTCI_LOCAL_TESTING.md

oshoval:
errors are expected and not related to this PR, waiting for the fix to be merged

rmohr:
> * We might want to try using slim mode on CI itself, because this way each lane will pull only the images it
>   needs instead pulling them all.
>   For example sig-network would pull CNAO images, but not all images such as CDI,
>   saving time to download and extract it.

I would not over-optimize. Then you suddenly have a test where you need CDI and it extra-pulls anyway. Also ensuring availability is harder, in case that images disappear and are no longer in a proxy.

We now finally have thanks to @brianmcarey  podman-in-docker, which allows us for CI to have read-only registries where we can pre-pull kubevirtci images, reducing io spikes (network bandwidth is no issue in CI realistically). See [#podman](https://github.com/kubevirt/project-infra/issues/1363#issue-928024234) for details if you are intersted.

rmohr:
I wonder a little bit about the motivation. How often do you pull down different providers? They only get updated every few days. Not saying it may not be an improvement if you are a power-user, but I rarely download more than 1-2 providers on a power-user-day, therefore I always favoured simplicity over adding additional complexity.

oshoval:
> > * We might want to try using slim mode on CI itself, because this way each lane will pull only the images it
> >   needs instead pulling them all.
> >   For example sig-network would pull CNAO images, but not all images such as CDI,
> >   saving time to download and extract it.
> 
> I would not over-optimize. Then you suddenly have a test where you need CDI and it extra-pulls anyway. Also ensuring availability is harder, in case that images disappear and are no longer in a proxy.

Np, just note that for example on sig-network we don't deploy CDI already for long time (but we do deploy ISTIO and CNAO),
Regarding disappearing images we still create both slim and not slim, and moved away from docker.io fwiw,
so it is less likely to happen and if it happens it is easy to move to the full image simply by unset the KUBEVIRT_SLIM which make it go back to default.

Will remove it from PR desc to make it simpler anyhow.
Thanks

rmohr:
> Np, just note that for example on sig-network we don't deploy CDI already for long time (but we do deploy ISTIO and CNAO),

Yep, the funny part is, that as soon as we start pre-pulling, it is cheaper to use the whole thing (because the images already part of the image and they can be just used for all CI jobs in parallel vs. having to pulling them to the container overlay on each run). That may not be obvious.

> Regarding disappearing images we still create both slim and not slim, and moved away from docker.io fwiw,

Yep all good, just wanted to point this out.

oshoval:
> I wonder a little bit about the motivation. How often do you pull down different providers? They only get updated every few days. Not saying it may not be an improvement if you are a power-user, but I rarely download more than 1-2 providers on a power-user-day, therefore I always favoured simplicity over adding additional complexity:
> 
> * no rate-limiting thinking, mirroring or proxying necessary
> * no additional paths in the providers with potential issues

Some people like @EdDev running it on their laptop which is more limited by resources
I find myself also provisioning / cluster-up pretty a lot. and each time a new version exist (as there is one on each kubevirtci merge / every kubevirt kubevirtci bump) it pulls another 15GB uncompressed image) wasting 5 minutes that can be saved.
Every prune of the images which is the common solution when disk is full is making you to repull the image.

For the developers of kubevirtci, the slim mode can allow to provision faster without hacks to skip pre pulling (2nd bullet)

For the users, the default is still there, but usually we dont need CDI, CNAO, ISTIO, Prometheus, just a basic provider,
so we can just export `KUBEVIRT_SLIM=true` and this will select the right slim image based on the same `KUBEVIRTCI_TAG`
Running cluster-up in 2.5 minutes instead of 7.5 minutes feels much better imo, and doesn't fill hd so fast

rmohr:
> I wonder a little bit about the motivation. How often do you pull down different providers? They only get updated every few days. Not saying it may not be an improvement if you are a power-user, but I rarely download more than 1-2 providers on a power-user-day, therefore I always favoured simplicity over adding additional complexity:
> 
> * no rate-limiting thinking, mirroring or proxying necessary
> * no additional paths in the providers with potential issues

@oshoval I have btw. no strong feeling either way on this. I am mostly curious to understand if you really have the need.

oshoval:
> > Np, just note that for example on sig-network we don't deploy CDI already for long time (but we do deploy ISTIO and CNAO),
> 
> Yep, the funny part is, that as soon as we start pre-pulling, it is cheaper to use the whole thing (because the images already part of the image and they can be just used for all CI jobs in parallel vs. having to pulling them to the container overlay through a proxy on each run). That may not be obvious.

Indeed but note please that even pulling the full image from proxy on CI (and untaring right ?) takes time that can be saved,
and we strive at least on sig-nework for example to not deploy anything that is not needed (i.e CDI, Prom)

> 
> > Regarding disappearing images we still create both slim and not slim, and moved away from docker.io fwiw,
> 
> Yep all good, just wanted to point this out.

Np thanks

awels:
Would this in help in the case we are running into right when trying to backport a fix. Basically we need a fixed version of CDI (fixes half the problem) and a fixed version of KubeVirt (fixes the other half). This works fine in main, but when we start backporting we ran into the problem that older branches have an older CDI that is not fixed and the functional tests are failing. We released a fixed older version of CDI so we can use it in KubeVirt, but we are not able to really use it due to CDI being bundled with kubevirtci and no good way of overriding it. @brybacki emailed kubevirt-dev about this scenario.

rmohr:
> > > Np, just note that for example on sig-network we don't deploy CDI already for long time (but we do deploy ISTIO and CNAO),
> > 
> > 
> > Yep, the funny part is, that as soon as we start pre-pulling, it is cheaper to use the whole thing (because the images already part of the image and they can be just used for all CI jobs in parallel vs. having to pulling them to the container overlay through a proxy on each run). That may not be obvious.
> 
> Indeed but note please that even pulling the full image from proxy on CI (and untaring right ?) takes time that can be saved, and we strive at least on sig-nework for example to not deploy anything that is not needed (i.e CDI, Prom)

Still, it would require you to have all kinds of slim-mode combinations to not have to pull through the proxy (pulling through the proxy would be the most expensive operation on time and disk-io on each ci run). It is cheaper to just have on image for all CI lanes then with everythign in there, even if you don't deploy CDI.

oshoval:
> Would this help in the case we are running into right when trying to backport a fix. Basically we need a fixed version of CDI (fixes half the problem) and a fixed version of KubeVirt (fixes the other half). This works fine in main, but when we start backporting we ran into the problem that older branches have an older CDI that is not fixed and the functional tests are failing. We released a fixed older version of CDI so we can use it in KubeVirt, but we are not able to really use it due to CDI being bundled with kubevirtci and no good way of overriding it. @brybacki emailed kubevirt-dev about this scenario.

If you use the SLIM version, you have total control of what CDI and kubevirt to deploy (kubevirt anyhow)
note that even now, you can set `KUBEVIRT_DEPLOY_CDI=false` and CDI won't be installed, allowing you to install your own CDI version. Make sense ?

oshoval:
> > Indeed but note please that even pulling the full image from proxy on CI (and untaring right ?) takes time that can be saved, and we strive at least on sig-nework for example to not deploy anything that is not needed (i.e CDI, Prom)
> 
> Still, it would require you to have all kinds of slim-mode combinations to not have to pull through the proxy (pulling through the proxy would be the most expensive operation on time and disk-io on each ci run). It is cheaper to just have on image for all CI lanes then with everythign in there, even if you don't deploy CDI.

Of course that creating a matrix of combination is not good
But i am not sure that installing basic + CNAO + ISTIO lets say (which is what we are doing on sig-network) is not faster than downloading the whole one,
I tried that locally with the local proxy and it seems faster (proxy PR in this PR desc)

rmohr:
> Of course that creating a matrix of combination is not good
> But i am not sure that installing basic + CNAO + ISTIO lets say (which is what we are doing on sig-network) is not faster than downloading the whole one,
> I tried that locally with the local proxy and it seems faster (proxy PR in this PR desc)

Again, the whole image would be downloaded to the whole node once and all CI jobs can use it in parallel without any download at all, which allowes improved disk cache sharing in the kernel. All I want to point out is that the cheapest variant is this then, although that may sound surprising initially.

K8S.md
```diff
@@ -36,7 +36,12 @@ SSH into a node
 ```                                                                        
 cluster-up/ssh.sh node01                                                   
 ```                                                                        
-                                                                           
+
+Use slim provider (without pre-pulled images of the optional components such as CDI, CNAO etc)
+```
+export KUBEVIRT_SLIM=true
```
brianmcarey, line 42: I am not sure if this is the best place for this information - reading the doc as a whole this update seems to break the flow and is not really relevant?

brianmcarey (commented):
I tend to agree with @rmohr here - I am not sure how much this is needed. I run kubevirtci providers quite regularly locally on my laptop and I haven't had disk space issues - but I guess I do run `system prune` every so often. In the CI environment the kubevirtci provider images are downloaded once to each of the nodes and just reused from there so I am not sure if the slim variant would be useful there. 

But I guess this will help developers who are restricted by storage or network performance.

oshoval:
> I tend to agree with @rmohr here - I am not sure how much this is needed. I run kubevirtci providers quite regularly locally on my laptop and I haven't had disk space issues - but I guess I do run `system prune` every so often. In the CI environment the kubevirtci provider images are downloaded once to each of the nodes and just reused from there so I am not sure if the slim variant would be useful there.
> 
> But I guess this will help developers who are restricted by storage or network performance.

It will affect only local users + people who are doing provision
It saves a lot of time / space locally, CI won't use it, as indeed shareable read only storage is helpful

K8S.md
oshoval: I can move it to cluster-up/cluster/K8S.md
and / or add a "misc" section in that file, just before the new text
as also the console help below it, is pretty misc

wdyt ?

awels:
> > Would this help in the case we are running into right when trying to backport a fix. Basically we need a fixed version of CDI (fixes half the problem) and a fixed version of KubeVirt (fixes the other half). This works fine in main, but when we start backporting we ran into the problem that older branches have an older CDI that is not fixed and the functional tests are failing. We released a fixed older version of CDI so we can use it in KubeVirt, but we are not able to really use it due to CDI being bundled with kubevirtci and no good way of overriding it. @brybacki emailed kubevirt-dev about this scenario.
> 
> If you use the SLIM version, you have total control of what CDI and kubevirt to deploy (kubevirt is not bundled anyhow in kubevirtci, maybe just on other repos IIRC). The manifests are still there, just the images aren't, so you can install your own manifests of course (`KUBEVIRT_DEPLOY_CDI` should be false). Note that even now, you can set `KUBEVIRT_DEPLOY_CDI=false` and CDI won't be installed, allowing you to install your own CDI version. Make sense ? (i hope i understood correct)

Right, but the backport PR for KubeVirt would now require in that branch to always not use the CDI inside kubevirtci image, and install the one that contains the fix that allows the functional test to pass. There is an email thread in kubevirt-dev that explains the problem in detail. This PR might not be the right solution though.

K8S.md
oshoval: > I think the console information is useful to have there for someone starting out with kubevirtci who may be having issues connecting to the nodes over ssh.

You mean to move `Attach to node console with screen and pty` to their as well ?
If so it can be done on a follow PR please, as it is not related to this PR
I will move the related part to that file.

brianmcarey: Sorry I meant to keep the console information where it is as it is useful there.

oshoval:
Moved the slim notes to `cluster-up/cluster/K8S.md`(according comments)

rmohr:
> I tend to agree with @rmohr here - I am not sure how much this is needed. I run kubevirtci providers quite regularly locally on my laptop and I haven't had disk space issues - but I guess I do run `system prune` every so often. In the CI environment the kubevirtci provider images are downloaded once to each of the nodes and just reused from there so I am not sure if the slim variant would be useful there.
> 
> But I guess this will help developers who are restricted by storage or network performance.

Yeah does not look like it complicates things too much. I have no issue with it.

oshoval:
drafing until kubevirtci is stable / PR is lgtmed

rhrazdil:
From my perspective, the code looks good. The added setting doesn't change CI flow, and saves a couple gigs for
regular users that spin up the cluster.
/lgtm

oshoval:
The failures are due to, saw it also not on this PR, will retest bit later
`iptables: Resource temporarily unavailable.`

publish.sh
```diff
@@ -23,8 +23,12 @@ for i in ${CLUSTERS}; do
     if [[ $i =~ ipv6 ]]; then
         NETWORK_STACK="ipv6"
     fi
+
     cluster-provision/gocli/build/cli provision cluster-provision/k8s/$i --network-stack ${NETWORK_STACK}
     docker tag ${TARGET_REPO}/k8s-$i ${TARGET_REPO}/k8s-$i:${KUBEVIRTCI_TAG}
+
+    cluster-provision/gocli/build/cli provision cluster-provision/k8s/$i --network-stack ${NETWORK_STACK} --slim
```
brianmcarey, line 30: This is going to increase the run time for the publish-kubevirtci job which I think has a timeout currently of 3 hours - @oshoval - from your testing does this additional provision add much time? 
https://github.com/kubevirt/project-infra/blob/4f229481a8feb975199efc7ee853534c78aab8af/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-postsubmits.yaml#L3-L11

brianmcarey (commented):
Looks good, just a question on the impact to publishing the providers?

publish.sh
oshoval: Atm we stand between 1:20 - 1:50 as we can see here
https://prow.ci.kubevirt.io/?job=publish-kubevirtci

It will add time, but not alot, because the images are small and there is no caching needed
lets say 5m x 4 to create the providers + 3m x 4 to push them, so it adds around 35m

Sometimes the whole lane is slower, and then it might be marginal
lets run it once and then see by how much to extend ?
It should pass according the benchmarks above

dhiller:
Sorry I am late to the party, and I totally understand wanting to have stripped down images for local development. Every time the images are fetched I get me a coffee or do something else.

But since this doesn't happen so often, I agree with Roman that this looks like an over optimization. Also this is another case of increased complexity where I don't know if it's worth supporting this.

OTOH I see the image size reduced by 2/3 which could really help for minimal requirements of optional components.

From a benchmark perspective I think that the total time that should be measured here would not only be the download of the actual image but also the download of the components that are required additionally, so the total time of cluster-up with the required components (ISTIO, CNAO) that i.e. you @oshoval need to have the cluster fully functional vs. the pre-pulled cluster-up with the optional components.

AKA

```
#export KUBEVIRT_SLIM=true
export KUBEVIRT_DEPLOY_ISTIO=true
export $KUBEVIRT_WITH_CNAO=true
...

docker system prune
time make cluster-up
```


However, if enough people support this change, I find it good to go and would agree to take it in.

oshoval:
> Sorry I am late to the party, and I totally understand wanting to have stripped down images for local development. Every time the images are fetched I get me a coffee or do something else.
> 
> But since this doesn't happen so often, I agree with Roman that this looks like an over optimization. Also this is another case of increased complexity where I don't know if it's worth supporting this.

We do see it happen pretty often, especially where people prune and such (but not only).
 
> 
> OTOH I see the image size reduced by 2/3 which could really help for minimal requirements of optional components.
> 
> From a benchmark perspective I think that the total time that should be measured here would not only be the download of the actual image but also the download of the components that are required additionally, so the total time of cluster-up with the required components (ISTIO, CNAO) that i.e. you @oshoval need to have the cluster fully functional vs. the pre-pulled cluster-up with the optional components.

Usually we don't need ISTIO, CNAO and so on when we are developing locally, so need to calculate
amortized time imho.

> 
> AKA
> 
> ```
> #export KUBEVIRT_SLIM=true
> export KUBEVIRT_DEPLOY_ISTIO=true
> export $KUBEVIRT_WITH_CNAO=true
> ...
> 
> docker system prune
> time make cluster-up
> ```
> 
> However, if enough people support this change, I find it good to go and would agree to take it in.

Great thanks, I think we would benefit from it, and as raised above by Roman, this change doesn't complicate
things so much.

publish.sh
oshoval: Merged, ptal

brianmcarey (approved):
/approve

The slim version is off by default but kubevirtci users may still benefit from the reduced image size if they enable it.

oshoval:
Thanks

kind is not related

/test check-up-kind-1.23-vgpu

PR #853, opened 2022-08-08
grafana: Use quay mirror

oshoval:
Kustomize `grafana` to use quay image.

Had to do the following fix for that:    
Currently manifests are copied both on `linux`
and `k8s` phases.
In case using phased mode, the changes
that are done on the `linux` phase would be overridden.
Some changes should be done on `linux` phase
because the images are pre pulled on this phase,
so it should be done before pre pulling.
Therefore copy the manifests only on the `linux` phase.

Additional cosmetic change:
Verify that the script exists just before running it.

### DIFF
--- a/cluster-provision/gocli/cmd/provision.go
+++ b/cluster-provision/gocli/cmd/provision.go
@@ -251,25 +251,14 @@ func provisionCluster(cmd *cobra.Command, args []string) (retErr error) {
 		return err
 	}
 
-	// Copy scripts to the VM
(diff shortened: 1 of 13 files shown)


dhiller:
nit: use quay.io/kubevirtci/grafana/grafana? This way the hierarchy would be preserved

image

oshoval:
> quay.io/kubevirtci/grafana/grafana

I am using the automatic mirror script that replace / with _ so there where the name came from
I don't think we have quay.io/x/y/z only quay.io/x/y:label
if xyz is legit, then it worth a follow PR to update first the auto mirror and then to change it imo

rhrazdil:
/lgtm
/hold

Looks good to me, but adding a hold for @dhiller to have a chance to react. I think it's reasonable to keep the mirror script simple, rather than deal with all of the special cases.

rhrazdil:
Removing the unnecessary hold as the PR is not approved
/hold cancel

oshoval:
Rebase

PR #859, opened 2022-08-18
Don't wait for ceph toolbox to be deployed: we no longer deploy it

maya-r:
[We are failing to cluster-up now.](https://github.com/kubevirt/kubevirt/pull/8292#issuecomment-1219208441)


oshoval (approved):
Thanks
Can you please update the PR desc with the info inline ?

maya-r:
cc @brianmcarey

brianmcarey (approved):
/approve

oshoval:
I think we need to allow parallel runs of check-provision-k8s-1.2x ?
else when there are few kubevirtci PRs, it pending long time

PR #860, opened 2022-08-21
local storage: Fix deploy flow

oshoval:
Since local storage is deployed now after ceph,
ceph can't update local storage labels.

Remove the labels from the local storage manifests,
and add them only if no other default storage was detected.
This way ceph won't need to update them at all.

### DIFF
--- a/cluster-up/cluster/k8s-provider-common.sh
+++ b/cluster-up/cluster/k8s-provider-common.sh
@@ -98,6 +98,11 @@ function deploy_cdi() {
 function deploy_local_storage() {
     if [ "$KUBEVIRT_DEPLOY_LOCAL_STORAGE" == "true" ]; then
         $kubectl create -f /provision/local-volume.yaml
+
+        default_storage_class=$($kubectl get storageclass -o jsonpath='{\.items[?\(@\.metadata\.annotations\.storageclass\\.kubernetes\\.io/is-default-class==\"true\"\)].metadata.name}')
+        if [ -z $default_storage_class ]; then
+            $kubectl patch storageclass/local -p \'{\"metadata\": {\"annotations\":{\"storageclass.kubernetes.io/is-default-class\":\"true\"}}}\'
+        fi
     fi
 }
 
--- a/cluster-provision/k8s/1.22/manifests/local-volume.yaml
+++ b/cluster-provision/k8s/1.22/manifests/local-volume.yaml
@@ -2,8 +2,6 @@ apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
  name: local
- annotations:
-  storageclass.kubernetes.io/is-default-class: "true"
 provisioner: kubernetes.io/no-provisioner
 volumeBindingMode: WaitForFirstConsumer
 reclaimPolicy: Delete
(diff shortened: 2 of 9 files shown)


oshoval:
/cc @brianmcarey 
Trying to fix 
https://github.com/kubevirt/kubevirt/pull/8292#issuecomment-1220393033

For some reason couldn't test it locally (ceph fails to install even before it reaches the PR changes)
so trying on CI

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -98,6 +98,9 @@ function deploy_cdi() {
 function deploy_local_storage() {
     if [ "$KUBEVIRT_DEPLOY_LOCAL_STORAGE" == "true" ]; then
         $kubectl create -f /provision/local-volume.yaml
+        if [ "$KUBEVIRT_STORAGE" == "rook-ceph-default" ] || [[ "$KUBEVIRT_PROVIDER_EXTRA_ARGS" =~ enable-ceph ]]; then
```
maya-r, line 101: Perhaps this should be formulated as:
- Don't set as default storage class on the YAML (neither true nor false)
- If there is no other default storage class by this point, set local as default storage class.
To avoid hard-coding the logic of ceph

oshoval: pushed change that does that + fixes the [failure](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/860/check-provision-k8s-1.24/1561257325266735104) (will order commits once it passes)

maya-r: can use:
```
default_storage_class = $($kubectl get storageclass -o jsonpath='{.items[?(@.metadata.annotations.storageclass\.kubernetes\.io/is-default-class=="true")].metadata.name}')
if [ -z $default_storage_class ];
```
to test if there is no default storage class, instead of having logic specific to ceph

oshoval: nice done

for some reason there is a failure here
not sure why yet

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/860/check-provision-k8s-1.24/1561265652461735936

oshoval:
The failure is due to timeout in check-cluster-up, I added deploying of ceph just for testing, doesnt need to be there.
Removing it from there, the fact that we reached there should be enough

maya-r:
I hope it's not because rook needs local storage

oshoval:
> I hope it's not because rook needs local storage

I didn't change the fact that local storage is deployed, just where it is deployed (and not due to this PR)
There is a timeout of 300 seconds there, this what failed fwiw

maya-r:
alright... not sure why ceph wouldn't deploy locally.
/lgtm

oshoval:
It seems that even with sig-network we use storage as part of SetupAlpineHostPath
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/860/check-provision-k8s-1.24/1561333256777895936

I am closing this PR and reverting the change of the storage flag for now

PR #862, opened 2022-08-21
kind: Remove local volume manifest

oshoval:
We don't use it on kind


oshoval:
will make sure first locally that it isn't deployed, because for example it might be used for SetupAlpineHostPath

PR #855, opened 2022-08-10
vm based providers: Introduce KUBEVIRT_DEPLOY_LOCAL_STORAGE

oshoval:
Add an env var KUBEVIRT_DEPLOY_LOCAL_STORAGE
which toggles deployment of local volume storage
(default value: true).

PR #863, opened 2022-08-21
Revert "vm based providers: Introduce KUBEVIRT_DEPLOY_LOCAL_STORAGE"

oshoval:
This reverts commit 0be2b24c92442865cde90014b88cd17631056ea7.
https://github.com/kubevirt/kubevirtci/pull/855

It cause problems because ceph update labels of local storage,
which is deployed later now.

While trying to fix [it](https://github.com/kubevirt/kubevirtci/pull/860), saw that even sig-network uses storage
(as part of SetupAlpineHostPath),
so there isn't actually a case where we don't need storage at all.
Therefore the variable isn't needed.


cluster-provision/k8s/1.23/k8s_provision.sh
```diff
@@ -42,9 +42,7 @@ systemctl daemon-reload
 systemctl enable crio && systemctl start crio
 systemctl enable kubelet && systemctl start kubelet
 
-# Version 1.39.7-x cause the VM to hang on random places
-NETWORK_MANAGER_VERSION=1.39.5-1
-dnf install -y NetworkManager-${NETWORK_MANAGER_VERSION}.el8.x86_64 NetworkManager-ovs-${NETWORK_MANAGER_VERSION}.el8.x86_64
```
maya-r, line 47: Picked up some extra diffs?

oshoval:
Clean unrelated commits

brianmcarey (approved):
/approve

PR #867, opened 2022-08-24
vm based providers: Expose a UDP port for DNS

oshoval:
Extend vm based providers to support exposing udp ports.

Expose a custom UDP port by the vm based providers.
This allows to expose a nodePort of a service, set it's node port to `31111`,
and then it will be reachable from the host itself via DNS default port `53`.

The flow is:
dnsMasq container exposes `127.0.0.1:53 -> 31111`
The traffic that reaches the container via destination port `31111` is DNATed
to the `control-plane node ip` at destination port `31111`.
We can create a NodePort on `control-plane node ip` port `31111`, pointing to
`kube-dns` (CoreDNS's service).
This way the CoreDNS pod will be reachable from the host itself via `127.0.0.1:53`

Additional changes:
Add debug packages to the base image - helps to develop and debug, instead installing
them every time that it is needed.

### DIFF
--- a/cluster-provision/centos8/scripts/vm.sh
+++ b/cluster-provision/centos8/scripts/vm.sh
@@ -73,16 +73,28 @@ else
   iptables -t nat -A OUTPUT -p tcp --dport 22${n} -j DNAT --to-destination 192.168.66.1${n}:22
 fi
 
+function create_ip_rules {
+  protocol=$1
+  shift
+  if [ "$ROOTLESS" -ne 1 ]; then
+    for port in "$@"; do
+      iptables -t nat -A PREROUTING -p ${protocol} -i eth0 -m ${protocol} --dport ${port} -j DNAT --to-destination 192.168.66.101:${port}
+    done
+  else
+    for port in "$@"; do
+      # Add DNAT rule for rootless podman (traffic originating from loopback adapter)
+      iptables -t nat -A OUTPUT -p ${protocol} --dport ${port} -j DNAT --to-destination 192.168.66.101:${port}
+    done
+  fi
+}
+
 # Route ports from container to VM for first node
 if [ "$n" = "01" ] ; then
-  for port in 6443 8443 80 443 30007 30008 31001; do
-    if [ "$ROOTLESS" -ne 1 ]; then
-      iptables -t nat -A PREROUTING -p tcp -i eth0 -m tcp --dport ${port} -j DNAT --to-destination 192.168.66.1${n}:${port}
-    else
-      # Add DNAT rule for rootless podman (traffic originating from loopback adapter)
-      iptables -t nat -A OUTPUT -p tcp --dport ${port} -j DNAT --to-destination 192.168.66.1${n}:${port}
-    fi
-  done
+  tcp_ports=( 6443 8443 80 443 30007 30008 31001 )
+  create_ip_rules "tcp" "${tcp_ports[@]}"
+
+  udp_ports=( 31111 )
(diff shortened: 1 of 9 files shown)


oshoval:
@brianmcarey 
I think we still have CRIO 1.24 problem ?
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/867/check-provision-k8s-1.24-ipv6/1564165955808923648

brianmcarey:
> @brianmcarey I think we still have CRIO 1.24 problem ? https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/867/check-provision-k8s-1.24-ipv6/1564165955808923648
> 
> cnao forever loop

Yes it looked like it picked up 1.24.1 - I will take a look at the mirror to see what happened.

cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -87,6 +87,10 @@ function _registry_volume() {
 function _add_common_params() {
     # shellcheck disable=SC2155
     local params="--nodes ${KUBEVIRT_NUM_NODES} --memory ${KUBEVIRT_MEMORY_SIZE} --cpu 6 --secondary-nics ${KUBEVIRT_NUM_SECONDARY_NICS} --random-ports --background --prefix $provider_prefix ${KUBEVIRT_PROVIDER} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
+
+    dns_host_port=53
```
qinqon, line 91: This can be problematic for stuff like NetorkManager with dnsmasq where the dns server is running at 127.0.0.1

oshoval: on fedora for example the dns service runs on 127.0.0.53

do we have a real example where it is problematic ? can we first meet that problem and then fix if needed ?
it will be need to only parameterize this parameter so the user can supply it
(and change tests that we will present with the new feature with `nslookup -p`, assuming this way we are going to test)

cluster-provision/centos8/scripts/vm.sh
```diff
@@ -73,16 +73,28 @@ else
...
-    else
-      # Add DNAT rule for rootless podman (traffic originating from loopback adapter)
-      iptables -t nat -A OUTPUT -p tcp --dport ${port} -j DNAT --to-destination 192.168.66.1${n}:${port}
-    fi
-  done
+  tcp_ports=( 6443 8443 80 443 30007 30008 31001 )
+  create_ip_rules "tcp" "${tcp_ports[@]}"
+
+  udp_ports=( 31111 )
+  create_ip_rules "udp" "${udp_ports[@]}"
```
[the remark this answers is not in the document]
oshoval: hmm but why ?
the if is on other condition totally
we can just use const for example in the function itself wdyt?

oshoval:
Addressed https://github.com/kubevirt/kubevirtci/pull/867#discussion_r1006712882
and rebased (2nd push)

cluster-up/cluster/ephemeral-provider-common.sh
maya-r: I had to edit my network settings because I was running dnsmasq locally already

oshoval: PR to fix that
https://github.com/kubevirt/kubevirtci/pull/903

PR #875, opened 2022-09-13
Retry removing containers that may have failed to delete due to dependent containers

brianmcarey:
During kubevirtci provisioning, the provision-dnsmasq container fails to
delete when run with podman as there are dependent containers still
running[1].

Adding a retry here successfully removes the provision-dnsmasq container
as there are no longer any dependent containers.

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/publish-kubevirtci/1565349289993965568#1:build-log.txt%3A4745

/cc @oshoval @dhiller @enp0s3 

### DIFF
--- a/cluster-provision/gocli/docker/docker.go
+++ b/cluster-provision/gocli/docker/docker.go
@@ -207,6 +207,8 @@ func NewCleanupHandler(cli *client.Client, cleanupChan chan error, errWriter io.
 	go func() {
 		createdContainers := []string{}
 		createdVolumes := []string{}
+		containersFailedToRemove := []string{}
+
 		defer close(done)
 
 		for {
@@ -229,6 +231,13 @@ func NewCleanupHandler(cli *client.Client, cleanupChan chan error, errWriter io.
 								io.Copy(os.Stderr, reader)
 							}
 						}
+						err := cli.ContainerRemove(ctx, c, types.ContainerRemoveOptions{Force: true})
+						if err != nil {
+							fmt.Fprintf(errWriter, "%v\n", err)
+							containersFailedToRemove = append(containersFailedToRemove, c)
+						}
+					}
+					for _, c := range containersFailedToRemove {
 						err := cli.ContainerRemove(ctx, c, types.ContainerRemoveOptions{Force: true})
 						if err != nil {
 							fmt.Fprintf(errWriter, "%v\n", err)


cluster-provision/gocli/cmd/rm.go
```diff
@@ -39,12 +39,12 @@ func rm(cmd *cobra.Command, _ []string) error {
...
 	}
 
-	var dnsmasq *types.Container
+	var dnsmasq []*types.Container
 nodnsmasq:
 	for i, c := range containers {
 		for _, name := range c.Names {
 			if strings.HasSuffix(name, "dnsmasq") {
-				dnsmasq = &containers[i]
+				dnsmasq = append(dnsmasq, &containers[i],)
```
oshoval, line 47: wouldn't it readd again and again the dnsmasq container because the continue in next line ?
(iterating on the containers again, trying to delete already deleted ones, and adding the already added ones to the list)
on the other hand, it would have happen also before if so, so i guess i am missing something

oshoval (commented):
Thanks,
did you check it solves the publish job ?

Asking because maybe the code that is running is NewCleanupHandler and not this one?

Also i am not sure it runs this code, because if it would it does have already `dnsmasq` as suffix
(unless the prefix is different, few lines above, and then it wont help)
and during provision we have only the "k8s-1.24-provision-dnsmasq" one, not both
so not sure we need an array actually
wdyt?

cluster-provision/gocli/docker/docker.go
```diff
@@ -232,6 +234,15 @@ func NewCleanupHandler(cli *client.Client, cleanupChan chan error, errWriter io.
 						err := cli.ContainerRemove(ctx, c, types.ContainerRemoveOptions{Force: true})
 						if err != nil {
 							fmt.Fprintf(errWriter, "%v\n", err)
+							failedContainers = append(failedContainers, c)
```
oshoval, line 237: What about skipping remove for dnsmasq container here, and removing it at the end instead deleting it, failing to delete and handing it please ?

brianmcarey: > What about skipping remove for dnsmasq container here, and removing it at the end instead deleting it, failing to delete and handing it please ?

`createdContainers` just contains a list of container IDs as strings so its hard to identify which container ID maps to the dnsmasq container. 
> 
> Btw do we know that we enter this loop ? asking because forceClean is false, so it means err != nil right ? I wonder why err != nil if so, but this is other story. (we do see that we fail to delete the dnsmasq so we do enter some cleanup logic, I didn't find any other than this)

publish-kubevirtci runs the provision so it hits `NewCleanupHandler` here - https://github.com/kubevirt/kubevirtci/blob/d14be236cab7d2f219398611dd0a8ade8ce8b35f/cluster-provision/gocli/cmd/provision.go#L131
which has `forceClean` as true.

cluster-provision/gocli/docker/docker.go
```diff
@@ -232,6 +234,15 @@ func NewCleanupHandler(cli *client.Client, cleanupChan chan error, errWriter io.
...
+						}
+					}
+					if failedContainers != nil {
```
oshoval, line 240: no need the condition, as the range would just run on empty list, isn't it ?

Maybe worth renaming to something like containerRemoveRetry / containerRemoveFailed ?

cluster-provision/gocli/docker/docker.go
oshoval: > > What about skipping remove for dnsmasq container here, and removing it at the end instead deleting it, failing to delete and handing it please ?
> 
> `createdContainers` just contains a list of container IDs as strings so its hard to identify which container ID maps to the dnsmasq container.

you can do something like filterByPrefix, but whatever you prefer

oshoval (approved):
Thanks

cluster-provision/gocli/docker/docker.go
```diff
@@ -207,6 +207,8 @@ func NewCleanupHandler(cli *client.Client, cleanupChan chan error, errWriter io.
 	go func() {
 		createdContainers := []string{}
 		createdVolumes := []string{}
+		containerRemoveRetry := []string{}
```
dhiller, line 210: I find this variable name a bit odd to read at first. Maybe we can rename this to reflect better what it holds? Suggestion:
```suggestion
		containersFailedToRemove := []string{}
```

brianmcarey: Yes I think your suggested name is better - I will make the update - thanks @dhiller !

oshoval (approved):

cluster-provision/gocli/docker/docker.go
```diff
@@ -229,6 +231,13 @@ func NewCleanupHandler(cli *client.Client, cleanupChan chan error, errWriter io.
 								io.Copy(os.Stderr, reader)
 							}
 						}
+						err := cli.ContainerRemove(ctx, c, types.ContainerRemoveOptions{Force: true})
+						if err != nil {
+							fmt.Fprintf(errWriter, "%v\n", err)
```
oshoval, line 236: since we are going to retry them, and expect that it will fail, maybe don't print it here, just if the retry fails ?

style nit: (not mandatory)
```
if err := ...; err != nil {
}
```

PR #879, opened 2022-09-21
Bump minor version of 1.24-ipv6 & 1.24-psa providers

brianmcarey:
This should be handled by the automated PR[1] but it is currently missed.

Bumping the minor version of the 1.24 provider variants.

[1] https://github.com/kubevirt/kubevirtci/pull/877


oshoval (approved):
Thanks
Indeed it is missing by the automation
Not sure we need to bother doing it manually in the future until it is fixed as part of the automation

brianmcarey:
> Thanks Indeed it is missing by the automation Not sure we need to bother doing it manually in the future until it is fixed as part of the automation

I will create an issue for the automation fix.

dhiller (approved):
/approve

oshoval:
is there a 1.24-psa presubmit lane ? even manual one ?

xpivarc:
> is there a 1.24-psa presubmit lane ? even manual one ?

Yes, there is.

xpivarc:
> /test check-provision-k8s-1.24-psa

Unfortunately, this will not work afaik.

oshoval:
> > /test check-provision-k8s-1.24-psa
> 
> Unfortunately, this will not work afaik.

It fails with sonobuoy, lets just skip sonobuoy on it then ?
we might want to make it always run, especially without sonobuoy, else if it fails it will fail the publish
but we can be brave and just fix in case there are problems

PR #881, opened 2022-09-22
vm based providers: Use ssh.sh instead of scp

oshoval:


oshoval:
it wa some of the problems but doesn't fully work yet

PR #872, opened 2022-08-31
kubevirtci 1.25 provider

dhiller:
This is a replacement for the automated PR #866 

Todo:

- [ ] apply workaround for [fetching coredns image issue](https://github.com/kubernetes/kubernetes/issues/112131#issuecomment-1232279438) 
- [ ] Remove deprecated elements (see [deprecation-guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/) )
  - [ ]  PodDisruptionBudget 

/cc @brianmcarey

PR #883, opened 2022-09-29
Deliver Kubevirt 1.25 provider

Barakmor1:
This is a follow-up PR to : https://github.com/kubevirt/kubevirtci/pull/872

- Had to update cluster-provision/k8s/1.25/manifests/cni.diff  to include moving PodDisruptionBudget from apiVersion: policy/v1beta1 to apiVersion: policy/v1 as it is not longer available in k8s1.25 in apiVersion: policy/v1beta1.
- Remove `kubectl {}kubeconfig=/etc/kubernetes/admin.conf taint nodes node01 node-role.kubernetes.io/master:NoSchedule{-}` from the provision process as this taint doesn't appear anymore.
- Bump rook-ceph v1.8.9 -> v1.10.1
rook-ceph fails on kubernetes 1.25 with version under v1.9.10
Please see https://github.com/rook/rook/pull/10937/files
- Update prometheus manifests to use policy/v1 with PodDisruptionBudget (same as the first bullet)
- Update whereabouts manifest to use policy/v1 CronJob as it is not longer available in k8s1.25 in apiVersion: policy/v1beta1 

Now waiting for the following issue in cnao to be resolved as it is blocking the provision process:
https://github.com/kubevirt/cluster-network-addons-operator/issues/1419
Once the issue is resolved we need to bump cano version with the fix to kubevirt-ci.

/cc @dhiller  @brianmcarey


Barakmor1:
@enp0s3

brianmcarey:
cc @brybacki  - just so you are aware it looks like we have to take a bump of ceph in order to get the 1.25 provider delivered
https://github.com/kubevirt/kubevirtci/pull/883/commits/0b46cd9850e888b1827393484ab9bc68997b6ab1

enp0s3:
New CNAO version is in the oven : https://github.com/kubevirt/cluster-network-addons-operator/pull/1424

brybacki:
> 

Good, and It would be great to check if that version can be used for other lanes. But that is probably an independent task. What do you think @awels

awels:
I suspect the bump is because of some resource version changing and not being available in 1.25 anymore (or maybe the security updates?). It would make sense to check if earlier versions can be updated as well.

xpivarc:
/close
https://github.com/kubevirt/kubevirtci/pull/887

PR #896, opened 2022-11-10
Support centos9 provisioner 

Barakmor1:
Adding support for centos stream 9 k8s 1.25 provisioner
The default OS of k8s 1.25 will stay centos8
To change it we will have to modify cluster-provision/k8s/1.25/base to centos9


dhiller (commented):
/approve

Great work, thank you @Barakmor1! I assume that we can then start with creating 1.26 right after this is merged, or do you think there's something missing?

cluster-provision/centos9/scripts/dnsmasq.sh
```diff
@@ -0,0 +1,43 @@
...
+  for s in $(seq 1 ${NUM_SECONDARY_NICS}); do
+    tap_name=stap$(($i - 1))-$(($s - 1))
+    ip tuntap add dev $tap_name mode tap user $(whoami)
+    ip link set $tap_name master br${s}
+    ip link set dev $tap_name up
+  done
+done
+
+# Make sure that all VMs can reach the internet
+iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
```
brianmcarey, line 36: Should we take this opportunity to change to nftables? I think it is default in centos stream 9 and iptables is just using nftables in the backend.

dhiller: How big of a change would this be? Could we not also do this on the 1.26 provider only, since the 1.25 provider is still using centos 8 as default?

brianmcarey: It probably would be better to carry this change out in a separate PR later - just a nice to have - it still works using iptables.

Barakmor1:
> /approve
> 
> Great work, thank you @Barakmor1! I assume that we can then start with creating 1.26 based on Centos 9 Stream right after this is merged, or do you think there's something missing?

Yes i think that your right Daniel :)

cluster-provision/centos9/scripts/dnsmasq.sh
Barakmor1: Looking into it Thanks for the note @brianmcarey

cluster-provision/k8s/1.25/provision.sh
```diff
@@ -102,7 +114,12 @@ baseurl=https://storage.googleapis.com/kubevirtci-crio-mirror/devel_kubic_libcon
 gpgcheck=0
 enabled=1
 EOF
-dnf install -y cri-o containers-common-1-23.module_el8.7.0+1106+45480ee0.x86_64
+if [[ "$release" == "centos8" ]]; then
+    dnf install -y cri-o containers-common-1-23.module_el8.7.0+1106+45480ee0.x86_64
+elif [[ "$release" == "centos9" ]]; then
+    dnf install -y cri-o
```
brianmcarey, line 120: cri-o will probably still fail to pull images if we don't do some sort of patch to `/etc/containers/policy.json` - I guess centos stream 9 has the newer version which cri-o is not able to parse correctly yet?

see https://github.com/cri-o/cri-o/issues/6197 for more details

brianmcarey (commented):
Looks good @Barakmor1  - just one question inline

brianmcarey (commented):
/lgtm

cluster-provision/k8s/1.25/provision.sh
brianmcarey: Ok this doesn't look to be an issue here - the containers-common from the cri-o mirror satisfies the requirements. 
```
[root@79482ac54e73 /]# dnf info containers-common-4:1-19.el8.28.9.noarch
Last metadata expiration check: 0:03:36 ago on Tue Nov 15 10:27:04 2022.
Installed Packages
Name         : containers-common
Epoch        : 4
Version      : 1
Release      : 19.el8.28.9
Architecture : noarch
Size         : 81 k
Source       : containers-common-1-19.el8.28.9.src.rpm
Repository   : @System
From repo    : devel_kubic_libcontainers_stable
Summary      : Common configuration and documentation for containers
License      : ASL 2.0
Description  : This package contains common configuration files and documentation for container
             : tools ecosystem, such as Podman, Buildah and Skopeo.
             :
             : It is required because the most of configuration files and docs come from projects
             : which are vendored into Podman, Buildah, Skopeo, etc. but they are not packaged
             : separately.

[root@79482ac54e73 /]# cat /etc/containers/policy.json
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports":
        {
            "docker-daemon":
                {
                    "": [{"type":"insecureAcceptAnything"}]
                }
        }
}
```

oshoval:
Thanks for the PR

@brianmcarey
@0xFelix 

The 3rd commit might be interesting also for the problems we saw when provisioning with podman but running with docker, since it fixed the SCP (not sure if it will fix all the problems after it, but worth to try)

Moreover there is a problem that the first run with podman after reboot, fails
(not related to this PR), it fails due to the SCP problem, so the 3rd commit might also fix that one

Barakmor1:
> Thanks for the PR
> 
> @brianmcarey @0xFelix
> 
> The 3rd commit might be interesting also for the problems we saw when provisioning with podman but running with docker, since it fixed the SCP (not sure if it will fix all the problems after it, but worth to try)
> 
> Moreover there is a problem that the first run with podman after reboot, fails (not related to this PR), it fails due to the SCP problem, so the 3rd commit might also fix that one

Hey Or,
I totally agree i don't see why we need this file:
https://github.com/kubevirt/kubevirtci/blob/1fbb690811caad6d6ee3b1bec22bca85f27a9ce3/cluster-provision/gocli/cmd/scp.go
this scp option in gocli is error prune and supporting this code is unnecessary in my opinion.

Maybe we should revive your PR: https://github.com/kubevirt/kubevirtci/pull/881

oshoval:
> > Thanks for the PR
> > @brianmcarey @0xFelix
> > The 3rd commit might be interesting also for the problems we saw when provisioning with podman but running with docker, since it fixed the SCP (not sure if it will fix all the problems after it, but worth to try)
> > Moreover there is a problem that the first run with podman after reboot, fails (not related to this PR), it fails due to the SCP problem, so the 3rd commit might also fix that one
> 
> Hey Or, I totally agree i don't see why we need this file: https://github.com/kubevirt/kubevirtci/blob/1fbb690811caad6d6ee3b1bec22bca85f27a9ce3/cluster-provision/gocli/cmd/scp.go this scp option in gocli is error prune and supporting this code is unnecessary in my opinion.
> 
> Maybe we should revive your PR: #881

This is not what i meant tbh ;]
I meant that maybe your PR fix scp, does it ?

Barakmor1:
> > > Thanks for the PR
> > > @brianmcarey @0xFelix
> > > The 3rd commit might be interesting also for the problems we saw when provisioning with podman but running with docker, since it fixed the SCP (not sure if it will fix all the problems after it, but worth to try)
> > > Moreover there is a problem that the first run with podman after reboot, fails (not related to this PR), it fails due to the SCP problem, so the 3rd commit might also fix that one
> > 
> > 
> > Hey Or, I totally agree i don't see why we need this file: https://github.com/kubevirt/kubevirtci/blob/1fbb690811caad6d6ee3b1bec22bca85f27a9ce3/cluster-provision/gocli/cmd/scp.go this scp option in gocli is error prune and supporting this code is unnecessary in my opinion.
> > Maybe we should revive your PR: #881
> 
> This is not what i meant tbh ;] I meant that maybe your PR fix scp, does it ?

Right,
I never tested it but it make sense.

oshoval:
> Right, I never tested it but it make sense.

else, cluster-up would fail, IIUC
(as scp used for copying stuff during cluster-up)

PR #901, opened 2022-11-17
Add options to run multi-numa nodes

PiotrProkop:
It would be beneficial to be able to run e2e tests on nodes with multiple NUMA nodes. 

I was wondering how to expose topology info of the nodes so it would be easy to retrieve in tests.
For now I have introduced new label `kubevirtci.io/numa_nodes` which will indicate how many numa_nodes is available on given hosts. We can use this label in e2e tests to retrieve the NUMA count like this:
```go
func GetNUMANodesCount(virtCli kubecli.KubevirtClient, nodeName string) int {
       node, err := virtCli.CoreV1().Nodes().Get(context.Background(), nodeName, k8smetav1.GetOptions{})
       Expect(err).ShouldNot(HaveOccurred())

       numaNodesLabel := "kubevirtci.io/numa_nodes"

       for label, val := range node.GetLabels() {
               if label == numaNodesLabel {
                       numaNodes, err := strconv.Atoi(val)
                       if err != nil { // if label value can't be converted to integer assume only 1 NUMA node
                               return 1
                       }

                       return numaNodes
               }
       }
       return 1
}
```

### DIFF
--- a/cluster-up/cluster/ephemeral-provider-common.sh
+++ b/cluster-up/cluster/ephemeral-provider-common.sh
@@ -5,6 +5,7 @@ set -e
 KUBEVIRT_WITH_ETC_IN_MEMORY=${KUBEVIRT_WITH_ETC_IN_MEMORY:-false}
 KUBEVIRT_WITH_ETC_CAPACITY=${KUBEVIRT_WITH_ETC_CAPACITY:-none}
 KUBEVIRT_DNS_HOST_PORT=${KUBEVIRT_DNS_HOST_PORT:-31111}
+KUBEVIRT_NUMA_NODES=${KUBEVIRT_NUMA_NODES:-1}
 
 export KUBEVIRTCI_PODMAN_SOCKET=${KUBEVIRTCI_PODMAN_SOCKET:-"/run/podman/podman.sock"}
 
@@ -90,6 +91,8 @@ function _add_common_params() {
     local params="--nodes ${KUBEVIRT_NUM_NODES} --memory ${KUBEVIRT_MEMORY_SIZE} --cpu 6 --secondary-nics ${KUBEVIRT_NUM_SECONDARY_NICS} --random-ports --background --prefix $provider_prefix ${KUBEVIRT_PROVIDER} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
 
     params=" --dns-port $KUBEVIRT_DNS_HOST_PORT $params"
+    # add NUMA node param
+    params=" --numa-nodes $KUBEVIRT_NUMA_NODES $params"
 
     if [[ $TARGET =~ windows_sysprep.* ]] && [ -n "$WINDOWS_SYSPREP_NFS_DIR" ]; then
         params=" --nfs-data $WINDOWS_SYSPREP_NFS_DIR $params"
--- a/cluster-provision/centos8/scripts/vm.sh
+++ b/cluster-provision/centos8/scripts/vm.sh
@@ -22,6 +22,8 @@ while true; do
     -s | --block-device-size ) BLOCK_DEV_SIZE="$2"; shift 2 ;;
     -n | --nvme-device-size ) NVME_DISK_SIZES+="$2 "; shift 2 ;;
     -t | --scsi-device-size ) SCSI_DISK_SIZES+="$2 "; shift 2 ;;
+    --numa ) NUMA_NODES+="$2 "; shift 2 ;;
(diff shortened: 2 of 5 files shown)


PiotrProkop:
Failed tests are mostly `DeadlineExceeded desc = context deadline exceeded"` 
/retest

PiotrProkop:
my change shouldn't affect `kind` clusters, so I guess it's just a flake
/retest

PiotrProkop:
this `vgpu` test doesn't like me :(
/retest

cluster-up/cluster/ephemeral-provider-common.sh
```diff
@@ -151,6 +151,15 @@ function _add_common_params() {
         params=" --docker-proxy=$KUBEVIRTCI_PROXY $params"
     fi
 
+    # MULTI_NUMA for now means 2 NUMA nodes, it assumes the memory can be divided by 2 and cpu count is 6
```
xpivarc, line 154: For consistency and for simplicity(at least from my view) we should move this to vm.sh/run.go. Here should be just a simple --numa flag that enables the configuration.

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -144,6 +144,11 @@ function up() {
     fi
     $kubectl label node -l $label node-role.kubernetes.io/worker=''
 
+    if [ "$KUBEVIRT_MULTI_NUMA" == "true" ]; then
```
xpivarc, line 147: Why do we need this?

xpivarc (commented):
Thanks for the PR.

cluster-up/cluster/k8s-provider-common.sh
PiotrProkop: This is not necessary to run multi NUMA node but it makes it easy to filter nodes which has more than one NUMA node, for example we can implement such function for test purposes:
```go
func GetMultiNUMANodesCount(virtCli kubecli.KubevirtClient, nodeName string) int {
       node, err := virtCli.CoreV1().Nodes().Get(context.Background(), nodeName, k8smetav1.GetOptions{})
       Expect(err).ShouldNot(HaveOccurred())

       numaNodesLabel := "kubevirtci.io/numa_nodes"

       for label, val := range node.GetLabels() {
               if label == numaNodesLabel {
                       numaNodes, err := strconv.Atoi(val)
                       if err != nil { // if label value can't be converted to integer assume only 1 NUMA node
                               return 1
                       }

                       return numaNodes
               }
       }
       return 1
}
```

cluster-up/cluster/ephemeral-provider-common.sh
PiotrProkop: so we would just have one env var `KUBEVIRT_NUMA_NODES` which will default to `1`? And the logic to create proper `qemu` args can be moved to `go` code. Sounds good!

xpivarc: Yes. This seems to me more intuitive and less complex.

cluster-up/cluster/k8s-provider-common.sh
xpivarc: The side effect is that the tests would be very specific to kubevirtci. Or to put it another way, others would need to modify their clusters. I do not object against it as I can see an example when one might have only a subset of nodes with `numa nodes>1` in the cluster and would need to signal it to testsuite.

PiotrProkop: another solution we can use is to install `node-feature-discovery` and rely on `NodeResourceTopology` which exposes [NUMA topology](https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/blob/master/pkg/apis/topology/v1alpha1/types.go#L39) via custom resources.

xpivarc: Actually, we might be exposing this already with the node labeller. Let me have a look. But I like what you suggest as that is more of a standard than making our own label.

cluster-up/cluster/ephemeral-provider-common.sh
PiotrProkop: I have moved the logic to `run.go`

cluster-up/cluster/k8s-provider-common.sh
PiotrProkop: If this go through I'll prepare another PR to add `nfd` to one of the provisioners

PR #907, opened 2022-11-22
New kubevirt k8s 1.26 rc provider

Barakmor1:
Creating new k8s 1.26 rc provider that runs with Centos Stream 9
whenever k8s 1.26 will have stable release the only change that 
we need to update the provider is to update the version in:
`cluster-provision/k8s/1.26/version`

Please note:
One of the commits include using previous version of kubeadm,kubectl,kubelet
when running with rc/alpha/beta version because when Kubernetes version is 
rc/alpha/beta version the same release isn't available for these packages.

According to:
https://kubernetes.io/releases/version-skew-policy/#kubelet
and 
https://kubernetes.io/releases/version-skew-policy/#kubectl
kubelet,kubectl version should work fine when they are only one version older than the api version

**Problem:**
but according to:
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#kubeadm-s-skew-against-the-kubernetes-version
kubeadm can be used with Kubernetes components that are the **same version** as kubeadm or **one version older**
and the package manager doesn't recognize rc versions of kubeadm(also of kubectl and kubelet but this isn't a problem as explained above).
**Solution:**
Therefore i had to switch the binary of kubeadm to the rc version and the reason i still download the closest stable version of kubeadm with dnf first and only than change the binary is for the dependencies packages(package manager manage this for us).

I know that this solution is a bit hacky but it's the best i found for that problem.
If you have a better idea of adjusting kubeadm version to the api version please let me know


dhiller:
@Barakmor1 I am going to prepare check-provision-k8s-1.26 for this one.

dhiller:
Strange, the check-provision for 1.26 says the version 1.25. Is that expected @Barakmor1 ?

https://prow.ci.kubevirt.io/log?container=test&id=1595439316920700928&job=check-provision-k8s-1.26

Barakmor1:
> Strange, the check-provision for 1.26 says the version 1.25. Is that expected @Barakmor1 ?
> 
> https://prow.ci.kubevirt.io/log?container=test&id=1595439316920700928&job=check-provision-k8s-1.26

The link is broken

dhiller:
@Barakmor1 it succeeded right away, great work!

Let's just see why it says that it's still 1.25, while it actually should be 1.26-rc.0...

Barakmor1:
As discussed privately with @dhiller 
According to 
https://kubernetes.io/docs/concepts/architecture/nodes/#info
`kubectl get nodes`  tell us either the kubelet version or the kube-proxy version.

So i added commit to add `kubectl version --short ` command to get info about the
control plane version.

dhiller (approved):
/approve

dhiller:
/override check-up-kind-1.23-sriov
since the SRIOV tests are flaky
since the Mediated Devices test is flaky

PR #908, opened 2022-11-22
Add FIPS-enabled provider

jean-edouard:
This is an exact copy of the 1.25 provider, with just a line added to provision.sh to enable FIPS mode


dhiller (commented):
Could you split that comment into the copy and the addition of the one line, so that we can see where it is?

Thank you!

cluster-provision/k8s/1.25-fips/provision.sh
```diff
@@ -0,0 +1,190 @@
...
+    kubelet-${version} \
+    kubectl-${version} \
+    kubernetes-cni
+
+kubeadm config images pull --kubernetes-version ${version}
+
+dnf install -y centos-release-nfv-openvswitch
+dnf install -y openvswitch2.16
+
+fips-mode-setup --enable
```
dhiller, line 170: Can't we just enable this inside the default provider with an ENV var?

jean-edouard: I wish, but this command requires a reboot, so I guess not?

dhiller: I don't remember that we reboot in the process of building the nodes before freezing them, so did you add that somewhere. Or is my memory failing me?

dhiller (commented):
/approve

dhiller (approved):

cluster-provision/k8s/1.25-fips/provision.sh
xpivarc: @dhiller I think we could run this in the run stage and I don't see reboot as a problem.

jean-edouard: @xpivarc do you mean we don't need to provision a new provider, and can just re-use an existing one, like the cgroupsv2 lane does for example?
I'm not sure what you mean by "run stage", but if there's a way to avoid creating a new image we should definitely use it!

PR #914, opened 2022-11-30
k8s-1.26: Update cni ipv6 diff file

oshoval:
Add missing section to the cni of ipv6.
This way we can create a single stack provider when needed.

### DIFF
--- a/cluster-provision/k8s/1.26/manifests/cni_ipv6.diff
+++ b/cluster-provision/k8s/1.26/manifests/cni_ipv6.diff
@@ -108,3 +108,12 @@
            env:
              # Choose which controllers to run.
              - name: ENABLED_CONTROLLERS
+@@ -3847,7 +3863,7 @@
+ 
+ # This manifest creates a Pod Disruption Budget for Controller to allow K8s Cluster Autoscaler to evict
+ 
+-apiVersion: policy/v1beta1
++apiVersion: policy/v1
+ kind: PodDisruptionBudget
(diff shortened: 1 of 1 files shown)


Barakmor1:
/lgtm 
Thank you

brianmcarey (approved):
/approve

oshoval:
Thanks

brianmcarey:

This change is related to the 1.26 provider.

oshoval:
> 

Note that it won't affect it, because it will affect only when a new folder named k8s-1.26-ipv6 is created
no need to create it yet, because there is already 1.25-ipv6 that we need to use (currently using 1.24)
and only one should exists at a time (once 1.25-ipv6 works we can delete 1.24-ipv6)

Thanks

Barakmor1:
> @Barakmor1 Note please that this failed (container create error) Not related to this PR imo, since this PR touched a file that is used only if the folder ends with ipv6 You can see it uses `+ patch /provision/cni.yaml /tmp/cni.diff` the non ipv6 diff (which is`/tmp/cni_ipv6.diff`) https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/914/check-provision-k8s-1.26/1600162586337218560

Yep something is wrong with centos9 version please have a look in:
https://github.com/kubevirt/kubevirtci/pull/924
@dhiller @brianmcarey

PR #924, opened 2022-12-08
Update Centos Stream 9 version to a stable version

Barakmor1:
Currently we have problem deploying the cni pods with the current centos 9 version:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/914/check-provision-k8s-1.26/1600162586337218560
getting this error in the pod's status:
`container create failed: time="2022-12-08T11:41:22Z" level=error msg="runc create failed: unable to start container process: error during container init: write /proc/self/attr/keycreate: invalid argument"`
with this newer version the problem doesn't exist.
Signed-off-by: bmordeha <bmodeha@redhat.com>


oshoval:
Nice, do you have a specific issue to link about it ?

jean-edouard:
Is this the latest stable version? Where do we find those numbers online? Thank you!

Barakmor1:
> Is this the latest stable version? Where do we find those numbers online? Thank you! Edit: looks like it matches `CentOS-Stream-Container-Base-9` versions in https://cloud.centos.org/centos/9-stream/x86_64/images/ In that case, why don't we use `20221206.0`?

Because it didn't work :\ 
this is the latest one that did work locally

jean-edouard:
> > Is this the latest stable version? Where do we find those numbers online? Thank you! Edit: looks like it matches `CentOS-Stream-Container-Base-9` versions in https://cloud.centos.org/centos/9-stream/x86_64/images/ In that case, why don't we use `20221206.0`?
> 
> Because it didn't work :\ this is the latest one that did work locally

Can you please elaborate? If there a problem with 20221206.0 we should file an issue, because it's likely to persist in future versions!

Barakmor1:
> > > Is this the latest stable version? Where do we find those numbers online? Thank you! Edit: looks like it matches `CentOS-Stream-Container-Base-9` versions in https://cloud.centos.org/centos/9-stream/x86_64/images/ In that case, why don't we use `20221206.0`?
> > 
> > 
> > Because it didn't work :\ this is the latest one that did work locally
> 
> Can you please elaborate? If there a problem with 20221206.0 we should file an issue, because it's likely to persist in future versions!

Yes ,
i got 
`Job for crio.service failed because the control process exited with error code. ` error
 after running 
`systemctl enable --now crio`

also tried the resolution in :
https://access.redhat.com/solutions/5009091
And it didn't work for me

jean-edouard:
> > > > Is this the latest stable version? Where do we find those numbers online? Thank you! Edit: looks like it matches `CentOS-Stream-Container-Base-9` versions in https://cloud.centos.org/centos/9-stream/x86_64/images/ In that case, why don't we use `20221206.0`?
> > > 
> > > 
> > > Because it didn't work :\ this is the latest one that did work locally
> > 
> > 
> > Can you please elaborate? If there a problem with 20221206.0 we should file an issue, because it's likely to persist in future versions!
> 
> Yes , i got `Job for crio.service failed because the control process exited with error code. ` error after running `systemctl enable --now crio`
> 
> also tried the resolution in : https://access.redhat.com/solutions/5009091 And it didn't work for me

Ok, thank you, please file a Bugzilla issue for it.
/approve

Barakmor1:
/hold 
Let's wait for the lane to finish

oshoval:
Please include in the BZ also the current version and the problem, because the fact it works might be just temporary until a new containerd or some RPM is released

oshoval:
It blocks publish
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/publish-kubevirtci/1600820546306576384

/cc @brianmcarey FYI

brianmcarey (commented):
/lgtm

brianmcarey:

This lane looks to be broken and not related to this change.
https://github.com/kubevirt/kubevirtci/issues/925

brianmcarey:
/unhold

check-provision-k8s-1.26  passed - https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/924/check-provision-k8s-1.26/1600864837548642304

PR #927, opened 2022-12-09
bump 1.26 to release

dhiller:
/cc @Barakmor1 @brianmcarey


dhiller:
Looks like we have a problem here, it says

```
kubeadm not found
```
[source](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/927/check-provision-k8s-1.26/1601103967780081664#1:build-log.txt%3A2990)

dhiller:
/hold since the 1.26 is not yet mandatory and doesn't run always.

See https://github.com/kubevirt/project-infra/pull/2509

dhiller:
Retried, since the repo seems to have the package for 1.26.0: https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64/repodata/filelists.xml

dhiller:
/unhold

Yay, 1.26 succeeded!

Barakmor1:
Nice!
I think that we can delete these function calls and the functions them self now that we no longer use the download manager:
 https://github.com/kubevirt/kubevirtci/blob/3e490743abaa5e6b3d369e46cf693611070de96b/cluster-provision/k8s/1.26/provision.sh#L5
 https://github.com/kubevirt/kubevirtci/blob/3e490743abaa5e6b3d369e46cf693611070de96b/cluster-provision/k8s/1.26/provision.sh#L186
 https://github.com/kubevirt/kubevirtci/blob/3e490743abaa5e6b3d369e46cf693611070de96b/cluster-provision/k8s/1.26/provision.sh#L200

dhiller:
> Nice! I think that we can delete these function calls and the functions them self now that we no longer use the download manager:
> 
> https://github.com/kubevirt/kubevirtci/blob/3e490743abaa5e6b3d369e46cf693611070de96b/cluster-provision/k8s/1.26/provision.sh#L5
> 
> 
> https://github.com/kubevirt/kubevirtci/blob/3e490743abaa5e6b3d369e46cf693611070de96b/cluster-provision/k8s/1.26/provision.sh#L18
> 
> 
> https://github.com/kubevirt/kubevirtci/blob/3e490743abaa5e6b3d369e46cf693611070de96b/cluster-provision/k8s/1.26/provision.sh#L186
> 
> 
> https://github.com/kubevirt/kubevirtci/blob/3e490743abaa5e6b3d369e46cf693611070de96b/cluster-provision/k8s/1.26/provision.sh#L200

@Barakmor1 thanks for your review! Good points!

I removed the functions and their calls.

I also now changed it to use the repodata that is available inside the gcs bucket, this way we can install the rpms via dnf directly and don't have to use `yum localinstall`.

PTAL!

Barakmor1:
> > Nice! I think that we can delete these function calls and the functions them self now that we no longer use the download manager:
> > https://github.com/kubevirt/kubevirtci/blob/3e490743abaa5e6b3d369e46cf693611070de96b/cluster-provision/k8s/1.26/provision.sh#L5
> > 
> > https://github.com/kubevirt/kubevirtci/blob/3e490743abaa5e6b3d369e46cf693611070de96b/cluster-provision/k8s/1.26/provision.sh#L18
> > 
> > https://github.com/kubevirt/kubevirtci/blob/3e490743abaa5e6b3d369e46cf693611070de96b/cluster-provision/k8s/1.26/provision.sh#L186
> > 
> > https://github.com/kubevirt/kubevirtci/blob/3e490743abaa5e6b3d369e46cf693611070de96b/cluster-provision/k8s/1.26/provision.sh#L200
> 
> @Barakmor1 thanks for your review! Good points!
> 
> I removed the functions and their calls.
> 
> I also now changed it to use the repodata that is available inside the gcs bucket, this way we can install the rpms via dnf directly and don't have to use `yum localinstall`.
> 
> PTAL.

Awsome,
But i do wonder if it would work for rc releases

cluster-provision/k8s/1.26/provision.sh
```diff
@@ -3,16 +3,7 @@
...
-  packages_version=$kubernetes_version
-  if [[ $kubernetes_version == *"alpha"* ]] || [[ $kubernetes_version == *"beta"* ]] || [[ $kubernetes_version == *"rc"* ]]; then
-    kubernetes_minor_version=$(echo $kubernetes_version | cut -d. -f2)
-    packages_major_version=$(echo $kubernetes_version | cut -d. -f1)
-    packages_minor_version=$((kubernetes_minor_version-1))
-    packages_patch_version=0
-    packages_version="$packages_major_version.$packages_minor_version.$packages_patch_version"
-  fi
-  echo $packages_version
+  echo "$(curl --fail -L https://storage.googleapis.com/kubernetes-release/release/stable.txt | sed 's/^v//')"
```
Barakmor1, line 6: I think that it might be a problem when the stable version will be higher than 1.26

dhiller: Can you elaborate, please?

I see that the stable is always lower than the latest:
```
$ curl --fail -L https://storage.googleapis.com/kubernetes-release/release/latest.txt | sed 's/^v//'
1.27.0-alpha.0%
$ curl --fail -L https://storage.googleapis.com/kubernetes-release/release/stable.txt | sed 's/^v//'
1.26.0%
```

Barakmor1: In the future when 1.28 will be release the 1.26 prisoner will try to use kubeadm,kubelet and kubectl with 1.28 version
and according to Kubernetes docs this is not safe:
 https://kubernetes.io/releases/version-skew-policy/#kubelet
https://kubernetes.io/releases/version-skew-policy/#kubectl
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#kubeadm-s-skew-against-the-kubernetes-version

dhiller: Meh, you are right! Thanks!

dhiller: @Barakmor1 I've updated `getKubernetesClosestStableVersion` such that it fetches the latest stable (instead of the 0) from the previous release. I think that brings us closer to what we want. Sorry for being so nitty about this :slightly_smiling_face:

Barakmor1:
Thank you!
/lgtm

brianmcarey (approved):
/approve

oshoval:
Would adding `?ignoreCache=1` to the previous link will help as well ?
it force uncached download from what i saw (on other links of gcs)

PR #930, opened 2022-12-12
Upgrade cri-o to version 1.25 for 1.25/6 providers

brianmcarey:

### DIFF
--- a/cluster-provision/k8s/1.25-ipv6/provision.sh
+++ b/cluster-provision/k8s/1.25-ipv6/provision.sh
@@ -97,7 +97,7 @@ export PATH=$ISTIO_BIN_DIR:$PATH
   chmod +x $ISTIO_BIN_DIR/istioctl
 )
 
-export CRIO_VERSION=1.24
+export CRIO_VERSION=1.25
 cat << EOF >/etc/yum.repos.d/devel_kubic_libcontainers_stable.repo
 [devel_kubic_libcontainers_stable]
 name=Stable Releases of Upstream github.com/containers packages (CentOS_8_Stream)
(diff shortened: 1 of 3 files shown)


jean-edouard (commented):
It surprises me that https://cri-o.io/ still tells you to use the packages from the **retired** openSUSE Kubic project!
It looks like we are using a google mirror for it, which is probably a good idea since the official repository will probably disappear at some point.
I probably don't want to know why Centos (and other distros) don't just include cri-o in their base repos...
My only real concern here is, do we get 1.25.0 or 1.25.1 with this change? If it's 1.25.1 then I'll approve. Please confirm. Thank you!

brianmcarey:
> It surprises me that https://cri-o.io/ still tells you to use the packages from the **retired** openSUSE Kubic project! It looks like we are using a google mirror for it, which is probably a good idea since the official repository will probably disappear at some point. I probably don't want to know why Centos (and other distros) don't just include cri-o in their base repos... My only real concern here is, do we get 1.25.0 or 1.25.1 with this change? If it's 1.25.1 then I'll approve. Please confirm. Thank you!

We mirror the latest stable release for `1.25` so yes I can confirm that this installs 1.25.1 - the job to mirror this actually pulls these from the kubic project and runs the following weekly  - https://github.com/kubevirt/project-infra/blob/11b6727636c83f9972090af3d28bdc6f7a9b725b/hack/mirror-crio.sh#L25

brianmcarey:

Changes not related to vgpu lane.

dhiller (commented):
/approve

dhiller (approved):

PR #936, opened 2022-12-18
vm image builder: Enable using podman

ormergi:
Enable using podman for creating customized VMs containers disks using the vm-image-builder.

### DIFF
--- a/cluster-provision/images/vm-image-builder/README.md
+++ b/cluster-provision/images/vm-image-builder/README.md
@@ -5,18 +5,18 @@ would like to create KubeVirt VM's with customized container-disk,
 without the hassle of installing and customizing the OS manually.
 
 Use cases:
-- VM's for Demos
+- VMs for Demos
 - Pre-configured Kubevirt VM image for tests, for example:
     - Image with qemu-guest-agent installed and configured.
-    - Image with sriov drivers for sriov-lane images.
-    - Image with DPDK package for testing DPDK applications on KubeVirt.
+    - Image with SR-IOV drivers.
+    - Image with DPDK dependencies for testing DPDK applications on KubeVirt.
 
 
 ## Prerequisites
 
 The following RPM packages need to be present on your machine:
+- podman/docker-ce
 - cloud-utils
-- docker-ce
 - libguestfs
 - libguestfs-tools-c
 - libvirt
@@ -47,9 +47,9 @@ instructions are distributed between the following three files:
 
 ```
 $ ls -1 example/
-cloud-config #cloud-init configuration for virt-customize
-image-url    #download URL of the base image
-os-variant   #operating system variant (for example fedora32)
+cloud-config # cloud-init configuration for virt-customize
+image-url    # download URL of the base image
+os-variant   # operating system variant (for example fedora32)
 ```
 
 To create a completely new containerdisk, best copy the `example` folder and
@@ -62,7 +62,7 @@ $ ./build-containerdisk.sh example
 $ ./publish-containerdisk.sh example "localhost:$(./cluster-up/cli.sh ports registry | tr -d '\r')"
 ```
 
-### Create Kubevirt VM's with the new image
+### Create Kubevirt VM with the new image
 
 In the VMI / VM yaml file, change `spec.volumes[].containerDisk.image` to the new image path.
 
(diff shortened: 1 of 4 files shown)


cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
```diff
@@ -1,18 +1,20 @@
 #!/bin/sh
 set -ex
 
+CRI=${CRI:-docker}
```
oshoval, line 4: can you please do an auto detect as we have on kubevirt ?
```
function detect_cri() {
    if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi
}

export CRI_BIN=${CRI_BIN:-$(detect_cri)}
```
same for the rest of the files here
and then no need to export it, unless someone wants to forcefully override

better to name it please CRI_BIN (this is the name we are using on kubevirtci (case insensitive) / kubevirt)

oshoval (commented):
Thanks

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
ormergi: Done

oshoval, line 4: here forgotten right ?
also please update the readme,
no need to export and set CRI_BIN (appears twice)

ormergi: Done

cluster-provision/images/vm-image-builder/README.md
```diff
@@ -73,3 +73,12 @@ $ kubectl apply -f <VMI yaml file>
...
 ```
+
+### Using podman
+CRI_BIN environment variable controls which container runtime to use
+```bash
+$ export CRI_BIN=podman
+$ ./build-containerdisk.sh example
+$ 
+$ ./publish-containerdisk.sh example quay.io/example/example-containerdisk:latest
+```
```
oshoval, line 84: we don't need this now right ?
as it is auto detected

you might want instead to add that the CRI_BIN can be overriden by exporting CRI_BIN

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
```diff
@@ -1,18 +1,24 @@
 #!/bin/sh
 set -ex
 
+function detect_cri() {
+    if podman ps >/dev/null 2>&1; then echo "using podman"; elif docker ps >/dev/null 2>&1; then echo "using docker"; fi
```
oshoval, line 5: you need to only echo podman or docker, without using, as this will be the binary name
same for the rest please

it wont work this way

cluster-provision/images/vm-image-builder/README.md
ormergi: Done

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
ormergi: Done

cluster-provision/images/vm-image-builder/README.md
oshoval: we still have it 
https://github.com/kubevirt/kubevirtci/pull/936/files#diff-5911465af9b7f64fa499beab92df12415c644505539d24ba87eaeb0f19f92f2aR76-R84

we dont need the export it, nor podman specific explanation
just a general one but thats we already have

cluster-provision/images/vm-image-builder/README.md
```diff
@@ -5,18 +5,18 @@ would like to create KubeVirt VM's with customized container-disk,
...
 
 Use cases:
-- VM's for Demos
+- VMs for Demos
 - Pre-configured Kubevirt VM image for tests, for example:
     - Image with qemu-guest-agent installed and configured.
-    - Image with sriov drivers for sriov-lane images.
-    - Image with DPDK package for testing DPDK applications on KubeVirt.
+    - Image with SR-IOV drivers.
+    - Image with DPDK dependencies for testing DPDK applications on KubeVirt.
```
oshoval, line 12: Not related to the first commit / PR, if you can split it please to other commit ?
same for line 50 and 65

oshoval:
If you want, continue the effort please here

Thanks

/cc @brianmcarey 
FYI, another place where we didn't parameterize yet

PR #941, opened 2022-12-27
WIP publish job: Optimize provision job time

ormergi:
Currently, there are 5 providers and 2 base images, where for each provider we build two flavors of images, regular (with pre-pulled images) and slim (no pre-puled images). 
That is a total of 12 images per each publish job that executes for each merged PR.

This PR parallelizes the provision jobs that are executed by the publish job. 

Run provision jobs in parallel in order to reduce CI time and have quicker feedback.
The amount of jobs that will run concurrently is controlled by CONCURRENT_PROVISION_JOBS_COUNT env var, default 2.

### DIFF
--- a/publish.sh
+++ b/publish.sh
@@ -2,6 +2,9 @@
 
 set -ex
 
+readonly ARTIFACTS=${ARTIFACTS:-${PWD}}
+readonly CONCURRENT_PROVISION_JOBS_COUNT="${CONCURRENT_PROVISION_JOBS_COUNT:-2}"
+
 KUBEVIRTCI_TAG=$(date +"%y%m%d%H%M")-$(git rev-parse --short HEAD)
 export KUBEVIRTCI_TAG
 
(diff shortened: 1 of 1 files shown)


oshoval:
you might need to increase memory / cpu for the job
and btw it might wont help, because it is too heavy
and can affect other jobs if it takes too many resources

publish.sh
```diff
@@ -2,35 +2,92 @@
...
+          jobs_batch_pids+=($!)
+          echo "[$(date --utc)] ${job_name_prefix} started ${job}, PID: $!.."
+      done
+      offset=$((offset+CONCURRENT_PROVISION_JOBS_COUNT))
+
+      # wait for jobs to finish
+      set +e
+      for job_pid in "${jobs_batch_pids[@]}"; do
+          echo "[$(date --utc)] waiting for ${job_name_prefix} job, PID: ${job_pid}.."
+          wait ${job_pid}
```
oshoval, line 36: you are waiting serially, better to wait to any that occur
so if it fails, you fail fast

ormergi:
We should put this to the test on CI env and examine the tradeoff between CI time and resource consumption of the publish job.

oshoval:
> I posted this PR so we can put it to the test on CI env and examine the tradeoff between CI time and resource consumption.

It is prone to be problematic, so i am not sure 
it can consume all the memory 

what you can do is to run the publish with some push to other repo and run it here and there manually on CI when other tasks alive, but to merge directly, i am not sure is a good idea

ormergi:
> It is prone to be problematic, so i am not sure it can consume all the memory, and hurt already running jobs we might see lots of flakes

AFAIR prow job pods are set with fixed resources and limits amount of resources, I dont expect it to slow down other jobs on a given node, I wouldn't be surprised if the provision job would take longer tough.

When we will have some results, we can consider allocating more resources for the publish job.

Maybe we need to measure disk IOps consumption to avoid the slowliness of dnf commands during provision and kube-api services during tests, WDYT?

> what you can do is to run the publish with some push to other repo and run it here and there manually on CI when other tasks alive, 

I will change the push target and see how it goes.
I wanted to get some feedback first before actually running it on CI, let me know if you think its "ok-to-test".
If it is, dow you know if it's possible to execute the publish job on an opened PR?

> but to merge directly, i am not sure is a good idea

It was never my intention, this PR is marked as a WIP :slightly_smiling_face:

ormergi:
I just checked and it seems that the publish / provision jobs have their memory resource requirements specified but not limits, which means they will end up consuming more than what was requested.
Adding limits will terminate the pod in case of exceeds the memory requires, I am not sure if it's a good practice to take on CI env.

oshoval:
We can't add limits, and it might indeed use too many resources
CI tasks usually are not meant to run stuff parallel IIRC (you can ask Roman about it)
Btw when benchmarking it is worth to calculate also the push, because it takes lots of time and might be even
a bottleneck when using parallel
you can push to other kubevirtci repo if needed, just not to the official one

oshoval:
Please make sure to not push gocli , nor to tag kubevirtci git, otherwise it will replace the current head of KUBEVIRTCI_TAG

PR #946, opened 2023-01-04
SELinux: add missing permission to centos8-based providers

jean-edouard:
Since Centos 8 does not provide a recent-enough version of container-selinux, use the alternative solution from the docs in kubevirt/kubevirt.

See the kubevirt/kubevirt docs as updated by https://github.com/kubevirt/kubevirt/pull/9021:
```
SELinux-enabled nodes need to have [Container-selinux](https://github.com/containers/container-selinux) version 2.193.0 or newer installed.

For nodes running an operating system that does not offer container-selinux v2.193.0 or newer, it is possible to use v2.170.0 or newer and either:
- Not use passt-based networking
- Manually add the missing SELinux rule for passt by running the following commands on each node:
echo '(allow container_t tmpfs_t (filesystem (mount)))' >/tmp/passt.cil
semodule -i /tmp/passt.cil
```

### DIFF
--- a/cluster-provision/k8s/1.23/provision.sh
+++ b/cluster-provision/k8s/1.23/provision.sh
@@ -118,6 +118,10 @@ registries = ['registry:5000']
 registries = []
 EOF
 
+# Add an SELinux rule missing from pre-v2.193.0 container-selinux, see kubevirt/kubevirt docs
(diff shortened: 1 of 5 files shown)


jean-edouard:
/close
Superseded by https://github.com/kubevirt/kubevirt/pull/9048

PR #955, opened 2023-01-31
Group dnf installs rather than having multiple installs

brianmcarey:
There are a number of dnf installs carried out through the `provision.sh` script which can be grouped together.

Grouping them together can make it easier to see all the additional packages that are being installed and it should also slightly speed up the provisioning.

### DIFF
--- a/cluster-provision/k8s/1.25/provision.sh
+++ b/cluster-provision/k8s/1.25/provision.sh
@@ -2,6 +2,18 @@
 
 set -ex
 
+function getKubernetesClosestStableVersion() {
+  kubernetes_version=$version
+  packages_version=$kubernetes_version
+  if [[ $kubernetes_version == *"alpha"* ]] || [[ $kubernetes_version == *"beta"* ]] || [[ $kubernetes_version == *"rc"* ]]; then
+    kubernetes_minor_version=$(echo $kubernetes_version | cut -d. -f2)
+    packages_major_version=$(echo $kubernetes_version | cut -d. -f1)
+    packages_minor_version=$((kubernetes_minor_version-1))
+    packages_version="$(curl --fail -L "https://storage.googleapis.com/kubernetes-release/release/stable-${packages_major_version}.${packages_minor_version}.txt" | sed 's/^v//')"
+  fi
+  echo $packages_version
+}
+
 if [ ! -f "/tmp/extra-pre-pull-images" ]; then
     echo "ERROR: extra-pre-pull-images list missing"
     exit 1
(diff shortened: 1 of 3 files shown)


enp0s3 (commented):
/approve
@brianmcarey Thank you!

cluster-provision/k8s/1.25/provision.sh
```diff
@@ -48,11 +60,43 @@ function pull_container_retry() {
...
+name=Kubernetes Release
+baseurl=https://storage.googleapis.com/kubernetes-release/release/v${packages_version}/rpm/x86_64/
+enabled=1
+gpgcheck=0
+repo_gpgcheck=0
+EOF
+
+dnf install -y centos-release-nfv-openvswitch
+
+dnf install --disableexcludes=kubernetes -y "kernel-modules-$(uname -r)" cloud-utils-growpart patch lvm2 iscsi-initiator-utils iproute-tc cri-o podman libseccomp-devel openvswitch2.16 kubectl-${packages_version} kubeadm-${packages_version} kubelet-${packages_version} kubernetes-cni
```
oshoval, line 97: I would prefer please to keep it separated to groups i.e
k8s
openvswitch 
It allows maintaining and see what happens easier, 
it can still be one next to each other in one place

nit: if desired - some of them might better to install as late as possible, so if stuck breaks before, during development no need to wait for the rest

about the repo files - maybe export the files to a folder and just copy them instead in-place ?
better readability


the general ones can be bundled
better with one on each line using same dnf command but with back slash at the end of the line if possible please

brianmcarey: > 1. I would prefer please to keep it separated to groups i.e
>    `k8s`
>    `openvswitch`
>    It allows maintaining and see what happens easier,
>    it can still be one next to each other in one place

I am not sure how this will make it easier - if there is an issue installing a package you will still have to check node logs to see what has happened. 

> 
> * The general ones can be bundled
> * Maybe there is a way to split to lines, use the same dnf command, but have some separator between groups ?
> 
> 2. Better with one on each line using same dnf command but with back slash at the end of the line if possible please

Yes definitely better for readability - will make the update for this.

> 3. About the repo files - maybe export the files to a folder and just copy them instead in-place ?
>    better readability (can be on other PR of course)

That could be something we could look at in another PR

> 4. 1.24 isn't done by purpose / other code / EOL ?

I think there were slight differences in the 1.24 provider so I decided to leave it as is and let the provider age out. I can take a look again at applying these changes there. 

> 
> Thanks

Thanks Or

oshoval: > > 1. I would prefer please to keep it separated to groups i.e
> >    `k8s`
> >    `openvswitch`
> >    It allows maintaining and see what happens easier,
> >    it can still be one next to each other in one place
> 
> I am not sure how this will make it easier - if there is an issue installing a package you will still have to check node logs to see what has happened.

but it is easier for the eye to catch the section and what belong to that failure, they are separate snippets
and it make it more friendly and welcoming imo in general to read the list

PR #957, opened 2023-02-06
vm based providers: Introduce Provision manager

oshoval:
Provision manager determines which providers to rebuild and which to retag using
previous released `KUBEVIRTCI_TAG`.

**Motivation:**
1. Freeze versions unless changed, less prone for regressions,
improves reproducibility and consistency.
3. Faster provisioning.
4. Decouple collision domains, if one provider is problematic, we can still build other ones,
without blocking them.

**Flow:**

It checks which files changed versus the latest `KUBEVIRTCI_TAG`.
For each changed file, it find the matching rule and accumulate a list of targets to rebuild.

https://github.com/kubevirt/project-infra/pull/2600
Will be used as presubmit that checks dry running, and also allows to see what will happen during
publish.

**Notes:**

* We have the k8s patch version bumping bot that will keep us updated when needed.
* If desired we can even skip presubmits using this mechanism in case the files of the PR doesn't affect
a specific provider.
* We can improve the rule list by creating some more specific rules for some of the files instead
for their whole directory but lets do it on follow PR.

**Additional changes:**

* Refactor `publish.sh` by splitting to functions and parameterize CRI binary.


oshoval:
Added last commit: "publish: Parametrize CRI binary"

brianmcarey (commented):
Thanks Or, looks good! 

This should add stability for providers that aren't changed. 

Are we in danger of testing against older centos stream 8 packages if an older provider such as k8s-1.24 doesn't receive many updates? 

Should we be aiming to test closer to the latest from centos? 

Would this be better to do in go as you are doing a lot of array handling here which would probably be easier in go?

oshoval:
> Thanks Or, looks good!
> 
> This should add stability for providers that aren't changed.
> 
> Are we in danger of testing against older centos stream 8 packages if an older provider such as k8s-1.24 doesn't receive many updates?

If it doesn't receieve updates (even not auto patch bump) it won't be rebuilt.
but we can just use the "force" method that appears on the readme to build a new provider if needed lets say once in a x month(s) imo - maybe even a cron job ?
There is a flag to run publish with override for cron jobs if needed
 
> 
> Should we be aiming to test closer to the latest from centos?

What do you mean ?
The tool just gives control, imo we can learn on the fly the best practice with fine tuning

> Would this be better to do in go as you are doing a lot of array handling and regex matching here which would probably be easier in go?

I prefer to keep it in bash for now, later we can refactor if needed,
it already works and pretty solid
There aren't that much process imo, just 180 lines of bash total where most of them aren't complex

brianmcarey:
> > Are we in danger of testing against older centos stream 8 packages if an older provider such as k8s-1.24 doesn't receive many updates?
> 
> If it doesn't receive updates (even not auto patch bump) there won't be a rebuild. But we can just use the "force" method that appears on the readme to build a new provider if needed lets say once in a x month(s) imo - maybe even a cron job ? There is a flag to run publish with override for cron jobs if needed
> 
> > Should we be aiming to test closer to the latest from centos?
> 
> What do you mean ? The tool just gives control, imo we can learn on the fly the best practices and fine tune.
> 
> > Would this be better to do in go as you are doing a lot of array handling and regex matching here which would probably be easier in go?
> 
> I prefer to keep it in bash for now, later we can refactor if needed, It already works and pretty solid. There aren't that much processing imo, just 180 lines of bash total where most of them aren't complex. For the regex i use CLI globbing, so it is native, regex is just a term in the lang that was introduced by the PR. Btw imo python is even better than go here.

Ok happy enough with the intial bash version - looks good to me.

brianmcarey (commented):
/lgtm

oshoval:
Thanks Brian

/cc @enp0s3 @dhiller 

Can you please take a look ?

publish.sh
```diff
@@ -1,62 +1,156 @@
...
-docker tag  alpine-cloud-init:devel ${TARGET_KUBEVIRT_REPO}/alpine-with-test-tooling-container-disk:devel
+function build_gocli() {
+  (cd cluster-provision/gocli && make container)
+  ${CRI_BIN} tag ${TARGET_REPO}/gocli ${TARGET_REPO}/gocli:${KUBEVIRTCI_TAG}
+}
 
-# Push all images
+function build_centos8_base_image() {
+  for target in ${IMAGES_TO_BUILD[@]}; do
+    if [[ ! "$target" =~ "centos9" ]]; then
```
dhiller, line 55: Note: we are intending to remove the `-centos9` suffix in the 1.27 provider version, so this and the one below will stop working then. Can we simplify and just build both base images?

publish.sh
```diff
@@ -1,62 +1,156 @@
...
+    if [[ "$target" =~ "centos9" ]]; then
+      (cd cluster-provision/centos9 && ./build.sh)
+      return
+    fi
+  done
+}
+
+function build_base_images() {
+  build_centos8_base_image
+  build_centos9_base_image
```
dhiller, line 73:
```suggestion
  (cd cluster-provision/centos8 && ./build.sh)
  (cd cluster-provision/centos9 && ./build.sh)
```

publish.sh
dhiller: (see comment below)

publish.sh
```diff
@@ -1,62 +1,156 @@
...
+
+function create_git_tag() {
+  git config user.name "kubevirt-bot"
+  git config user.email "rmohr+kubebot@redhat.com"
+  git tag ${KUBEVIRTCI_TAG}
+  git push ${TARGET_GIT_REMOTE} ${KUBEVIRTCI_TAG}
+  echo "INFO: pushed new tag $KUBEVIRTCI_TAG"
+}
+
+function main() {
```
dhiller, line 146: Seeing the `main` function would make me think that this had better be written in a real language.

Yes, I know, I have done the same, but if we are going to increase the complexity of the publish process quite a bit, we should also use the right tools. This way we at least can add testing for more complex stuff, which is not as easy to do in `bash`.

hack/pman/pman.sh
```diff
@@ -0,0 +1,186 @@
+#!/bin/bash -e
```
dhiller, line 1: This one definitely is too complex to be written in `bash`. It needs proper testing.

publish.sh
oshoval: Lets start with bash please if possible, and later on change lang (python) and add tests upon needs ?
referring both here and https://github.com/kubevirt/kubevirtci/pull/957#discussion_r1130792832

hack/pman/pman.sh
oshoval: see please https://github.com/kubevirt/kubevirtci/pull/957#discussion_r1130801292

oshoval:
Thanks

Changes:
Rebased
Added k3d rule for the new provider
Addressed the comments(about base image

publish.sh
oshoval: See please https://github.com/kubevirt/kubevirtci/pull/957#discussion_r1130808869

hack/pman/pman.sh
dhiller: My experience tells me that the move to a proper language will never happen.

dhiller: BTW, why add Python when go is already available>

oshoval: I already have an action item to move to python so it will happen
I prefer to do this tool in its own repo as python, it is more suitable, and has very nice pytest
Here we will just consume it (as it will be a container image)
this way more projects might able to use it

publish.sh
dhiller: > In functions aspect, I just refactored it, we are using `main` everywhere but again a follow PR can update it to other lang
> 
> btw publish is a pretty straight forward script, so straight forward that python / go is overkill for publish itself it almost just run plain bash commands again and again, but this is just my opinion
> 
> Moving to python can be nice, but it require lots of work and steps, that imo suitable for follow ups. It is a global feature, not specifically for scripts of this PR.
> 
> In order to test it, i added DRY_RUN and replaced the CRI commands with echo, to see all is as planned
> 
> EDIT Btw `publish.sh` can even consume it such that it will calculate both lists (of rebuild and retag) this way it will reduce a bit the code needed for publish.sh.
> 
> Extracted the refactor to a standalone [PR](https://github.com/kubevirt/kubevirtci/pull/980).

First of all: this is _not_ refactoring.

_Refactoring is a disciplined technique for restructuring an existing body of code, altering its internal structure without changing its external behavior._
source: https://refactoring.com/

Without tests you can't be sure that you did not change it's external behaviour.

Second: we discussed the topic and we agreed that we don't want to introduce more bash code into the codebase. Also we don't want to introduce another language into the code base, since Go is already available.

hack/pman/pman.sh
dhiller: Another option would be that we integrate it into the existing gocli which is present already. There would be no overhead. Also I can't review Python code, so I would be against introducing it here if the need is not there.

publish.sh
oshoval: About the refactor:

This is the refactoring I mean https://github.com/kubevirt/kubevirtci/pull/980
not this PR, just the 3 commits, until today we didn't have coverage on kubevirtci.
Lets start this effort bit by bit please, not in this specific PR,
I have other tasks that I am committed to right now, I will have to drop this task for now
if I see that it involves a big change that wasn't planned.

Python is more suitable here as it is a script, please let me at least introduce a minimal POC,
so we can discuss it.
kubevirtci itself has lots of scripts, go isn't suitable for it all, it will be totally overkill to run go for a script like this one https://raw.githubusercontent.com/kubevirt/kubevirtci/741e74861ce87570b61abc1cec40e16ad05d7bb2/publish.sh (the splitting to function of publish.sh, you can see how easily readable it is, and how the commands are basic there).

Having Python in a container image will allow to use Python were suitable.
It is better also please to include other kubevirtci developers and myself in such decision, I wasn't part of it,
and I think that I contribute to kubevirtci.

hack/pman/pman.sh
oshoval: We have people that know to review it, and we can all learn it
I am learning it on the fly, and it is an amazing language.

I think we should have a broader forum to decide please, as I wasn't part of the decision even.
Can we do a meeting about it ? wdyt ?

oshoval:
Sumup
Converting pman to go (checking if it can be part of gocli)
for this specific effort publish will stay bash as we agreed

PR #960, opened 2023-02-13
Enable Seccomp configuration

xpivarc:


brianmcarey (approved):
/lgtm

/approve

This allows the conformance tests to run successfully against the 1.26 provider with PSA enabled.

cluster-provision/k8s/check-cluster-up.sh
```diff
@@ -94,6 +94,12 @@ export KUBEVIRTCI_GOCLI_CONTAINER=quay.io/kubevirtci/gocli:latest
 
             ${ksh} wait -n kubevirt kv kubevirt --for condition=Available --timeout 15m
 
+            if [ "${KUBEVIRT_PSA:-"false"}" == "true" ]; then
```
oshoval, line 97: No need the default value please because we have it here already
https://github.com/kubevirt/kubevirtci/blob/31e5860db1cabffa0597a3b338ac441ee1fdc346/cluster-up/hack/common.sh#L25

Easier to read without it imo

xpivarc: `./check-cluster-up.sh: line 97: KUBEVIRT_PSA: unbound variable`

Not sure what you are suggesting...

oshoval: ah ok it doesnt import this file, as it is check-up https://github.com/kubevirt/kubevirtci/blob/31e5860db1cabffa0597a3b338ac441ee1fdc346/cluster-up/hack/common.sh

so it is better please IMO to set KUBEVIRT_PSA to either false or true in the section of check-cluster-up where we set all the rest, but not in the condition itself please imo

PR #963, opened 2023-02-14
Remove check for localhost node name in up.sh

brianmcarey:
There is a case where external kubernetes cluster may have been installed without a hostname defined which leads to a node name like `localhost.localdomain` - this is the case with the AMD SEV cluster.

I have not hit this case using any of the other kubevirtci providers so I think this check is no longer needed. 

/cc @dhiller @xpivarc @enp0s3 


dhiller:
@qinqon do you remember in what case that check has been needed? I don't.

dhiller (commented):
/approve

I think this check is no longer needed, but let's wait for others to chime in here a bit.

brianmcarey:
No longer hit this issue after changing the node name on the AMD cluster. Not sure if this is still needed but the change is no longer required so closing for now.

PR #967, opened 2023-02-20
Add WFFC variation of rook-ceph-block

akalenyu:
We might want to test WFFC fancy storage sometime instead of just immediate binding ones


akalenyu:
/cc @ShellyKa13 @awels 
This way we could test wffc fancy storage and possibly stop using "local" as well

ShellyKa13 (commented):
/lgtm
sounds good!

PR #968, opened 2023-02-20
centos8: set SELinux boolean that is now required for chardev access by containers

jean-edouard:


xpivarc:
@brianmcarey FYI, relates to https://github.com/kubevirt/kubevirt/pull/9208

oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/968/check-provision-k8s-1.24/1627702552218832896

Not related to this PR
`failed to delete CustomResourceDefinition(e2e-test-crd-publish-openapi-5167-crd): Internal error occurred: etcdserver: request timed out` - we might want to consider using ramfs for etcd as we are doing with kind

oshoval:
We don't have a presubmit job for 1.26 anymore it seems, but publish does run it, 
so in case there are problems, publish might fail

oshoval:
> @brianmcarey FYI, relates to [kubevirt/kubevirt#9208](https://github.com/kubevirt/kubevirt/pull/9208)

Worth to add please on PR desc a failure example

brianmcarey (approved):
/lgtm

/approve

EdDev:
@jean-edouard , it seems we tweak the nodes with this configuration for the kubevirtci.

Who is suppose to tweak it for production nodes?

EdDev:
> The fact that each distro has its own SELinux policy, and that policies are randomly subject to change, makes it really hard on us! (Centos 8 allowed chardev access to container processes until recently...)

I think that formally we only support CentOS Stream for nodes, it is the only distribution we test.

The problem is that we bump it from time to time, catching bugs like this one.
Here is another example: https://bugzilla.redhat.com/2172090

However, patching our CI may just mask the problem.

oshoval:
> > The fact that each distro has its own SELinux policy, and that policies are randomly subject to change, makes it really hard on us! (Centos 8 allowed chardev access to container processes until recently...)
> 
> I think that formally we only support CentOS Stream for nodes, it is the only distribution we test.
> 
> The problem is that we bump it from time to time, catching bugs like this one. Here is another example: https://bugzilla.redhat.com/2172090

I am planning to do two things:
1. https://github.com/kubevirt/project-infra/pull/2600
More controlled patching
2. Create a bit easier way to test a non released provider with all kubevirt, so hopefully it will alllow us
to test things before they appear on the bump

> 
> However, patching our CI may just mask the problem.

oshoval:
I wonder if by any change, once it is fixed on kubevirt if it can shed some light about WA for 
https://github.com/kubernetes-sigs/kind/issues/2999
(anyhow, we are moving to k3d for now it seems soon)

PR #969, opened 2023-02-21
Remove the centos8 based k8s-1.26 provider

brianmcarey:
The centos9 1.26 provider is the main target for kubevirt testing now[1]

THe centos8 based provider is no longer required and can be removed which should speed up publishing of kubevirtci

[1] https://github.com/kubevirt/project-infra/pull/2610


oshoval (approved):
Thanks

Do we want to rename 1.26-centos9 to 1.26 ?

Do we want maybe to keep 1.26 centos 8 in case some bugs occur and we want to quickly check if its centos specific ? (well for this we can always try to use old provider if still supported on the other hand)

dhiller (approved):
/approve

PR #970, opened 2023-02-21
robust stack: Fix presubmit sanity check

oshoval:
The previous logic selected only the latest provider, 
excluding the variants that include dash (i.e -centos9).

Atm we run 1.26-centos9 instead 1.26, so the sanity check won't run.
Fix it by running it for all supported providers (1.25+).
This is a fast sanity check (only cluster-up and check IPs) and safer this way.

Also fix a bug that the condition compared `KUBEVIRT_PROVIDER` which 
is in the form `k8s-x.yz` with plain version i.e `x.yz`.

Note:
Maybe in the future we can consider running it even before the dual stack, so if it fails
it gives quick report, as it doesn't include conformance, while dual stack does.

### DIFF
--- a/cluster-provision/k8s/check-cluster-up.sh
+++ b/cluster-provision/k8s/check-cluster-up.sh
@@ -18,8 +18,6 @@ function cleanup() {
     make cluster-down
 }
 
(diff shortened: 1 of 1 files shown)


oshoval:
/cc @brianmcarey 

Waiting to see jobs done and will check the logic runs as expected

oshoval:
validated, all 3 providers work as expected (1.24 skip, rest are running single stack sanity check cluster-up)

brianmcarey (approved):
/lgtm 

/approve

PR #971, opened 2023-02-27
Fix some typos

zhouhao3:


cluster-up/README.md
```diff
@@ -1,6 +1,6 @@
 # How to use cluster-up
 
-This directory provides a wrapper around gocli. It can be vendored into other
+This directory provides a wrapper around gocli. It can be vended into other
```
brianmcarey, line 3: I think `vendored` is ok here - the rest look good to me.

zhouhao3: updated, thanks.

brianmcarey (approved):
/lgtm

/approve

oshoval:
Thanks for the PR

@dhiller  @brianmcarey 
Note please that with provision manager we won't rebuild provider when md files are changed
moreover since less providers would be built, less jobs are prone to fail as they now fail and the whole 3h might have to be redone

PR #975, opened 2023-03-02
centos8: update SELinux policy version

jean-edouard:
The latest Centos 8 release ships with the `selinux-policy-3.14.3-108` package, which removes (mistakenly I assume) permissions that are critical to us.
The permissions are back in later versions of the Centos 8 `selinux-policy` package. However, there hasn't been a new "release" of Centos 8 Stream that includes the fixes.
2 of the permissions in question are:
```
allow container_t container_file_t:chr_file { ioctl };
allow iptables_t cgroup_t dir { ioctl };
```
Updating `selinux-policy` at provision time should bring both permissions back.

Updating to the new version also allows us to revert the commit that sets the `container_use_devices` SELinux boolean.

### DIFF
--- a/cluster-provision/k8s/1.24/provision.sh
+++ b/cluster-provision/k8s/1.24/provision.sh
@@ -140,9 +140,6 @@ repo_gpgcheck=0
 gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
 EOF
 
-# Set the SELinux boolean that allows containers to use char devices like /dev/null
-semanage boolean --modify --on container_use_devices
-
 # Install Kubernetes packages.
 dnf install --skip-broken --nobest --nogpgcheck --disableexcludes=kubernetes -y \
     kubeadm-${version} \
--- a/cluster-provision/k8s/1.24/provision.sh
+++ b/cluster-provision/k8s/1.24/provision.sh
@@ -76,6 +76,10 @@ echo 'ACTION=="add|change", SUBSYSTEM=="block", KERNEL=="vd[a-z]", ATTR{queue/ro
 # To prevent preflight issue related to tc not found
 dnf install -y tc
 
+# The selinux-policy package shipped with the latest Centos stream "release" can be outdated.
+# Example: Centos 8 20220913.0 ships with selinux-policy-3.14.3-108, which misses crucial permissions
+dnf -y update selinux-policy
+
 # Install istioctl
 export PATH=$ISTIO_BIN_DIR:$PATH
 (
--- a/cluster-provision/k8s/1.25/provision.sh
+++ b/cluster-provision/k8s/1.25/provision.sh
@@ -80,6 +80,11 @@ echo 'ACTION=="add|change", SUBSYSTEM=="block", KERNEL=="vd[a-z]", ATTR{queue/ro
 
 # To prevent preflight issue related to tc not found
 dnf install -y iproute-tc
+
+# The selinux-policy package shipped with the latest Centos stream "release" can be outdated.
+# Example: Centos 8 20220913.0 ships with selinux-policy-3.14.3-108, which misses crucial permissions
+dnf -y update selinux-policy
+
 # Install istioctl
 export PATH=$ISTIO_BIN_DIR:$PATH
 (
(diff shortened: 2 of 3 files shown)


cluster-provision/k8s/1.24/provision.sh
```diff
@@ -76,6 +76,10 @@ echo 'ACTION=="add|change", SUBSYSTEM=="block", KERNEL=="vd[a-z]", ATTR{queue/ro
 # To prevent preflight issue related to tc not found
 dnf install -y tc
 
+# The selinux-policy package shipped with the latest Centos stream "release" can be outdated.
+# Example: Centos 8 20220913.0 ships with selinux-policy-3.14.3-108, which misses crucial permissions
+dnf -y update selinux-policy
```
oshoval, line 81: can we add a check that if the permissions exists fail before we update ?
so once it is fixed we can drop this, but not a must if we will remember / open issue about it that will remind us

jean-edouard: > can we add a check that if the permissions exists fail before we update ? so once it is fixed we can drop this, but not a must if we will remember / open issue about it that will remind us

That requires installing `setools`, which comes with 129 dependent packages...
TBH I feel like it kinda makes sense to run with the latest SELinux policy always, but I'm fine opening an issue if that's not desirable.

> 
> Beside that lgtm Thanks was it tested with kubevirt jobs manually ?

I did make sure that `dnf -y update selinux-policy` updates to a version that brings back the permissions we care about, but I did not run `make functest` on it.

oshoval: Imho having an issue is better please, bundle is preferable and easier to maintain

about `functest` would you please do ssh.sh to the VMs on the current provider, to install it as a patch and run for example this lane `pull-kubevirt-e2e-k8s-1.24-sig-network` ?
unless you know for sure this is the problem and we that we can go blindly

oshoval: Issue is fine thanks

About `DisableCustomSELinuxPolicy` do we need to change kubevirt then ?

oshoval: /cc @AlonaKaplan @phoracek

AlonaKaplan: We can try to move all the passt tests to an unpriveleged namespace (and disable the custom selinux policy). I believe it should work.

EdDev:
> # The selinux-policy package shipped with the latest Centos stream "release" can be outdated.
> # Example: Centos 8 20220913.0 ships with selinux-policy-3.14.3-108, which misses crucial permissions

I am missing something here. If we fetch a new refreshed image now, won't it be include in it?

cluster-provision/k8s/1.24/provision.sh
jean-edouard: The passt tests should actually run fine with the latest Centos 8 under container_t. In fact, Centos 8 now seems to behave mostly like Centos 9, where loading a custom policy fails...

jean-edouard:
> > The selinux-policy package shipped with the latest Centos stream "release" can be outdated.
> > Example: Centos 8 20220913.0 ships with selinux-policy-3.14.3-108, which misses crucial permissions
> 
> I am missing something here. If we fetch a new refreshed image now, won't it be include in it?

Correct, the packages get updated regularly but the images get refreshed rarely.

EdDev:
> > I am missing something here. If we fetch a new refreshed image now, won't it be include in it?
> 
> Correct, the packages get updated regularly but the images get refreshed rarely.

Ok, so even if this PR gets in, I guess we better refresh the image in order to avoid this patching later.
We have so many automatic bump scripts, I thought we have one for the images as well.

oshoval:
@brianmcarey 
Hi, can you please take a look ?  (comments inline)
It will allow us hopefully to unblock kubevirtci bumping

Thanks

dhiller (approved):
/approve

cluster-provision/k8s/1.24/provision.sh
oshoval: we dont need to do anything with kubevirt about `DisableCustomSELinuxPolicy` ?
Passt should pass out of the box ?

jean-edouard: We most likely do need to set that feature gate on all lanes.
Waiting to see what happens in https://github.com/kubevirt/kubevirt/pull/9208, with https://github.com/kubevirt/kubevirt/pull/9373 standing by.

PR #977, opened 2023-03-07
Enable deploying Multus V3 from manifests

ormergi:
Kubeivrt sig-network lanes test Kubevirt integration with Multus CNI, including VMs with secondary networks and connectivity.
We use [CNAO](https://github.com/kubevirt/cluster-network-addons-operator) to deploy Multus, which currently deploys V4.0.0-alpha (think plugin form) which is not a GA version.

Testing Kubevirt with GA Multus (V3.X, thin plugin form) is necessary for sig-network lanes integrity and having these lanes similar to D/S.

This PR set the VM based providers' cluster-up flow to conditionally deploy Multus V3 from manifests instead of deploying Multus V4 through CNAO.
*The '`MultusDynamicNetwork`' controller won't be deployed either because it depends on Multus V4.

This behavior is controlled by the new '`KUBEVIRT_WITH_MULTUS_V3`' environment variable and will take effect when it is set to 'true'.

This PR also add Multus sanity check for check-provision lanes in order to detect Multus-related breaking changes earlier before they land at kubevirt/kubevirt.

This is an intermediate state until Multus V4 is GA'ed.
Once we have CNAO that deploys Multus V4 GA we can revert these changes and have CNAO deploy Multus exclusively.

### DIFF
--- a/cluster-provision/k8s/check-cluster-up.sh
+++ b/cluster-provision/k8s/check-cluster-up.sh
@@ -30,6 +30,7 @@ export KUBEVIRTCI_GOCLI_CONTAINER=quay.io/kubevirtci/gocli:latest
     export KUBEVIRT_MEMORY_SIZE=5520M
     export KUBEVIRT_NUM_SECONDARY_NICS=2
     export KUBEVIRT_WITH_CNAO=true
+    export KUBEVIRT_WITH_MULTUS_V3=true
     export KUBEVIRT_DEPLOY_ISTIO=true
     export KUBEVIRT_DEPLOY_PROMETHEUS=true
     export KUBEVIRT_DEPLOY_PROMETHEUS_ALERTMANAGER=true
@@ -51,6 +52,14 @@ export KUBEVIRTCI_GOCLI_CONTAINER=quay.io/kubevirtci/gocli:latest
     ${ssh} node01 -- ip l show eth2
     ${ssh} node02 -- ip l show eth1
     ${ssh} node02 -- ip l show eth2
+
+    # Verify Multus v3 image is used
+    ${ksh} get ds -n kube-system kube-multus-ds -o yaml | grep multus-cni:v3
+
+    # Sanity check that Multus able to connect secondary networks
+    ${ksh} create -f "$DIR/test-multi-net.yaml"
+    ${ksh} wait pod test-multi-net --for condition=ready=true
+    ${ksh} delete -f "$DIR/test-multi-net.yaml"
 
     pre_pull_image_file="$DIR/${provision_dir}/extra-pre-pull-images"
     if [ -f "${pre_pull_image_file}" ]; then
--- a/cluster-up/cluster/k8s-provider-common.sh
+++ b/cluster-up/cluster/k8s-provider-common.sh
@@ -57,12 +57,23 @@ function deploy_cnao() {
         $kubectl create -f /opt/cnao/operator.yaml
 
         if [ "$KUBVIRT_WITH_CNAO_SKIP_CONFIG" != "true" ]; then
-            $kubectl create -f /opt/cnao/network-addons-config-example.cr.yaml
+            create_network_addons_config
         fi
 
         # Install whereabouts on CNAO lanes
         $kubectl create -f /opt/whereabouts
     fi
+}
+
+function create_network_addons_config() {
+    local nac="/opt/cnao/network-addons-config-example.cr.yaml"
+    if [ "$KUBEVIRT_WITH_MULTUS_V3" == "true" ]; then
+        local no_multus_nac="/opt/cnao/no-multus-nac.yaml"
+        $ssh node01 -- "awk '!/multus/' ${nac} | sudo tee ${no_multus_nac}"
+        nac=${no_multus_nac}
+    fi
+
+     $kubectl apply -f ${nac}
 }
 
 function wait_for_cnao_ready() {
@@ -71,6 +82,18 @@ function wait_for_cnao_ready() {
         if [ "$KUBVIRT_WITH_CNAO_SKIP_CONFIG" != "true" ]; then
             $kubectl wait networkaddonsconfig cluster --for condition=Available --timeout=200s
         fi
+    fi
+}
+
+function deploy_multus() {
+    if [ "$KUBEVIRT_WITH_MULTUS_V3" == "true" ]; then
+        $kubectl create -f /opt/multus/multus.yaml
+    fi
+}
+
+function wait_for_multus_ready() {
+    if [ "$KUBEVIRT_WITH_MULTUS_V3" == "true" ]; then
+        $kubectl rollout status -n kube-system ds/kube-multus-ds --timeout=200s
     fi
 }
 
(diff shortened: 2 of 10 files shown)


cluster-provision/k8s/check-cluster-up.sh
```diff
@@ -52,6 +53,16 @@ export KUBEVIRTCI_GOCLI_CONTAINER=quay.io/kubevirtci/gocli:latest
     ${ssh} node02 -- ip l show eth1
     ${ssh} node02 -- ip l show eth2
 
+    # Verify Multus v3 image is used
+    readonly expected_multus_v3_image="ghcr.io/k8snetworkplumbingwg/multus-cni:v3.8"
+    actual_multus_image=$(${ksh} get ds -n kube-system kube-multus-ds -o jsonpath='{.spec.template.spec.containers[0].image}')
+    [ "${actual_multus_image}" == "${expected_multus_v3_image}" ]
```
qinqon, line 59: Let's just simplify this and skip checking the "minor" part of the semver
```suggestion
    ${ksh} get ds -n kube-system kube-multus-ds -o yaml |grep multus-cni:v3
```

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -57,14 +57,28 @@ function deploy_cnao() {
...
 
         if [ "$KUBVIRT_WITH_CNAO_SKIP_CONFIG" != "true" ]; then
-            $kubectl create -f /opt/cnao/network-addons-config-example.cr.yaml
+            if [ "$KUBEVIRT_WITH_MULTUS_V3" == "true" ]; then
+                # set CNAO to not deploy Multus components to prevent conflicts with deploying Multus from manifests.
+                # FIXME: Remove this workaround when CNAO provides Multus v4 GA version.
+                create_no_multus_cluster_network_addons_cr
+            else
+                $kubectl create -f /opt/cnao/network-addons-config-example.cr.yaml
+            fi
```
qinqon, line 66: Let's move the condition inside a function so this is more clear
```suggestion
            compose_network_addons_config | $kubectl apply -f -
```

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -57,14 +57,28 @@ function deploy_cnao() {
...
 }
 
+function create_no_multus_cluster_network_addons_cr() {
+    local -r default_nac="/opt/cnao/network-addons-config-example.cr.yaml"
+    local -r no_multus_nac="/opt/cnao/network-addons-config-no-multus.yaml"
+    $ssh node01 -- "awk '!/multus/' ${default_nac} | sudo tee ${no_multus_nac}"
+
+    $kubectl create -f ${no_multus_nac}
+}
+
```
qinqon, line 81:
```suggestion
function compose_network_addons_config() {
    local nac="/opt/cnao/network-addons-config-example.cr.yaml"
    if [ "$KUBEVIRT_WITH_MULTUS_V3" == "true" ]; then 
        awk '!/multus/' ${nac} 
    else 
        cat $nac
    fi
}

```

qinqon (changes requested):
We should check just "v3"

/hold
Create test PR at kubevirt/kubevirt to check if sig network lanes are fine with this.

cluster-provision/k8s/check-cluster-up.sh
ormergi: done

cluster-up/cluster/k8s-provider-common.sh
ormergi: done

ormergi:
[Changes](https://github.com/kubevirt/kubevirtci/compare/27880dded027b05504a681af0f936444a43435cc..3e382458ff5bfd2654a31ec05d4e8f49e7aa8cf5): resolve [comments](https://github.com/kubevirt/kubevirtci/pull/977#pullrequestreview-1330207232) 

Testing this PR changes on kubevirt/kubevirt 
https://github.com/kubevirt/kubevirt/pull/9388

ormergi:
The cluster is created sucessfully with Multus v3, the hotplug NICs tests are failing because the feature is currently depends on Multus v4.
We will introduce a flag to the test suite to run these conditionally.
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/9388/pull-kubevirt-e2e-k8s-1.26-sig-network/1633504075087613952#1:build-log.txt%3A832

ormergi:
sig-network lane is green when skipping hotplug NICs tests.
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/9388/pull-kubevirt-e2e-k8s-1.26-sig-network/1633781069075451904

We will introduce periodic and optional presubmit jobs for testing the hotplug NICs with Multus v4.

ormergi:
Remove the hold as we verified this PR changes on kubevirt/kubevirt#9388
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/9388/pull-kubevirt-e2e-k8s-1.26-sig-network/1633781069075451904

/unhold

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -57,14 +57,25 @@ function deploy_cnao() {
...
         fi
 
         # Install whereabouts on CNAO lanes
         $kubectl create -f /opt/whereabouts
     fi
 }
 
+function create_network_addons_config() {
+    local nac="/opt/cnao/network-addons-config-example.cr.yaml"
+    if [ "$KUBEVIRT_WITH_MULTUS_V3" == "true" ]; then
```
oshoval, line 70: I think that if both `KUBVIRT_WITH_CNAO_SKIP_CONFIG` and `KUBEVIRT_WITH_MULTUS_V3` are defined
it will still deploy the CR even that `KUBVIRT_WITH_CNAO_SKIP_CONFIG` says do not, right ?

ormergi: No, because `create_network_addons_config()` is called when `KUBVIRT_WITH_CNAO_SKIP_CONFIG` is **not** true.

oshoval: Right
so if one wants to use `KUBVIRT_WITH_CNAO_SKIP_CONFIG` but to deploy V3
he would need to do it manually or there is a way still ?

cluster-provision/k8s/check-cluster-up.sh
```diff
@@ -30,6 +30,7 @@ export KUBEVIRTCI_GOCLI_CONTAINER=quay.io/kubevirtci/gocli:latest
     export KUBEVIRT_MEMORY_SIZE=5520M
     export KUBEVIRT_NUM_SECONDARY_NICS=2
     export KUBEVIRT_WITH_CNAO=true
+    export KUBEVIRT_WITH_MULTUS_V3=true
```
oshoval, line 33: kubevirt will use V4 right ?
if so it means that on kubevirtci we are deploying and testing only V3 which isn't 
the interesting case of kubevirt
this way a problem with V4 won't be detected before bump at kubevirt

ormergi: Posting this for visibility:

Yes, currently v4 is used, the next step we are going to take is to change kubevirt to use Multus v3 by default (by changing the relevant lanes to specify `KUBEVIRT_WITH_MULTUS_V3=true`).
This is why I also changed this test to use Multus v3.

PR #983, opened 2023-03-23
kind: amend env variable CONFIG_WORKER_CPU_MANAGER

zhlhahaha:
This variable is defined in cluster-up/cluster/kind/common.sh
which is CONFIG_WORKER_CPU_MANAGER rather than
CONFIG_CPU_MANAGER

### DIFF
--- a/cluster-up/cluster/kind-1.23/provider.sh
+++ b/cluster-up/cluster/kind-1.23/provider.sh
@@ -35,7 +35,7 @@ function up() {
     cp $KIND_MANIFESTS_DIR/kind.yaml ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
     _add_kubeadm_cpu_manager_config_patch
     _add_extra_mounts
-    export CONFIG_CPU_MANAGER=true
+    export CONFIG_WORKER_CPU_MANAGER=true
     kind_up
 
     configure_registry_proxy


zhlhahaha:
cc: @oshoval

brianmcarey (commented):
Can you add the reason why this change is needed to the commit message?

oshoval:
Thanks @zhlhahaha 
Sorry that missed it, we don't maintain this provider

Note that we might soon depracted support of kind in favor for k3d,
would k3d be good enough for you in such case?

You can see the new k3d-1.25-sriov provider, where you can `export DEPLOY_SRIOV=false` in order to 
deploy only k3d.

PR desc / commit typo `varialbe`

zhlhahaha:
> Can you add the reason why this change is needed to the commit message?

Of course, this variable name supposed to be same as the following one, which make cpumanager related configuration would be set in kind.yaml for worker node.
https://github.com/kubevirt/kubevirtci/blob/2957a2fa8f5920e06fe57dbc5a5fed0ce7678cb6/cluster-up/cluster/kind/common.sh#L10

zhlhahaha:
> Thanks @zhlhahaha Sorry that missed it, we don't maintain this provider so we don't have automation for it.

I see, now we use this provider for e2e tests on Arm64 server.

> Also note that this provider can't stay 1.23, as we support only the last 3 k8s versions.

Yes, you are right. I see the [issue](https://github.com/kubernetes-sigs/kind/issues/2999). for the cpumanager issue. I am curious why the issue occus.

> Note that we might soon deprecate support of kind in favor for k3d, would k3d be good enough for you in such case?

I am not familiar with k3d. I may need to verify if it run well on Arm64 server and if it is suitable for CICD pipeline on Arm64.

> You can see the new `k3d-1.25-sriov` provider, where you can `export DEPLOY_SRIOV=false` in order to deploy only k3d. We strive to support less providers.

Yes, it make sense.

oshoval:
> > Can you add the reason why this change is needed to the commit message?
> 
> Of course, this variable name supposed to be same as the following one, which make cpumanager related configuration would be set in kind.yaml for worker node.
> 
> https://github.com/kubevirt/kubevirtci/blob/2957a2fa8f5920e06fe57dbc5a5fed0ce7678cb6/cluster-up/cluster/kind/common.sh#L10

Please add the explanation to the commit message  / PR desc and we can merge this PR for now
later on please let us know if k3d is suitable for you so we can drop this provider.
Is it fine ?

oshoval:
Thanks, just a typo "varialbe" on PR desc and commit

oshoval (approved):
Thanks

brianmcarey (approved):
/approve

Issue #984, opened 2023-03-23
K3D on Arm64 server Verification


zhlhahaha:
- [x] k3d on bare metal server
- [x] kubeVirt on k3d cluster
- [x] k3d in bootstrap image
- [x] kubeVirt and k3d in bootstrap image

oshoval:
You know there is k3d-1.25-sriov right ?
that you can use as baseline with `export DEPLOY_SRIOV=false` just making sure so you won't need to duplicate work

zhlhahaha:
> You know there is k3d-1.25-sriov right ? that you can use as baseline with `export DEPLOY_SRIOV=false` just making sure so you won't need to duplicate work

Yes, I give it a try yesterday, and it failed to start on Arm64 server because of the cni. I am not sure why it not works. I will take a look tomorrow.

oshoval:
Thanks

Btw why don't you use vm based providers instead ?
What about adapting them to support Arm64 ? might be better?

oshoval:
> Arm64

https://lwn.net/Articles/921783/ wdyt ?

oshoval:
> Hi @oshoval , I have verified the k3d, on Arm64 server and in a nested container environment (bootstrap image) Good news is k3d can successfully start on Arm64 server and nested container environment, and 81 E2E tests pass and 1 failed. I did following modification:
> 
> 1. use the default CNI rather than calico (any specific reason to use calico CNI?)

Thank you

We must install manually the CNI, let it either be Calico or Flannel, because without it, multus doesn't work.
See please https://github.com/kubevirt/kubevirtci/pull/972#issuecomment-1455822191
and the comments below on cons of it.

> 2. make it run as a one node cluster (only have k3d-k3d-server-0 node)

We need at least 2 nodes (atm we use 3) because we test migration.

Due to 1 and 2
Lets have a different folder for k3d-1.25 ? 
It seems SR-IOV need different config.
I prefer to keep the providers untangled.

> 
> Here are some issues:
> 
> 1. vmi-killer seems not works in the cluster
> 2. tests/reporter seems not works well
> 3. E2E tests failed occationally in mutli-node k3d cluster
> 
> I am still checking these issue. And I also need to verify if stability of the k3d provider.

Thank you

zhlhahaha:
> We must install manually the CNI, let it either be Calico or Flannel, because without it, multus doesn't work.

[...]

PR #985, opened 2023-03-23
Control-plane nodes which are marked as schedulable are also given worker role.

prnaraya:
Currently, fix_node_labels is marking control-plane nodes as schedulable and removing the NoSchedule taint. Since these nodes are now behaving as worker nodes, they should also be given the worker node role.

### DIFF
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -163,13 +163,15 @@ function _get_pods() {
 function _fix_node_labels() {
     # Due to inconsistent labels and taints state in multi-nodes clusters,
     # it is nessecery to remove taint NoSchedule and set role labels manualy:
-    #   Control-plane nodes might lack 'scheduable=true' label and have NoScheduable taint.
+    #   Control-plane nodes might lack 'schedulable=true' label and have NoScheduable taint.
+    #	Control-plane nodes which have 'schedulable=true' should have worker role label
     #   Worker nodes might lack worker role label.
     master_nodes=$(_get_nodes | grep -i $MASTER_NODES_PATTERN | awk '{print $1}')
     for node in ${master_nodes[@]}; do
         # removing NoSchedule taint if is there
         if _kubectl taint nodes $node node-role.kubernetes.io/master:NoSchedule-; then
             _kubectl label node $node kubevirt.io/schedulable=true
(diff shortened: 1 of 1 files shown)


jean-edouard (approved):

cluster-up/cluster/kind/common.sh
```diff
@@ -163,13 +163,15 @@ function _get_pods() {
...
-    #   Control-plane nodes might lack 'scheduable=true' label and have NoScheduable taint.
+    #   Control-plane nodes might lack 'schedulable=true' label and have NoScheduable taint.
+    #	Control-plane nodes which have 'schedulable=true' should have worker role label
     #   Worker nodes might lack worker role label.
     master_nodes=$(_get_nodes | grep -i $MASTER_NODES_PATTERN | awk '{print $1}')
     for node in ${master_nodes[@]}; do
         # removing NoSchedule taint if is there
         if _kubectl taint nodes $node node-role.kubernetes.io/master:NoSchedule-; then
             _kubectl label node $node kubevirt.io/schedulable=true
+            _kubectl label node $node node-role.kubernetes.io/worker=""
```
oshoval, line 174: This can break SR-IOV
see please https://github.com/kubevirt/kubevirt/pull/9505#issuecomment-1488299304

oshoval:
Note that fails, even before reaching the scenario I think it will 
https://github.com/kubevirt/kubevirt/pull/9505#issuecomment-1488323858

We don't have automation for kind-1.23 / kind-1.23-sriov, and should drop them because they are too old
We can even consider removing them from kubevirtci as they are not needed for it anymore
(SR-IOV one is just temporary there in case k3d has suprises)
See please https://github.com/kubevirt/kubevirtci/issues/984 about k3d replacement suggestion.

oshoval:
Please see https://github.com/kubevirt/kubevirtci/pull/990

oshoval:
Hi, I think that the easiest solution, 
assuming you are using `kind-1.23` (until we deprecate it in favor of k3d),
is to include this labeling as part of `kind-1.23` provider itself (not in the common file), would it help you ?
Imo we shouldn't affect the kind SR-IOV provider atm.

oshoval:
> Hi, please see #1043

Note that it won't include the fix, as the SR-IOV lane is dedicated to SR-IOV, see please discussion in the mentioned PR.

PR #988, opened 2023-03-28
Adding Kubernetes 1.27 (rc0) provider

jean-edouard:
Copied from 1.26-centos9. No difference apart from k8s version and coredns version.
Using rc0 for now since that's the latest 1.27 tag available.


oshoval:
Thanks
Can you please split to 2 commits:
1. Duplicate the ref folder as is
2. Do the changes 

Easier to see just the changes this way (saw that they are mentioned on PR desc of course)

jean-edouard:
> Thanks Can you please split to 2 commits:
> 
> 1. Duplicate the ref folder as is
> 2. Do the changes
> 
> Easier to see just the changes this way (saw that they are mentioned on PR desc of course)

[`4ae9839` (#988)](https://github.com/kubevirt/kubevirtci/pull/988/commits/4ae983916c6fafc7532c5744d95b6d020093ef3c)

> 
> Do we want to name it centos9 as well ? (or no need because it doesn't have centos8 one)

@dhiller recommended no -centos9 suffix

> Is there a presubmit lane for it that we can rehearse ? So we know the post submit is safe.

https://github.com/kubevirt/project-infra/pull/2685

cluster-provision/k8s/1.27/k8s_provision.sh
```diff
@@ -161,8 +161,8 @@ envsubst < $kubeadm_raw_ipv6 > $kubeadm_manifest_ipv6
 until ip address show dev eth0 | grep global | grep inet6; do sleep 1; done
 
 # workaround for 1.25 trying to pull coredns: https://github.com/kubernetes/kubernetes/issues/112131#issuecomment-1232279438
```
oshoval, line 163: nit: maybe worth to change to 1.25+
but whatever you think

oshoval (approved):
Thanks

oshoval:
Btw if we had a mechanism to
```
export K8S_VERSION=1.27
make cluster-up
```
We would save a provider (the provider can prepull all the needed stuff, which as we can see 99% is common between 
1.26 and 1.27)

wdyt @dhiller @brianmcarey  ?

dhiller:
check-provision-k8s-1.27

dhiller:
> Btw if we had a mechanism to
> 
> ```
> export K8S_VERSION=1.27
> make cluster-up
> ```
> 
> We would save a provider (the provider can prepull all the needed stuff, which as we can see 99% is common between 1.26 and 1.27)
> 
> wdyt @dhiller @brianmcarey ?

Quite a while ago we separated each of the providers for good reason, since every provider needed a bit of change here and there, but everyone in different places. I don't think it's worth the effort to go back to the old route again.

oshoval:
> > We would save a provider (the provider can prepull all the needed stuff, which as we can see 99% is common between 1.26 and 1.27)
> > wdyt @dhiller @brianmcarey ?
> 
> Quite a while ago we separated each of the providers for good reason, since every provider needed a bit of change here and there, but everyone in different places. I don't think it's worth the effort to go back to the old route again.

We indeed separated, but even that there was one combined code, there were still 3 different providers,
here I suggest to have only one provider which is robust.
About changes here and there, we can maybe think about some abstraction layer, so as far as we can,
the changes will be done concentrated, and what ever can be done for all would be done for all  ?
For example waiting to ipv6 was needed only for single stack, but in order to have less difference,
I did it also for dual stack when we had two providers, and it worth it, because now we have one robust stack provider.
Same for the rest.

cluster-provision/k8s/1.27/k8s_provision.sh
```diff
@@ -161,8 +161,8 @@ envsubst < $kubeadm_raw_ipv6 > $kubeadm_manifest_ipv6
...
-podman pull k8s.gcr.io/coredns/coredns:v1.9.3
-podman image tag k8s.gcr.io/coredns/coredns:v1.9.3 k8s.gcr.io/coredns:v1.9.3
+podman pull k8s.gcr.io/coredns/coredns:v1.10.1
+podman image tag k8s.gcr.io/coredns/coredns:v1.10.1 k8s.gcr.io/coredns:v1.10.1
```
oshoval, line 165: Can we derive the tag needed automatically, so we don't need to maintain this code maybe?

Don't we need to update extra-pre-pull-images ?
In the past we had to IIRC, maybe the mechanism was changed

dhiller (commented):
/approve

Good to go from my POV, other refinements can be done in a follow up.

/hold to let the discussion settle, @jean-edouard remove if you think it's ready

dhiller:
> > > We would save a provider (the provider can prepull all the needed stuff, which as we can see 99% is common between 1.26 and 1.27)
> > > wdyt @dhiller @brianmcarey ?
> > 
> > 
> > Quite a while ago we separated each of the providers for good reason, since every provider needed a bit of change here and there, but everyone in different places. I don't think it's worth the effort to go back to the old route again.
> 
> We indeed separated, but even when there was one combined code, there were still 3 different providers, here I suggest to have only one provider which is robust. About changes here and there, we can maybe think about some abstraction layer, so as far as we can, the changes will be done concentrated, and what ever can be done for all would be done for all ? For example waiting to ipv6 was needed only for single stack, but in order to have less difference, I did it also for dual stack when we had two providers, and it worth it, because now we have one robust stack provider. Same for the rest.

I am not completely against removing duplication, but for this use case I think what we have is good enough.

Our highest priority is delivering new providers as fast as possible (in fact we are already two weeks behind, when looking at the [planning doc](https://github.com/kubevirt/sig-release/pull/1/files) ). It already turned out that with separate files per provider we could act much faster since all the complexity of `if this provider then configure that` is entirely removed.

This is key to reducing the complexity per provider, which makes delivering faster, and to me this seems much more valuable than eliminating duplication, which I think is bearable.

dhiller:
@jean-edouard I think this is good to go, so I am removing the hold

/unhold

dhiller:
@brianmcarey after this one is merged we should make the check-provision-k8s-1.27 required

PR #994, opened 2023-04-03
k3d: refactor _create_cluster and add k3d-1.25 provider

zhlhahaha:
To enable the k3d provider to adapt for SR-IOV e2e testing and Arm64 e2e testing, a flexible configuration for creating a k3d cluster is necessary. The following changes have been made:
    1. "_cluster_nodes_args" allows the number of servers and agents to be set.
    2. "_cni_args" sets the CNI for the k3d cluster.
    3. "_device_args" sets the mounted device from the host.
    4. When "DISABLE_DEFAULT_SERVICES" is set to true, all k3d default services (Traefik, ServiceLB, Metrics-Server) will be disabled.

Add k3d-1.25 provider which is based on k3d-1.25-sriov, but with the removal of SR-IOV related settings to simplify it as much as possible.

The discussion is in https://github.com/kubevirt/kubevirtci/issues/984

### DIFF
--- a/cluster-up/cluster/k3d/common.sh
+++ b/cluster-up/cluster/k3d/common.sh
@@ -29,6 +29,9 @@ function detect_cri() {
 export CRI_BIN=${CRI_BIN:-$(detect_cri)}
 KUBEVIRT_NUM_SERVERS=${KUBEVIRT_NUM_SERVERS:-1}
 KUBEVIRT_NUM_AGENTS=${KUBEVIRT_NUM_AGENTS:-2}
+DISABLE_DEFAULT_SERVICES=${DISABLE_DEFAULT_SERVICES:-false}
+K3D_CNI=${K3D_CNI:-calico}
+VFIO_ENABLED=${VFIO_ENABLED:-true}
 
 export KUBEVIRTCI_PATH
 export KUBEVIRTCI_CONFIG_PATH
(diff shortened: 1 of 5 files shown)


zhlhahaha:
@qinqon @oshoval

oshoval:
Thanks

I think we should use manifests approach once we solve the problem about it
Can you meanwhile please just create the new folder as you did, and use a dedicated file instead of common.sh ?
The current approach isn't generic enough and not straight forward.
See for example `_generate_k3d_args` which toggle between disabling the default services (needed for SR-IOV),
and between the changes needed for the non SR-IOV one.

Also was it tested on SRIOV ? it seems it didn't change the provider accordingly.

oshoval (changes requested):
Marking as request changes
Thanks

zhlhahaha:
> I think we should use manifests approach once we solve the problem about it Can you meanwhile please just create the new folder as you did, and use a dedicated file instead of common.sh ? The current approach isn't generic enough and not straight forward. See for example `_generate_k3d_args` which toggle between disabling the default services (needed for SR-IOV), and between the changes needed for the non SR-IOV one.

Yes, I agree with you. Manifests is much easier to maintain. Once the manifests issue solved, we can migrate to it. At least, base on it we can know where need to be modified when using it for Arm64 E2E test.

> Also was it tested on SRIOV ? it seems it didn't change the provider accordingly.

It should be works for SRIOV, as applying all default setting, it will generate same k3d args as it used to be.

oshoval:
> Yes, I agree with you. Manifests is much easier to maintain. Once the manifests issue solved, we can migrate to it. At least, base on it we can know where need to be modified when using it for Arm64 E2E test.

What about using default network "bridge" for podman for ARM ?
If you manage to do so, we can use manifests approach soon, because on CI we don't support yet podman,
and this is the bug that blocks us from using manifests atm.

zhlhahaha:
> Marking as request changes Thanks

Ok

zhlhahaha:
> What about using default network "bridge" for podman for ARM ? 

I think "bridge" network should be fine for ARM. 

> If you manage to do so, we can use manifests approach soon

I give a quick try for manifests on Arm64, and get some errors, but I think your suggestion is right.

> Because on CI  we don't support yet podman, and this is the bug that blocks us from using manifests atm.

In the `bootstrap` image, we have already used podman. And Arm64 E2E tests are supposed to run in the bootstrap image.

oshoval:
> > What about using default network "bridge" for podman for ARM ?
> 
> I think "bridge" network should be fine for ARM.
> 
> > If you manage to do so, we can use manifests approach soon
> 
> I give a quick try for manifests on Arm64, and get some errors, but I think your suggestion is right.
> 
> > Because on CI  we don't support yet podman, and this is the bug that blocks us from using manifests atm.
> 
> In the `bootstrap` image, we have already used podman. And Arm64 E2E tests are supposed to run in the bootstrap image.

Ye I mean on SR-IOV bootstrap we don't,
this is why I am saying that if your bootstrap use the default network bridge,
We can use the manifests (once we fix the problems you also mentioned above, I am not familiar with the details).

zhlhahaha:
> Ye I mean on SR-IOV bootstrap we don't, this is why I am saying that if your bootstrap use the default network bridge, We can use the manifests (once we fix the problems you also mentioned above, I am not familiar with the details).

I tried this on bare metal server which using docker rather than podman, so I think it is some other issue.

cluster-up/cluster/k3d/common.sh
```diff
@@ -116,38 +120,67 @@ function _label_agents() {
...
+--k3s-arg "--kubelet-arg=cpu-manager-policy=static@agent:*" \
+--k3s-arg "--kubelet-arg=kube-reserved=cpu=500m@agent:*" \
+--k3s-arg "--kubelet-arg=system-reserved=cpu=500m@agent:*" \
+-v /lib/modules:/lib/modules@agent:* \
+"
+    echo "${args}"
+}
+
+
 function _create_cluser() {
```
oshoval, line 174: Possible idea until we have manifests,
add a function to `common.sh` `named _create_arm_cluster`  which has all the changes needed
and then according the ARCH you can know which function to trigger
wdyt ?

zhlhahaha: If this script is only meant to serve CI/CD purposes, then it should suffice. However, if it's intended for use by normal users, it may lead to confusion since there is no CPU arch specific setting, but rather a naming convention with _create_arm_cluster.

In my opinion, if we want this script to have wider usage, it should be constructed like building blocks. We can have the basic parts in common.sh and put specific settings in another folder. In other words, it should be able to create a basic k3d cluster, and any other specific purposes can be built on top of it.

The core issue is figuring out which parts should be included in the basic k3d cluster and which should be put into the specific folders.

zhlhahaha: How about make this simple, just have a environmental variable `K3D_SINGLE_NODE` which can be true or false. If it is set to true, apply the adapted setting for Arm64 E2E testing, otherwise use sriov setting?

Here is an issue, if the E2E testing for Arm64 needs multi-node cluster, we need to change the script again.

oshoval: We use this lane for `SR-IOV` mostly and should
allow local development which just spin a cluster without `SR-IOV`
so the non `SR-IOV` should allow more than one node in the future, and not just `ARM` as you raised.
We can of course consider doing it simple and in steps.

I prefer that we continue next week please, after our vacation,
won't manage to review it before.
Thanks

zhlhahaha: > I prefer that we continue next week please, after our vacation, won't manage to review it before. Thanks

Ok, no problem. Thanks @oshoval

oshoval (commented):
Thanks
Reviewed again,
Added few comments and then we can take it even before
we have the manifests because it is good as it is until then.

cluster-up/cluster/k3d/common.sh
```diff
@@ -5,13 +5,33 @@ set -e
...
+esac
+
 function detect_cri() {
     if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi
 }
 
 export CRI_BIN=${CRI_BIN:-$(detect_cri)}
 KUBEVIRT_NUM_SERVERS=${KUBEVIRT_NUM_SERVERS:-1}
 KUBEVIRT_NUM_AGENTS=${KUBEVIRT_NUM_AGENTS:-2}
+DISABLE_DEFAULT_SERVICES=${DISABLE_DEFAULT_SERVICES:-false}
```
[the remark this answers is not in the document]
zhlhahaha: So we set `DISABLE_DEFAULT_SERVICES` to be true here, right?

> and if someone want to disable other he can set it ?

I am not quite understand this line.

cluster-up/cluster/k3d/common.sh
```diff
@@ -116,38 +120,67 @@ function _label_agents() {
...
+    fi
+    echo "${args}"
+}
+
+function _generate_k3d_args() {
+    args=$(_cluster_nodes_args)$(_cni_args)$(_device_args)
+    if [ ${DISABLE_DEFAULT_SERVICES} == true ] ; then
+        args=${args}" --k3s-arg --disable=traefik,servicelb,metrics-server@server:* "
+    else
+        args=${args}" --k3s-arg --disable=traefik@server:0 "
```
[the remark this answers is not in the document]
zhlhahaha: Shall we put `--no-lb` into this function?
```
if [ ${DISABLE_DEFAULT_SERVICES} == true ] ; then
        args=${args}" --k3s-arg --disable=traefik,metrics-server@server:* --no-lb "
fi
```

oshoval:
Before we continue, in case kind will be used for ARM, 
we might want to wait with this PR and use manifests when we have them ?
There is no reason to block k3d using manifests if we don't use it also for ARM (since podman is what blocking us now)

oshoval:
> > Before we continue, in case kind will be used for ARM, we might want to wait with this PR and use manifests when we have them ? There is no reason to block k3d using manifests if we don't use it also for ARM (since podman is what blocking us now)
> 
> Yes, I agree with you, manifests is better.

Is it fine to close the PR ?
In case we will need it we can back to it and address the comments
Thanks a lot for your effort

zhlhahaha:
> Is it fine to close the PR ? In case we will need it we can back to it and address the comments Thanks a lot for your effort

Ok, no problem.

PR #691, opened 2021-10-12
Add setup to emulate SCSI devices

alicefr:
Use emulated scsi devices in the provisioned nodes for development and testing.

Set the variable KUBEVIRT_PROVIDER_EXTRA_ARGS="--scsi 1G" for creating a SCSI disk during the node creation.

This PR adds a similar setup for SCSI devices like  #672 for NVMe devices.
Example
```bash
$ export KUBEVIRT_PROVIDER_EXTRA_ARGS="--scsi 1G" 
$ make cluster-up
[...]
$ cluster-up/ssh.sh node01
WARNING: KUBEVIRTCI_GOCLI_CONTAINER is set and will take precedence over the also set KUBEVIRTCI_TAG
selecting docker as container runtime
2021/10/12 10:08:26 Waiting for host: 192.168.66.101:22
2021/10/12 10:08:26 Connected to tcp://192.168.66.101:22
$ lsscsi
[0:0:0:0]    disk    QEMU     QEMU HARDDISK    2.5+  /dev/sda
$ lsblk /dev/sda 
NAME MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
sda    8:0    0   1G  0 disk 
```

PR #996, opened 2023-04-06
Provide usbs into VMs

xpivarc:
In order to test usb pass-through we need
few usb devices available in the VM.

Usage:
export KUBEVIRT_PROVIDER_EXTRA_ARGS="--usb 20M --usb 30M"

### DIFF
--- a/cluster-provision/centos9/scripts/vm.sh
+++ b/cluster-provision/centos9/scripts/vm.sh
@@ -22,6 +22,7 @@ while true; do
     -s | --block-device-size ) BLOCK_DEV_SIZE="$2"; shift 2 ;;
     -n | --nvme-device-size ) NVME_DISK_SIZES+="$2 "; shift 2 ;;
     -t | --scsi-device-size ) SCSI_DISK_SIZES+="$2 "; shift 2 ;;
+    -u | --usb-device-size ) USB_SIZES+="$2 "; shift 2 ;;
     -- ) shift; break ;;
     * ) break ;;
   esac
@@ -140,6 +141,16 @@ for size in ${SCSI_DISK_SIZES[@]}; do
   qemu-img create -f raw $disk $size
   let "disk_num+=1"
 done
+
+
+disk_num=0
+for size in ${USB_SIZES[@]}; do
+  echo "Creating disk "$size" for USB disk emulation"
+  disk="/usb-"${disk_num}".img"
+  qemu-img create -f raw $disk $size
+  let "disk_num+=1"
+done
+
 
 exec qemu-system-x86_64 -enable-kvm -drive format=qcow2,file=${next},if=virtio,cache=unsafe ${block_dev_arg} \
   -device virtio-net-pci,netdev=network0,mac=52:55:00:d1:55:${n} \
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -58,6 +58,7 @@ var soundcardPCIIDs = []string{"8086:2668", "8086:2415"}
 var cli *client.Client
 var nvmeDisks []string
 var scsiDisks []string
+var usbDisks []string
 
 type dockerSetting struct {
 	Proxy string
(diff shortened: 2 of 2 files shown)


cluster-provision/centos9/scripts/vm.sh
```diff
@@ -151,5 +157,10 @@ exec qemu-system-x86_64 -enable-kvm -drive format=qcow2,file=${next},if=virtio,c
   -vnc :${n} -cpu host,migratable=no,+invtsc -m ${MEMORY} -smp ${CPU} \
   -serial pty -M q35,accel=kvm,kernel_irqchip=split \
   -device intel-iommu,intremap=on,caching-mode=on -device intel-hda -device hda-duplex -device AC97 \
+  -device qemu-xhci,id=bus1 -device qemu-xhci,id=bus2  \
```
alicefr, line 160: Do we want to have it always enabled? Otherwise, we could add an option as its the case for SCSI and NVMe emulation.

xpivarc: We can do it on demand I don't have a problem with it.

victortoso:
As Alice mentioend, not sure if we want it always enabled but 3 usb storage devices with xhci controller should be enough to basic tests of usb passthrough.

alicefr:
> As Alice mentioend, not sure if we want it always enabled but 3 usb storage devices with xhci controller should be enough to basic tests of usb passthrough.

@xpivarc do you mind replicate what I did for [nvme](https://github.com/kubevirt/kubevirtci/pull/672) and [SCSI](https://github.com/kubevirt/kubevirtci/pull/691)? This allows to add any arbitrary numbers for devices

xpivarc:
@victortoso @alicefr PTAL

cluster-provision/gocli/cmd/run.go
```diff
@@ -114,6 +115,7 @@ func NewRunCommand() *cobra.Command {
 	run.Flags().Bool("enable-psa", false, "Pod Security Admission")
 	run.Flags().Bool("single-stack", false, "enable single stack IPv6")
 	run.Flags().Bool("enable-audit", false, "enable k8s audit for all metadata events")
+	run.Flags().StringArrayVar(&usbDisks, "usb", []string{}, "size of the emulate USB disk to pass to the node")
```
victortoso, line 118: My question is more about the end result for testing usb devices (emulated) in kubevirt.

So, the default is empty. Will we have an image that we can use in CI with something like `export KUBEVIRT_PROVIDER_EXTRA_ARGS="--usb 20M --usb 30M"`  ?

cluster-provision/gocli/cmd/run.go
```diff
@@ -475,6 +477,23 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 			}
 		}
 
+		var vmArgsUSBDisks []string
+		const bus = " -device qemu-xhci,id=bus%d"
+		const drive = " -drive if=none,id=stick%d,format=raw,file=/usb-%d.img"
+		const dev = " -device usb-storage,bus=bus%d.0,drive=stick%d"
```
victortoso, line 483: Didn't test but [documentation](https://www.qemu.org/docs/master/system/devices/usb.html) says we don't need to set `bus=` when using qemu-xhci

cluster-provision/gocli/cmd/run.go
alicefr: @victortoso if we need we could set the flag and the environmental variables when creating the kubevirtci instance on the CI.

alicefr:
Looks good, thanks @xpivarc 
/lgtm

cluster-provision/gocli/cmd/run.go
xpivarc: Yes, we can do that in automation

cluster-provision/gocli/cmd/run.go
xpivarc: Depends on # of controllers afaik.

xpivarc:
@brianmcarey @dhiller

brianmcarey (approved):
/approve

PR #1009, opened 2023-05-17
Cnao: Bump CNAO version in order to have KSD pre pulled

oshoval:


oshoval:
lets create a fresh one once kubevirtci is stable

PR #1011, opened 2023-05-19
downgrade kernel_version="4.18.0-448.el8.x86_64" for centos stream 8 providers

dhiller:
Also:
* bump k8s providers minor versions
* move k8s.gcr.io to registry.k8s.io

Closes #999


cluster-provision/k8s/1.24/provision.sh
```diff
@@ -47,8 +47,26 @@ function pull_container_retry() {
     fi
 }
 
+# FIXME: remove after version >= kernel-4.18.0-492.el8
```
oshoval, line 50: I think this is the major change on this PR and should appear please on the PR desc

can we consider using koji ?

https://discussion.fedoraproject.org/t/downgrading-to-a-previous-kernel-version/72820/7
https://fedoramagazine.org/install-kernel-koji

Thanks

dhiller: Koji is not available of the Centos Stream 8 repo:
```sh
[root@ibm-p8-kvm-03-guest-02 ~]# dnf install koji
...
No match for argument: koji
Error: Unable to find a match: koji
```

dhiller: What I just found out is that there's a kojihub for Centos Stream 8 and it has a build with the new kernel [here](https://kojihub.stream.centos.org/koji/buildinfo?buildID=33049)

FYI @brianmcarey

dhiller: I have created a PR using the box from CentOS kojihub here: https://github.com/kubevirt/kubevirtci/pull/1012

dhiller:
@brianmcarey I'd say this one is good enough. The 1.24 has had failures in the conformance tests. The new box [doesn't look stable](https://geek-and-poke.com/geekandpoke/2011/8/22/hudsonjenkins-status-monitor.html). WDYT?

brianmcarey:
> @brianmcarey I'd say this one is good enough. The 1.24 has had failures in the conformance tests. The new box [doesn't look stable](https://geek-and-poke.com/geekandpoke/2011/8/22/hudsonjenkins-status-monitor.html). WDYT?

Yes I agree - we can take this to unblock merges to the kubevirtci repo. The 1.24 provider is due to be removed in the near future anyway.

brianmcarey (approved):
/approve

oshoval:
> > Thanks, it might worth to consider pinning kernel also for centos9 based providers? We don't need to suffer kernel changes, we can bump once in a while wdyt?
> 
> We had this discussion a while ago whether we should pin versions vs. consuming the latest. I'd say let's not pin and consume the latest (which has underwent a testing process and only then put on a mirror) and catch eventual issues through our CI. Otherwise we would need to frequently bump the pins and that would require a process by itself.
> 
> WDYT?

We need at least to make the pin mechanism in a function, that we can call in case we have fatas like we had now,
We can consider the bump mechanism imo, it will give more control and determinism.

Adding @EdDev

PR #1012, opened 2023-05-22
Use box from koji

dhiller:
As an alternative to downgrading the kernel (see https://github.com/kubevirt/kubevirtci/pull/1011), we try to use the latest build that contains the fix we need.

Also:
* bump k8s providers minor versions
* move k8s.gcr.io to registry.k8s.io

Closes https://github.com/kubevirt/kubevirtci/pull/999
Closes https://github.com/kubevirt/kubevirtci/pull/1011


cluster-provision/centos8/Dockerfile
```diff
@@ -23,7 +23,7 @@ COPY scripts/download_box.sh /
 
 RUN echo "Centos8 version $centos_version"
 
-ENV CENTOS_URL https://cloud.centos.org/centos/8-stream/x86_64/images/CentOS-Stream-Vagrant-8-$centos_version.x86_64.vagrant-libvirt.box
+ENV CENTOS_URL https://kojihub.stream.centos.org/kojifiles/packages/CentOS-Stream-Vagrant/8/$centos_version/images/CentOS-Stream-Vagrant-8-$centos_version.x86_64.vagrant-virtualbox.box
```
brianmcarey, line 26: I think we would need the libvirt.box here instead of the box built for virtualbox?

dhiller: Good catch, thanks!

dhiller:
> Very nice This build is safe ? (you raised on some comment there are some unsafe ones)
> 
> About the image repo replace, need to make sure please no change will get overridden when there is a bump of that manifest, in case it will, need to either change scripts if there are or maybe add some notes in a doc about it ?
> 
> Thanks

Do you mean that a build might be overwritten? IIUC koji builds are immutable when done.

dhiller:
> > > Very nice This build is safe ? (you raised on some comment there are some unsafe ones)
> > > About the image repo replace, need to make sure please no change will get overridden when there is a bump of that manifest, in case it will, need to either change scripts if there are or maybe add some notes in a doc about it ?
> > > Thanks
> > 
> > 
> > Do you mean that a build might be overwritten? IIUC koji builds are immutable when done.
> 
> I mean the 2nd commit changes, "move k8s.gcr.io to registry.k8s.io"

Ah, good point, thanks for clarification. IMHO the deprecation of the old registry should have been communicated over and over, but we still might not be safe from someone re-introducing the old one.

Need to think about how best to communicate this.

oshoval:
> OTOH this will raise errors right away. I did the change since I was reminded of CI that the old registry is defunct. Maybe that is good enough, WDYT @oshoval ?

Totally

dhiller:
/close

#1011 was merged

PR #1015, opened 2023-05-31
vm based providers: Add enable audit flag

oshoval:
In order to see rbac allow / forbid actions by k8s, add flag which allows audit logging.

Can be used by `export KUBEVIRT_ENABLE_AUDIT=true` before `cluster-up`,
and then use:
`./cluster-up/ssh.sh node01 "sudo cat /var/log/k8s-audit/k8s-audit.log"`
It is the same log file that always been, just logs all metadata events.


brianmcarey (approved):
/lgtm
/approve

PR #1016, opened 2023-06-01
Downgrade to previous minor release of kubernetes

brianmcarey:
Drop to previous minor release for 1.25, 1.26 and 1.27 providers.

There is an issue with the latest minor releases that is causing some sig-network tests to fail[1] so lets drop back a minor version to allow us to deliver kubevirtci changes to kubevirt again.

The automated PR will rebump the minor version so this change is temporary until a fix is identified. 

[1] https://github.com/kubevirt/kubevirt/pull/9616#issuecomment-1566168918

/cc @xpivarc @ormergi @dhiller


dhiller (approved):
/approve

PR #1019, opened 2023-06-05
WIP DNM kind: Try to use kind for sig-network

oshoval:
Quick and dirty, try to use kind for sig-network

First commit (can be split later)
In order to ease readability, arrange to functions. 
Add a flag `DEPLOY_SRIOV` (default: true), in order to allow skipping deploying SR-IOV.
This is turn would allow to spin a kind cluster, and use it for non SR-IOV tasks.


oshoval:
Doesn't worth the effort

PR #1020, opened 2023-06-06
provision-manager: Fix publish script

oshoval:
A @ was missing so the looping on build and push didnt work.


dhiller (approved):
/lgtm
/approve

PR #1021, opened 2023-06-06
WIP: try to revert pman

oshoval:


oshoval:
@brianmcarey 
we see that it started to pass even with pman at least twice

https://prow.ci.kubevirt.io/?job=check-provision-k8s-1.26-centos9

PR #1023, opened 2023-06-07
vm based: Skip presubmit according provision manager

oshoval:
Skip kubevirtci presubmits in case the changes doesn't affect the
specific provider (according pman).

This time it will test them all because `provision.sh` itself is an "all" rule according pman.
In order to provision locally when no changes exists,
one will need to use `BYPASS_PMAN_CHANGE_CHECK` from now on.

### DIFF
--- a/cluster-provision/k8s/provision.sh
+++ b/cluster-provision/k8s/provision.sh
@@ -7,6 +7,7 @@ PHASES_DEFAULT="linux,k8s"
 PHASES="${PHASES:-$PHASES_DEFAULT}"
 CHECK_CLUSTER="${CHECK_CLUSTER:-false}"
 export SLIM="${SLIM:-false}"
+BYPASS_PMAN_CHANGE_CHECK=${BYPASS_PMAN_CHANGE_CHECK:-false}
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 provision_dir="$(basename "$(pwd)")"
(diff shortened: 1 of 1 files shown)


oshoval:
Thanks
Addressed comments

dhiller (approved):
/approve

oshoval:
/test check-provision-manager 

Weird it hanged

oshoval:
Can you please override check-provision-k8s-1.26-centos9  ? it seems it hanged
and it isnt important for this PR

cluster-provision/k8s/provision.sh
```diff
@@ -15,8 +16,18 @@ export base
 
 cd $DIR
 
-export KUBEVIRT_CGROUPV2="${CGROUPV2}"
+make -C ../gocli cli
```
brianmcarey, line 19: Would it be possible to use an existing gocli image instead of building here? Not sure if we need to add an extra build here.

oshoval: we always build it here already and we must, (see few lines below)
it takes a moment and must be build as part of the provision as it use `provision` command of gocli and the pman command, so if there are changes we need to take them into account for both commands

brianmcarey: Ok sorry I missed that. Looks good to me.

brianmcarey:
/override check-provision-k8s-1.26-centos9

Not related to this change and just delaying it.

PR #1025, opened 2023-06-12
cnao, vm based: Bump CNAO to v0.86.0

oshoval:
This will allow KSD to be prepulled.
Note that CNAO and KMP Rbacs were changed (reduced to be more precise).


oshoval:
/hold

lets understand first what is wrong with https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2380
just to be safe

PR #1027, opened 2023-06-13
cluster kind: update kind-1.23 to kind-1.27

zhlhahaha:


brianmcarey (commented):
For adding a provider - its generally better to copy the existing provider in the first commit and then make the required updates. It can be difficult to update a provider like this as there maybe lanes in kubevirt that still rely on kind-1.23 but the update PR will fail because the provider no longer exists. 

The old provider can be removed easily in a follow up PR.

zhlhahaha:
> For adding a provider - its generally better to copy the existing provider in the first commit and then make the required updates. It can be difficult to update a provider like this as there maybe lanes in kubevirt that still rely on kind-1.23 but the update PR will fail because the provider no longer exists.
> 
> The old provider can be removed easily in a follow up PR.

Yes, you are right. Updated.

cluster-up/cluster/kind-1.27/provider.sh
```diff
@@ -0,0 +1,51 @@
...
+
+if [ $CLUSTER_NAME == $DEFAULT_CLUSTER_NAME ]; then
+    export HOST_PORT=$DEFAULT_HOST_PORT
+else
+    export HOST_PORT=$ALTERNATE_HOST_PORT
+fi
+
+function set_kind_params() {
+    export KIND_VERSION="${KIND_VERSION:-0.19.0}"
+    export KIND_NODE_IMAGE="${KIND_NODE_IMAGE:-quay.io/kubevirtci/kindest-node:v1.27.1}"
```
oshoval, line 18: Please use the original image, as it doesn't have rate limits anymore
and won't need mirroring.

also the mirrored image did problems for me.

cluster-up/cluster/kind-1.27/README.md
```diff
@@ -0,0 +1,45 @@
...
+Provides a pre-deployed k8s cluster with version 1.27.1 that runs using [kind](https://github.com/kubernetes-sigs/kind) The cluster is completely ephemeral and is recreated on every cluster restart.
+The KubeVirt containers are built on the local machine and are then pushed to a registry which is exposed at
+`localhost:5000`.
+
+## Bringing the cluster up
+
+The following needs to be executed as root.
+
+```bash
+export KUBEVIRT_PROVIDER=kind-1.27
```
oshoval, line 12: You might want to remove the versioning so it won't need to be updated every now and then please
(all over the file)

oshoval (commented):
Thanks

oshoval:
Can you please remove the previous folder ?
kind-1.23 (unless I missed that it was removed or so)

cluster-up/cluster/kind-1.27/README.md
zhlhahaha: I am not quit sure about this. Do you means remove the line `export KUBEVIRT_PROVIDER=kind-1.27` in `README.md` or avoid use specific kind version in `README.md`? I think this should be fine.

zhlhahaha:
> Can you please remove the previous folder ? kind-1.23 (unless I missed that it was removed or so) unless you want to do it on a different PR note that we doing it on steps because we have to bump in kubevirt, 

Yes, we will remove kind-1.23 in follow up PR. as you can see @brianmcarey 's comments https://github.com/kubevirt/kubevirtci/pull/1027#pullrequestreview-1476928735

> but imho if it doesn't posses risks for you you can do it in one shot

I have a little concern that if we remove kind-1.23 in this PR, there would be some time gap among sync kubevirtci, kubevirt, and updating the arm64 e2e test job in project-infra. The time gap may lead to crash of e2e test on Arm64. By keeping the kind-1.23, we can make sure the migration of e2e test on Arm64 smoothly from kind-1.23 to kind-1.27.

cluster-up/cluster/kind-1.27/README.md
oshoval: Ye just use please "x.yz", so we never need to update versioning there
Thans

cluster-up/cluster/kind-1.27/provider.sh
```diff
@@ -0,0 +1,51 @@
...
+
+if [ $CLUSTER_NAME == $DEFAULT_CLUSTER_NAME ]; then
+    export HOST_PORT=$DEFAULT_HOST_PORT
+else
+    export HOST_PORT=$ALTERNATE_HOST_PORT
+fi
+
+function set_kind_params() {
+    export KIND_VERSION="${KIND_VERSION:-0.19.0}"
+    export KIND_NODE_IMAGE="${KIND_NODE_IMAGE:-kindest/node:v1.27.1}"
```
oshoval, line 18: We usually use please hash, not sure if the label is changed, but safer to use hash, as it is for sure predictable

`kindest/node:v1.27.1@sha256:b7d12ed662b873bd8510879c1846e87c7e676a79fefc93e17b2a52989d3ff42b`
took from https://github.com/kubernetes-sigs/kind/releases of course

zhlhahaha: updated

oshoval (commented):
Thanks

oshoval:
Weird it didn't catch the approval

/lgtm

PR #1028, opened 2023-06-13
provision-manager: Switch to fedora

oshoval:
It is possible that gocli Alpine is the cause for the hangs during provision we see lately.
It started with the pman PR, and also running pman itself hanged once.
Lets move to fedora, to see if it is related.
We can also consider using go-git and then we can remove the OS totally.

### DIFF
--- a/cluster-provision/gocli/build/Dockerfile
+++ b/cluster-provision/gocli/build/Dockerfile
@@ -1,6 +1,6 @@
-FROM alpine:3.18.0
+FROM fedora:35
 
-RUN apk add git
+RUN dnf install -y git
 WORKDIR /workdir
 RUN git config --global --add safe.directory /workdir
 


oshoval:
we see it is not related to the image
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1028/check-provision-k8s-1.27/1668682223009665024

I believe it is related then to rpms / network manager / kernel
(unless it is related to gocli vendoring that happened on provision manager PR, but i am skeptic)

Btw I suggest that we will have a mechanism that allows to freeze rpms, in order to investigate and mitigate such cases.
We can do it by dnf install that doesnt delete the rpms, and then we can reuse the files and their versions when needed.
At least that we will have the selection when to do it, for example in the current case it could help us know if it is the rpms that cause it.
At the past we had a problem with NetworkManager that caused a bug pretty much like we have now
Wdyt?

@brianmcarey @dhiller FYI
closing this PR

PR #1030, opened 2023-06-14
pman: Update info about BYPASS_PMAN_CHANGE_CHECK

oshoval:
Ease user in case running provision locally, when no changes to the target exist.

### DIFF
--- a/KUBEVIRTCI_LOCAL_TESTING.md
+++ b/KUBEVIRTCI_LOCAL_TESTING.md
@@ -24,6 +24,10 @@ cd $KUBEVIRTCI_DIR
 (cd cluster-provision/k8s/1.25; ../provision.sh)
 ```
 
+Note: 
+If you see "INFO: skipping provision of x.yz because according provision-manager it hadn't changed"
+please use `export BYPASS_PMAN_CHANGE_CHECK=true` to bypass provision-manager changes check and force provision.
+
 ### prepare for using the new provisioned cluster
 
 ```bash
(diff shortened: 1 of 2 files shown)


oshoval:
Failures here not related to the PR

brianmcarey:
> @brianmcarey Do we want a way to tell pman force skip all ? for example in such PR it would help I think it worth it, wdyt?
> 
> Btw what pman doesn't give us yet also is to skip kind if kind wasn't changed, it is because it is not written in such way atm. only legit targets are vm based providers. If we want to extend it we can on a follow effort.

The problem I can see with regularly skipping the check-provision jobs is if a couple start failing - it will be hard to identify a timeline for when the failures start as they may have been skipped for a few weeks and not run. We don't periodically test the provisions.

brianmcarey (approved):
/approve

enp0s3:
/hold
looks like something bad happening on `check-provision-k8s-1.27`, not related to this PR changes

oshoval:
> > @brianmcarey Do we want a way to tell pman force skip all ? for example in such PR it would help I think it worth it, wdyt?
> > Btw what pman doesn't give us yet also is to skip kind if kind wasn't changed, it is because it is not written in such way atm. only legit targets are vm based providers. If we want to extend it we can on a follow effort.
> 
> The problem I can see with regularly skipping the check-provision jobs is if a couple start failing - it will be hard to identify a timeline for when the failures start as they may have been skipped for a few weeks and not run. We don't periodically test the provisions.

I mean only for PRs like this one, we will use it manually on PRs that change just docs / kind,
we have pretty live PRs that do trigger presubmit
setting it will be one timer per PR where it was set

oshoval:
`Error response from daemon: container create: conflict in specified volumes - multiple volumes at "/var/run/disk": duplicate mount destination`

Someone is familiar with this maybe ?

brianmcarey (commented):
/unhold 

/lgtm

/retest-required

oshoval:
Rebased, lets see if it solves the provision problem
(even that theoretically CI does rebase)

brianmcarey (commented):
/lgtm

PR #1031, opened 2023-06-15
kind: Introduce a way to skip kind based providers presubmits

oshoval:
If `cluster-up/cluster/kind/skip_tests` is changed versus the upstream/main,
it will skip the kind based providers presubmits tests.


oshoval:
/hold

assuming upstream/main is declared on CI as well , lets see

ormergi:
Could you please explain why this change is needed?
From what I understand with this change one could post a PR and skip CI gating tests.

oshoval:
> Could you please explain why this change is needed? From what I understand with this change one could post a PR and skip CI gating tests.

It allows that if you for example change only markdown files or vm based providers files, it will skip kind based providers presubmits if you manually set it.
The better thing will be indeed to extend provision manager such that it will cover also kind based providers, and then it will be done automatically basically

lets close it, and if needed extend provision manager, because it can add busy work that doesnt worth it

PR #1033, opened 2023-06-25
Allow deploying custom CDI version

akalenyu:
This may be useful in cases where a backported
CDI fix needs to be exercised in kubevirt for example, to ensure kubevirt integration with it.

So we would:
- Release CDI with the backport
- Open a PR with tests in kubevirt/kubevirt and deploy this custom CDI to ensure integration


lyarwood (commented):
/lgtm

Yeah that would be useful, thanks!

cluster-up/hack/common.sh
```diff
@@ -32,6 +32,7 @@ KUBEVIRT_DEPLOY_GRAFANA=${KUBEVIRT_DEPLOY_GRAFANA:-false}
 KUBEVIRT_CGROUPV2=${KUBEVIRT_CGROUPV2:-false}
 KUBEVIRT_DEPLOY_CDI=${KUBEVIRT_DEPLOY_CDI:-false}
 KUBEVIRT_DEPLOY_CDI_LATEST=${KUBEVIRT_DEPLOY_CDI_LATEST:-false}
```
brianmcarey, line 34: Do we need to keep this variable? Could we not just use? 

`KUBEVIRT_CUSTOM_CDI_VERSION="latest"`

brianmcarey (commented):
Looks good @akalenyu  - just one question. Not sure if we still need `KUBEVIRT_DEPLOY_CDI_LATEST` ?

cluster-up/hack/common.sh
akalenyu: Yes, we don't want to always go with "latest", the default is "", i.e. use whatever CDI is in kubevirtci at that point
kubevirt/kubevirt only uses released versions

brianmcarey: But in the case of wanting to test against latest - could you not just use `KUBEVIRT_CUSTOM_CDI_VERSION="latest"`  instead of needing to set `KUBEVIRT_DEPLOY_CDI_LATEST="true"` 

It would just give us a chance to remove one env variable.

akalenyu: oh. I kept that to not break anything that may be already using it..

lyarwood: https://github.com/search?q=org%3Akubevirt%20KUBEVIRT_DEPLOY_CDI_LATEST&type=code

We should be good to drop it, as the person that introduced it I really don't mind, it's just as easy to set `KUBEVIRT_CUSTOM_CDI_VERSION="latest"`  as @brianmcarey highlights.

brianmcarey: Yes -  I did a quick search through project-infra and there no mention of `KUBEVIRT_DEPLOY_CDI_LATEST`

akalenyu: makes sense, done

cluster-up/hack/common.sh
```diff
@@ -31,7 +31,7 @@ KUBEVIRT_DEPLOY_PROMETHEUS_ALERTMANAGER=${KUBEVIRT_DEPLOY_PROMETHEUS_ALERTMANAGE
 KUBEVIRT_DEPLOY_GRAFANA=${KUBEVIRT_DEPLOY_GRAFANA:-false}
 KUBEVIRT_CGROUPV2=${KUBEVIRT_CGROUPV2:-false}
 KUBEVIRT_DEPLOY_CDI=${KUBEVIRT_DEPLOY_CDI:-false}
-KUBEVIRT_DEPLOY_CDI_LATEST=${KUBEVIRT_DEPLOY_CDI_LATEST:-false}
+KUBEVIRT_CUSTOM_CDI_VERSION=${KUBEVIRT_CUSTOM_CDI_VERSION}
```
brianmcarey, line 34: Does this need a default or is it ok? What happens when not set?

akalenyu: We'll just use whatever was in kubevirtci at that point
(which gets bumped periodically by these: https://github.com/kubevirt/kubevirtci/pull/911)

brianmcarey (approved):
/approve

cluster-up/hack/common.sh
lyarwood: I was going to question if we even need this but it does make the variable more discoverable IMHO so I'm cool with it.

PR #1036, opened 2023-06-29
remove PSA booleans since its always enabled

enp0s3:
Signed-off-by: enp0s3 <ibezukh@redhat.com>

The existence of KUBEVIRT_PSA with default of `false` is confusing. I tackled this issue when trying
to use KCI in a new project, having in mind that KUBEVIRT_PSA is false I was thinking the PSA is off in my cluster.
Since PSA is on by default since 1.23, and we don't to disable it in any of our job configs, I suggest to remove the
variable in order to eliminate the confusion.


cluster-provision/k8s/1.25/k8s_provision.sh
```diff
@@ -146,7 +146,6 @@ plugins:
       usernames: []
       runtimeClasses: []
       # Hopefuly this will not be needed in future. Add your favorite namespace to be ignored and your operator not broken
-      # You also need to modify psa.sh
```
xpivarc, line 149: We need to change the above enforcement level.

xpivarc (changes requested):

cluster-provision/k8s/1.25/k8s_provision.sh
enp0s3: @xpivarc you mean `enforce: "privileged"`? to which level?

xpivarc: Yes, it needs to be changed to "restricted" as audit

enp0s3: @xpivarc Thanks. Its difficult to find docs that explain these labels, had to go deep to [sig-auth](https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement#summary).

xpivarc: fyi there is nice doc https://kubernetes.io/docs/concepts/security/pod-security-standards/

enp0s3: Yeah I saw it but it doesn't explain the meaning of `enforce, audit, warn`

xpivarc: Oh, you are right. The bits are not on one page https://kubernetes.io/docs/concepts/security/pod-security-admission/

enp0s3:
Need to adjust sonobouy tests to configure privileged NS for the tests

enp0s3:
In order for the tests to pass we need to configure Kubevirt CR for Seccomp, as we do in [e2e](https://github.com/kubevirt/kubevirt/blob/b1c1893a24891543385d2cac24de6945345f17eb/tests/testsuite/kubevirtresource.go#L86). Also, following HCO [logic](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/6728bbd9d6e56e2a0632e35b1d8bfc9882600c13/controllers/hyperconverged/hyperconverged_controller_test.go#L236) I undertstand that its legit to pre-configure it.

brianmcarey:
@enp0s3 @xpivarc whats the status here? It would be good to get this cleaned up. Is there anything I can help with?

enp0s3:
@brianmcarey Hey. The conformance tests failure needs to be investigated, I won't have BW this week.

brianmcarey:
/retest-required

Lets get a fresh run and then I will try to have a look.

dhiller:
@enp0s3 I agree that we want to remove the KUBEVIRT_PSA switch completely, but I also understand that we always want to enable the PSA policy as described in `psa.sh` per provider. Therefore the modification of `psa.sh` would need to happen always, no?

PR #1042, opened 2023-07-09
vm based providers: Allow using custom images URL

oshoval:
If `KUBEVIRTCI_PROVISION_CHECK` is not used,
you can set the following:
`KUBEVIRTCI_CONTAINER_REGISTRY` (default: `quay.io`)
`KUBEVIRTCI_CONTAINER_ORG` (default: `kubevirtci`)
`KUBEVIRTCI_CONTAINER_SUFFIX` (default: according gocli tag)
in order to use a custom image.

Usage example (will fetch quay.io/username/k8s-1.25:latest, in case KUBEVIRT_PROVIDER=k8s-1.25)
```
export KUBEVIRTCI_CONTAINER_REGISTRY=quay.io
export KUBEVIRTCI_CONTAINER_ORG=username
export KUBEVIRTCI_CONTAINER_SUFFIX=latest
```

### DIFF
--- a/KUBEVIRTCI_LOCAL_TESTING.md
+++ b/KUBEVIRTCI_LOCAL_TESTING.md
@@ -35,6 +35,10 @@ please use `export BYPASS_PMAN_CHANGE_CHECK=true` to bypass provision-manager ch
 export KUBEVIRTCI_PROVISION_CHECK=1
 ```
 
+If `KUBEVIRTCI_PROVISION_CHECK` is not used,
+you can set `KUBEVIRTCI_CONTAINER_REGISTRY` (default: `quay.io`), `KUBEVIRTCI_CONTAINER_ORG` (default: `kubevirtci`) and `KUBEVIRTCI_CONTAINER_SUFFIX` (default: according gocli tag),
+in order to use a custom image.
+
 Note:
 In case you updated gocli and need to test it locally as well, export additionally:
 ```bash
--- a/cluster-up/cluster/ephemeral-provider-common.sh
+++ b/cluster-up/cluster/ephemeral-provider-common.sh
@@ -101,8 +101,22 @@ function _add_common_params() {
 
     if [ -n "${KUBEVIRTCI_PROVISION_CHECK}" ]; then
         params=" --container-registry=quay.io --container-suffix=:latest $params"
-    elif [[ ${KUBEVIRT_SLIM} == "true" ]]; then
-        params=" --slim $params"
+    else
+        if [ -n $KUBEVIRTCI_CONTAINER_REGISTRY ]; then
+            params=" --container-registry=$KUBEVIRTCI_CONTAINER_REGISTRY $params"
+        fi
+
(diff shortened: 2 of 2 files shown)


xpivarc:
What is the use case for this?

oshoval:
For example if you want to push to your own repository and then use it, (and not just the one that you can use with
`KUBEVIRTCI_PROVISION_CHECK` after provisioning it)

Later on we can allow presubmit on CI to push pre released providers with a special temp tag,
and using this code, we can allow known images (restrict others, see below partial example) to be tested versus kubevirt, before kubevirtci is released

```
            if [ "$CI" == "true" ]; then
                echo "KUBEVIRT_CONTAINER_REGISTRY must be left empty for safeness" 
                # same need to be done for KUBEVIRT_CONTAINER_ORG
                exit 1
            fi
```

KUBEVIRTCI_LOCAL_TESTING.md
```diff
@@ -35,6 +35,10 @@ please use `export BYPASS_PMAN_CHANGE_CHECK=true` to bypass provision-manager ch
 export KUBEVIRTCI_PROVISION_CHECK=1
 ```
 
+If `KUBEVIRTCI_PROVISION_CHECK` is not used,
+you can set `KUBEVIRT_CONTAINER_REGISTRY` (default: `quay.io`), `KUBEVIRT_CONTAINER_ORG` (default: `kubevirtci`) and `KUBEVIRT_CONTAINER_SUFFIX` (default: according gocli tag),
```
brianmcarey, line 39: Could we rename the environment variables from `KUBEVIRT` to `KUBEVIRTCI` ?

brianmcarey (commented):
Thanks @oshoval - looks useful. Just one comment inline on the naming of the env variables.

KUBEVIRTCI_LOCAL_TESTING.md
oshoval: yes and no imho, it is indeed KUBEVIRTCI
but all the other parameters even that they belong to KUBEVIRTCI, have prefix KUBEVIRT, so i don't think we should do just to those ?

brianmcarey: The rest of the env vars are more related to how kubevirt is deployed and what is deployed alongside it whereas the env vars you are adding are focused on published kubevirtci images. 

I just think that `KUBEVIRT_CONTAINER_REGISTRY` would lead me to think of the image registry used for the kubevirt images.

oshoval: right, done

oshoval:
btw not related to this PR, but the vgpu is flaky for long time (unless a real bug) 

https://search.ci.kubevirt.io/?search=Should+override+default+mdev+configuration+on+a+specific+node&maxAge=336h&context=1&type=bug%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

brianmcarey (approved):
/approve

oshoval:
seems the provision started to hang again

Issue #991, opened 2023-03-29
Investigate the possible solutions for #990 (reverting kind worker assignment for control plane)


oshoval:
This PR #990 reverts https://github.com/kubevirt/kubevirtci/pull/985 due to a few reasons (see please [PR](https://github.com/kubevirt/kubevirtci/issues/990) desc).

We should investigate the potential solutions which allow us to bring the PR back, and keep all the tests passing.

Note that this was by design always on the provider, we had 1 control and 2 workers.
There can be ways to fix it, but they need to take into account PF assignments,
Or to change only the non SR-IOV provider (which imo is the best solution, without need to
reduce number of nodes for SR-IOV unless we find out it doesn't have side effects).
Note also that soon we plan to [replace](https://github.com/kubevirt/kubevirtci/issues/984) kind with k3d totally.
For more info please see discussion here 
https://github.com/kubevirt/kubevirtci/pull/983#issuecomment-1480887214

PR #1043, opened 2023-07-10
kind, SR-IOV: Arrange to functions

oshoval:
In order to ease readability, arrange to functions.

### DIFF
--- a/cluster-up/cluster/kind-1.27-sriov/provider.sh
+++ b/cluster-up/cluster/kind-1.27-sriov/provider.sh
@@ -13,6 +13,13 @@ else
     export HOST_PORT=$ALTERNATE_HOST_PORT
 fi
 
+function print_available_nics() {
+    # print hardware info for easier debugging based on logs
+    echo 'Available NICs'
+    ${CRI_BIN} run --rm --cap-add=SYS_RAWIO quay.io/phoracek/lspci@sha256:0f3cacf7098202ef284308c64e3fc0ba441871a846022bb87d65ff130c79adb1 sh -c "lspci | egrep -i 'network|ethernet'"
+    echo ""
+}
+
 function set_kind_params() {
     export KIND_VERSION="${KIND_VERSION:-0.18.0}"
     export KIND_NODE_IMAGE="${KIND_NODE_IMAGE:-kindest/node:v1.27.1@sha256:9915f5629ef4d29f35b478e819249e89cfaffcbfeebda4324e5c01d53d937b09}"
@@ -44,12 +51,13 @@ function configure_registry_proxy() {
     KIND_BIN="$kind_binary_path" PROXY_HOSTNAME="$ci_proxy_hostname" $configure_registry_proxy_script
 }
 
-function up() {
-    # print hardware info for easier debugging based on logs
-    echo 'Available NICs'
-    ${CRI_BIN} run --rm --cap-add=SYS_RAWIO quay.io/phoracek/lspci@sha256:0f3cacf7098202ef284308c64e3fc0ba441871a846022bb87d65ff130c79adb1 sh -c "lspci | egrep -i 'network|ethernet'"
-    echo ""
+function deploy_sriov() {
+    print_available_nics
+    ${KUBEVIRTCI_PATH}/cluster/$KUBEVIRT_PROVIDER/config_sriov_cluster.sh
+    print_sriov_data
+}
 
+function up() {
     cp $KIND_MANIFESTS_DIR/kind.yaml ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
     export CONFIG_WORKER_CPU_MANAGER=true
     kind_up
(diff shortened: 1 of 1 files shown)


cluster-up/cluster/kind-1.27-sriov/README.md
```diff
@@ -67,6 +67,11 @@ export KUBECTL_PATH="/usr/bin/kubectl"
 This allows users to test or use custom images / different kind versions before making them official.
 See https://github.com/kubernetes-sigs/kind/releases for details about node images according to the kind version.
 
+### Running kind without SR-IOV
```
ormergi, line 70: Since this is an SR-IOV provider, the way I see it, there is no sense in running it without all the configurations it does for SR-IOV.
If you want to allow spinning up basic KIND cluster for kubevirt, I suggest introducing some kind of basic provider to allow that.
One that have all basic configurations necessary for kubevirt (e.g.: registry).

It doesn't have to be a formal provider, at least a separate file for spinning up a basic cluster, where each `provider.sh` script will call it and do extra configuration in its context.
I think it will simplify the provider's script by separating responsibilities (spinning up the cluster vs configuring it) and no need to act conditionally upon cluster kind (if SRIOV... do).

oshoval: We already have kind-x which is maintained by @zhlhahaha for ARM
but it is dedicated to ARM needs iiuc (in the past we discussed if we can combine them, and it was concluded it is better not, because this way it is decoupled)

imho, having this flag here won't hurt, sometimes you need just a cluster of kind, without SR-IOV features
and this simple method allows it

oshoval: We already have kind-x which is maintained by @zhlhahaha for ARM
but it is dedicated to ARM needs iiuc (in the past we discussed if we can combine them, and it was concluded it is better not, because this way it is decoupled)

imho, having this flag here won't hurt, sometimes you need just a cluster of kind, without SR-IOV features
and this simple method allows it

another benefit is that it fixes https://github.com/kubevirt/kubevirtci/issues/991

EdDev (changes requested):
This PR is doing two things: Arranging functions and adding a controversy flag.
I think the function part has agreement and can be pushed forward.

I share @ormergi concern about the flag added. It makes little sense to see such a flag under a folder/lane that is aimed to serve SR-IOV. While I can understand it may have usage, it is more or less a hack.

There is a `kind` folder, under which you may consider adding functions that the SR-IOV can use, and a deployment that does not include SR-IOV to be kept there.

oshoval:
Kept only the function arrangements

i honestly find the flag helpful in some cases, for example during the kind provider development
if it possible to open a follow PR for that, would be great
thanks

cluster-up/cluster/kind-1.27-sriov/provider.sh
```diff
@@ -13,6 +13,13 @@ else
     export HOST_PORT=$ALTERNATE_HOST_PORT
 fi
 
+function print_available_nics() {
+    # print hardware info for easier debugging based on logs
+    echo 'Available NICs'
+    ${CRI_BIN} run --rm --cap-add=SYS_RAWIO quay.io/phoracek/lspci@sha256:0f3cacf7098202ef284308c64e3fc0ba441871a846022bb87d65ff130c79adb1 sh -c "lspci | egrep -i 'network|ethernet'"
```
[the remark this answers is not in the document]
oshoval: /cc @phoracek 
FYI that we going to change it

cluster-up/cluster/kind-1.27-sriov/provider.sh
```diff
@@ -44,12 +51,13 @@ function configure_registry_proxy() {
     KIND_BIN="$kind_binary_path" PROXY_HOSTNAME="$ci_proxy_hostname" $configure_registry_proxy_script
 }
 
-function up() {
-    # print hardware info for easier debugging based on logs
-    echo 'Available NICs'
-    ${CRI_BIN} run --rm --cap-add=SYS_RAWIO quay.io/phoracek/lspci@sha256:0f3cacf7098202ef284308c64e3fc0ba441871a846022bb87d65ff130c79adb1 sh -c "lspci | egrep -i 'network|ethernet'"
-    echo ""
+function deploy_sriov() {
```
ormergi, line 54: I would call it something like `prepare_sriov_cluster` because it does more than deploy SR-IOV components.

cluster-up/cluster/kind-1.27-sriov/provider.sh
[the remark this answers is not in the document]
ormergi: Did it every helped us in a way?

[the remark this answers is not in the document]
oshoval: I bet it did in the first times, this is why it was added, according the comments as well
@phoracek  can suggest more maybe

cluster-up/cluster/kind-1.27-sriov/provider.sh
oshoval: imho deploy verb encapsulate more than preparation, which stands usually for the first phase of something

PR #1049, opened 2023-07-14
Add container stop to ensure it stops after shutdown

brianmcarey:
/cc none


cluster-provision/gocli/cmd/provision.go
```diff
@@ -274,6 +275,11 @@ func provisionCluster(cmd *cobra.Command, args []string) (retErr error) {
 	if err != nil {
 		return err
 	}
+	timeout := time.Second * 10
```
oshoval, line 278: Thanks, but i am not sure that we don't workaround some other problem with it,
we have on line R269 `ssh.sh sudo shutdown now -h` which stops the VM, once the VM is stopped, as it was the entrypoint of the container, the container will stop by itself as far as i remember,
adding a stop means we intervene with it, if we can debug it on CI once it happens, to understand why the VM hangs might be better?

even if at the end we take it, maybe worth to check what is a legitimate timeout, maybe 10 seconds is not enough for the VM to gracefully shutdown,
we also need to make sure that it doesn't have consequences (well, i believe that there is a problem with some rpm/kernel that is hidden, and even if we dont add the stop might appear sometimes, for example if node get restarted, which we don't usually do, so in this aspect i guess we are fine)

brianmcarey: > Thanks, but i am not sure that we don't workaround some other problem with it, we have on line R269 `ssh.sh sudo shutdown now -h` which stops the VM, once the VM is stopped, as it was the entrypoint of the container, the container will stop by itself as far as i remember, adding a stop means we intervene with it, if we can debug it on CI once it happens, to understand why the VM hangs might be better?

I haven't been unable to reproduce this issue locally but when I did catch it occurring on the cluster - I did not see anything in the vm console that indicated that the shutdown failed to complete. However the qemu process does seem to hang.
> 
> i wonder if the stop can hang as well by itself when the problem occurs, unless it is `stop --force`
> 
> even if at the end we take it, maybe worth to check what is a legitimate timeout, maybe 10 seconds is not enough for the VM to gracefully shutdown, we also need to make sure that it doesn't have consequences (well, i believe that there is a problem with some rpm/kernel that is hidden, and even if we dont add the stop might appear sometimes, for example if node get restarted, which we don't usually do, so in this aspect i guess we are fine)
> 
> maybe we can enable some logging, that if it hangs, then using the stop, we would able to see the log when it starts again during `cluster-up` ? also a console during the shutdown can show why it is hanged maybe
> 
> EDIT - having said all this, did you see it improves the failures ? if so, just worth to takes it and at least give more timeout, and maybe enable some logging so we would able to see on the next boot if some problem occurred maybe? (assuming the stop doesn't hang)

I ran this a number of times against the lanes and they all completed successfully. I think it is worth a try - it shouldn't cause any harm anyway.

dhiller (commented):
My only concern here is that writes to disk might be left out and we end up in a corrupt state, but since I guess then the node will not start up at all in the check-cluster-up we should be good. WDYT?

oshoval:
> My only concern here is that writes to disk might be left out and we end up in a corrupt state, but since I guess then the node will not start up at all in the check-cluster-up we should be good. WDYT?

I also not fond of this
for the reason you said, and also because it worth to ask qemu devs about it,
the provider should be stable, and such things might cause very hard to find noise
also what about measuring how many time before the stop we should let please ?
in the worst case we can try and see if there are noise, and if so try to keep in mind and revert,
problem is that noise can start only after a year and we wont remember.
on the other hang, i bet it is some rpm / kernel problem, that with time will be fixed.

I have an alternate solution until we fix it from root if desired,
there is phased provision,
we can use it even on CI presubmit / post submit,
assuming that the problem happens during k8s_provision.sh and not provision.sh
which btw can give hint what happens in case it is true
we will provision `linux` and then have a timeout and retry on the `k8s` phase
until it passed correctly without the manual stop and without hanging.

On one hand, it will make the provider bit safer, on the other hand doesn't solve the root cause,
but we are not using shut down / restart with kubevirtci so we are fine as long as the scope of the problem is those two.

Btw this approach also can make the post submit provision more stable (we can consider adding retry also to the `linux` phase if desired), but possible bit longer in case of retry, which at the end save time and resources and worth it.

Wdyt?

dhiller:
The problem seems to have gone from sometimes to always now.

Examples:
* https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1074/check-provision-k8s-1.28/1696685969467510784
* https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1072/check-provision-k8s-1.26-centos9/1696489914406801408

dhiller:
/approve

/hold 

awaiting discussion today about the issue

oshoval:
@brianmcarey 
Can you please sumup the offline debug sessions so we have it for future ref ?

brianmcarey:
This looks like a different issue from what we have seen previously as it happens every time and can be easily reproduced locally. 

Once the shutdown is issued there is a kernel panic that causes a restart rather than a shutdown. 

```
[  588.755094] systemd-shutdown[1]: All filesystems, swaps, loop devices, MD devices and DM devices detached.
[  588.768743] systemd-shutdown[1]: Syncing filesystems and block devices.
[  588.769211] systemd-shutdown[1]: Powering off.
[  588.771224] ACPI: PM: Preparing to enter system sleep state S5
[  588.771784] reboot: Power down
[  588.772047] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000000
[  588.772689] CPU: 0 PID: 1 Comm: systemd-shutdow Not tainted 5.14.0-361.el9.x86_64 #1
[  588.773305] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS 1.16.2-1.fc38 04/01/2014
[  588.773938] Call Trace:
[  588.774118]  <TASK>
[  588.774285]  dump_stack_lvl+0x34/0x48
[  588.774556]  panic+0xea/0x2e4
[  588.774775]  do_exit.cold+0x15/0x15
[  588.775034]  __do_sys_reboot+0x1e5/0x220
[  588.775336]  ? __rseq_handle_notify_resume+0x32/0x50
[  588.776593]  ? exit_to_user_mode_loop+0xd0/0x130
[  588.778584]  ? do_writev+0x6f/0x120
[  588.781724]  do_syscall_64+0x59/0x90
[  588.783697]  ? exc_page_fault+0x62/0x150
[  588.785017]  entry_SYSCALL_64_after_hwframe+0x72/0xdc
[  588.786410] RIP: 0033:0x7fd8e0d45427
[  588.788046] Code: 0b 00 f7 d8 64 89 02 b8 ff ff ff ff eb b8 0f 1f 44 00 00 f3 0f 1e fa 89 fa be 69 19 12 28 bf ad de e1 fe b8 a9 00 00 00 0f 05 <48> 3d 00 f0 ff ff 77 01 c3 48 8b 15 c9 49 0b 00 f7 d8 64 89 02 b8
[  588.791579] RSP: 002b:00007ffeb7f5ddc8 EFLAGS: 00000202 ORIG_RAX: 00000000000000a9
[  588.793047] RAX: ffffffffffffffda RBX: 0000000000000000 RCX: 00007fd8e0d45427
[  588.794497] RDX: 000000004321fedc RSI: 0000000028121969 RDI: 00000000fee1dead
[  588.795944] RBP: 00007ffeb7f5dfd0 R08: 0000000000000000 R09: 00007ffeb7f5d1c0
[  588.797207] R10: 00007ffeb7f5d380 R11: 0000000000000202 R12: 0000000000000000
[  588.798390] R13: 00007ffeb7f5de18 R14: 00007ffeb7f5de20 R15: 0000000000000000
[  588.799576]  </TASK>
[  588.800411] Kernel Offset: 0x1e800000 from 0xffffffff81000000 (relocation range: 0xffffffff80000000-0xffffffffbfffffff)
[  588.801804] Rebooting in 10 seconds..
[    0.000000] Linux version 5.14.0-361.el9.x86_64 (mockbuild@x86-05.stream.rdu2.redhat.com) (gcc (GCC) 11.4.1 20230605 (Red Hat 11.4.1-2), GNU ld version 2.35.2-42.el9) #1 SMP PREEMPT_DYNAMIC Thu Aug 24 13:40:45 UTC 2023
```

oshoval:
so i think we need to pin previous kernel and not just let it happen, unless it happens only for shutdown ?

brianmcarey:
> thanks so i think we need to pin previous kernel and not just let it happen, unless it happens only for shutdown ?

Pinning to the previous kernel would not be straight forward here - we would have to reboot into the older kernel - we can't just do the same as here  - https://github.com/kubevirt/kubevirtci/pull/1011/commits/2f1728cccec87449b67e094aa4da05d9c5b66a03

oshoval:
> > thanks so i think we need to pin previous kernel and not just let it happen, unless it happens only for shutdown ?
> 
> Pinning to the previous kernel would not be straight forward here - we would have to reboot into the older kernel - we can't just do the same as here - [2f1728c](https://github.com/kubevirt/kubevirtci/commit/2f1728cccec87449b67e094aa4da05d9c5b66a03)

np, also if it is only on shutdown we are fine,
but just curious why we cant ?

oshoval:
Now that https://github.com/kubevirt/kubevirtci/pull/1075 is merged,
maybe will be good to paste here for having the history, what you see / saw when it hangs due to the previous problem this PR meant to solve please?

oshoval:
i think the bug gone now right ?
seems it was some rpm / os
so in this case i think we don't need to merge this PR

brianmcarey:
/close

No longer seeing this problem.

PR #1051, opened 2023-07-16
Revert "Update github.com/docker/docker to 20.10.24+incompatible

oshoval:
This reverts commit 36bc2512bea5c8a819795bc35fd0c8d723aab52d.

It seems it breaks kubevirt / HCO
https://github.com/kubevirt/kubevirt/pull/10103

(need to validate)


oshoval:
this is not the bad PR, need to bisect
it might be even new CI podman or so

PR #1052, opened 2023-07-17
Improve err msg on cluster image pull

xpivarc:


oshoval:
tried to run this locally,
but the error comes from the daemon maybe ?
`Error response from daemon: invalid reference format`
same print with and without, unless i missed something

note that it can easily simulated locally, so it might help to debug

xpivarc:
> tried to run this locally, but the error comes from the daemon maybe ? `Error response from daemon: invalid reference format` same print with and without, unless i missed something
> 
> note that it can easily simulated locally, so it might help to debug

Yes, it comes from the daemon but the `gocli` is the client and therefore we should be able to decorate the err we get from it.
Weird that you don't see anything more than the error. This indicates it is a different command that is failing.

oshoval:
https://github.com/kubevirt/kubevirtci/pull/1054
should fix the bug

brianmcarey (commented):
I think this is still worth adding

/lgtm

brianmcarey (approved):
/approve

PR #1050, opened 2023-07-14
Update github.com/docker/docker to 20.10.24+incompatible in gocli

brianmcarey:
This should resolve a couple of security alerts previously raised by dependabot[1]

[1] https://github.com/kubevirt/kubevirtci/pull/995

/closes #995 

/cc @dhiller @xpivarc

PR #1053, opened 2023-07-17
Try to revert the commits since the latest working kubevirtci

oshoval:
Revert
* 36bc2512b (tag: 2307142054-36bc251) Update github.com/docker/docker to 20.10.24+incompatible in gocli (#1050)
* f5c9dda61 (tag: 2307140228-f5c9dda) vm based providers: Allow using custom images URL (#1042)

See https://github.com/kubevirt/kubevirt/pull/10103#issuecomment-1637084438

We believe it won't help since nothing major was changed,
but lets use it as a rebuild, and then think deeper


oshoval:
sounds for me not a good idea tbh

oshoval:
closing unless you insist

PR #511, opened 2020-12-30
(WIP) sriov, Add shellcheck linter

oshoval:
Signed-off-by: Or Shoval <oshoval@redhat.com>

PR #1054, opened 2023-07-17
vm based providers: Fix cluster-up script

oshoval:
bash `-n` must come either with quotes or "[[" (which eliminate the need to use quotes).

It will ensure bash evaluate it as
```
if [ -n "$KUBEVIRTCI_CONTAINER_REGISTRY" ]; then
   echo abc
fi
```
instead of 
```
if [ -n ]; then
   echo abc
fi
```
on the 2nd case it will print even if the variable is empty or not defined, which is not what we need.

Fixes "Error response from daemon: normalizing image: normalizing name for compat API: invalid reference format"

### DIFF
--- a/cluster-up/cluster/ephemeral-provider-common.sh
+++ b/cluster-up/cluster/ephemeral-provider-common.sh
@@ -102,15 +102,15 @@ function _add_common_params() {
     if [ -n "${KUBEVIRTCI_PROVISION_CHECK}" ]; then
         params=" --container-registry=quay.io --container-suffix=:latest $params"
     else
-        if [ -n $KUBEVIRTCI_CONTAINER_REGISTRY ]; then
+        if [[ -n ${KUBEVIRTCI_CONTAINER_REGISTRY} ]]; then
             params=" --container-registry=$KUBEVIRTCI_CONTAINER_REGISTRY $params"
         fi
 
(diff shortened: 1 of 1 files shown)


oshoval:
/cc @brianmcarey @dhiller 
can you please take a look ?
it fixes the problem of kubevirtci bump / latest versions

dhiller (approved):
/approve

dhiller:
> /approve

@oshoval thank you for the fix!

ormergi (commented):
Good catch!

ormergi (approved):

ormergi:
It would be nice if we had a basic shell-check gate to prevent such issues :nerd_face:

oshoval:
> It would be nice if we had a basic shell-check gate to prevent such issues nerd_face

started here
https://github.com/kubevirt/kubevirtci/pull/511

note that some of the changes it suggest cause flakiness (sometimes pass sometimes doesn't)
and some are not convenient, so the first time is a mess to ignore and see what is needed and what not
so it is tricky, but doable, all about priorities

brianmcarey (commented):
/lgtm

oshoval:
can we override the lanes ?
the other do pass, and it is the same script for all

oshoval:
@brianmcarey 
can you please override the lane? it test it several times already (stuck during provision)

PR #1057, opened 2023-07-18
test img, fedora: add iperf3 tool

maiqueb:
iperf3 is an open-source speed test and network performance measurement tool.

We plan on using this tool to assure that VM live-migration works (i.e. established TCP connections are not broken when live-migrating).

We cannot rely on ping / nc since each client request would just create a new TCP connection.

### DIFF
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -43,7 +43,7 @@ write_files:
         bind 'set enable-bracketed-paste off'
 runcmd:
   - sudo systemctl daemon-reload
-  - sudo dnf install -y kernel-modules-$(uname -r) qemu-guest-agent stress-ng dmidecode virt-what vm-dump-metrics tpm2-tools nc sg3_utils
+  - sudo dnf install -y kernel-modules-$(uname -r) qemu-guest-agent stress-ng dmidecode virt-what vm-dump-metrics tpm2-tools nc sg3_utils iperf3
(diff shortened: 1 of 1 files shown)


oshoval (approved):

brianmcarey (approved):
/approve

maiqueb:
is this image generation automated @oshoval ? 

From what I hear from @qinqon it is possible it isn't. 

The alpine one is from what I hear.

oshoval:
> > is this image generation automated @oshoval ?
> > From what I hear from @qinqon it is possible it isn't.
> > The alpine one is from what I hear.

don't remember tbh, the replacement used to be manual even if the image creation is automatic
time passed since last time i messed with it

ormergi:
> We cannot rely on ping / nc since each client request would just create a new TCP connection.

Isn't the connection that created by the client requests persisted? 
  I think its expected the connection to disconnect on migration, maybe iperf re-establish the connection under the hood and give the impression the connection stay live

maiqueb:
> > We cannot rely on ping / nc since each client request would just create a new TCP connection.
> 
> Isn't the connection that created by the client requests persisted? I think its expected the connection to disconnect on migration, maybe iperf re-establish the connection under the hood and give the impression the connection stay live

? No. If the other side doesn't close the connection, and the 5 tuple remains the same, you're good, the connection will remain.

brianmcarey:
@maiqueb  - here is the new image
`quay.io/kubevirtci/fedora-with-test-tooling:v20230718-9af0be4`

maiqueb:
thanks @brianmcarey !!!!

PR #1059, opened 2023-07-20
Add the new k8s-1.28 provider

brianmcarey:
Includes:
- Copy of previous k8s-1.27 provider
- Updates to 1.28
- Removal of AdvancedAuditing feature gate from kubeadm.conf[1]

[1] https://github.com/kubernetes/kubernetes/blob/fa88c0b7796170eeff5686ae1d7d0f2f3f0df5de/CHANGELOG/CHANGELOG-1.27.md?plain=1#L2243


dhiller (approved):

xpivarc:
> /approve
> 
> /hold since the check is not yet required.

@dhiller Do we need the job to be required?

dhiller:
> > /approve
> > /hold since the check is not yet required.
> 
> @dhiller Do we need the job to be required?

No, usually we make the check required after the provider is stable.

xpivarc:
> > > /approve
> > > /hold since the check is not yet required.
> > 
> > 
> > @dhiller Do we need the job to be required?
> 
> No, usually we make the check required after the provider is stable.

Can we remove the hold then?

brianmcarey:
> > > > /approve
> > > > /hold since the check is not yet required.
> > > 
> > > 
> > > @dhiller Do we need the job to be required?
> > 
> > 
> > No, usually we make the check required after the provider is stable.
> 
> Can we remove the hold then?

Better to keep the hold in case the 1.28 lane itself fails - this provider would end up merging even though it might not be 100%.

I can remove the hold when I am happy that the lane is passing and looks good.

dhiller:
> > > > > > /approve
> > > > > > /hold since the check is not yet required.
> > > > > 
> > > > > 
> > > > > @dhiller Do we need the job to be required?
> > > > 
> > > > 
> > > > No, usually we make the check required after the provider is stable.
> > > 
> > > 
> > > Can we remove the hold then?
> > 
> > Better to keep the hold in case the 1.28 lane itself fails - this provider would end up merging even though it might not be 100%.
> > 
> > I can remove the hold when I am happy that the lane is passing and looks good.
> 
> Thanks for the reply. I see your point. I was thinking that if anything would be wrong it would be better to have it in Kubevirt because most likely we would need to fix Kubevirt itself.

No, because of the check cluster up fails it is of no use to merge this. This my holding of the PR.

oshoval:
Thanks
Note that rules.yaml should be changed as well please
we can remove the need to do so once centos8 is gone from the providers
or even now if we simplify that it will always build both centos even if any provider is changed (waste some time)

oshoval:
I also see here double tag `* 21cdd093d (tag: 2307201839-21cdd09, tag: 2307201615-21cdd09) Add the new k8s-1.28 provider (#1059)`

Not sure why, maybe because there was a queued [postsubmit](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/publish-kubevirtci/1682097892468199424) job

it might cause problems sometimes, but i believe it works this time, because the first tag is detected and the bundle is healthy this time.

PR #1062, opened 2023-07-24
pman: Update rules.yaml with k8s-1.28

oshoval:
Once we remove centos8 folder we won't need to do as such anymore.


brianmcarey (approved):
/approve

PR #1064, opened 2023-07-26
Add kubelet to the replaced binaries for kubernetes pre releases

brianmcarey:
We should use the same version of kubelet as we do for kubeadm for prerelease providers.


xpivarc:
/lgtm
I did not verify this

brianmcarey (commented):
/approve

brianmcarey:
/override check-provision-k8s-1.26-centos9 

Changes made to 1.28 provider

PR #1065, opened 2023-07-27
Remove `--os-type` flag from customize-image.sh as it is deprecated

brianmcarey:
The following warning appears when building the test VM images:

"WARNING --os-type is deprecated and does nothing. Please stop using it"

/cc @enp0s3


brianmcarey (commented):
/approve

PR #1066, opened 2023-08-03
Enable unprivileged_userfaultfd on centos 9 to support post-copy migration

iholder101:
For more info, please look at
https://github.com/openshift/machine-config-operator/pull/3724

### DIFF
--- a/cluster-provision/k8s/1.26-centos9/k8s_provision.sh
+++ b/cluster-provision/k8s/1.26-centos9/k8s_provision.sh
@@ -13,6 +13,10 @@ cat <<EOT >/etc/sysconfig/kubelet
 KUBELET_EXTRA_ARGS=--cgroup-driver=systemd --runtime-cgroups=/systemd/system.slice  --fail-swap-on=false --kubelet-cgroups=/systemd/system.slice
 EOT
 
+# Enable userfaultfd for centos9 to support post-copy live migration.
+# For more info: https://github.com/openshift/machine-config-operator/pull/3724
+echo "vm.unprivileged_userfaultfd = 1" > /etc/sysctl.d/enable-userfaultfd.conf
+
 # Needed for kubernetes service routing and dns
 # https://github.com/kubernetes/kubernetes/issues/33798#issuecomment-250962627
 modprobe bridge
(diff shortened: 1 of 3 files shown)


cluster-provision/centos9/Dockerfile
```diff
@@ -32,3 +32,7 @@ RUN curl -L -o /initrd.img http://mirror.stream.centos.org/9-stream/BaseOS/x86_6
 RUN curl -L -o /vmlinuz http://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz
 
 COPY scripts/* /
+
+# Enable userfaultfd for non-root users to enable post-copy migration.
+# For more info, please look at: https://github.com/openshift/machine-config-operator/pull/3724
+RUN sysctl vm.unprivileged_userfaultfd=1
```
rmohr, line 38: I think you want to run this command in the actual guest, so you have to move it to the provision scripts (check for `sysctl`, we should have some commands already at the corresponding places.

iholder101: Thank you @rmohr.
The thing is that we probably want to run this only on centos 9, and I'm not sure what's the right place to put this

rmohr: you can probably check in `kubevirtci/cluster-provision/k8s/1.27/k8s_provision.sh` for the OS the script is executed on. Not sure what's the simples way but `lsb_release`, `/etc/centos-version`, ... would be options to explore.

rmohr: If we talks about a new kernel default, it may be better to check for the kernel verson though ...

iholder101: I thought it should be under `cluster-provision/centos9` somewhere, not under the k8s scripts, but in second thought 1.27 is always being provisioned with centos9.

Thanks @rmohr! let me reshape the PR

iholder101:
@rmohr Thanks again, PTAL

rmohr:
/hold

feel free to unhold when you think this is ready.

xpivarc (commented):
/lgtm
works for me

brianmcarey (commented):
/lgtm

/unhold

PR #1071, opened 2023-08-22
kind, vgpu: Remove old kind-1.25-vgpu

oshoval:
We are using kind-1.27-vgpu instead.


brianmcarey (approved):
/approve

thanks for the cleanup @oshoval

PR #1083, opened 2023-09-11
gocli: provision: tell qemu to avoid rebooting

jean-edouard:
For some reason, sometimes VM shutdown leads to a reboot, breaking the provisioning process. That would happen if for example the VM looks like it crashed instead of just shutdown. Telling qemu to never reboot should fix the problem, and avoid many provisionning failures.

### DIFF
--- a/cluster-provision/gocli/cmd/provision.go
+++ b/cluster-provision/gocli/cmd/provision.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
-	"strconv"
 	"strings"
 
 	"github.com/alessio/shellescape"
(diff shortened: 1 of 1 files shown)


dhiller:
@jean-edouard I don't get what this PR wants to achieve? Can you elaborate?

jean-edouard:
> @jean-edouard I don't get what this PR wants to achieve? Can you elaborate?

Yes, hi! I noticed that, both on the check-provision lanes and on my box, the provisioning script was flaky, often hanging on "waiting for the node to stop".
So I investigated, and found out that while we issued a "shutdown now" command, the VM rebooted instead, causing the VM to never stop.
Since I couldn't figure out why the VM decided to reboot instead of shutting down, I thought about telling qemu to never reboot instead.
And that completely fixed the provisioning script on my box.
However, the `check-provision` lanes still fail on this PR, at the same place and maybe even more often than usual, which has me quite confused...
This PR might need some tweaking, I'm not sure, but I do believe we need this fix, which is why I left it open.

oshoval:
Please see https://github.com/kubevirt/kubevirtci/pull/1049

oshoval:
i think the bug gone now right ?
seems it was some rpm / os
so in this case i think we don't need to merge this PR

oshoval:
fwiw
https://github.com/kubevirt/kubevirtci/pull/1049#issuecomment-1834306634

/cc @brianmcarey

PR #1085, opened 2023-09-26
centos9: Bump version

oshoval:
Previous doesnt exists anymore


brianmcarey (approved):
/approve

we should maybe look at automating this bump in the future.

PR #1092, opened 2023-11-01
Bump CentOS Stream 9 version to 20231030.0

brianmcarey:
The old version is no longer available which causes provisioning to fail[1]

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1091/check-provision-k8s-1.28/1719674070099824640

/cc @oshoval @dhiller @xpivarc


dhiller (approved):
/approve

PR #1101, opened 2023-12-07
Bump provisioning base image to fedora 39

brianmcarey:
/cc @dhiller


cluster-provision/centos9/Dockerfile
```diff
@@ -1,5 +1,5 @@
 
-FROM quay.io/kubevirtci/fedora@sha256:a5ae95e717f0e79b9f0d6b58a20666ebbbc4cbc96abf18d4ab7746e387f12f6b
+FROM quay.io/kubevirtci/fedora@sha256:e3a6087f62f288571db14defb7e0e10ad7fe6f973f567b0488d3aac5e927035a
```
dhiller, line 2: We don't use anything related to docker or containerd, right? Since there's an [issue with containerd](https://github.com/coreos/fedora-coreos-tracker/issues/1578) that is not yet fixed in F39.

dhiller (commented):
One question inline.

dhiller (approved):
/approve

PR #1102, opened 2023-12-08
Remove the local cluster provider as no longer used

brianmcarey:
/cc @xpivarc @dhiller


xpivarc (approved):
I support this. We do not use this provider as of now and we are certainly not maintaining it. The provider is not covered by CI (at least to my knowledge) and we do not configure this provider the same way we configure others. It seems reasonable to remove this as one can easily use this directly and still work with Kubevirt.

dhiller (approved):
/approve

/hold

Wait for two days for other objections, then let's remove the hold.

brianmcarey:
/unhold

I think this is ok to go in now.

PR #1103, opened 2023-12-11
Install nftables on nodes as required by some network tests

brianmcarey:
The install of podman was removed previously[1] as the workaround it was used for was no longer needed.

nftables was installed as a dependency of podman so now it is missing from the nodes and the sig-network lanes are failing[2]

[1] https://github.com/kubevirt/kubevirtci/commit/b60c664ad09b672350fe1377cdf3af1d097de76f
[2] https://github.com/kubevirt/kubevirt/pull/10858#issuecomment-1847283993

/cc @dhiller @EdDev @oshoval


oshoval (approved):
Thanks

EdDev (approved):
I am surprised it is not installed by default.

Thanks!

dhiller (commented):
/approve

PR #1104, opened 2023-12-13
Update fedora test VMs from version 35 to 39

brianmcarey:
Fedora 35 has been EOL for some time now - there is no point testing
against it. The e2e tests should make use of fedora 39 VMs


cluster-provision/images/vm-image-builder/example/image-url
```diff
@@ -1 +1 @@
-https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/32/Cloud/x86_64/images/Fedora-Cloud-Base-32-1.6.x86_64.qcow2
```
dhiller, line 1: :open_mouth: 32? :+1:

dhiller (commented):
/approve

PR #1113, opened 2024-01-10
Update centos stream version to 20240108.0 tag

brianmcarey:
From time to time old centos stream 9 images are removed and cleaned up,
this can cause publishes of kubevirtci[1] and presubmits to fail.

Updating to 20240108.0

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/publish-kubevirtci/1745004368974843904#1:build-log.txt%3A2263


oshoval:
i think we should pin a version, we can create a bump helper / automation
otherwise we are too prone for bugs

brianmcarey:
@oshoval in the interest of getting these lanes passing again - I have bumped the version to 20240108.0 and got rid of the change to latest.

oshoval (approved):
Thanks

dhiller (approved):
/approve

PR #1114, opened 2024-01-11
Update bump-centos-version.sh for centos stream 9

brianmcarey:
/cc @dhiller 

A periodic prowjob will call this script and create a PR.


oshoval (approved):
This script is good to have,
but the periodic isn't imo (commented on the dedicated PR)

Thanks

dhiller (approved):
/approve

PR #1117, opened 2024-01-12
document podman.socket prerequisite for cluster-up

### LINKED ISSUE #1116: Document use of podman.socket in make cluster-up
`make cluster-up` in the case where `podman` is being used, requires that the podman runtime socket be enabled.  

If a user is on a host that is enabled for podman but not docker, and they try just `make cluster-up` it will fail with an error that says `/run/podman/podman.sock not found` -- which, in the absence of any indication in the over-simplified `make cluster-up` instruction to enable the podman socket with `systemctl start podman.socket` requires that the user dig through the code and related documentation just to test whether or not kubevirt or kubevirt CDI works.   On the face of it, it would seem to be just plain broken.

cfilleke:
see issue #1116

This addresses issue #3052 in [kubevirt-cdi](https://github.com/kubevirt/containerized-data-importer) as well
see comments in [the related PR](https://github.com/kubevirt/containerized-data-importer/pull/3053)

The enablement of podman socket on the test host should be documented in cluster-up README.  
Without podman socket cluster-up fails 

Fixes #1116

we ran into this setting up an external CI/CD system for kubevirt CDI on s390x but it applies to all platforms where `podman` is in use over `docker`.  

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-up/README.md
+++ b/cluster-up/README.md
@@ -1,4 +1,22 @@
-# How to use cluster-up
+# cluster-up
+
+## Prerequisites: podman or docker
+
+cluster-up requires that either podman or docker be installed on the host.
+
+If podman is being used, it is also necessary to enable podman socket with:
+
+```
+sudo systemctl enable podman.socket
+sudo systemctl start podman.socket
+```
+
+for more information see:
+
+https://github.com/kubevirt/kubevirtci/blob/main/PODMAN.md
+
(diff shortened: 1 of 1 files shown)


brianmcarey (approved):
/ok-to-test

/approve

thanks @cfilleke !

PR #1120, opened 2024-01-19
Remove version check on running single stack sanity check

brianmcarey:
Cluster providers < 1.27 are no longer supported so this case is no longer required.

The single stack sanity check should run on all check-provisions.

/cc @oshoval


oshoval (approved):

brianmcarey (commented):
/approve

PR #1164, opened 2024-03-31
Convert the vm.sh script into a go binary

### LINKED ISSUE #1160: Convert the vm.sh script into a go binary
**Is your feature request related to a problem? Please describe**:
Create a go cli that replaces the [vm.sh](https://github.com/kubevirt/kubevirtci/blob/main/cluster-provision/centos9/scripts/vm.sh) script.

Kuruyia:
This adds a new Go CLI based on Cobra that will replace the [vm.sh](https://github.com/kubevirt/kubevirtci/blob/main/cluster-provision/centos9/scripts/vm.sh) script used when provisioning a cluster to set up the VMs that will contain the k8s cluster nodes. This is done in order to increase the maintainability of this script in the future.

Fixes #1160 

The CLI flags were kept as-is so this can be used as a drop-in replacement. 

I've already started to split the functions into multiple files so it is (hopefully) already a bit cleaner than the current shell script.

**Checklist**

**Release note**:

```release-note
NONE
```


alicefr:
@Kuruyia thanks for the work! Can you implement a couple of unit tests for the functions?

alicefr:
@dhiller @brianmcarey do you have any idea how we could test this?

cluster-provision/centos9/vmcli/cmd/root.go
```diff
@@ -0,0 +1,379 @@
...
+
+	return iptables.AppendUnique("nat", "PREROUTING", "-p", "tcp", "-i", "eth0", "-m", "tcp", "--dport", dportFlag, "-j", "DNAT", "--to-destination", toDestinationFlag)
+}
+
+// Add the iptables rule to forward ports from the container to the guest
+func addIptablesRules(iptables *iptables.IPTables, rootless bool, protocol string, ports []int) error {
+	if rootless {
+		for _, port := range ports {
+			// Add DNAT rule for rootless podman (traffic originating from loopback adapter)
+			toDestinationFlag := fmt.Sprintf("192.168.66.101:%d", port)
```
alicefr, line 109: Since you used the same constant value in mulitple places, I would define a const for it

cluster-provision/centos9/vmcli/cmd/qemu/img.go
```diff
@@ -0,0 +1,47 @@
...
+// Wrapper around the output of the "qemu-img info" command
+type QemuImgInfo struct {
+	// The virtual size of the disk image
+	VirtualSize uint64 `json:"virtual-size"`
+}
+
+// Creates a virtual disk image
+func CreateDisk(path string, format string, size uint64) error {
+	cmd := exec.Command("qemu-img", "create", "-f", format, path, strconv.FormatUint(size, 10))
+	cmd.Stderr = os.Stderr
```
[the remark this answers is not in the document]
Kuruyia: I didn't know if we wanted to show the output of those commands, but in hindsight more verbosity can't hurt.

cluster-provision/centos9/vmcli/cmd/qemu/system.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/google/uuid"
+)
+
+// The name of the QEMU main executable
+const qemuExec = "qemu-system-%s"
```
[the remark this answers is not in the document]
Kuruyia: This script is actually running in a [Fedora 39](https://github.com/kubevirt/kubevirtci/blob/d69cd4e5c4591a8275a176cb5aa9b0b2fb53bcd1/cluster-provision/centos9/Dockerfile#L2) container, so this is why it currently works.

Do we want to add a check for the current OS and use `qemu-kvm` if available instead?

Kuruyia:
> @Kuruyia thanks for the work! Can you implement a couple of unit tests for the functions?

Hey, thanks for your review!

I'll address your change requests and add the unit tests this week. I guess we're only testing the functions that don't have any side effects for now?

It'll be more involved to test functions that execute `qemu-img`, try to create a `/dev/kvm` device, ...

cluster-provision/centos9/vmcli/cmd/qemu/system.go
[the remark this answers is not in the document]
alicefr: Ah I see. I think we should get rid of fedora in the future. It is up to you, either you can a put a comment with a TODO and we can address this in a second PR or you can add it here.

alicefr:
> > @Kuruyia thanks for the work! Can you implement a couple of unit tests for the functions?
> 
> Hey, thanks for your review!
> 
> I'll address your change requests and add the unit tests this week. I guess we're only testing the functions that don't have any side effects for now?
> 
> It'll be more involved to test functions that execute `qemu-img`, try to create a `/dev/kvm` device, ...

Since we didn't have any tests for the bash script, I would just add some simple ones. 
For example, some that check the qemu command line. The rest is a bit difficult because this run multiple binaries, for example, the part that creates the the iptables, I would skip it for now.

Another place where you can add unit tests is CalcNextDisk`. In this case, you could for example declare type and a function type and overwrite it in the unit tests for `files, err := os.ReadDir(searchDir)`. Something like
```golang
type readDirFunc func(n int) ([]os.DirEntry, error)
type DiskUtil struct {
  readDirFunc readDir
}

func NewDiskUtil() *DiskUtil {
  return &DiskUtil{
      readDir = os.ReadDir
  }
}
```
and in the tests something like this: 
```golang
type mockDirEntry struct {
   name string
}

func newmockDirEntry(name string) os.DirEntry {
   return &mockDirEntry{name:name}
}

func (m *mockDirEntry) Name() string {
  return m.name
}
[...] // implement the rest of the function for DirEntry

func mockReadDir(n int) ([]os.DirEntry, error) {
    return osDirEntry{
       newmockDirEntry("disk0.qcow2"),
       newmockDirEntry("disk1.qcow2"),
    }, nil
}

diskUtil := &DiskUtil{
      readDir = mockReadDir
  }
// test
diskUtil.CalcNextDisk(...)
```

cluster-provision/centos9/vmcli/cmd/qemu/system.go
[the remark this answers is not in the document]
Kuruyia: I think it'd be better to address this in a second PR, so this one doesn't get too big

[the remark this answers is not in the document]
alicefr: Agreww

dhiller:
> @dhiller @brianmcarey do you have any idea how we could test this?


@Kuruyia thank you for your work here!

@alicefr I agree with what you suggested. Since we just didn't (or rather couldn't) unit-test anything here, let's try to increase coverage of testing where we can - anything is better than nothing, as it is currently.

lyarwood:
/cc

FWIW https://github.com/kubevirt/kubevirtci/pull/1174 is about to merge and will need to be covered here. I wasn't aware of this PR prior to starting this work sorry otherwise I would've extended it myself.

Kuruyia:
> /cc
> 
> FWIW #1174 is about to merge and will need to be covered here. I wasn't aware of this PR prior to starting this work sorry otherwise I would've extended it myself.

No problem, I was working on the unit tests today so I'll work on adding your modifications as well afterwards, thanks for letting me know.

Kuruyia:
Hey! Sorry for the delay, I just added the unit tests as you asked, and ported here the NUMA feature that was recently added to `vmcli.sh` as well.

Here's what I think is left to do:
- run those unit tests in the CI
- integrate this new CLI in place of the `vmcli.sh` shell script
- support RHEL and RHEL-based distributions (when starting QEMU)

As this PR is getting bigger, it might be better to offload those tasks to separate PRs imho.

alicefr:
> Hey! Sorry for the delay, I just added the unit tests as you asked, and ported here the NUMA feature that was recently added to `vmcli.sh` as well.
> 
> Here's what I think is left to do:
> 
> * run those unit tests in the CI
> * integrate this new CLI in place of the `vmcli.sh` shell script
> * support RHEL and RHEL-based distributions (when starting QEMU)
> 
> As this PR is getting bigger, it might be better to offload those tasks to separate PRs imho.

Yes, I think we could replace the bash script in a second PR. For example, we might need to think also how to transition form it to the golang binary. We might want to have a version with both to be able to switch to the old version if there are any problem. 
However, one thing that is necessary are the build of the binary and to copy it inside the provisioner image.

Kuruyia:
> Yes, I think we could replace the bash script in a second PR. For example, we might need to think also how to transition form it to the golang binary. We might want to have a version with both to be able to switch to the old version if there are any problem.
> However, one thing that is necessary are the build of the binary and to copy it inside the provisioner image.

Yup no problem, what do you think about adding a new stage at the beginning of [the centos9 Dockerfile](https://github.com/kubevirt/kubevirtci/blob/main/cluster-provision/centos9/Dockerfile) to build this binary, and then copy it in the current (single) stage?

cluster-provision/centos9/vmcli/cmd/qemu/img_test.go
```diff
@@ -0,0 +1,68 @@
+package qemu
+
+import "testing"
```
alicefr, line 3: We usually ginkgo in kubevirt as test framework

cluster-provision/centos9/Dockerfile
```diff
@@ -1,5 +1,16 @@
...
+FROM golang:1.22 AS vmcli-builder
 
-FROM quay.io/kubevirtci/fedora@sha256:e3a6087f62f288571db14defb7e0e10ad7fe6f973f567b0488d3aac5e927035a
+WORKDIR /app
+
+COPY vmcli/go.mod vmcli/go.sum ./
+RUN go mod download
+
+COPY vmcli/ ./
+RUN CGO_ENABLED=0 GOOS=linux go build -o /vmcli ./cmd/cli
```
lyarwood, line 10: nit - Can you write the binary out to something like `_bin/vmcli`?

cluster-provision/centos9/Dockerfile
```diff
@@ -1,5 +1,16 @@
+# Build the vmcli Go CLI
+FROM golang:1.22 AS vmcli-builder
 
-FROM quay.io/kubevirtci/fedora@sha256:e3a6087f62f288571db14defb7e0e10ad7fe6f973f567b0488d3aac5e927035a
+WORKDIR /app
+
+COPY vmcli/go.mod vmcli/go.sum ./
+RUN go mod download
```
lyarwood, line 7: Can we just vendor dependencies instead of this step?

cluster-provision/centos9/vmcli/cmd/root.go
```diff
@@ -0,0 +1,392 @@
...
+dockerize -wait tcp://192.168.66.1%02[1]d:22 -timeout 300s &>/dev/null
+ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no vagrant@192.168.66.1%02[1]d -i vagrant.key -p 22 -q $@
+`
+
+	// The path to the legacy provisioned disk image
+	provisionedDiskPath = "provisioned.qcow2"
+	// The path to the first primary disk image
+	firstDiskPath = "disk01.qcow2"
+	// The default size of the primary disk
+	defaultDiskSize = 50 * 1024 * 1024 * 1024 // 50GiB
```
lyarwood, line 34: nit - use k8s.io/apimachinery/pkg/api/resource ?

cluster-provision/centos9/vmcli/cmd/qemu/system.go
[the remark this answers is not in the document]
lyarwood: Why do we need to get rid of Fedora in this context?

lyarwood (commented):
This is looking great, some quick comments but this should be ready to come out of draft and possibly land shortly. 

Can you provide a [Makefile](https://github.com/kubevirt/kubevirtci/blob/4151e66316dfab1d8bf22f9c999c6620ddd09f7d/cluster-provision/gocli/Makefile) that we could then call from a [presubmit](https://github.com/kubevirt/project-infra/blob/c223f019fe6ce8ada54e6cb0a416883d21cc5457/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml#L160-L184) as we do for gocli?

cluster-provision/centos9/vmcli/cmd/qemu/system.go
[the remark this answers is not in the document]
alicefr: we don't need to. But we have converted everything to centos stream. It seems a bit inconsistent to me that we still keep fedora. Aynway, this can be a topic for another PR.

Kuruyia:
> This is looking great, some quick comments but this should be ready to come out of draft and possibly land shortly.
> 
> Can you provide a [Makefile](https://github.com/kubevirt/kubevirtci/blob/4151e66316dfab1d8bf22f9c999c6620ddd09f7d/cluster-provision/gocli/Makefile) that we could then call from a [presubmit](https://github.com/kubevirt/project-infra/blob/c223f019fe6ce8ada54e6cb0a416883d21cc5457/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml#L160-L184) as we do for gocli?

Do you want a Makefile that replaces the [`build.sh`](https://github.com/kubevirt/kubevirtci/blob/239678e4ccf7eac19748be3bb80def74c3956b6b/cluster-provision/centos9/build.sh) script?

alicefr:
@Kuruyia please squash the commits that changes the vmcli into a single one, for example, the ones that inline the errors.

alicefr:
@Kuruyia I think now what we need it is a plan to integrate this into main. 
/cc @dhiller @brianmcarey I will also bring this topic to the next community meeting

cluster-provision/centos9/vmcli/cmd/utils/disk.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Construct the name of the new disk
+	nextDiskPath := fmt.Sprintf("/disk%02d.qcow2", lastDiskNb+1)
+
+	if forcedNextDiskPath != "" {
+		nextDiskPath = forcedNextDiskPath
+	}
+
+	if lastDiskNb == 0 {
+		lastDiskPath = "box.qcow2"
```
alicefr, line 72: nit: can we already initialized the `lastDiskPath` direcly with this value or can it also be empty?

alicefr: From the code if nothing happen then lastDiskNb stays 0 and the lastDiskPath becomes "box.qcow2"

cluster-provision/centos9/vmcli/cmd/utils/disk_test.go
```diff
@@ -0,0 +1,78 @@
...
+				nextDisk, _, err := diskUtil.CalcNextDisk("/", "")
+				Expect(err).NotTo(HaveOccurred())
+				Expect(nextDisk).To(Equal("/disk01.qcow2"))
+			})
+
+			It("returns the default disk as the current one", func() {
+				_, currentDisk, err := diskUtil.CalcNextDisk("/", "")
+				Expect(err).NotTo(HaveOccurred())
+				Expect(currentDisk).To(Equal("box.qcow2"))
+			})
```
alicefr, line 36: Why do we need 2 different test to verify this? You can simply reduce to one and verify the 2 variables

cluster-provision/centos9/vmcli/cmd/utils/disk_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+			When("the next disk name is forced", func() {
+				const forcedDisk = "/forced.qcow2"
+
+				It("returns the forced disk", func() {
```
alicefr, line 41: I would incorporate this test with the previous one by using a DescribeTable

cluster-provision/centos9/vmcli/cmd/utils/env_test.go
```diff
@@ -0,0 +1,166 @@
...
+				nodeNb, err := envUtil.GetNodeNb()
+				Expect(err).NotTo(HaveOccurred())
+				Expect(nodeNb).To(Equal(19))
+			})
+		})
+	})
+
+	Describe("Waiting for the TAP interface", func() {
+		When("the interface does not exist", func() {
+			const tapNum = 1
```
alicefr, line 63: nit: just define it in the Describe so you can avoid to repeat the constant

cluster-provision/centos9/vmcli/cmd/utils/env_test.go
```diff
@@ -0,0 +1,166 @@
...
+			BeforeEach(func() {
+				createMockFileWithData(fs, utils.ContainerenvPath, "engine=\"podman-4.3.1\"\nrootless=0\n")
+			})
+
+			It("returns that the environment is not rootless", func() {
+				rootless, err := envUtil.IsRootless()
+				Expect(err).NotTo(HaveOccurred())
+				Expect(rootless).To(BeFalse())
+			})
+		})
```
alicefr, line 152: This looks like it could be reduce and merged together into a DescribeTable. You can pass a function as parameter and execute it based on the different entries

alicefr:
> > This is looking great, some quick comments but this should be ready to come out of draft and possibly land shortly.
> > Can you provide a [Makefile](https://github.com/kubevirt/kubevirtci/blob/4151e66316dfab1d8bf22f9c999c6620ddd09f7d/cluster-provision/gocli/Makefile) that we could then call from a [presubmit](https://github.com/kubevirt/project-infra/blob/c223f019fe6ce8ada54e6cb0a416883d21cc5457/github/ci/prow-deploy/files/jobs/kubevirt/kubevirtci/kubevirtci-presubmits.yaml#L160-L184) as we do for gocli?
> 
> Do you want a Makefile that replaces the [`build.sh`](https://github.com/kubevirt/kubevirtci/blob/239678e4ccf7eac19748be3bb80def74c3956b6b/cluster-provision/centos9/build.sh) script?

@Kuruyia I don't quite understand why you need the build.sh. AFAIU, you added an additional building step and it should be automatically executed when we build the centos stream 9 node.

@lyarwood there is a difference here between gocli and vmcli. We build gocli on the host and then copy the binary, while here vmcli is built into the container. IMO, the second approach is better and we should also do the same for gocli. It is rare, but building gocli depends on the go version installed on the host. It could happen that the go version is too old. 
Building the binary in the container guarantees that we don't have any golang version mismatch.

cluster-provision/centos9/vmcli/cmd/utils/disk.go
Kuruyia: Indeed, nice catch, my conversion from the shell script was a bit rough here

cluster-provision/centos9/vmcli/cmd/utils/disk_test.go
Kuruyia: I personally like to have more "atomic" tests, but yeah I'll merge both given that they are small

Kuruyia:
> @Kuruyia please squash the commits that changes the vmcli into a single one, for example, the ones that inline the errors.

I squashed all the commits for the vmcli into a single one, and have another commit for the modifications done on the Dockerfile. Please tell me if you want a more granular squash.

> @Kuruyia I don't quite understand why you need the build.sh. AFAIU, you added an additional building step and it should be automatically executed when we build the centos stream 9 node.

I don't really understand what targets are needed inside this Makefile. I assumed it would replace the `build.sh` script as the example @lyarwood provided shows the gocli Makefile used for building the container image.

cluster-provision/centos9/vmcli/cmd/utils/disk_test.go
alicefr: Avoiding tests duplication makes the code more compact and in general easier to fix and read, but there are exceptions of course.

Kuruyia:
I fixed the name of the base golang image, and tested cluster provisioning under Fedora 40 with Podman. It built successfully, so hopefully this time's the charm for the CI as well :)

/retest

alicefr:
hi @Kuruyia thanks for the persistence. I'm in favor of this change. 
Kind reminder to @xpivarc @brianmcarey

Kuruyia:
> hi @Kuruyia thanks for the persistence. I'm in favor of this change. Kind reminder to @xpivarc @brianmcarey

No problem.

It looks like the `check-up-kind-1.*-vgpu` tests failed, but looking at those test histories, they just tend to do that?
I see that the required `check-provision-*` tests also didn't run on the last commit (`0d8d7fa`). Could you just rerun those?

oshoval:
Thanks for your contribution.

> It looks like the `check-up-kind-1.*-vgpu` tests failed, but looking at those test histories, they just tend to do that? I see that the required `check-provision-*` tests also didn't run on the last commit (`0d8d7fa`). Could you just rerun those?

Centos9 folder which is the only changed folder doesnt affect kind-X providers.

Please consider that check-provision uses podman, but the post submit itself uses docker, worth to check it as well,
maybe for example by creating a draft PR that use docker to provision and run rehearse there.

PR #1193, opened 2024-05-20
kind-ovn: Introduce new kind-ovn provider

oshoval:
Provider allows running kubevirt with ovn-k8s tests on CI.

Fixes #

**Checklist**

**Release note**:

```release-note

```

### DIFF
--- a/cluster-up/cluster/kind-ovn/provider.sh
+++ b/cluster-up/cluster/kind-ovn/provider.sh
@@ -1,0 +1,88 @@
+#!/bin/bash -ex
+# This file is part of the KubeVirt project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2024 Red Hat, Inc.
+#
+
+KIND_VERSION=0.19.0
+export KIND_IMAGE=kindest/node
+export K8S_VERSION=v1.28.0@sha256:dad5a6238c5e41d7cac405fae3b5eda2ad1de6f1190fa8bfc64ff5bb86173213
+
+KIND_PATH=${KIND_PATH:-"${KUBEVIRTCI_CONFIG_PATH}/${KUBEVIRT_PROVIDER}/_kind"}
+CLUSTER_PATH=${CLUSTER_PATH:-"${KUBEVIRTCI_CONFIG_PATH}/${KUBEVIRT_PROVIDER}/_ovnk"}
+CLUSTER_NAME=${KUBEVIRT_PROVIDER}
+
+function calculate_mtu() {
+    overlay_overhead=58
+    current_mtu=$(cat /sys/class/net/$(ip route | grep "default via" | head -1 | awk '{print $5}')/mtu)
+    expr $current_mtu - $overlay_overhead
+}
+
+MTU=${MTU:-$(calculate_mtu)}
+
+PLATFORM=$(uname -m)
+case ${PLATFORM} in
+x86_64* | i?86_64* | amd64*)
+    ARCH="amd64"
+    ;;
+aarch64* | arm64*)
+    ARCH="arm64"
+    ;;
+*)
+    echo "invalid Arch, only support x86_64, aarch64"
+    exit 1
+    ;;
+esac
+
+function fetch_kind() {
+    mkdir -p $KIND_PATH
+    current_kind_version=$($KIND_PATH/kind --version |& awk '{print $3}')
+    if [[ $current_kind_version != $KIND_VERSION ]]; then
+        echo "Downloading kind v$KIND_VERSION"
+        curl -LSs https://github.com/kubernetes-sigs/kind/releases/download/v$KIND_VERSION/kind-linux-${ARCH} -o "$KIND_PATH/kind"
+        chmod +x "$KIND_PATH/kind"
+    fi
+    export PATH=$KIND_PATH:$PATH
+}
+
+function prepare_config() {
+    echo "STEP: Prepare provider config"
+    cat >$KUBEVIRTCI_CONFIG_PATH/$KUBEVIRT_PROVIDER/config-provider-$KUBEVIRT_PROVIDER.sh <<EOF
+master_ip=127.0.0.1
+kubeconfig=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
+kubectl=kubectl
+docker_prefix=127.0.0.1:5000/kubevirt
+manifest_docker_prefix=localhost:5000/kubevirt
+EOF
+}
+
+function up() {
+    cluster::install
+    fetch_kind
+    pushd $CLUSTER_PATH/contrib ; ./kind.sh --cluster-name $CLUSTER_NAME --multi-network-enable -mtu $MTU --local-kind-registry --enable-interconnect; popd
+    cp ~/$CLUSTER_NAME.conf "${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig"
+    prepare_config
+}
+
+function down() {
+    kind delete cluster --name $CLUSTER_NAME
+}
+
+function _kubectl() {
+    export KUBECONFIG=$(./cluster-up/kubeconfig.sh)
+    kubectl "$@"
+}
+
+source ${KUBEVIRTCI_PATH}cluster/kind-ovn/install-ovn.sh
--- a/cluster-up/cluster/kind-ovn/README.md
+++ b/cluster-up/cluster/kind-ovn/README.md
@@ -1,0 +1,29 @@
+# OVN K8S in a Kind cluster
+
+Provides a k8s cluster that runs using [KinD](https://github.com/kubernetes-sigs/kind)
+The cluster is completely ephemeral and is recreated on every cluster restart. The KubeVirt containers are built on the
+local machine and are then pushed to a registry which is exposed at
+`localhost:5000`.
+
+## Bringing the cluster up
+
+```bash
+export KUBEVIRT_PROVIDER=kind-ovn
+make cluster-up
+```
+
+## Bringing the cluster down
+
+```bash
+export KUBEVIRT_PROVIDER=kind-ovn
+make cluster-down
+```
+
+## FAQ
+
+In case the cluster deployment fails, you need to make sure you have enough watches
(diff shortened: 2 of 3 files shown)


cluster-up/cluster/kind-ovn/provider.sh
```diff
@@ -0,0 +1,92 @@
...
+
+function up() {
+    cluster::install
+    fetch_kind
+
+    if [[ ${CI} == "true" ]]; then
+        sysctl fs.inotify.max_user_watches=1048576
+        sysctl fs.inotify.max_user_instances=512
+        dnf install -y openssl
+    fi
```
oshoval, line 75: will be removed once we can use the updated legacy image
https://github.com/kubevirt/project-infra/pull/3431

oshoval:
Thanks for the help here @qinqon

cluster-up/cluster/kind-ovn/provider.sh
```diff
@@ -0,0 +1,91 @@
...
+#
+
+KIND_VERSION=0.19.0
+export KIND_IMAGE=kindest/node
+export K8S_VERSION=v1.28.0@sha256:dad5a6238c5e41d7cac405fae3b5eda2ad1de6f1190fa8bfc64ff5bb86173213
+
+KIND_PATH=${KIND_PATH:-"${KUBEVIRTCI_CONFIG_PATH}/${KUBEVIRT_PROVIDER}/_kind"}
+CLUSTER_PATH=${CLUSTER_PATH:-"${KUBEVIRTCI_CONFIG_PATH}/${KUBEVIRT_PROVIDER}/_ovnk"}
+
+MTU=${MTU:-"1392"}
```
qinqon, line 26: Let's try to calculate it taking the MTU of the interface from the default gw route

```bash
overlay_overhead=58
current_mtu=$(cat /sys/class/net/$(ip route | grep "default via" | head -1 | awk '{print $5}')/mtu)
overlay_mtu=$(expr $current_mtu - $overlay_overhead)
```

cluster-up/cluster/kind-ovn/provider.sh
```diff
@@ -0,0 +1,91 @@
...
+
+function up() {
+    cluster::install
+    fetch_kind
+
+    if [[ ${CI} == "true" ]]; then
+        sysctl fs.inotify.max_user_watches=1048576
+        sysctl fs.inotify.max_user_instances=512
+        dnf install -y openssl
+    fi
```
qinqon, line 75: I remember needing this at my laptop too, not only CI.

cluster-up/cluster/kind-ovn/provider.sh
```diff
@@ -0,0 +1,91 @@
...
+
+    pushd $CLUSTER_PATH/contrib ; ./kind.sh --multi-network-enable -mtu $MTU -lr --enable-interconnect; popd
```
qinqon, line 77: We should also specify a different `--cluster-name` so it do not collide with the default ovn-k cluster, maybe `--cluster-name=kubevirtci`

qinqon, line 77: Let's also use descriptive names here, so instead of `-lr` do `--local-registry`

cluster-up/cluster/kind-ovn/provider.sh
```diff
@@ -0,0 +1,91 @@
...
+    cp ~/ovn.conf "${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig"
+    prepare_config
+}
+
+function down() {
+    echo "Not supported yet by this provider"
```
qinqon, line 83: The implementation of this is `kind delete cluster --name $CLUSTER_NAME`

qinqon (changes requested):

cluster-up/cluster/kind-ovn/provider.sh
oshoval: but why ? there is no need, one can just export MTU in case he has smaller MTU
and it is rare
we can just add readme instead
and make the code simpler

cluster-up/cluster/kind-ovn/provider.sh
oshoval: user would just need to do it once manually imo, we can add a readme
(locally i didnt but this is just a specific machine)

cluster-up/cluster/kind-ovn/provider.sh
maiqueb: If something changes on the prow infra, your code will break, you won't know why, you'll have to debug, spend time on it, eventually figure out the MTU changed, and update one magic value by another magic value.

Quique's suggestion will work under those circumstances. 

> it is rare

? How so ? Why do you think this is strange ?

cluster-up/cluster/kind-ovn/provider.sh
maiqueb: Configuring the max user instances is something I have to do locally (I use podman).
Not sure about the max user watches.

cluster-up/cluster/kind-ovn/provider.sh
maiqueb: Good point; I'd go with the lane's name though.

cluster-up/cluster/kind-ovn/provider.sh
qinqon: > but why ? there is no need, one can just export MTU in case he has smaller MTU and it is rare we can just add readme instead and make the code simpler

@oshoval by default this should try to calculate it with something like the bash snippet and also give users a scape hatch in case the calculations are wrong.

cluster-up/cluster/kind-ovn/provider.sh
qinqon: > user would just need to do it once manually imo, we can add a readme, wdyt ? (locally i didnt but this is just a specific machine)
> 
> btw maybe worth without the watches on CI ? or you saw you must have them on CI ?

This is idempotent configuration, not big deal with doing it every time, let's just remove the CI env var and that's it.

oshoval: note this is not something we are doing on any other provider,
i mean if some operations like installing stuff, config (for example disable_ipv6=0) etc
we require the user to do it just once, i think it is wrong and might require root (maybe we need root anyhow)
but if you insist i will do that
note that soon we will have a CI image that have 3 lines already, so it will be required only locally
which is even worse

cluster-up/cluster/kind-ovn/provider.sh
oshoval: > ? How so ? Why do you think this is strange ?

based on the machines i saw, usually MTU is not that small
anyhow i will do that

cluster-up/cluster/kind-ovn/provider.sh
maiqueb: I really don't follow your argument @oshoval

oshoval:
trying something, please dont review it yet
trying the new image quay.io/kubevirtci/golang-legacy:v20240523-1a2c9b8
that should have the 3 lines (watches / opensssl) built in

cluster-up/cluster/kind-ovn/provider.sh
oshoval: ok tested with the new CI image `quay.io/kubevirtci/golang-legacy:v20240523-1a2c9b8`
we don't need on CI those lines anymore because now they are part of the image
```
        sysctl fs.inotify.max_user_watches=1048576
        sysctl fs.inotify.max_user_instances=512
        dnf install -y openssl
```
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/3426/rehearsal-pull-kubevirt-e2e-net-integration-tests/1793555893502087168

so now we don't need those on CI, lets talk about locally
until today, all the providers we created / have, don't install packages that are needed on the root level
(referring to openssl that was required), it is like if the provider need `wget` it won't install, the user need to.
and they also don't set configuration, for example disable_ipv6 must be 0, and it is the user responsibility to set it, the provider doesn't do it it each time

So please reconsider, do you still want to include those lines that now only required locally,
or to just add a readme that say what is needed (or even just add it on the PR desc) ?
fwiw we don't get into resolution in the readme of what should be dnf installed, and even about some of the configuration (such as disable_ipv6=0) we don't usually.

maiqueb: I really don't care. 

As I see it, the 2 options are:
- hand a command to the user that just works. downside is ?... I don't see one.
- hand a command to the user that depends on the configuration of their environment. If / once the commands fails on their environment, the user will need to understand what's happening / look for a readme, and understand the dependencies, then act upon them

If the kubevirtci way is to do the latter, by all means, apply that. I know what I would rather have as a user.

oshoval:
ready to review again please
addressed comments

using updated legacy image that set watches / install openssl already
bumped OVN k8s to include the e2e tests that were merged
testing the changes on CI for kubevirtci and kubevirt
https://github.com/kubevirt/project-infra/pull/3426
https://github.com/kubevirt/project-infra/pull/3430

note that on kubevirt we don't need `export DOCKER_PREFIX=localhost:5000` anymore,
as the provider writes the config file that kubevirt uses

oshoval:
rebased, and amend commit msg

brianmcarey (commented):
@oshoval can you add a short README with the steps for running it locally just so that we have it captured?

cluster-up/cluster/kind-ovn/provider.sh
```diff
@@ -0,0 +1,91 @@
...
+export K8S_VERSION=v1.28.0@sha256:dad5a6238c5e41d7cac405fae3b5eda2ad1de6f1190fa8bfc64ff5bb86173213
+
+KIND_PATH=${KIND_PATH:-"${KUBEVIRTCI_CONFIG_PATH}/${KUBEVIRT_PROVIDER}/_kind"}
+CLUSTER_PATH=${CLUSTER_PATH:-"${KUBEVIRTCI_CONFIG_PATH}/${KUBEVIRT_PROVIDER}/_ovnk"}
+CLUSTER_NAME=${KUBEVIRT_PROVIDER}
+
+function calculate_mtu() {
+    overlay_overhead=58
+    current_mtu=$(cat /sys/class/net/$(ip route | grep "default via" | head -1 | awk '{print $5}')/mtu)
+    echo $(expr $current_mtu - $overlay_overhead)
```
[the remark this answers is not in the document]
oshoval: we do need, else it tried to exec the value and fails
...provider.sh: line 30: 1442: command not found

oshoval:
addressed comments of @qinqon @brianmcarey 
thx

brianmcarey (commented):
Looks good to me - just one question inline.

cluster-up/cluster/kind-ovn/provider.sh
```diff
@@ -0,0 +1,91 @@
...
+}
+
+MTU=${MTU:-$(calculate_mtu)}
+
+PLATFORM=$(uname -m)
+case ${PLATFORM} in
+x86_64* | i?86_64* | amd64*)
+    ARCH="amd64"
+    ;;
+ppc64le)
```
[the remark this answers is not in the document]
oshoval: just carry over, will remove

oshoval:
addressed comments

brianmcarey (approved):
/approve

PR #1209, opened 2024-06-25
Library based ssh

aerosouund:
Replaces ssh.sh in run.go with golang library based ssh that goes directly to the node rather than the node container as a middle man

Fixes #

**Checklist**

**Release note**:

```release-note
Use library based ssh in run.go
```


aerosouund:
@xpivarc @brianmcarey @dhiller

cluster-provision/gocli/utils/ssh/ssh.go
```diff
@@ -0,0 +1,86 @@
...
+		log.Fatalf("Failed to create SSH session: %v", err)
+	}
+	defer session.Close()
+
+	var stderr bytes.Buffer
+	var stdout bytes.Buffer
+
+	session.Stdout = os.Stdout
+	session.Stderr = os.Stderr
+	if !stdOut {
```
xpivarc, line 76: In case we have `stdOut=false` we don't return anything, right? This should be fixed and we should always return something...

cluster-provision/gocli/cmd/run.go
```diff
@@ -576,15 +585,22 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
...
+		}
+
+		success, err = docker.Exec(cli, nodeContainer(prefix, nodeName), []string{"/bin/bash", "-c", "scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i vagrant.key " + fmt.Sprintf("-r /scripts vagrant@192.168.66.10%d:/home/vagrant/scripts", x+1)}, os.Stdout)
+		if err != nil {
+			return err
+		}
+
+		if !success {
+			return fmt.Errorf("Copying scripts directory for node %s failed", nodeName)
+		}
```
xpivarc, line 599: This seems to appear here from nowhere.  Please elaborate why/how...

cluster-provision/gocli/cmd/run.go
```diff
@@ -598,35 +614,30 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 			if err != nil {
 				return fmt.Errorf("parsing proxy settings for node %s failed", nodeName)
 			}
-			success, err = docker.Exec(cli, nodeContainer(prefix, nodeName), []string{"/bin/bash", "-c", fmt.Sprintf("cat <<EOF >/scripts/docker-proxy.sh %s", proxyConfig)}, os.Stdout)
-			if err != nil {
-				return fmt.Errorf("write failed for proxy provision script for node %s", nodeName)
+
+			if _, err = sshClient.JumpSSH(fmt.Sprintf("cat <<EOF > ~/scripts/docker-proxy.sh %s", proxyConfig), false, true); err != nil {
```
xpivarc, line 618: Should this be executed in context of the VM?

cluster-provision/gocli/utils/ssh/ssh.go
aerosouund: Not sure what you fully mean, in case stdOut is false we return the command output as the first variable in the function and stdErr as the second
otherwise we return stdout.String() but it doesn't matter since in any case the output you want is printed on os.Stdout. does this answer your question? if it does do you suggest any changes ?

cluster-provision/gocli/cmd/run.go
aerosouund: Yes, as the script on the vm is not modified properly, its a template. this script also doesn't exist on the gocli it exists only on the node container
so the other option is to run this on the node container then scp it again. it my opinion its better to just run this on the vm since the script is already scp'd in the initial scp

cluster-provision/gocli/cmd/run.go
aerosouund: if you think this comment is enough please resolve the conversation

cluster-provision/gocli/utils/ssh/ssh.go
aerosouund: as of now, no
all callers don't depend on the returned value since in this pr what is being replaced is script runs. also, previously before this pr the return value was never important. all callers did `docker.Exec` with `os.Stdout`. 
however in the context of my full rewrite PR some callers do, when turning bind vfio to bash for example:
`addr, err := o.sshClient.JumpSSH(o.sshPort, o.nodeIdx, "lspci -D -d "+o.pciID, true, false)`

in this context printing to stdout isn't all that important, since this value is important in the context of the script logic only

xpivarc: Thanks for the elaboration. I would suggest to drop the argument for now and introduce it later when you need.

cluster-provision/gocli/cmd/run.go
```diff
@@ -591,6 +604,20 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 			return err
 		}
 
+		// the scripts required for running the provider exist in the node container, in order to execute them directly on the node they must be copied to the node first
+		devNull, _ := os.Open(os.DevNull)
```
xpivarc, line 608: You are looking for `io.Discard`

cluster-provision/gocli/cmd/run.go
```diff
@@ -591,6 +604,20 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
...
+		success, err = docker.Exec(cli, nodeContainer(prefix, nodeName), []string{"/bin/bash", "-c", "scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i vagrant.key " + fmt.Sprintf("-r /scripts vagrant@192.168.66.10%d:/home/vagrant/scripts", x+1)}, devNull)
+		if err != nil {
+			return err
+		}
+
+		// move the scripts to the same location they were in in the nodecontainer to enable command abstraction
+		for _, cmd := range []string{"sudo mkdir /scripts", "sudo cp -r /home/vagrant/scripts/* /scripts"} {
```
xpivarc, line 615: Why can't we do this in previous command?

xpivarc (commented):
Few more questions/comment but looks good to me

cluster-provision/gocli/cmd/run.go
aerosouund: you mean `sudo mkdir /scripts && sudo cp -r /home/vagrant/scripts/* /scripts` ?

cluster-provision/gocli/utils/ssh/ssh.go
```diff
@@ -0,0 +1,88 @@
...
+	session.Stdout = os.Stdout
+	session.Stderr = os.Stderr
+
+	if len(cmd) > 0 {
+		firstCmdChar := cmd[0]
+		// indicates the command is a script or a script with params
+		if string(firstCmdChar) == "/" || string(firstCmdChar) == "-" {
+			cmd = "sudo /bin/bash " + cmd
+		}
+	}
```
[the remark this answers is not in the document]
aerosouund: i don't want to unnecessarily append `/bin/bash` to every command in case its not a script or a script with params in case any command might not work with bin bash for any reason.
even now if it will work for the current existing commands to just use bin bash i want to generalize the interface for later usage by other code or PRs

cluster-provision/gocli/cmd/run.go
xpivarc: I mean the previous command (scp) `-r /scripts vagrant@192.168.66.10%d:/home/vagrant/scripts`  moves the scripts to the VM and here we move them within the VM. Why can't we set the location right with the scp?

aerosouund: Because editing the root directory requires sudo privileges and the root user does not have an ssh key so you cant scp as him

cluster-provision/gocli/utils/ssh/ssh.go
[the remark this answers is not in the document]
xpivarc: @aerosouund This could be good to document on the interface.

xpivarc:
@dhiller @brianmcarey Can you help me merge this one (missing approver right on this repo :) )

dhiller (commented):
Kudos @aerosouund for tackling this immense task, great job overall!

I've added a couple of nits inline. The only real question I have is whether it's really necessary to add `stretchr/testify` where we are using Ginkgo already here (and everywhere else).

cluster-provision/gocli/cmd/run_test.go
```diff
@@ -0,0 +1,64 @@
+package cmd
```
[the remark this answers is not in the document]
aerosouund: There's no particular reason, i'm mostly just familiar with testify. I am very open to using ginkgo if it can solve my need to test for SSH command execution using a mock

cluster-provision/gocli/cmd/run.go
```diff
@@ -586,212 +605,178 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
...
+}
 
-	if cephEnabled {
-		nodeName := nodeNameFromIndex(1)
-		success, err := docker.Exec(cli, nodeContainer(prefix, nodeName), []string{
-			"/bin/bash",
-			"-c",
-			"ssh.sh sudo /bin/bash < /scripts/rook-ceph.sh",
-		}, os.Stdout)
+func provisionNode(sshClient sshutils.SSHClient, n *nodesconfig.NodeLinuxConfig) error {
```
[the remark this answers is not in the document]
aerosouund: Sure, In my opinion the config is just a config. you should declare a method on a struct when the struct is whats doing the work, for example `client.Apply` or `docker.Exec` as it shows to who's reading the code the relation between the two.

In subsequent PR's i will be introducing a `KubevirtProvider` type, which can implement this method and it will maintain this relationship i am talking about

if you think otherwise i'd love to hear your take

cluster-provision/gocli/cmd/cmd_suite_test.go
```diff
@@ -1,13 +1,15 @@
...
 import (
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/stretchr/testify/suite"
 )
 
 func TestCmd(t *testing.T) {
+	suite.Run(t, new(TestSuite))
```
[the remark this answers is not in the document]
xpivarc: Note: The code is not covered right now. So anything is better than nothing

cluster-provision/gocli/docker/adapter.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func NewDockerAdapter(cli *client.Client, nodeName string) *DockerAdapter {
+	return &DockerAdapter{
+		nodeName:     nodeName,
+		dockerClient: cli,
+	}
+}
+
+func (d *DockerAdapter) SSH(cmd string) error {
```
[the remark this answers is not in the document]
xpivarc: >Also there's refactoring potential, since I've seen it being used repeatedly inside a loop over a string slice, maybe it would make sense to change the func into variadic type

The first goal is to set abstraction and showcase how we can unit test this. I agree that there can be more done but we need to prioritize and go step by step

cluster-provision/gocli/cmd/run.go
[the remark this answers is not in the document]
dhiller: You could i.e. call the func `Apply` if it's a receiver method. This would then read

```
myNodeConfig.Apply(sshClient ...)
```

but yeah, we can leave that for later.

cluster-provision/gocli/cmd/cmd_suite_test.go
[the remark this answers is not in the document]
dhiller: Yes, test coverage is important - but adding another testing framework is opening doors for more confusion.

dhiller (commented):
Thank you for this work @aerosouund ! 

/approve

cluster-provision/gocli/cmd/cmd_suite_test.go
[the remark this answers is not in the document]
aerosouund: i have submitted the testing commit again using ginkgo this time

brianmcarey (commented):
/approve

thanks @aerosouund !

brianmcarey:
/override check-provision-alpine-with-test-tooling

Not related to these changes.

cluster-provision/gocli/cmd/run.go
```diff
@@ -869,48 +850,29 @@ func getDevicePCIID(pciAddress string) (string, error) {
...
-	}
-
-	success, err = docker.Exec(cli, node, []string{"/bin/bash", "-c", fmt.Sprintf("ssh.sh sudo mount -t tmpfs -o size=%s tmpfs %s", mountSize, etcdDataDir)}, os.Stdout)
-	if !success || err != nil {
-		return fmt.Errorf("create tmpfs mount '%s' for etcd data on node %s failed: %v", etcdDataDir, node, err)
+func prepareEtcdDataMount(sshClient libssh.Client, etcdDataDir string, mountSize string) error {
+	cmds := []string{
+		fmt.Sprintf("mkdir -p %s", etcdDataDir),
+		fmt.Sprintf("test -d %s", etcdDataDir),
+		fmt.Sprintf("mount -t tmpfs -o size=%s tmpfs %s", mountSize, etcdDataDir),
```
xpivarc, line 868: @aerosouund This needs to run with sudo, see failure in https://github.com/kubevirt/kubevirt/pull/12310

PR #1216, opened 2024-07-08
Fix missing whitespace in check script

Davo911:
`cluster-up/check.sh` fails, as the whitespace is missing in an if condition 

**Release note**:

```release-note
NONE
```


xpivarc (approved):

brianmcarey (approved):
/approve

PR #1217, opened 2024-07-11
Opts package

### LINKED ISSUE #1110: kubevirtci, cluster-up: move cluster spin up logic into gocli
We currently have the logic of spinning up a KubeVirtCI cluster inside the cluster-up folder of kubevirt/kubevirtci.

This is sourced into kubevirt/kubevirt. An update of kubevirtci updates the sourced folder. Other users of kubevirtci clone the repo on the fly.

Let's move all the logic of cluster-up into gocli, this way we only need to update the tag of the gocli image for a cluster update.

aerosouund:

Replace all bash script calls in the run command with the golang native opts package

Part of: #1110 

***Special Notes***
This branch is based off `library-based-ssh`. PR: #1209 
it introduces the same commits but after library based ssh gets merged it will only introduce commits of the opts package

**Checklist**

**Release note**:

```release-note
Move bash scripts to the gocli
```


aerosouund:
cc: @dhiller @brianmcarey @xpivarc @acardace

dhiller (commented):
Hey @aerosouund, this is only part of the review I promised - I'll take another look tomorrow. However, hope you find the feedback helpful already.

cluster-provision/gocli/opts/bind-vfio/bind-vfio_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/mock/gomock"
+	kubevirtcimocks "kubevirt.io/kubevirtci/cluster-provision/gocli/utils/mock"
+)
+
+func TestBindVfio(t *testing.T) {
```
dhiller, line 11: nit: use Ginkgo

cluster-provision/gocli/opts/cdi/cdi.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"embed"
+	"regexp"
+
+	k8s "kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/k8s"
+	"kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/libssh"
+)
+
+//go:embed manifests/*
+var f embed.FS
```
dhiller, line 12: prefer to use file embed

cluster-provision/gocli/opts/cdi/cdi_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/mock/gomock"
+	k8s "kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/k8s"
+	kubevirtcimocks "kubevirt.io/kubevirtci/cluster-provision/gocli/utils/mock"
+)
+
+func TestCdiOpt(t *testing.T) {
```
dhiller, line 12: Ginkgo

cluster-provision/gocli/opts/cnao/cnao.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"embed"
+
+	k8s "kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/k8s"
+	"kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/libssh"
+)
+
+//go:embed manifests/*
+var f embed.FS
```
dhiller, line 11: file embed

cluster-provision/gocli/opts/cnao/cnao_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/mock/gomock"
+	k8s "kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/k8s"
+	kubevirtcimocks "kubevirt.io/kubevirtci/cluster-provision/gocli/utils/mock"
+)
+
+func TestCnaoOpt(t *testing.T) {
```
dhiller, line 12: Ginkgo

cluster-provision/gocli/opts/docker-proxy/docker-proxy.go
```diff
@@ -0,0 +1,48 @@
...
+		sshClient: sc,
+	}
+}
+
+func (o *DockerProxyOpt) Exec() error {
+	override, err := f.ReadFile("conf/override.conf")
+	if err != nil {
+		return err
+	}
+	script := strings.ReplaceAll(string(override), "$PROXY", o.proxy)
```
dhiller, line 30: how about using text.template here instead?

cluster-provision/gocli/opts/aaq/aaq.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return err
+	}
+	if o.customVersion != "" {
+		pattern := `v[0-9]+\.[0-9]+\.[0-9]+(.*)?$`
+		regex, err := regexp.Compile(pattern)
+		if err != nil {
+			return err
+		}
+		operatorNewVersion := regex.ReplaceAllString(string(operator), o.customVersion)
```
dhiller, line 45: how about text.template?

cluster-provision/gocli/opts/istio/istio.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/sirupsen/logrus"
+	istiov1alpha1 "istio.io/operator/pkg/apis/istio/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	k8s "kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/k8s"
+	"kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/libssh"
+)
+
+//go:embed manifests/*
+var f embed.FS
```
dhiller, line 17: use file embed

cluster-provision/gocli/opts/istio/istio.go
```diff
@@ -0,0 +1,93 @@
...
+
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, operator)
+		if err != nil {
+			return err
+		}
+		if operator.Status != nil && operator.Status.Status == 3 {
+			break
+		}
+		logrus.Info("Istio operator didn't move to Healthy status, sleeping for 10 seconds")
+		time.Sleep(time.Second * 10)
```
dhiller, line 85: use exponential backoff

cluster-provision/gocli/opts/istio/istio_test.go
```diff
@@ -0,0 +1,22 @@
+package istio
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
dhiller, line 6: Gingko

cluster-provision/gocli/opts/ksm/ksm_test.go
```diff
@@ -0,0 +1,27 @@
+package ksm
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
dhiller, line 7: Ginkgo

cluster-provision/gocli/opts/multus/multus.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"embed"
+
+	k8s "kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/k8s"
+	"kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/libssh"
+)
+
+//go:embed manifests/*
+var f embed.FS
```
dhiller, line 11: single file embed

cluster-provision/gocli/opts/multus/multus_test.go
```diff
@@ -0,0 +1,21 @@
+package multus
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
dhiller, line 6: Ginkgo

cluster-provision/gocli/opts/node01/node01.go
```diff
@@ -0,0 +1,58 @@
...
+	cgroupv2, err := f.ReadFile("conf/00-cgroupv2.conf")
+	if err != nil {
+		return err
+	}
+	advAudit, err := f.ReadFile("conf/adv-audit.yaml")
+	if err != nil {
+		return err
+	}
+	cmds := []string{
+		`[ -f /home/vagrant/single_stack ] && export kubeadm_conf="/etc/kubernetes/kubeadm_ipv6.conf" && export cni_manifest="/provision/cni_ipv6.yaml" || { export kubeadm_conf="/etc/kubernetes/kubeadm.conf"; export cni_manifest="/provision/cni.yaml"; }`,
```
dhiller, line 33: maybe best to have kept the shell script via embed and used it as a template, and then send it to execute via ssh as a whole for now?

cluster-provision/gocli/opts/prometheus/prometheus.go
```diff
@@ -0,0 +1,133 @@
...
+func NewPrometheusOpt(c k8s.K8sDynamicClient, grafanaEnabled, alertmanagerEnabled bool) *PrometheusOpt {
+	return &PrometheusOpt{
+		grafanaEnabled:      grafanaEnabled,
+		alertmanagerEnabled: alertmanagerEnabled,
+		client:              c,
+	}
+}
+
+func (o *PrometheusOpt) Exec() error {
+	defaultManifests := []string{
```
dhiller, line 28: To avoid boilerplate here, would it make sense to write the manifests to a temp file system and apply them one by one, iterating over the manifests with [WalkDir](https://pkg.go.dev/io/fs@master#WalkDir)?

dhiller (commented):
2nd review - I mostly found nits, as before. :+1:

cluster-provision/gocli/pkg/k8s/k8s.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+}
+
+func (c *K8sDynamicClientImpl) Get(gvk schema.GroupVersionKind, name, ns string) (*unstructured.Unstructured, error) {
+	resourceClient, err := c.initResourceClientForGVKAndNamespace(gvk, ns)
+	if err != nil {
+		return nil, err
+	}
+
+	obj, err := resourceClient.Get(context.TODO(), name, v1.GetOptions{})
```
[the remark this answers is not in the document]
aerosouund: while i originally didn't plan for context to be used but the k8s client-go package uses it. so it can be a nice motivation if we later want to adopt context across the repo. for the time being i'll swap it to `context.Background`

cluster-provision/gocli/pkg/k8s/k8s.go
```diff
@@ -0,0 +1,222 @@
...
+		if err != nil {
+			return err
+		}
+		logrus.Infof("Object %v applied successfully", obj.GetName())
+	}
+
+	return nil
+}
+
+func (c *K8sDynamicClientImpl) createWithRetries(resourceClient dynamic.ResourceInterface, obj *unstructured.Unstructured) error {
```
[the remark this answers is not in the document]
aerosouund: I agree with the exponential backoff and differentiating between errors

The reason why i am retrying is because my observation when testing this is sometimes the api server becomes unavailable momentarily but eventually becomes stable. so instead of failing altogether i am retrying to see if there's an actual problem or no

cluster-provision/gocli/pkg/libssh/ssh.go
```diff
@@ -84,10 +98,119 @@ func (s *SSHClientImpl) Command(cmd string) error {
 			cmd = "sudo /bin/bash " + cmd
 		}
 	}
+	logrus.Infof("[node %d]: %s", s.nodeIdx, cmd)
 
 	err = session.Run(cmd)
 	if err != nil {
-		return fmt.Errorf("Failed to execute command: %v, %v", cmd, err)
+		if !stdOut {
+			err = fmt.Errorf(stderr.String())
```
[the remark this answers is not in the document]
aerosouund: we can, but the reason i did this is because its already printed to the output. so to avoid the user seeing the error twice

cluster-provision/gocli/opts/aaq/aaq.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+	"regexp"
+
+	"github.com/sirupsen/logrus"
+	k8s "kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/k8s"
+	"kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/libssh"
+)
+
+//go:embed manifests/*
+var f embed.FS
```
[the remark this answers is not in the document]
aerosouund: For the ones that use a small number of files thats alot better indeed.
But for one that has many files i think it would look a bit unappealing to have 10+ variables initialized at the start of the file. for example.. the kubevirt provider PR introduces a provision k8s opt. which depends on 21 files

what do you think ?

cluster-provision/gocli/opts/node01/node01.go
aerosouund: Thats fine, but i am trying as much as i can to not create any new bash scripts as this is what we wanted to depart from originally. it might be reasonable if its a long command like that

what do you think ?

cluster-provision/gocli/opts/prometheus/prometheus.go
aerosouund: Yeah that would be alot better indeed i will implement this

aerosouund:
@dhiller i have addressed most points, let me go over everything again in case i missed anything then i will ping you when i think its in a good shape :)

cluster-provision/gocli/opts/aaq/aaq.go
aerosouund: same as the comment for docker proxy

cluster-provision/gocli/opts/docker-proxy/docker-proxy.go
aerosouund: i think for what i want its too complicated
since i would have to export the field i want to replace then also create a resulting string builder and provide a pointer to it in template.Execute

unless there's a simpler way that i don't know about ?

aerosouund:
@dhiller 
all points addressed :)
please take a look again and let me know

p.s: i swapped the implementation of copy remote file to an implementation based on `github.com/bramvdbogaerde/go-scp`. The whole reason we needed to do what we did originally is because you couldn't scp as root the files that require that level of permission. but since the introduction of the root key opt this is no longer the case

cluster-provision/gocli/pkg/libssh/ssh.go
```diff
@@ -16,13 +21,18 @@ var sshKey []byte
 // is going to be passed. any leading ways to configure the script like /bin/bash or anything is left to the caller to account for as an implementation detail
 type Client interface {
 	Command(cmd string) error
+	CommandWithNoStdOut(cmd string) (string, error)
+	CopyRemoteFile(remotePath, localPath string) error
```
xpivarc, line 25: Could we replace the `localPath` with `io.Writer` ?

cluster-provision/gocli/pkg/libssh/ssh.go
```diff
@@ -52,23 +62,14 @@ func NewSSHClient(port uint16, idx int, root bool) (*SSHClientImpl, error) {
...
-	defer client.Close()
-
-	conn, err := client.Dial("tcp", fmt.Sprintf("192.168.66.10%d:22", s.nodeIdx))
-	if err != nil {
-		return fmt.Errorf("Error establishing connection to the next hop host: %s", err)
+	if s.client == nil {
+		err := s.initClient()
+		if err != nil {
+			return err
+		}
```
xpivarc, line 69: Why was this change made? I am not seeing any explanation.

cluster-provision/gocli/pkg/libssh/ssh.go
```diff
@@ -84,10 +85,122 @@ func (s *SSHClientImpl) Command(cmd string) error {
...
+	logrus.Infof("[node %d]: %s", s.nodeIdx, cmd)
 
 	err = session.Run(cmd)
 	if err != nil {
 		return fmt.Errorf("Failed to execute command: %v, %v", cmd, err)
 	}
 	return nil
 }
+
+func (s *SSHClientImpl) CommandWithNoStdOut(cmd string) (string, error) {
```
xpivarc, line 97: It seems we have duplicates now, can we somehow have common parts?

cluster-provision/gocli/pkg/libssh/ssh.go
xpivarc, line 25: nit: `remotePath` to `remotePathToCopy`

cluster-provision/gocli/pkg/libssh/ssh.go
```diff
@@ -16,13 +21,18 @@ var sshKey []byte
 // is going to be passed. any leading ways to configure the script like /bin/bash or anything is left to the caller to account for as an implementation detail
 type Client interface {
 	Command(cmd string) error
+	CommandWithNoStdOut(cmd string) (string, error)
+	CopyRemoteFile(remotePath, localPath string) error
+	SCP(destPath string, contents fs.File) error
```
xpivarc, line 26: Please use `io.Reader` instead of `fs.File`

xpivarc (commented):
Leaving partial review

cluster-provision/gocli/pkg/k8s/k8s.go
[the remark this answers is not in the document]
aerosouund: daniel had asked me the same

The reason why i am retrying is because my observation when testing this is sometimes the api server becomes unavailable momentarily but eventually becomes stable. so instead of failing altogether i am retrying to see if there's an actual problem or no

cluster-provision/gocli/pkg/libssh/ssh.go
aerosouund: we can
but my rationale behind this was to make this ssh client as similar as can be to the ssh commands. and these work with paths and directories and abstract the os internals. without this the responsibility of file creation is moved to the caller

cluster-provision/gocli/pkg/libssh/ssh.go
aerosouund: this code (the client instantiation) is duplicated around all ssh methods.
both Daniel and Antonio on an earlier review thought its better to reuse the connection. and i agree with this as well. as it avoids the overhead of connection establishment with every command

cluster-provision/gocli/utils/main.go
```diff
@@ -0,0 +1 @@
+package utils
```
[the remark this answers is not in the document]
aerosouund: there isn't anymore. see latest code

cluster-provision/gocli/pkg/libssh/ssh.go
xpivarc: I think the flexibility is better in this case.

cluster-provision/gocli/pkg/libssh/ssh.go
xpivarc: It would be great to make this change explicit by making it a separate commit

cluster-provision/gocli/pkg/libssh/ssh.go
aerosouund: @xpivarc 
there's a problem with this

```golang
func (s *SSHClientImpl) executeCommand(cmd string, outBuffer, errBuffer io.Writer) (string, error) {
	session, err := s.client.NewSession()
	if err != nil {
		return "", err
	}
	defer session.Close()

	session.Stdout = outBuffer
	session.Stderr = errBuffer

	if len(cmd) > 0 {
		firstCmdChar := cmd[0]
		// indicates the command is a script or a script with params
		if string(firstCmdChar) == "/" || string(firstCmdChar) == "-" {
			cmd = "sudo /bin/bash " + cmd
		}
	}
	logrus.Infof("[node %d]: %s", s.nodeIdx, cmd)

	err = session.Run(cmd)
	if err != nil {
		err = fmt.Errorf(errBuffer.String())
		return "", fmt.Errorf("Failed to execute command: %v, %v", cmd, err)
	}
	return outBuffer.String(), nil
}
```

if you try to create a method that can take an arbitrary io writer a problem arises because how you handle both cases is dependent on what the io writer actually is. a file, or a buffer. you can combat this by type assertion. but technically an io.Discard is a writer too, should you account for it ?
it gives the feeling that the method can can an io writer but in fact will only work an os file or a buffer


```golang
func (s *SSHClientImpl) executeCommand(cmd string, outWriter, errWriter io.Writer) (string, error) {
	session, err := s.client.NewSession()
	if err != nil {
		return "", err
	}
	defer session.Close()

	session.Stdout = outWriter
	session.Stderr = errWriter

	if len(cmd) > 0 {
		firstCmdChar := cmd[0]
		// indicates the command is a script or a script with params
		if string(firstCmdChar) == "/" || string(firstCmdChar) == "-" {
			cmd = "sudo /bin/bash " + cmd
		}
	}
	logrus.Infof("[node %d]: %s", s.nodeIdx, cmd)

	if _, ok := outWriter.(*os.File); ok {
		err = session.Run(cmd)
		if err != nil {
			return "", fmt.Errorf("Failed to execute command: %v, %v", cmd, err)
		}
		return "", nil
	}

	if outbuf, ok := outWriter.(*bytes.Buffer); ok {
		errbuf, ok := errWriter.(*bytes.Buffer)
		if !ok {
			return "", fmt.Errorf("Unable to cast error writer as a buffer")
		}
		err = session.Run(cmd)
		if err != nil {
			err = fmt.Errorf(errbuf.String())
			return "", fmt.Errorf("Failed to execute command: %v, %v", cmd, err)
		}
		return outbuf.String(), nil
	}
	return "", fmt.Errorf("Invalid writers passed, must use a buffer or an os file")
}
```

let me know your thoughts

xpivarc:
```
diff --git a/cluster-provision/gocli/pkg/libssh/ssh.go b/cluster-provision/gocli/pkg/libssh/ssh.go
index 16f35300..d7539fa2 100644
--- a/cluster-provision/gocli/pkg/libssh/ssh.go
+++ b/cluster-provision/gocli/pkg/libssh/ssh.go
@@ -64,6 +64,10 @@ func NewSSHClient(port uint16, idx int, root bool) (*SSHClientImpl, error) {
 // SSH performs two ssh connections, one to the forwarded port by dnsmasq to the local which is the ssh port of the control plane node
 // then a hop to the designated host where the command is desired to be ran
 func (s *SSHClientImpl) Command(cmd string) error {
+	return s.executeCommand(cmd, os.Stdout, os.Stderr)
+}
+
+func (s *SSHClientImpl) executeCommand(cmd string, outWriter, errWriter io.Writer) error {
 	if s.client == nil {
 		err := s.initClient()
 		if err != nil {
@@ -77,8 +81,8 @@ func (s *SSHClientImpl) Command(cmd string) error {
 	}
 	defer session.Close()
 
-	session.Stdout = os.Stdout
-	session.Stderr = os.Stderr
+	session.Stdout = outWriter
+	session.Stderr = errWriter
 
 	if len(cmd) > 0 {
 		firstCmdChar := cmd[0]
@@ -91,43 +95,20 @@ func (s *SSHClientImpl) Command(cmd string) error {
 
 	err = session.Run(cmd)
 	if err != nil {
-		return fmt.Errorf("Failed to execute command: %v, %v", cmd, err)
+		return fmt.Errorf("failed to execute command: %s", cmd)
 	}
 	return nil
 }
 
 func (s *SSHClientImpl) CommandWithNoStdOut(cmd string) (string, error) {
-	if s.client == nil {
-		err := s.initClient()
-		if err != nil {
-			return "", err
-		}
-	}
-
-	session, err := s.client.NewSession()
-	if err != nil {
-		return "", err
-	}
-	defer session.Close()
 	var stdout, stderr bytes.Buffer
-	session.Stdout = &stdout
-	session.Stderr = &stderr
 
-	if len(cmd) > 0 {
-		firstCmdChar := cmd[0]
-		// indicates the command is a script or a script with params
-		if string(firstCmdChar) == "/" || string(firstCmdChar) == "-" {
-			cmd = "sudo /bin/bash " + cmd
-		}
-	}
-	logrus.Infof("[node %d]: %s", s.nodeIdx, cmd)
-
-	err = session.Run(cmd)
+	err := s.executeCommand(cmd, &stdout, &stderr)
 	if err != nil {
-		err = fmt.Errorf(stderr.String())
-		return "", fmt.Errorf("Failed to execute command: %v, %v", cmd, err)
+		return "", fmt.Errorf("%w, %s", err, stderr.String())
 	}
 	return stdout.String(), nil
+
 }
 
 func (s *SSHClientImpl) SCP(fileName string, contents io.Reader) error {
```

xpivarc: How about this?

aerosouund: hmmm, yeah that makes alot of sense.
i will add that

cluster-provision/gocli/opts/aaq/testconfig.go
```diff
@@ -0,0 +1,9 @@
+package aaq
+
+import (
+	kubevirtcimocks "kubevirt.io/kubevirtci/cluster-provision/gocli/utils/mock"
+)
+
+func AddExpectCalls(sshClient *kubevirtcimocks.MockSSHClient) {
```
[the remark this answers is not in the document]
aerosouund: Yes, in `run_test.go`

cluster-provision/gocli/opts/rookceph/manifests/cluster-test.yaml
```diff
@@ -0,0 +1,50 @@
+#################################################################################################################
+# Define the settings for the rook-ceph cluster with common settings for a small test cluster.
```
brianmcarey, line 2: These manifests are just copied from the existing ones?

brianmcarey (commented):
Looks good @aerosouund - I just have a couple of questions inline.

xpivarc (commented):
Functionally this looks good to me. We could refactor even more but I think we need to balance here in order to converge

/approve
@brianmcarey I left lgtm for you

xpivarc:
/hold 
To not merge accidentally

cluster-provision/gocli/go.mod
```diff
@@ -1,65 +1,172 @@
...
+	k8s.io/kubectl => k8s.io/kubectl v0.28.4
+	k8s.io/kubelet => k8s.io/kubelet v0.28.4
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.28.4
+	k8s.io/metrics => k8s.io/metrics v0.28.4
+	k8s.io/node-api => k8s.io/node-api v0.28.4
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.28.4
+	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.28.4
+	k8s.io/sample-controller => k8s.io/sample-controller v0.28.4
+	kubevirt.io/qe-tools => kubevirt.io/qe-tools v0.1.8
+)
```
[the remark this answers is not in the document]
aerosouund: these were used by some of the dependencies introduced, particularly the kubernetes dependencies that we import to get their schemas. prometheus, rook, etc..
without the replaces when you fetch those k8s packages they fail because they try to download invalid versions of their own dependencies

cluster-provision/gocli/opts/istio/istio.go
```diff
@@ -0,0 +1,118 @@
...
+		return err
+	}
+
+	if err := o.client.Apply(obj); err != nil {
+		return err
+	}
+
+	cmds := []string{
+		"source /var/lib/kubevirtci/shared_vars.sh",
+		"PATH=/opt/istio-1.15.0/bin:$PATH istioctl --kubeconfig /etc/kubernetes/admin.conf --hub quay.io/kubevirtci operator init",
```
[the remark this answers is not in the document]
aerosouund: Perhaps yes, i wrote it like that because we don't take custom versions. and the version was hardcoded into the bash code as well

cluster-provision/gocli/opts/prometheus/manifests/README.md
```diff
@@ -0,0 +1,33 @@
+# Prometheus
+
```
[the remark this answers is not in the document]
aerosouund: They are copied yes
They are introduced in the opts package commit
since each opt depends on its manifests and conf to run. so it makes sense to add the code and the dependencies together

cluster-provision/gocli/opts/rookceph/manifests/cluster-test.yaml
aerosouund: yes
see the comment i left on the other conversation where you mentioned this

cluster-provision/gocli/opts/aaq/aaq.go
```diff
@@ -0,0 +1,68 @@
...
+				logrus.Info(err.Error())
+				continue
+			}
+			if err := o.client.Apply(obj); err != nil {
+				return fmt.Errorf("error applying manifest %s", err)
+			}
+		}
+	}
+
+	if err := o.sshClient.Command("kubectl --kubeconfig=/etc/kubernetes/admin.conf wait --for=condition=Ready pod --timeout=180s --all --namespace aaq"); err != nil {
```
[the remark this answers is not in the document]
aerosouund: yeah good idea

cluster-provision/gocli/opts/provision/provision.go
```diff
@@ -0,0 +1,53 @@
...
+		`echo 'ACTION=="add|change", SUBSYSTEM=="block", KERNEL=="vd[a-z]", ATTR{queue/rotational}="0"' > /etc/udev/rules.d/60-force-ssd-rotational.rules`,
+		"dnf install -y iproute-tc",
+		"mkdir -p /opt/istio-1.15.0/bin",
+		`curl "https://storage.googleapis.com/kubevirtci-istioctl-mirror/istio-1.15.0/bin/istioctl" -o "/opt/istio-1.15.0/bin/istioctl"`,
+		`chmod +x /opt/istio-1.15.0/bin/istioctl`,
+		"dnf install -y container-selinux",
+		"dnf install -y libseccomp-devel",
+		"dnf install -y centos-release-nfv-openvswitch",
+		"dnf install -y openvswitch2.16",
+		"dnf install -y NetworkManager NetworkManager-ovs NetworkManager-config-server",
```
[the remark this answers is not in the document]
aerosouund: This can be a nice follow up

cluster-provision/gocli/opts/multus/multus.go
```diff
@@ -0,0 +1,48 @@
...
+		if err != nil {
+			logrus.Info(err.Error())
+			continue
+		}
+		if err := o.client.Apply(obj); err != nil {
+			return fmt.Errorf("error applying manifest %s", err)
+		}
+	}
+	if err := o.sshClient.Command("kubectl --kubeconfig=/etc/kubernetes/admin.conf rollout status -n kube-system ds/kube-multus-ds --timeout=200s"); err != nil {
+		return err
```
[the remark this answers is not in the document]
aerosouund: I prefer to leave the flexibility to each opt in that area, because while there are similarities there are small differences for each, same for walking the directory.
some apply a single manifest, some iterate over a manifest to split it and apply whats in it, some walk a directory and apply a manifest.. etc

in general this can also be a nice follow up if we detect similarities that can be grouped

brianmcarey:
@xpivarc including service manifests like prometheus in gocli means that we loose some flexibility in deploying different versions of services to different cluster providers. I'm not sure how much of an issue that might be.

xpivarc:
> @xpivarc including service manifests like prometheus in gocli means that we loose some flexibility in deploying different versions of services to different cluster providers. I'm not sure how much of an issue that might be.

@brianmcarey 
Do we use it now? I am not aware... but I doubt that we loose it. The version can be templated and then it is only question of where do we encode what version is tight to what provider...

brianmcarey:
> > @xpivarc including service manifests like prometheus in gocli means that we loose some flexibility in deploying different versions of services to different cluster providers. I'm not sure how much of an issue that might be.
> 
> @brianmcarey Do we use it now? I am not aware... but I doubt that we loose it. The version can be templated and then it is only question of where do we encode what version is tight to what provider...

I have a case where I just want to update the monitoring stack on the k8s-1.31 provider and I don't think I will be able to achieve this with how the manifests are handled at the moment. Here is the PR in question - https://github.com/kubevirt/kubevirtci/pull/1228

Of course this update to the monitoring stack could probably be applied to all providers without a problem but there are performance jobs that would be sensitive to changes to this stack.

aerosouund:
@brianmcarey 
thats a valid point, can you tell me what are these performance jobs that might be sensitive to the change ?
i want to measure the impact and check if there is any way to avoid the degradation
or potentially explore building stack versioning into the PR

brianmcarey:
> @brianmcarey thats a valid point for sure, can you tell me what are these performance jobs that might be sensitive to the change ? i want to measure the impact and check if there is any way to avoid the degradation or potentially explore building stack versioning into the PR

Metric names are prone to change - which could cause these performance jobs to stop recording metrics of interest. 
Performance jobs like - https://testgrid.k8s.io/kubevirt-periodics#periodic-kubevirt-e2e-k8s-1.29-sig-performance

aerosouund:
@brianmcarey @xpivarc 
it seems that (according to my very limited understanding about that part of kubevirt) that we mainly depend on just two metrics, and these metrics are generated by kubevirt itself in those tests you linked

`kubevirt_vmi_phase_transition_time_from_creation_seconds_bucket`
`rest_client_requests_total`

this is according to this documentation snippet i found in `docs/perf-scale-benchmarks.md`:

1. The collected metrics are categorized into two buckets, performance and scale
   1. Performance Metrics: This tells users how KubeVirt stack is performing. Examples include
      `vmiCreationToRunningSecondsP50` and `vmiCreationToRunningSecondsP95`. This helps users understand how KubeVirt 
       performance evolved over the releases; depending on the user deployment, the numbers will vary, because a real
       production workload could use other KubeVirt extension points like the device plugins, custom scheduler, 
       different version of kubelet etc. These numbers are just a guidance for how the KubeVirt codebase is performing 
       with minimal VMIs, provided all other variables(hardware, kubernetes version, cluster-size etc) remain the same.
   1. Scalability metrics: This helps users understand the KubeVirt scaling behaviors. Examples include, 
      `PATCH-pods-count` for VMI, `PATCH-virtualmachineinstances-count` for VMI and `UPDATE-virtualmachineinstances-count`
      for VMI. These metrics are measured on the client side to understand the load generated to apiserver by the 
      KubeVirt stack. This will help users and developers understand the cost of new features going into KubeVirt. It
      will also make end-users aware about the most expensive calls coming from KubeVirt in their deployment and 
      potentially act on it.  
      
Based on this we can conclude that its safe to update the stack to any version we want, since we rely on kubevirt native metrics
But also to be safe i want to ask this to someone from sig performance, is there anyone you can refer me to ? 
If you also know any information contradicting to this let me know

xpivarc:
We also have a monitoring lane. I think that we could merge this if there is no multiple versions maintained now and we will add the support in the follow up. @brianmcarey WDYT?

aerosouund:
@brianmcarey 
this failure might have been because I updated the go version locally, I reverted it to 1.21. but locally my toolchain is 1.22.3
and I believe some packages actually require 1.22
but lets try to compile it as such now again

aerosouund:
its still trying to download 1.22 for some reason
```
CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "-X 'kubevirt.io/kubevirtci/cluster-provision/gocli/images.SUFFIX=:'" -o /home/prow/go/src/github.com/kubevirt/kubevirtci/cluster-provision/gocli/build/cli ./cmd/cli
go: downloading go1.22.3 (linux/amd64)
go: download go1.22.3: golang.org/toolchain@v0.0.1-go1.22.3.linux-amd64: verifying module: checksum database disabled by GOSUMDB=off
make: *** [Makefile:22: cli] Error 1
```

even when go mod is now
```
module kubevirt.io/kubevirtci/cluster-provision/gocli

go 1.21
```

aerosouund:
Detected the issue on a machine with no go 1.22
Some packages that require 1.22 explicitly in vendor/modules.txt made the go toolchain upgrade itself while downloading the dependencies
The dependent packages are :
`kubevirt.io/containerized-data-importer-api v1.59.1-0.20240610172909-253d5a6e7f95`
`k8s.io/pod-security-admission v0.22.0-beta.0.0.20240531013614-68e02f3c6735`
`k8s.io/client-go v0.30.1`
`k8s.io/apimachinery v0.30.3`
`k8s.io/api v0.30.3`
`istio.io/api v1.22.0`
`github.com/rook/rook/pkg/apis v0.0.0-20240801164516-9c98467f3c32`

What is the possibility of bumping the go version during build ? if there is none, i will try to downgrade these packages

aerosouund:
Not fully sure as well, it might just be that the gosumdb is off, on all machines i managed to compile this on it had
`GOSUMDB='sum.golang.org'`
is there any reason why its off in the first place ?

aerosouund:
@brianmcarey 
I have verified that not using gosumdb was the problem, i have added it to the makefile and fixed the merge conflict
from my perspective.. i don't know why we wouldn't use the gosumdb atleast in building the gocli. if there are any problems with this let me know

cluster-provision/gocli/opts/prometheus/manifests/prometheus/prometheus-prometheus.yaml
```diff
@@ -0,0 +1,49 @@
...
+  namespace: monitoring
+spec:
+  alerting:
+    alertmanagers:
+    - apiVersion: v2
+      name: alertmanager-main
+      namespace: monitoring
+      port: web
+  externalLabels: {}
+  image: quay.io/prometheus/prometheus:v2.26.0
```
brianmcarey, line 20: Can you update the monitoring manifests to match the latest versions that were updated in this PR - https://github.com/kubevirt/kubevirtci/pull/1228

brianmcarey (commented):
Looks good @aerosouund  - we just need to consolidate the monitoring manifests after a recent update here - https://github.com/kubevirt/kubevirtci/pull/1228

cluster-provision/k8s/fetch-latest-cdi.sh
```diff
@@ -69,7 +69,13 @@ cdi_version="${tag_name/#v/}"
         curl -O -s -L -f "$file"
     done
 
-    while IFS= read -r -d '' file; do
-        mv "$file" "${file/#\.\/cdi-/\.\/cdi-$cdi_version-}"
-    done < <(find . -type f -name 'cdi*.yaml' -print0)
+    # Renaming files to cr.yaml and operator.yaml
```
brianmcarey, line 72: What is the need for this renaming? 

Can we not just take the updated files directly?

aerosouund: we can, i just named them this way because the downloaded files are called cdi-cr, and cdi-operator
however cdi is implied

aerosouund:
@brianmcarey 
it still fails to build because of the CDI api's required version
`go: kubevirt.io/containerized-data-importer-api in vendor/modules.txt requires go >= 1.22.3 (running go 1.22.2)`

cluster-provision/k8s/fetch-latest-cdi.sh
brianmcarey: its probably better just to take the manifests as they are released saves us having to rename etc.

brianmcarey:
> @brianmcarey it still fails to build because of the CDI api's required version `go: kubevirt.io/containerized-data-importer-api in vendor/modules.txt requires go >= 1.22.3 (running go 1.22.2)`

Yes I think we need to update the go version in the bootstrap image that is running those jobs - let me take a look

aerosouund:
@brianmcarey 
updated the monitoring stack and the scripts

aerosouund:
@brianmcarey @xpivarc 
The previous issues were eliminated but there was another issue with ipv6 handling, this has now been dealt with
I also managed to perfectly replicate the CI setup on my testing environment to catch those bugs. i was previously testing using `make cluster-up` which obviously missed a few things

I also cleaned up the commits
Please let's rerun the tests

aerosouund:
@brianmcarey 
1.30 passed, can we please run the rest ?

aerosouund:
@brianmcarey 
Any other areas you think we should address before an lgtm ?

cluster-provision/gocli/go.mod
```diff
@@ -4,62 +4,168 @@ go 1.21
...
 	gopkg.in/inf.v0 v0.9.1 // indirect; gindirect
 )
+
+replace (
+	github.com/kubernetes-incubator/external-storage => github.com/libopenstorage/external-storage v0.20.4-openstorage-rc3
+	github.com/libopenstorage/secrets => github.com/rook/secrets v0.0.0-20240315053144-3195f6906937
+	github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.1-0.20240131200429-02120488a4c0
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20230406152840-ce21e3fe5da2
+	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20230324103026-3f1513df25e0
+	github.com/openshift/library-go => github.com/mhenriks/library-go v0.0.0-20230310153733-63d38b55bd5a
```
xpivarc, line 144: ` github.com/mhenriks/library-go` this seems wrong

aerosouund: @xpivarc 
didn't seem to be needed, removed it and the code still compiled

aerosouund:
@dhiller 
apologies, i pushed code to address [this](https://github.com/kubevirt/kubevirtci/pull/1217#discussion_r1726604538) and didn't see you requested a test

please rerequest

cluster-provision/gocli/pkg/libssh/ssh.go
```diff
@@ -276,3 +240,24 @@ func (s *SSHClientImpl) CopyRemoteFile(remotePathToCopy string, target io.Writer
 
 	return err
 }
+
+func (s *SSHClientImpl) initClient() error {
```
dhiller, line 221: This is not thread safe, you should ensure that the initialization is synchronized.

A naive implementation would look like this:

```suggestion
var initMutex sync.Mutex

func (s *SSHClientImpl) initClient() error {
	initMutex.Lock()
	defer initMutex.Unlock()
	if s.client != nil {
		return nil
	}
```

dhiller (changes requested):
Thank you @aerosouund , so far I have only found one issue wrt. ssh client initialization sync, comment inline.

cluster-provision/gocli/pkg/libssh/ssh.go
dhiller: Apologies, my bad - the initMutex should be part of the instance struct

aerosouund: @dhiller okay let me implement this and i will ping you for a retest shortly

cluster-provision/gocli/pkg/libssh/ssh.go
```diff
@@ -54,33 +57,24 @@ func NewSSHClient(port uint16, idx int, root bool) (*SSHClientImpl, error) {
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
 	}
 	return &SSHClientImpl{
-		config:  c,
-		sshPort: port,
-		nodeIdx: idx,
+		config:    c,
+		sshPort:   port,
+		initMutex: sync.Mutex{},
```
dhiller, line 62: nit: this is not required

dhiller (commented):
/approve

I had found just one nit, everything else looks good to me.

brianmcarey:
It looks like we are hitting issues with cluster-up since this PR was merged - https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/12677/pull-kubevirt-e2e-k8s-1.29-sig-storage/1828330116103540736#1:build-log.txt%3A327

https://github.com/kubevirt/kubevirt/pull/12677

aerosouund:
@brianmcarey checking

aerosouund:
@brianmcarey this is fixed in #1255 
please take a look at it

aerosouund:
we also added #1253 for quicker detection of such issues in the future, this script is meant to run as a CI lane on kubevirt CI

hack/bump-cdi.sh
```diff
@@ -41,11 +41,26 @@ if [ $# -gt 0 ]; then
...
     else
-        echo "Updated cdi manifests for $provider_dir"
+        echo "Updated cdi manifests for $cdi_dir"
     fi
-done < <(find ./cluster-provision/k8s -mindepth 1 -maxdepth 1 -type d -regex '^.*[0-9]\.[0-9]+.*$' -regextype 'posix-extended' -print0)
+
+    # Bump versions in the two files within the directory
+    for file in "$cdi_dir"/*; do
+        # Replace version numbers in the file (example: bumping from 1.0.0 to 1.1.0)
+        sed -i 's/1\.0\.0/1.1.0/g' "$file"
```
brianmcarey, line 59: This sed doesn't look to be correct - and the bump-cdi job is no longer working following this update - https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirtci-bump-cdi/1828968549343825920

aerosouund: checking

aerosouund:
@brianmcarey: https://github.com/kubevirt/kubevirtci/pull/1258

brianmcarey:
@aerosouund  - there looks to be some issue with the istio deployment that is being picked up by the network lanes on the bump kubevirtci PR - https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/12677/pull-kubevirt-e2e-k8s-1.31-sig-network/1829308375628779520

aerosouund:
checking

aerosouund:
FIX: https://github.com/kubevirt/kubevirtci/pull/1262

oshoval:
Seems `KUBVIRT_WITH_CNAO_SKIP_CONFIG` is missing ?
Can you please suggest if it was dropped or replaced 
We do need it from time to time and might need it soon again

aerosouund:
@oshoval 
No i think i missed it, i will add a fix for this pretty soon

oshoval:
Thanks
we do use this flag on https://github.com/kubevirt/macvtap-cni
and might need to bump kubevirtci there soon

/cc @EdDev fyi

oshoval:
> Thanks we do use this flag on https://github.com/kubevirt/macvtap-cni and might need to bump kubevirtci there soon
> 
> /cc @EdDev fyi

Any chance this can be prioritized please ?
@xpivarc @brianmcarey @dhiller 
It might block some major efforts soonish
Thanks

aerosouund:
Yeah no worries
I can sort this out in the next two days

aerosouund:
@oshoval 
FYI: https://github.com/kubevirt/kubevirtci/pull/1283

aerosouund:
@brianmcarey @dhiller @xpivarc 
I have a question regarding the CNI & Kubeadm manifests for different kubernetes VM providers (`cni.diff`, `cni_ipv6.diff`, `kubeadm.conf`, `kubeadm_ipv6.conf`).
Do we want to retain flexibility for each VM provider to deploy its preferred version of these ? or do we unify them across all providers ?

oshoval:
imo we should retain, because if there will be problems in just one provider and we combine, we will be in problem
it is also against the untangling effort that was done in the past

maybe if you combine but leave an option to inject custom it will be best
reminds a bit this idea https://github.com/kubevirt/kubevirtci/issues/1301

PR #1230, opened 2024-07-23
KV provider

aerosouund:

Create the KubevirtProvider type and make it the main entity for cluster actions

Part of: #1110 

***Special Notes***
This branch is based off `opts`. PR: #1217
it introduces the same commits but after opts gets merged it will only introduce commits of the kubevirt provider

**Checklist**

**Release note**:

```release-note
Create kubevirt provider, move run and provision logic to it
```


aerosouund:
cc: @acardace @xpivarc @dhiller @brianmcarey

aerosouund:
I think a reason for this flake, the one where the CI job keeps waiting on the node to stop is that the node shuts down before reporting the status back to the ssh agent which makes it not know whether this happened or no

this code previously looked like this
```golang
	_cmd(cli, nodeContainer(prefix, nodeName), "ssh.sh sudo shutdown now -h", "shutting down the node")
```
and now it looks like this (this PR introduces this change)

```golang
    if err = sshClient.Command("sudo shutdown now -h"); err != nil {
        return err
    }
```
which results in a golang error, since previously the error wasn't being handled

`failed to execute command: sudo shutdown now -h`

and so what i suggest we do is that we don't handle the error, i don't know if we also need to force the code to continue executing after a certain duration but this may not be necessary. since previously when ssh.sh would fail the gocli wouldn't get this failure and the CI run hangs. but since we now ssh directly its guaranteed that if there's an error you will get it

I will push a commit with this suggestion

cluster-provision/gocli/opts/k8sprovision/conf/crio-yum.repo
```diff
@@ -0,0 +1,6 @@
+[isv_kubernetes_addons_cri-o_stable_v1.28]
+name=CRI-O v1.28 (Stable) (rpm)
```
brianmcarey, line 2: I think we will need some version handling here - cri-o version should match the k8s version being deployed.

xpivarc:
> I think a reason for this flake, the one where the CI job keeps waiting on the node to stop is that the node shuts down before reporting the status back to the ssh agent which makes it not know whether this happened or no
> 
> this code previously looked like this
> 
> ```go
> _cmd(cli, nodeContainer(prefix, nodeName), "ssh.sh sudo shutdown now -h", "shutting down the node")
> ```
> 
> and now it looks like this (this PR introduces this change)
> 
> ```go
>     if err = sshClient.Command("sudo shutdown now -h"); err != nil {
>         return err
>     }
> ```
> 
> which results in a golang error, since previously the error wasn't being handled
> 
> `failed to execute command: sudo shutdown now -h`
>

Is this from the `_cmd`?

> and so what i suggest we do is that we don't handle the error, i don't know if we also need to force the code to continue executing after a certain duration but this may not be necessary. since previously when ssh.sh would fail the gocli wouldn't get this failure and the CI run hangs. but since we now ssh directly its guaranteed that if there's an error you will get it
> 
> I will push a commit with this suggestion

cluster-provision/gocli/opts/k8sprovision/conf/crio-yum.repo
aerosouund: yeah agreed, after inspecting the existing state of the code. some providers share crio version but occasionally one provider might deviate. will account for this

aerosouund:
@xpivarc 
> Is this from the `_cmd`?
yes, i believe so. since when it was executed through `ssh.sh` it would timeout
and if executed and expected to handle the error it crashes

aerosouund:
@brianmcarey @dhiller @xpivarc 
The PR is finally ready, highly appreciate a review when you have time

cluster-provision/gocli/pkg/libssh/ssh.go
```diff
@@ -224,9 +227,24 @@ func (s *SSHClientImpl) executeCommand(cmd string, outWriter, errWriter io.Write
...
-	client, err := ssh.Dial("tcp", net.JoinHostPort("127.0.0.1", fmt.Sprint(s.sshPort)), s.config)
+
+	var (
+		client *ssh.Client
+		err    error
+	)
+
+	operation := func() error {
+		client, err = ssh.Dial("tcp", net.JoinHostPort("127.0.0.1", fmt.Sprint(s.sshPort)), s.config)
+		return err
```
brianmcarey, line 238: Why did you remove the more verbose error message here?

cluster-provision/gocli/providers/base_provider_test.go
```diff
@@ -35,6 +37,18 @@ var _ = Describe("Node Provisioning", func() {
 		}
 
 		k8sClient = k8s.NewTestClient(reactors...)
+		kp = NewKubevirtProvider("k8s-1.30", "", &client.Client{}, []KubevirtProviderOption{
```
brianmcarey, line 40: Could we get k8s-1.30 dynamically from what folders are existing? This will fail when removing the k8s-1.30 provider?

cluster-provision/gocli/providers/provision.go
```diff
@@ -0,0 +1,214 @@
...
+		if err := kp.copyDirectory(ctx, kp.Docker, node.ID, dir, "/"); err != nil {
+			return fmt.Errorf("failed copying additional kernel arguments into the container: %v", err)
+		}
+	}
+
+	logrus.Infof("Commiting the node as %s", target)
+	_, err = kp.Docker.ContainerCommit(ctx, node.ID, container.CommitOptions{
+		Reference: target,
+		Comment:   "PROVISION SUCCEEDED",
+		Author:    "gocli",
```
brianmcarey, line 204: Should the author be "The KubeVirt Team"?

cluster-provision/gocli/pkg/libssh/ssh.go
aerosouund: That's fine, I can reintroduce it. I probably just removed it unintentionally

cluster-provision/gocli/providers/base_provider_test.go
aerosouund: No its unrelated to the folder and off the top of my mind it won't fail if we drop 1.30 altogether
the version validation logic has been moved to the run command, and since this is being defined in line then it wont be obstructed by this logic

`cluster-provision/gocli/cmd/run.go:143`
```golang
	allowedK8sVersions := []string{"k8s-1.28", "k8s-1.29", "k8s-1.30", "k8s-1.31"}
	var validVersion bool
	for _, v := range allowedK8sVersions {
		if k8sVersion == v {
			validVersion = true
		}
	}
	if !validVersion {
		return fmt.Errorf("Invalid k8s version passed, please use one of k8s-1.28, k8s-1.29, k8s-1.30 or k8s-1.31")
	}
```

cluster-provision/gocli/providers/provision.go
aerosouund: Yeah, we can change it

PR #1237, opened 2024-08-01
Remove usage of "filters" in OWNERS

xpivarc:
The documentation states:
"WARNING: The approve plugin does not currently respect filters. Until that is fixed, filters should only be used for the labels key (as shown in the above example)"

From now on we can only use OWNERS files per directory.


brianmcarey (approved):
/approve

PR #1290, opened 2024-10-07
fedora: Add tcpdump to tooling image

qinqon:
At some scenarios the egress link is broken during development and we have to inspect packages, on those scenarios we cannot install tcpdump and having it out of the box is a great help.


maiqueb (approved):

brianmcarey (commented):
/approve 

thanks @qinqon !

PR #1286, opened 2024-09-27
kind-1.28: add conformance.json

dhiller:
**What this PR does / why we need it**:

Adds `conformance.json` so that we can run conformance tests with kind provider.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

PR #1295, opened 2024-10-08
Revert "Automatic bump of CentOS Stream to latest (#1288)"

dhiller:

This reverts commit 16d40d2a563a61e866cc409c0b2cf1ef66f3c4a4.

Above commit has a CentOS bump that is likely a suspect of breaking SRIOV.

Fixes #

Noticed in https://github.com/kubevirt/kubevirtci/pull/1286#issuecomment-2399557829


brianmcarey (commented):
This is only used by the virtual providers - not used by the kind providers so it shouldn't effect the kind sriov lane. 

As we run the nightlies maybe something changed in kubevirt between the 1st and the 5th.

dhiller:
> This is only used by the virtual providers - not used by the kind providers so it shouldn't effect the kind sriov lane.
> 
> As we run the nightlies maybe something changed in kubevirt between the 1st and the 5th.

Good point - but I only see commits on the 1st, after that the next commit is from 8th

PR #1298, opened 2024-10-11
k8s: Make it work even if kubectl is not installed on the machine

### LINKED ISSUE #1293: kubectl.sh,bug: possibly broken - run `kubectl.sh` without having `kubectl` installed on your machine
Check whether using kubectl.sh without having kubectl installed on local machine still works.

> it used to be the case, a few years back, that you could run `kubectl.sh` without having `kubectl` installed on your machine, since the binary would be pulled from the running cluster. At some point that broke, so I resigned myself to installing `kubectl` on my machine despite not needing it outside of KubeVirt development. If at all feasible, going back to the previous status quo would be very appreciated. Maybe that's already the case and I simply haven't noticed?

_Originally posted by @andreabolognani in https://github.com/kubevirt/kubevirtci/issues/1277#issuecomment-2397408354_

andreabolognani:

Makes it possible to bring up a cluster on a machine where `kubectl` is not installed.

**Which issue(s) this PR fixes**:

Fixes #1293

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-up/cluster/k8s-provider-common.sh
+++ b/cluster-up/cluster/k8s-provider-common.sh
@@ -66,11 +66,6 @@ function up() {
 
     ${_cli} scp --prefix $provider_prefix /etc/kubernetes/admin.conf - >${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
 
-    # Set server and disable tls check
-    export KUBECONFIG=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
-    kubectl config set-cluster kubernetes --server="https://$(_main_ip):$(_port k8s)"
-    kubectl config set-cluster kubernetes --insecure-skip-tls-verify=true
-
     # Workaround https://github.com/containers/conmon/issues/315 by not dumping the file to stdout for the time being
     if [[ ${_cri_bin} = podman* ]]; then
         k8s_version=$(kubectl get node node01 --no-headers -o=custom-columns=VERSION:.status.nodeInfo.kubeletVersion)
@@ -80,6 +75,11 @@ function up() {
     fi
 
     chmod u+x ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl
+
+    # Set server and disable tls check
+    export KUBECONFIG=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
+    ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl config set-cluster kubernetes --server="https://$(_main_ip):$(_port k8s)"
+    ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl config set-cluster kubernetes --insecure-skip-tls-verify=true
 
     # Make sure that local config is correct
     prepare_config


andreabolognani:
I should note that I'm forcing the use of Docker by having

```
KUBEVIRT_CRI=docker
KUBEVIRTCI_RUNTIME=docker
```

in my environment, but the change should be logically sound regardless of runtime.

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -81,6 +76,11 @@ function up() {
 
     chmod u+x ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl
 
+    # Set server and disable tls check
+    export KUBECONFIG=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
+    ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl config set-cluster kubernetes --server="https://$(_main_ip):$(_port k8s)"
+    ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl config set-cluster kubernetes --insecure-skip-tls-verify=true
```
victortoso, line 82: We get `$kubectl`  variable from `cluster/ephemeral-provider-common.sh`, it is used in this file in other places too. Just changing that would be enough or you do need to move it down?

I've also noticed the `kubectl` usage on the `if podman` branch above, probably similar issue would happen there too

andreabolognani: Good catch.

There's definitely a chicken-and-egg problem here: in the case of podman, `${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl` (pointed to by the `$kubectl` variable) is downloaded for a specific k8s version in order to match the running cluster, but that in turn requires calling `kubectl`...

We could look into whether https://github.com/containers/conmon/issues/315, which is mentioned as the reason for this workaround existing, is still a problem in practice or whether we can just unify the code and avoid the additional call to `kubectl`.

Alternatively, it might be possible to obtain the value of `.status.nodeInfo.kubeletVersion` without needing `kubectl` by using `curl` against some HTTP endpoint? I'm not sure how feasible that would be.

Moving the calls down would still be necessary though, because if we left things as they are right now we would be calling `${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl` before having had the chance to create the file in question...

victortoso: So, on the podman topic, my guess is this has been fixed based on the comment... https://github.com/google/oss-fuzz/pull/4774#issuecomment-1308132677 

Looking at https://github.com/containers/conmon/issues/315#issuecomment-1027157864 seems that this was reproducible with conmon 2.1.0 but on the other comment it was fixed in 2.1.5

I _think_ we are doing some good tests with `podman` so if this breaks something, we will know quite fast too.

andreabolognani: The special handling for podman is gone since.

I've looked into using `$kubectl` but I'm not entirely sure whether doing so would be appropriate in this function. There's also a `_kubectl` function that looks like it might apply, but again I'm unclear on whether it would fit the context. So I ultimately decided to stick with the minimal, obviously correct change.

victortoso:
/cc @brianmcarey @nestoracunablanco @alicefr 

Is this PR acceptable or should we consider mandatory having kubectl installed?
I personally think it is fine and probably expected even, with our bazel build system, to use container's kubectl

nestoracunablanco:
@victortoso, I have some thoughts regarding consistency. Requiring users (and CI) to have kubectl installed on their machines doesn't seem too painful to me. However, having code that references kubectl from [${KUBEVIRT_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubectl](https://github.com/kubevirt/kubevirtci/blob/main/cluster-up/cluster/ephemeral-provider-common.sh#L255) creates inconsistency in some places.
I believe the proposed change is beneficial and could serve as a good first step in a refactoring process to ensure greater consistency across the codebase.

brianmcarey:
> @dhiller @brianmcarey can you approve this please?

Was this already handled by https://github.com/kubevirt/kubevirtci/pull/1437 ?

andreabolognani:
@brianmcarey indeed it was! Thanks for pointing that out.

I'll be honest: the fact that this PR has been sitting pretty much completely ignored for 8 months, despite multiple pings on my side both here and on the related issue, is disappointing. The changes I needed have made it in after all, which is good, but at least one other person had to be affected by the issue and independently figure out how to address it, and that's just wasteful.

Hopefully we ensure that first-time contributors specifically don't ever have this experience, because I bet it would discourage them from ever touching the project again if they did.

PR #1251, opened 2024-08-20
k8s-1.30: Enable feature-gate of DRA alpha feature

victortoso:
**What this PR does / why we need it**:

This is related to ongoing projects to test Dynamic Resource Allocation (DRA) in KubeVirt:

https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/

I'm enabling it in k8s-1.30 as per ongoing PCI passthrough driver developed by Sibasish Behera as part of his GSoC 2024 project, see:

https://github.com/kubevirt/dra-pci-driver

**Special notes for your reviewer**:

This patch mimics what was being done by hand in the dra-pci-driver, described [here](https://github.com/TheRealSibasishBehera/dra-pci-driver/blob/main/doc/KV_SETUP.md#step-2-enable-dynamic-resource-allocation)

I validated that it worked by [checking](https://github.com/TheRealSibasishBehera/dra-pci-driver/blob/main/doc/KV_SETUP.md#step-3-verification-for-resource-api) `resourceclasses` exist within this image.

When the Node is up:

```
toso@tapioca ~/s/k/kubevirt (main)> cluster-up/kubectl.sh get resourceclasses
selecting docker as container runtime
No resources found
```

After installing `dra-pci-driver`

```
toso@tapioca ~/s/k/kubevirt (main)> cluster-up/kubectl.sh get resourceclasses
selecting docker as container runtime
NAME              DRIVERNAME                 AGE
pci.kubevirt.io   pci.resource.kubevirt.io   4m23s
```

**Checklist**

This checklist is not enforcing, but it's a reminder of items that could be relevant to every PR.
Approvers are expected to review this list.

- [ ] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [ ] PR: The PR description is expressive enough and will help future contributors
- [ ] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [ ] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
[...]

PR #1299, opened 2024-10-11
Kubelet drop in

victortoso:
**What this PR does / why we need it**:

We are using deprecated kubelet's command line options which have config alternative options. Seems to be known since b89e9a64dc269b23d1f4e2d1318f46ebe11b5b21. This, combined with the goal of enabling k8s feature gates such as DRA in https://github.com/kubevirt/kubevirtci/pull/1251, the goal of this PR is to improve kubelet's config options and use [drop-in](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/#kubelet-conf-d) feature.

Commit logs clarify further the changes.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

This PR only focused on k8s-1.30 to validate the change. I can do here or in a new PR for k8s-1.31 too.
/cc @lyarwood 

**Checklist**

This checklist is not enforcing, but it's a reminder of items that could be relevant to every PR.
Approvers are expected to review this list.

- [ ] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [ ] PR: The PR description is expressive enough and will help future contributors
- [ ] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [ ] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [ ] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough). New features and bug fixes require at least on e2e test
- [ ] Documentation: A [user-guide update](https://github.com/kubevirt/user-guide/) was considered and is present (link) or not required. You want a user-guide update if it's a user facing feature / API change.
- [ ] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was considered

**Release note**:

```release-note
NONE
```

Issue #1264, opened 2024-09-03
[Feature Request]: Use KubevirtCI with a local customized Kubernetes repository


iholder101:
**Is your feature request related to a problem? Please describe**:
I want to bring up a KubevirtCI cluster with a local Kubernetes repository.
If this would be possible, it will allow:
1. Running KubevirtCI with a Kubernetes branch that hadn't been released yet, for example, running with Kubernetes' current `master` branch.
2. Running KubevirtCI with a customized development version of Kubernetes. In other words, it would allow changing Kubernetes' code and test the changes with Kubevirt before they've been merged or accepted by Kubernetes.

As a bonus, I think that the second use-case above will also be extremely valuable for Kubernetes developers in general, as currently there is no tool that let's you easily bring up a development Kubernetes cluster based of VMs.

**Describe the solution you'd like**:
Ideally, I picture the following workflow:
1. Clone Kubernetes repository.
2. (optionally) Make code changes to Kubernetes.
3. Use an environment variable to configure running KubevirtCI with a local Kubernetes repo, for example: `export USE_LOCAL_KUBERNETES_REPO=/path/to/my/k8s/repo`
4. Use `make cluster-up && make cluster-sync` as usual.

**Describe alternatives you've considered**:
It's possible to use `kind` for a development Kubernetes environment, then install Kubevirt on it. However, kind's node is based on containers which bring many limitations compared to VM based nodes.

**Additional context**:
As written above, this has potential to help Kubernetes developers in general, not only Kubevirt devs.

vladikr:
hmm... looks like the "local" provider was removed. https://github.com/kubevirt/kubevirtci/pull/1102
It did pretty much what you are asking for.
I used it extensively in the very distant past.

PR #1345, opened 2025-01-14
Add ability to enable Kubernetes feature gates and runtimeconfig in KubeVirt CI

Barakmor1:
There is a need to test new capabilities, such as [the ImageVolume POC](https://github.com/kubevirt/kubevirt/pull/13656/commits/5bed0f6b2962450f44a90bc7b070a0d4987cf3dd)  and [DRA](https://github.com/kubevirt/kubevirtci/pull/1251).

This change allows enabling/disabling Kubernetes feature gates and runtime config in the KubeVirt CI  . Users can now set the `K8S_FEATURE_GATES` environment variable before running `make cluster-up` to enable specific feature gates. For example: 

```
export K8S_FEATURE_GATES="<FeatureGate1>=true,<FeatureGate2>=false"
```
Users can also set runtime configuration by setting
the `K8S_API_RUNTIME_CONFIG` environment variable before
running `make cluster-up`.
    
For example:
 
```
export K8S_API_RUNTIME_CONFIG="RuntimeConfig=true"

```

This modifies the KubeController, KubeAPI, and Kubelet configurations files, restarts the components, and waits for them to be ready with the feature gates applied.

Fixes #

**Checklist**

**Release note**:

```release-note
Add ability to enable Kubernetes feature gates and runtime config in KubeVirt CI
```

### DIFF
--- a/KUBEVIRTCI_LOCAL_TESTING.md
+++ b/KUBEVIRTCI_LOCAL_TESTING.md
@@ -65,6 +65,18 @@ export KUBEVIRT_DEPLOY_PROMETHEUS_ALERTMANAGER=true
 export KUBEVIRT_DEPLOY_GRAFANA=true
 ```
 
+#### start cluster with Kubernetes feature gates
+To enable/disable Kubernetes feature gates, export the following variable before running `make cluster-up`:
+
+```bash
+export K8S_FEATURE_GATES="<FeatureGate1>=true,<FeatureGate2>=false"
+```
+
+for feature gates that require runtime config the following variable should be exported:
+```bash
+export K8S_API_RUNTIME_CONFIG="<RuntimeConfig>=true"
+```
+
 #### start cluster with swap enabled
 To enable swap, please also export the following variables before running `make cluster-up`:
 ```bash
--- a/cluster-provision/gocli/opts/k8scomponents/kubelet.go
+++ b/cluster-provision/gocli/opts/k8scomponents/kubelet.go
@@ -1,0 +1,78 @@
+package k8scomponents
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/libssh"
+)
+
+type kubeletComponent struct {
+	featureGates string
+	cmds         []string
+	sshClient    libssh.Client
+}
+
+func newKubeletComponent(sc libssh.Client, featureGates string) *kubeletComponent {
+	return &kubeletComponent{
+		sshClient:    sc,
+		featureGates: featureGates,
+	}
+}
+
+func (k *kubeletComponent) verifyComponent() error {
+	output, err := k.sshClient.CommandWithNoStdOut(searchFeatureGatesInFile("/var/lib/kubelet/config.yaml"))
+	if err != nil {
+		return err
+	}
+	fgsExist := len(output) > 0
+	if fgsExist {
+		return fmt.Errorf(featureGateExistInKubeletError)
+	}
+	return nil
+}
+
+func (k *kubeletComponent) prepareCommandsForConfiguration() error {
+	if k.featureGates != "" {
+		k.cmds = append(k.cmds, addFeatureGatesFieldToKubeletConfigCommand)
+		keyValuePairs := strings.Split(k.featureGates, ",")
+		var formattedFeatureGates []string
+		for _, pair := range keyValuePairs {
+			formattedFeatureGates = append(formattedFeatureGates, strings.Replace(pair, "=", ": ", 1))
+		}
+		for _, fg := range formattedFeatureGates {
+			k.cmds = append(k.cmds, addFeatureGatesToKubeletConfigCommand(fg))
+		}
+		k.cmds = append(k.cmds, kubeletRestartCommand)
+	}
+
+	return nil
+}
+
+func (k *kubeletComponent) runCommandsToConfigure() error {
+	return runCommands(k.cmds, k.sshClient)
+}
+
+func (k *kubeletComponent) waitForComponentReady() error {
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+	timeoutChannel := time.After(3 * time.Minute)
+	select {
+	case <-timeoutChannel:
+		return fmt.Errorf("timed out after 3 minutes waiting for node to be ready")
+	case <-ticker.C:
+		output, err := k.sshClient.CommandWithNoStdOut(getNodeReadyStatusCommand)
+		if err == nil && !strings.Contains(output, "false") {
+			return nil
+		}
+		if err != nil {
+			fmt.Printf("Modifying kubelet configuration, API server not responding yet, err: %v\n", err)
+		}
+	}
+	return nil
+}
+
+func (k *kubeletComponent) requiredOnlyForMaster() bool {
+	return false
+}
--- a/cluster-provision/gocli/cmd/nodesconfig/nodeconfig.go
+++ b/cluster-provision/gocli/cmd/nodesconfig/nodeconfig.go
@@ -4,6 +4,8 @@ package nodesconfig
 type NodeLinuxConfig struct {
 	NodeIdx         int
 	K8sVersion      string
+	FeatureGates    string
+	RuntimeConfig   string
 	FipsEnabled     bool
 	DockerProxy     string
 	EtcdInMemory    bool
--- a/cluster-provision/gocli/cmd/nodesconfig/opts.go
+++ b/cluster-provision/gocli/cmd/nodesconfig/opts.go
@@ -88,6 +88,18 @@ func WithSwap(swap bool) LinuxConfigFunc {
 	}
 }
 
+func WithFeatureGates(featureGates string) LinuxConfigFunc {
+	return func(n *NodeLinuxConfig) {
+		n.FeatureGates = featureGates
+	}
+}
+
+func WithRuntimeConfig(runtimeConfig string) LinuxConfigFunc {
+	return func(n *NodeLinuxConfig) {
+		n.RuntimeConfig = runtimeConfig
+	}
+}
+
 func WithKsmEnabled(ksmEnabled bool) LinuxConfigFunc {
 	return func(n *NodeLinuxConfig) {
 		n.KsmEnabled = ksmEnabled
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -38,6 +38,7 @@ import (
 	dockerproxy "kubevirt.io/kubevirtci/cluster-provision/gocli/opts/docker-proxy"
 	etcdinmemory "kubevirt.io/kubevirtci/cluster-provision/gocli/opts/etcd"
 	"kubevirt.io/kubevirtci/cluster-provision/gocli/opts/istio"
+	"kubevirt.io/kubevirtci/cluster-provision/gocli/opts/k8scomponents"
 	"kubevirt.io/kubevirtci/cluster-provision/gocli/opts/ksm"
 	"kubevirt.io/kubevirtci/cluster-provision/gocli/opts/multus"
 	"kubevirt.io/kubevirtci/cluster-provision/gocli/opts/nfscsi"
@@ -80,8 +81,8 @@ EOF
 	etcdDataDir         = "/var/lib/etcd"
 	nvmeDiskImagePrefix = "/nvme"
 	scsiDiskImagePrefix = "/scsi"
-	QEMU_DEVICE_S390X  = "virtio-net-ccw"
-	QEMU_DEVICE_X86_64 = "virtio-net-pci"
+	QEMU_DEVICE_S390X   = "virtio-net-ccw"
+	QEMU_DEVICE_X86_64  = "virtio-net-pci"
 )
 
 var soundcardPCIIDs = []string{"8086:2668", "8086:2415"}
@@ -112,6 +113,8 @@ func NewRunCommand() *cobra.Command {
 	run.Flags().UintP("secondary-nics", "", 0, "number of secondary nics to add")
 	run.Flags().String("qemu-args", "", "additional qemu args to pass through to the nodes")
 	run.Flags().String("kernel-args", "", "additional kernel args to pass through to the nodes")
+	run.Flags().String("feature-gates", "", "k8s feature gates to enable")
+	run.Flags().String("runtime-config", "", "k8s api runtime config")
 	run.Flags().BoolP("background", "b", true, "go to background after nodes are up")
 	run.Flags().Bool("random-ports", true, "expose all ports on random localhost ports")
 	run.Flags().Bool("slim", false, "use the slim flavor")
@@ -221,6 +224,16 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 		return err
 	}
 	kernelArgs, err := cmd.Flags().GetString("kernel-args")
+	if err != nil {
+		return err
+	}
+
+	featureGates, err := cmd.Flags().GetString("feature-gates")
+	if err != nil {
+		return err
+	}
+
+	runtimeConfig, err := cmd.Flags().GetString("runtime-config")
 	if err != nil {
 		return err
 	}
@@ -578,7 +591,7 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 	qemuArgs += " -serial pty"
 
 	var qemuNetDevice = getNetDeviceByArch()
-	
+
 	wg := sync.WaitGroup{}
 	wg.Add(int(nodes))
 	// start one vm after each other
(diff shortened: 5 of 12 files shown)


cluster-provision/gocli/opts/featuregate/featuregate.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (o *featureGateOpt) Exec() error {
+	output, err := o.sshClient.CommandWithNoStdOut(searchFeatureGatesInFile("/var/lib/kubelet/config.yaml"))
+	if err != nil {
+		return err
+	}
+	fgsExist := len(output) > 0
+	if fgsExist {
+		return fmt.Errorf(featureGateExistInKubeletError)
+	}
```
iholder101, line 51: Is it possible to add features in this case / override the existing ones?
However this shouldn't block this PR.

cluster-provision/gocli/opts/featuregate/featuregate.go
```diff
@@ -0,0 +1,150 @@
...
+	componentKubeAPIServer     component = "kube-apiserver"
+	componentKubeControllerMgr component = "kube-controller-manager"
+	componentKubeScheduler     component = "kube-scheduler"
+
+	searchFeatureGatesInFileFormat                    = "awk '/feature-gates/' %s"
+	getComponentCommandFormat                         = "kubectl --kubeconfig=/etc/kubernetes/admin.conf get pods -n kube-system -l component=%s -o jsonpath='{.items[0].spec.containers[*].command}'"
+	getComponentReadyContainersFormat                 = "kubectl --kubeconfig=/etc/kubernetes/admin.conf get pods -n kube-system -l component=%s -o=jsonpath='{.items[0].status.containerStatuses[*].ready}'"
+	addFeatureGatesToComponentCommandFormat           = `sudo sed -i 's/- %s/- %s\n    - %s/g' /etc/kubernetes/manifests/%s.yaml`
+	searchComponentsFilesCommand                      = "ls /etc/kubernetes/manifests"
+	addFeatureGatesFieldToKubeletConfigCommand        = "sudo /bin/su -c \"echo -e 'featureGates:' >> /var/lib/kubelet/config.yaml\""
```
iholder101, line 28: Using both `sudo` and `/bin/su` is redundant, can you please remove one of them?

cluster-provision/gocli/opts/featuregate/featuregate.go
```diff
@@ -0,0 +1,150 @@
...
+	addFeatureGatesToKubeletConfigCommandFormatFormat = `sudo sed -i 's/featureGates:/featureGates:\n  %s: true/g' /var/lib/kubelet/config.yaml`
```
iholder101, line 29: In Kubelet, the `--feature-gates` flag is defined as `a set of key=value pairs` [1].
I think that it's actually a good idea. The problem with your implementation is that it's impossible to turn a feature **off** which can be useful for beta features which are usually turned on by default.

[1] https://github.com/kubernetes/kubernetes/blob/f3cbd79db7f0c86a2d3602fdff6b174543d2cf1c/cmd/kubelet/app/options/options.go#L441

cluster-provision/gocli/opts/featuregate/featuregate.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return err
+	}
+
+	output, err = o.sshClient.CommandWithNoStdOut(searchComponentsFilesCommand)
+	if err != nil {
+		return err
+	}
+	onMasterNode := len(output) > 0
+	if onMasterNode { //should modify components files only on master node
```
iholder101, line 75: What about single node clusters? Is the single node a master?

iholder101 (commented):
Thank you @Barakmor1! Looks awesome!
I think this PR is really important as it will help us test alpha/beta features a lot of time in advance which is a huge advantage in terms of development.

This is a relatively shallow review as I'm not an expert in this area.

p.s. If in the future we'll also be able to support https://github.com/kubevirt/kubevirtci/issues/1264, combining these two capabilities could be absolutely great!

cluster-provision/gocli/opts/featuregate/featuregate.go
victortoso: I think we might need some custom code per feature gate too, as they might have different requirements for activation. For DRA in k8s 1.30 you can use [this diff ](https://github.com/kubevirt/kubevirtci/pull/1251/files) to know what was needed to be activated.

Barakmor1: > In Kubelet, the --feature-gates flag is defined as a set of key=value pairs [1].
I think that it's actually a good idea. The problem with your implementation is that it's impossible to turn a feature off which can be useful for beta features which are usually turned on by default.

> [1] https://github.com/kubernetes/kubernetes/blob/f3cbd79db7f0c86a2d3602fdff6b174543d2cf1c/cmd/kubelet/app/options/options.go#L441

Good catch! I used the wrong name for the variable. What I did was change the kubelet's config file and restart the kubelet. You can undo this by removing the feature gate and restarting the kubelet again. :)


> I think we might need some custom code per feature gate too, as they might have different requirements for activation. For DRA in k8s 1.30 you can use [this diff ](https://github.com/kubevirt/kubevirtci/pull/1251/files)to know what was needed to be activated.

From what I understand, we can apply the feature gate to the relevant components: for the kubelet, through the featureGate field in the config file, and for the API server, kube-controller, and kube-scheduler, using the `--feature-gate` flag. I believe this should work for any feature gate.

Am I missing anything?

cluster-provision/gocli/opts/featuregate/featuregate.go
Barakmor1: AFAIK it is

cluster-provision/gocli/opts/featuregate/featuregate.go
Barakmor1: By default, no feature gates should be enabled in KubeVirt CI. This is just a sanity check to ensure that no one is modifying the feature gates in earlier stages.

cluster-provision/gocli/opts/featuregate/featuregate.go
victortoso: > Am I missing anything?

for DRA in k8s 1.30, in the API server, it was also needed to change the `runtime-config`, [see](https://github.com/kubevirt/kubevirtci/pull/1251/files#diff-212e992eddef20aa4a45572b89f594cd18c540c82f505cd9e4d8545498f1eb66R23). I'm not sure how common that is, but I'd expect that newer features might request some modifications that are not covered by setting `featureGates:`

By all means, if just setting the field is enough for most of the cases, we can improve this latter.

My hope is that we could use [drop-in](https://github.com/kubevirt/kubevirtci/pull/1299) replacement over components when we want to change the defaults, be it adding or removing things...

victortoso: Just to be 100% sure... I've tested your PR on k8s 1.30 (provision + gocli)
Adding the feature gate bits worked as expected but we can't get `resourceclass` 

```
toso@tapioca ~/s/k/kubevirtci (k8sfg)> ./cluster-up/kubectl.sh get resourceclasses
error: the server doesn't have a resource type "resourceclasses"
```

But by changing the `runtime-config` it does.

```
toso@tapioca ~/s/k/kubevirtci (k8sfg)> ./cluster-up/kubectl.sh get resourceclasses
No resources found
```


(All the other feature gate checks that I did, just to make sure the bits were in)
```
toso@tapioca ~/s/k/kubevirtci (k8sfg)> export KUBEVIRTCI_PROVISION_CHECK=1 K8S_FEATURE_GATES="DynamicResourceAllocation" KUBEVIRTCI_GOCLI_CONTAINER=quay.io/kubevirtci/gocli:latest
...
toso@tapioca ~/s/k/kubevirtci (k8sfg)> cluster-up/ssh.sh node01 "sudo cat /etc/kubernetes/manifests/kube-apiserver.yaml" | grep -i feature
    - --feature-gates=DynamicResourceAllocation=true
toso@tapioca ~/s/k/kubevirtci (k8sfg)> cluster-up/ssh.sh node01 "sudo cat /etc/kubernetes/manifests/kube-apiserver.yaml" | grep -i runtime
      type: RuntimeDefault
toso@tapioca ~/s/k/kubevirtci (k8sfg)> clear^C
toso@tapioca ~/s/k/kubevirtci (k8sfg)> cluster-up/ssh.sh node01 "sudo cat /etc/kubernetes/manifests/kube-controller-manager.yaml" | grep -i feature
    - --feature-gates=DynamicResourceAllocation=true
toso@tapioca ~/s/k/kubevirtci (k8sfg)> cluster-up/ssh.sh node01 "sudo cat /etc/kubernetes/manifests/kube-scheduler.yaml" | grep -i feature
    - --feature-gates=DynamicResourceAllocation=true
toso@tapioca ~/s/k/kubevirtci (k8sfg)> cluster-up/ssh.sh node01 "sudo cat /var/lib/kubelet/config.yaml" | grep -i dyn
  DynamicResourceAllocation: true
```
```

Comment (Barakmor1) [reply cluster-provision/gocli/opts/featuregate/featuregate.go:29]: @victortoso Thanks for explaining i'll try to improve it :)

**Comment by jean-edouard** [comment]
> /cc @alicefr @victortoso @jean-edouard @brianmcarey

This PR makes sense to me. It's just surprising and unfortunate that we need that much code just to add a couple feature gates to k8s.
Let's just try and keep the code as simple as possible.
Please ping me once every review comment has been addressed and I'll approve.

**Comment by iholder101** [comment]
> > /cc @alicefr @victortoso @jean-edouard @brianmcarey
> 
> This PR makes sense to me. It's just surprising and unfortunate that we need that much code just to add a couple feature gates to k8s. Let's just try and keep the code as simple as possible. Please ping me once every review comment has been addressed and I'll approve.

@Barakmor1 Maybe there's a way to reduce code and make things a bit cleaner.
@victortoso used the concept of config file drop-ins [1] as part of https://github.com/kubevirt/kubevirtci/pull/1299 which can also be used here. Briefly, it allows having multiple config files that would be merged later by kubelet. So instead of changing the config, you can add a new config (e.g. `05-k8s-feature-gates.conf`) with the following contents:
```yaml
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
featureGates:
- NodeSwap
```

WDYT?

[1] https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/#kubelet-conf-d

Barakmor1:
@iholder101 @victortoso Please have another look

Barakmor1:
> > > /cc @alicefr @victortoso @jean-edouard @brianmcarey
> > 
> > 
> > This PR makes sense to me. It's just surprising and unfortunate that we need that much code just to add a couple feature gates to k8s. Let's just try and keep the code as simple as possible. Please ping me once every review comment has been addressed and I'll approve.
> 
> @Barakmor1 Maybe there's a way to reduce code and make things a bit cleaner. @victortoso used the concept of config file drop-ins [1] as part of #1299 which can also be used here. Briefly, it allows having multiple config files that would be merged later by kubelet. So instead of changing the config, you can add a new config (e.g. `05-k8s-feature-gates.conf`) with the following contents:
> 
> ```yaml
> apiVersion: kubelet.config.k8s.io/v1beta1
> kind: KubeletConfiguration
> featureGates:
> - NodeSwap
> ```
> 
> WDYT?
> 
> [1] https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/#kubelet-conf-d

Hey, unfortunately, I couldn’t find an easy way to update Kubernetes components with feature gates and flags after the cluster is already deployed. 

Most of the work in the PR is about changing the configurations of the kube-apiserver, kube-controller, kube-scheduler, and kubelet, and then making sure these changes are applied correctly. The final step is to ensure the components are ready again before moving on with other kubevirt CI tasks.

Using the drop-in would only change how the kubelet configuration is updated, but we’d still need to check that the changes are applied and verify that kubelet is ready. So, I don’t think this  would simplify the code. 
Please let me know if I missed something or if you have another idea, please let me know, and I’d be happy to try it out! :)

KUBEVIRTCI_LOCAL_TESTING.md
```diff
@@ -65,6 +65,18 @@ export KUBEVIRT_DEPLOY_PROMETHEUS_ALERTMANAGER=true
...
+#### start cluster with Kubernetes feature gates
+To enable/disable Kubernetes feature gates, export the following variable before running `make cluster-up`:
+
+```bash
+export K8S_FEATURE_GATES="<FeatureGate1>=true,<FeatureGate2>=false"
+```
+
+for feature gates that require runtime config the following variable should be exported:
+```bash
+export K8S_API_RUNTIME_CONFIG="<RuntimeConfig>=true"
```
victortoso, line 77: What does it mean, setting it to true?

cluster-provision/gocli/opts/featuregate/featuregate.go
Barakmor1: Done, now we can disable featureGates as well :)

cluster-provision/gocli/opts/featuregate/featuregate.go
Barakmor1: I added a new var for the runtime config, i also included this information in the PR description.

Please note that now we also need to add =true suffix to the vars.

victortoso (commented):
> Using the drop-in would only change how the kubelet configuration is updated, but we’d still need to restart kubelet and verify that kubelet is ready. So, I don’t think this would simplify the code.

Right, so the kubevirtci image is already created when we do something like `K8S_FEATURE_GATES=DRA make cluster-up` and we would need to restart kubelet after copying the config file, on each worker node.

Still, we do have some logic that does something like that, see [configure_cpu_manager](https://github.com/kubevirt/kubevirtci/blob/main/cluster-up/cluster/k8s-provider-common.sh#L42) introduced in 2981e7e8369609b983361b01df0354bbc8562409 

I'm not sure what is the best way forward. I think that, k8s FG requirements can change between k8s versions and if we want to help developers try those with kubevirtci, we should probably have to handle custom k8s configs and manage to update the cluster's components with it.

Drop-in would be great, but afaict not all k8s components have it and even kubelet support is beta. What about using kubeadm config?

KUBEVIRTCI_LOCAL_TESTING.md
Barakmor1: According to https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/ we can also disable runtime-config

victortoso: `--runtime-config <comma-separated 'key=value' pairs>`

Ah, got it.

cluster-provision/gocli/opts/k8scomponents/kubelet.go
```diff
@@ -0,0 +1,78 @@
...
+		output, err := k.sshClient.CommandWithNoStdOut(getNodeReadyStatusCommand)
+		if err == nil && !strings.Contains(output, "false") {
+			return nil
+		}
+		if err != nil {
+			fmt.Printf("Modifying kubelet configuration, API server not responding yet, err: %v\n", err)
+		}
+	}
+	return nil
+}
```
iholder101, line 74: Since this function is duplicated, you can extract it and get a function as a parameter that would do the actual validation. This way you'll be able to use it for all components.

iholder101 (commented):
Thanks @Barakmor1, looks great!
Can this PR be tested with a feature gate somehow to ensure it's working before we merge it? Perhaps create a cloned draft PR that enables an FG?

Barakmor1:
> Can this PR be tested with a feature gate somehow to ensure it's working before we merge it? Perhaps create a cloned draft PR that enables an FG?

This would only take affect if `K8S_FEATURE_GATES` is set

Barakmor1:
/reopen
closed by mistake

cluster-provision/gocli/opts/k8scomponents/kubelet.go
```diff
@@ -0,0 +1,78 @@
...
+		k.cmds = append(k.cmds, addFeatureGatesFieldToKubeletConfigCommand)
+		keyValuePairs := strings.Split(k.featureGates, ",")
+		var formattedFeatureGates []string
+		for _, pair := range keyValuePairs {
+			formattedFeatureGates = append(formattedFeatureGates, strings.Replace(pair, "=", ": ", 1))
+		}
+		for _, fg := range formattedFeatureGates {
+			k.cmds = append(k.cmds, addFeatureGatesToKubeletConfigCommand(fg))
+		}
+		k.cmds = append(k.cmds, kubeletRestartCommand)
```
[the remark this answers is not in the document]
Barakmor1: I agree, although I don't think this should block the merge, considering other places ([swap](https://github.com/kubevirt/kubevirtci/blob/3335da3f3c76d4fc926fd2a996fc5b4d5a086c93/cluster-provision/gocli/opts/swap/swap.go#L52) and [cpu-manager](https://github.com/kubevirt/kubevirtci/blob/main/cluster-up/cluster/k8s-provider-common.sh#L42)) where we are already modifying kubelet without the drop-in method. However, I see there is consensus on modifying kubelet with the drop-in beta feature, which is not yet supported until we add the --config flag to the kubeadm command during node provisioning which will be done in [this PR](https://github.com/kubevirt/kubevirtci/pull/1299) of @victortoso . 
I'll wait until it's merged and then use the drop-in method.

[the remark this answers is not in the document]
iholder101: > which is not yet supported until we add the --config flag to the kubeadm command during node provisioning which will be done in https://github.com/kubevirt/kubevirtci/pull/1299 of @victortoso .
> I'll wait until it's merged and then use the drop-in method.

Not sure there's a need to wait, you can just add the `--config-dir` in this PR.

> I agree, although I don't think this should block the PR

I agree. Although I'm for it, if you feel strongly about it I don't think it should block the PR.
Worst case we can do that as a follow-up.

victortoso:
@Barakmor1 https://github.com/kubevirt/kubevirtci/pull/1299 is now merged. Feel free to ping me when you get a new version avaiable for review

Barakmor1:
@brianmcarey It seems there's an issue with kubevirtCI i can't connect to provisioned cluster locally.

oshoval:
it changes kubeadm.conf right?
if so, maybe worth to consider supplying custom kubeadm.conf for cluster-up ?
this way you will have all the force you need
unless you need different variant of clusters and dont want the whole manifest for each

PR #1348, opened 2025-01-19
kind, check-cluster-up: Enable Kubevirt CPUManager FG when SR-IOV provider is tested

ormergi:
The "check-up-kind-sriov" is turned optional and **not gating** because it constantly failing, due to following tests failures:
```
SRIOV VMI connected to single SRIOV network should have cloud-init meta_data with tagged interface and aligned cpus to sriov interface numa node for VMIs with dedicatedCPUs
SRIOV VMI connected to single SRIOV network [test_id:3959]should create a virtual machine with sriov interface and dedicatedCPUs
```

The mentioned tests causing the lane to fail following removal of programmatic skips in kubevirt/kubevirt tests https://github.com/kubevirt/kubevirt/pull/13144, affecting the mentioned tests.
Previously the mentioned tests were skipped silently (bad) and now, following the programmatic skip removal, fail loudly.
The root cause for the failures (or previous skips) is tests depends on Kubevirt's CPUManager feature but its not enabled at all, see below notes section for more details *.

This PR fixes the lane by enabling Kubevirt's CPUManager features when SR-IOV provider is tested.

Fixes #

* When Kubevirt's CPUManager feature is on, it will label supporting nodes with cpumanager=ture label (done by the  heartbeat controller).
The failing tests, creates VMs with dedicated-CPUs option, Kubevirt will label such VM's virt-launcher pod with node-selector signifying `cpumanager=true` label.
The end result is the tested VMs fail to become ready on time due to impossible scheduling; VMs has `cpumanager=ture` node selector, but no node has `cpumanager=true` label.

**Checklist**

**Release note**:

```release-note
kind/check-cluster-up.sh enable Kubevirt's CPUManager feature when the SR-IOV provider is tested.
```

### DIFF
--- a/cluster-up/cluster/kind/check-cluster-up.sh
+++ b/cluster-up/cluster/kind/check-cluster-up.sh
@@ -69,6 +69,11 @@ export CRI_BIN=${CRI_BIN:-$(detect_cri)}
     fi
     ${kubectl} wait -n kubevirt kv kubevirt --for condition=Available --timeout 15m
 
+    if [[ "$KUBEVIRT_PROVIDER" =~ "sriov" ]]; then
+      # Some SR-IOV tests require Kubevirt CPUManager feature
+      ${kubectl} patch kubevirts -n kubevirt kubevirt --type=json -p='[{"op": "replace", "path": "/spec/configuration/developerConfiguration/featureGates","value": ["CPUManager"]}]'
+    fi
+
     echo "Run latest nighly build Kubevirt conformance tests"
     kubevirt_plugin="--plugin ${nightly_build_base_url}/${latest}/conformance.yaml"
     SONOBUOY_EXTRA_ARGS="${SONOBUOY_EXTRA_ARGS} ${kubevirt_plugin}"


ormergi:
SR-IOV lane is green and ran the tests that failed all over https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1348/check-up-kind-sriov/1880962417102426112

cluster-up/cluster/kind/check-cluster-up.sh
```diff
@@ -69,6 +69,11 @@ export CRI_BIN=${CRI_BIN:-$(detect_cri)}
     fi
     ${kubectl} wait -n kubevirt kv kubevirt --for condition=Available --timeout 15m
 
+    if [[ "$KUBEVIRT_PROVIDER" =~ "sriov" ]]; then
+      # Some SR-IOV tests require Kubevirt CPUManager feature
+      ${kubectl} patch kubevirts -n kubevirt kubevirt --type=json -p='[{"op": "replace", "path": "/spec/configuration/developerConfiguration/featureGates","value": ["CPUManager"]}]'
```
orelmisan, line 74: Please consider adding a feature gate to the end of the existing list, instead of replacing the whole list, as it will enable future expansion.

orelmisan (commented):
Thank you for the PR @ormergi.
Could you please give a few words about why is the PR needed?

cluster-up/cluster/kind/check-cluster-up.sh
ormergi: In case additional FG should be enabled I think there should be single a single patch call with all necessary FGs.
I didnt exported the FG name to var because its the only one at the moment.

ormergi:
> Thank you for the PR @ormergi. Could you please give a few words about why is the PR needed?

Done

orelmisan:
> > Thank you for the PR @ormergi. Could you please give a few words about why is the PR needed?
> 
> Done

Thank you.
Could you please add it to the commit message as well?

cluster-up/cluster/kind/check-cluster-up.sh
EdDev: IMO this should not come as a hard dependency of the SR-IOV provider.
Do you see a problem with directly marking the need to have `CPUManager` as input from the caller?

Also, the patching is odd to me too.
- Why do you use `replace` and not a simple `add`?
- Why do you think it is better to assume there is only one FG? It will just make it harder for the next contributor to add other FGs in general.

ormergi: I dont see problem with setting CPUManager feature to always on, in fact in kubevirt/kubevirt its always on.
We can go with that. Let me know what you think.

Regarding the patch, I used "replace" because "add" didn't work for me in away I can add FG with one-liner.
When Kubevirt CR `developerConfiguration.featureGates` is not initialized, it will require patch for initializing it and another one for adding the FG.
Using "replace" the way  I did enable having one-liner.

Passing the provider name is the most simple way I could find to solve it and get the lane gating as soon as possible.
We can introduce some additional env var to hold a FG list, on a follow up PR in case it will be needed (I tried to keep things simple).

dhiller: @ormergi I am not convinced that here is the right place to patch kubevirt? IMHO it should go into github.com/kubevirt/kubevirt/tree/main/hack/cluster-deploy.sh. WDYT?

ormergi: @dhiller WDYT?

brianmcarey: This looks good to me as these tests are not failing in kubevirt - we just need to ensure that we have the cluster correctly configured in the kubevirtci presubmit which is what this script is used for so I think it is in the right place. 

Can we move this forward to get the sriov kubevirtci presubmit back to being required?

dhiller: > @dhiller Changing deploy-cluster.sh affect cluster-sync flow, I am not sure its we should do that. Please note "check-up-kind-sriov" calls "kind/check-cluster-up.sh" directly, and "kind/check-cluster-up.sh" deploy kubevirt (from nightly release yamls).

@ormergi makes sense, thanks for the explanation!

dhiller (approved):
/approve

brianmcarey (commented):
/lgtm

PR #1354, opened 2025-01-21
fix: gocli startup in s390x architecture

nestoracunablanco:
The gocli container image is multi-architecture compatible; however, it attempts to download the k8s-1.30 image, which is not available for the s390x architecture. The k8s-1.30-slim image, on the other hand, is compatible. To enable support for s390x, we can enforce the use of the --slim parameter for this architecture.

**Release note**:

```release-note
NONE
```


chandramerla:
/lgtm 
With this change we are making for `make cluster-up` on s390x to always pass `--slim` argument to `gocli run` command invoked

brianmcarey (commented):
/approve 

thanks @nestoracunablanco

PR #1376, opened 2025-02-10
In Provision Manager, call "git fetch --tags" command to avoid git diff fail

chandramerla:
Without change - In provision manager while doing git diff to a tag obtained from latest file in gcs, if that tag is not present on local clone, we are exiting with a message to run git fetch by the user. 

Now with this change- we are once 'trying' to run the git fetch tags command after add upstream remote, before asking the user to do that.

This is mainly required to avoid failures in publish-kubevirtci-s390x job builds below, as while a build of this job is running, x86 equivalent job publish-kubevirtci might have updated latest file in GCS with new tag and tagged a commit in main, but as the publish-kubevirtci-s390x build is already in progress, its tags are outdated and needs a fetch:

[publish-kubevirtci-s390x/1886496178602774528](https://storage.googleapis.com/kubevirt-prow/logs/publish-kubevirtci-s390x/1886496178602774528/build-log.txt)
[publish-kubevirtci-s390x/1886111630970851328](https://storage.googleapis.com/kubevirt-prow/logs/publish-kubevirtci-s390x/1886111630970851328/build-log.txt)
[publish-kubevirtci-s390x/1885392389216931840](https://storage.googleapis.com/kubevirt-prow/logs/publish-kubevirtci-s390x/1885392389216931840/build-log.txt)
[publish-kubevirtci-s390x/1884869941908213760](https://storage.googleapis.com/kubevirt-prow/logs/publish-kubevirtci-s390x/1884869941908213760/build-log.txt)

Fixes #

/cc @brianmcarey 

**Checklist**

**Release note**:
```release-note
NONE
```

### DIFF
--- a/cluster-provision/gocli/cmd/provision_manager.go
+++ b/cluster-provision/gocli/cmd/provision_manager.go
@@ -314,21 +314,36 @@ func getKubevirtciTag() (string, error) {
 }
 
 func runCommand(command string, args []string) (string, error) {
-	var stdout, stderr bytes.Buffer
+	var stderr1 bytes.Buffer
+	remoteAddCmd := exec.Command("git", "remote", "add", "upstream", "https://github.com/kubevirt/kubevirtci.git")
+	remoteAddCmd.Stderr = &stderr1
+	if err := remoteAddCmd.Run(); err != nil {
+		logrus.Debugf("Failed to add remote named upstream: %v, stderr: %s", err, stderr1.String()) //upstream remote already may exist
+	} else {
+		logrus.Debug("Successfully added remote named upstream.")
+	}
 
+	var stderr2 bytes.Buffer
+	fetchCmd := exec.Command("git", "fetch", "upstream", "main", "--tags")
+	fetchCmd.Stderr = &stderr2
+	if err := fetchCmd.Run(); err != nil {
+		logrus.Debugf("Failed to fetch tags from upstream: %v, stderr: %s", err, stderr2.String())
+	}
+
+	var stdout3, stderr3 bytes.Buffer
 	cmd := exec.Command(command, args...)
-	cmd.Stdout = &stdout
-	cmd.Stderr = &stderr
+	cmd.Stdout = &stdout3
+	cmd.Stderr = &stderr3
 	err := cmd.Run()
 	if err != nil {
-		if strings.Contains(stderr.String(), "unknown revision or path not in the working tree") {
+		if strings.Contains(stderr3.String(), "unknown revision or path not in the working tree") {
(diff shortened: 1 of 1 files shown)


cluster-provision/gocli/cmd/provision_manager.go
```diff
@@ -314,21 +314,36 @@ func getKubevirtciTag() (string, error) {
 }
 
 func runCommand(command string, args []string) (string, error) {
-	var stdout, stderr bytes.Buffer
+	var stderr1 bytes.Buffer
+	remoteAddCmd := exec.Command("git", "remote", "add", "upstream", "https://github.com/kubevirt/kubevirtci.git")
```
brianmcarey, line 318: Will this fail if the user already has a remote called "upstream"?

brianmcarey (commented):
I am not sure how much we need this as it can impact the local user flow. 

We should try to make these updates to the s390x publish flow.

chandramerla:
> I am not sure how much we need this as it can impact the local user flow.
> 
> We should try to make these updates to the s390x publish flow.

Actually this won't make **much** difference to local user flow. As we just try adding upstream and fetching tags and just ignore any failure that occur during the same and later we anyway continue doing it same way as it is currently ie., asking user to run fetch tags command.

I see we currently tell user to run "git fetch upstream main --tags" and also now the code attempts the same once, before informing users. This helps local user flow also to be less dependent on user actions. 

Now from impact on the current user flow view, it actually makes user remote tracking branch upstream/main to get updated to latest commit. I think here we can actually fetch ONLY tags with the command without mention of branch name main like "git fetch upstream --tags". Because at the end we only want local branch to be compared with tag, not with main.
So, if I update the command in my new code and also in the log message to "git fetch upstream --tags", then it doesn't make any difference to the user environment IMO. WDYT?

cluster-provision/gocli/cmd/provision_manager.go
```diff
@@ -314,21 +314,36 @@ func getKubevirtciTag() (string, error) {
...
+	remoteAddCmd.Stderr = &stderr1
+	if err := remoteAddCmd.Run(); err != nil {
+		logrus.Debugf("Failed to add remote named upstream: %v, stderr: %s", err, stderr1.String()) //upstream remote already may exist
+	} else {
+		logrus.Debug("Successfully added remote named upstream.")
+	}
+
+	var stderr2 bytes.Buffer
```
brianmcarey, line 326: can we not just reuse sdtout and stderr for each of these rather than stderr3, stderr2 etc?

brianmcarey (approved):
/approve 

thanks @chandramerla

PR #1378, opened 2025-02-11
chore: update default kubevirt_provider.

nestoracunablanco:
This change enables the execution of `make cluster-up` on s390x without the need to set any additional environment variables.

**Release note**:

```release-note
NONE
```


nestoracunablanco:
@oshoval @qinqon, do you have any thoughts on this PR? Thank you in advance!

oshoval:
as it improves the flow for 390x, lgtm, thanks

/lgtm

brianmcarey (commented):
Its probably better to set the env variable for s390x jobs as we cannot always guarantee that the s390x provider version will match this default version as both are subject to change. 

We generally change this when the default provider is being removed and update to the latest provider version. 

No issues with this change would just recommend against relying on the default provider version for the s390x jobs. 

/approve

PR #1385, opened 2025-02-26
common: fix permission denied

alicefr:
Use the kubectl from the k8s provisioner and set the expected permissions on kubectl and kubeconfig.

**Release note**:

```release-note
NONE
```


alicefr:
Close in favor of 1386

PR #1384, opened 2025-02-26
Revert "fix: segmentation fault when using podman (#1368)"

brianmcarey:
**What this PR does / why we need it**:

This reverts commit f58328a78330dff7e7c62d10c867b6ca27ee6baa.

A number of users have reported issues running dev clusters locally with docker since this change.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

/cc @dhiller 

**Checklist**

This checklist is not enforcing, but it's a reminder of items that could be relevant to every PR.
Approvers are expected to review this list.

- [ ] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [ ] PR: The PR description is expressive enough and will help future contributors
- [ ] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [ ] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [ ] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough). New features and bug fixes require at least on e2e test
- [ ] Documentation: A [user-guide update](https://github.com/kubevirt/user-guide/) was considered and is present (link) or not required. You want a user-guide update if it's a user facing feature / API change.
- [ ] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was considered

**Release note**:

```release-note

```

PR #1386, opened 2025-02-27
Fix/seg fault permission denied

nestoracunablanco:
Following up on the discussion regarding #1384 and the conversation in [Slack](https://kubernetes.slack.com/archives/C0163DT0R8X/p1740566069460129), the current analysis indicates that both changes are necessary.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-up/cluster/k8s-provider-common.sh
+++ b/cluster-up/cluster/k8s-provider-common.sh
@@ -47,14 +47,20 @@ function up() {
     fi
     eval ${_cli:?} run $params
 
-    ${_cli} scp --prefix $provider_prefix /etc/kubernetes/admin.conf - >${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
+    # Workaround https://github.com/containers/conmon/issues/315 by not dumping file content to stdout
+    scp_suffix="- >${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER"
+    if [[ ${_cri_bin} = podman* ]]; then
+        scp_suffix="/kubevirtci_config"
+    fi
+
+    ${_cli} scp --prefix $provider_prefix /etc/kubernetes/admin.conf ${scp_suffix}/.kubeconfig
 
     # Set server and disable tls check
     export KUBECONFIG=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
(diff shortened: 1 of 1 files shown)


Barakmor1:
I think we should verify that this works with rootless Docker as well, because the test lanes of Kubevirtci don't

nestoracunablanco:
Hi @chandramerla, this issue is affecting users. In short, can the check-provision-k8s-1.31-s390x error be safely ignored? For more details, please refer to the PR and the Slack conversation linked in the description.

nestoracunablanco:
> I think we should verify that this works with rootless Docker as well, because the test lanes of Kubevirtci don't

@Barakmor1, I have limited time today and may not be able to verify it until 16:00 UTC+1. Could someone else take over the verification with rootless Docker? Thank you!

cluster-up/cluster/k8s-provider-common.sh
```diff
@@ -47,16 +47,29 @@ function up() {
...
+    if [[ ${_cri_bin} = podman* ]]; then
+        args="-v ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER:/kubevirtci_config"
+    fi
+    # Fix permissions for kubeconfig and kubectl
+    ${_cri_bin} run --privileged --rm $args \
+        --entrypoint /bin/sh ${_cli_container} \
+        -c "chmod 755 ${scp_suffix}/.kubectl"
+    ${_cri_bin} run --privileged --rm $args \
+        --entrypoint /bin/sh ${_cli_container} \
+        -c "chmod 766 ${scp_suffix}/.kubeconfig"
```
chandramerla, line 67: @nestoracunablanco I didn't quite follow the all messages in the thread. So, with this PR we are enabling podman cri to work correctly and assuming podman is in root mode. Isn't it? Will the container run commands make any difference to docker? If not, does it make sense to move it to if block above checked if cri is podman?

nestoracunablanco: Hi @chandramerla, you don’t need to worry too much about this. To summarize, some users were affected by recent changes, and one of my pull requests was initially thought to be the cause. However, after conducting some tests, it appears that may not be the case, or further analysis is required. I’m currently working on merging my previous changes with a new commit from @alicefr to see if that resolves the issue.  Thank you for taking the time to look into the s390x ci job :)

nestoracunablanco:
Closed in favor of #1387

PR #1391, opened 2025-03-11
fix: exit early when nested virtualization is not enabled

nestoracunablanco:
This change saves developers and maintainers time checking logs.

Context: In our s390x CI cluster, sometimes nested virtualization was disabled. This approach will help us identify such occurrences more quickly in the future.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-up/check.sh
+++ b/cluster-up/check.sh
@@ -53,8 +53,10 @@ if is_enabled "$KVM_NESTED"; then
 	echo "[ OK ] $KVM_ARCH nested virtualization enabled"
 else
 	echo "[ERR ] $KVM_ARCH nested virtualization not enabled"
+	exit 1
 fi
 
 if is_enabled "$KVM_HPAGE" && [ "$(uname -m)" = "s390x" ]; then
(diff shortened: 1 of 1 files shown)


cluster-up/check.sh
```diff
@@ -53,8 +53,10 @@ if is_enabled "$KVM_NESTED"; then
 	echo "[ OK ] $KVM_ARCH nested virtualization enabled"
 else
 	echo "[ERR ] $KVM_ARCH nested virtualization not enabled"
+	exit 1
```
oshoval, line 56: /hold
i am not sure it is fine to do so, if kubevirtci supports emulation then it shouldnt exit

please verify that deeply first

nestoracunablanco: Started a discussion in the mailing list. Thanks :)

oshoval: thanks, imho slack / this PR can fit better for such discussion
also if using mailing list, worth to add the PR link please

nestoracunablanco: Closing it based on the discussion in the mailing list.

PR #1393, opened 2025-03-11
fix: exit early when nested virtualization is not enabled in s390x

nestoracunablanco:
This change saves developers and maintainers time checking logs.

Context: In our s390x CI cluster, sometimes nested virtualization was disabled. This approach will help us identify such occurrences more quickly in the future.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-up/check.sh
+++ b/cluster-up/check.sh
@@ -53,8 +53,13 @@ if is_enabled "$KVM_NESTED"; then
 	echo "[ OK ] $KVM_ARCH nested virtualization enabled"
 else
 	echo "[ERR ] $KVM_ARCH nested virtualization not enabled"
+	# Exits early only in s390x. There may be some use cases with emulation for another architectures.
+	if [ "$KVM_ARCH" == "s390x" ]; then
+		exit 1
+	fi
 fi
 
 if is_enabled "$KVM_HPAGE" && [ "$(uname -m)" = "s390x" ]; then
 	echo "[ERR ] $KVM_HPAGE KVM hugepage enabled. It needs to be disabled while nested virtualization is enabled for s390x"
+	exit 1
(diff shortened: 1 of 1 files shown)


oshoval (commented):
Thanks
i cant say i like it
it adds bit complexity for specific cases and divergence between the archs
i would say the lanes themselves need to check it, not kubevirtci

assuming kubevirtci meant to support with and without,
it isnt a good API to allow it to just some archs and not on others,
also maintainability wise

nestoracunablanco:
Hi @oshoval I must admit, I’m not particularly fond of it either. I would love to explore other areas in the code where we can implement early failure while maintaining readability. Given the filename "check.sh," I initially thought this would be the perfect place for some pre-flight checks.

Regarding your points:
- the issue of allowing certain architectures while restricting others: there will always be differences in what one architecture permits compared to another. In the case of s390x, maintaining support only led to complications, as the tests don’t even start.
- divergence between architectures: even if we were to add checks in every unit test, it wouldn’t resolve the underlying issues. It’s clear that the architectures diverge, and this commit addresses that.

If you have any suggestions on how we can improve the situation, I’m more than willing to collaborate and identify other areas in the code where we can correct it.

oshoval:
Personally i am against this change in this file, sorry

You can for example add it on the bootstrap / prow job manifest etc

oshoval:
@EdDev might have good input here please

nestoracunablanco:
Outcome of the conversation with @EdDev: We can address the issue more generically by using an environment variable, such as REQUIRE_NESTED_VIRTUALIZATION. I will test this solution and push the new changes.

nestoracunablanco:
Closed in favor of #1397

PR #1397, opened 2025-03-14
feat: enable variable for requiring nested virtualization

nestoracunablanco:
When this variable is enabled, the cluster-up process will fail early. This approach helps to avoid disrupting existing workflows that rely on emulation.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-up/check.sh
+++ b/cluster-up/check.sh
@@ -53,6 +53,10 @@ if is_enabled "$KVM_NESTED"; then
 	echo "[ OK ] $KVM_ARCH nested virtualization enabled"
 else
 	echo "[ERR ] $KVM_ARCH nested virtualization not enabled"
+	if [ $KUBEVIRT_NESTED_VIRTUALIZATION_REQUIRED == "true" ]; then
+		echo "[ERR ] nested virtualization required, exiting..."
+		exit 1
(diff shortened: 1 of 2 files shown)


EdDev (approved):
Makes sense to me.
If there is a better place to put this, I hope others can comment about.

brianmcarey (commented):
thanks for this @nestoracunablanco 

/approve

dhiller (changes requested):
/approve

Although IIUC this is intended to enable the special case of kubevirt s390x lanes to fail early, which would normally go into k/kubevirt automation/test.sh , I find this solution simple and elegant enough to accept.

One nit I have though: env vars in this project are prefixed with `KUBEVIRT_` 

/hold

nestoracunablanco:
@dhiller, after running the tests three times, I've noticed that the errors related to pulling images appear to be consistent. I don't believe these issues are related to the change. Are there similar errors occurring in other PRs as well?

dhiller (commented):
/lgtm

@nestoracunablanco we are nearing stabilizing the check-provision, please bear with us.

brianmcarey (commented):
/hold cancel 

/retest

PR #1442, opened 2025-05-13
fix: Isolate kubeconfig and control CRI via KIND_EXPERIMENTAL_PROVIDER in kind provider

yaroslavborbat:
### What this PR does
This PR improves the Kind cluster creation process used for running KubeVirt locally by addressing two issues:

1. **Runtime Selection Control:**:  
  We now rely on `KIND_EXPERIMENTAL_PROVIDER` to control the container runtime used by Kind, ensuring consistent use of Podman or Docker when needed.
2. **Kubeconfig Isolation**:  
   Instead of using `kind get kubeconfig` after creation, we now pass `--kubeconfig` directly to `kind create`.  
   This ensures test clusters do not accidentally interact with the user's default kubeconfig (`~/.kube/config`), improving safety and isolation.

These changes make the local KubeVirt setup more predictable, stable and secure during development and testing.

**Checklist**

- [x] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required

**Release note**:

```release-note
Isolate kubeconfig and control CRI via KIND_EXPERIMENTAL_PROVIDER in kind provider
```

### DIFF
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -3,10 +3,18 @@
 set -e
 
 function detect_cri() {
-    if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi
+    if podman ps >/dev/null 2>&1; then
+        echo podman
+    elif docker ps >/dev/null 2>&1; then
+        echo docker
+    else
+        echo "Error: no container runtime detected. Please install Podman or Docker." >&2
+        exit 1
+    fi
 }
 
 export CRI_BIN=${CRI_BIN:-$(detect_cri)}
+export KIND_EXPERIMENTAL_PROVIDER=${CRI_BIN}
 CONFIG_WORKER_CPU_MANAGER=${CONFIG_WORKER_CPU_MANAGER:-false}
 # only setup ipFamily when the environmental variable is not empty
 # avaliable value: ipv4, ipv6, dual
@@ -193,11 +201,9 @@ function _fix_node_labels() {
 }
 
 function setup_kind() {
-    $KIND --loglevel debug create cluster --retain --name=${CLUSTER_NAME} --config=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml --image=$KIND_NODE_IMAGE \
+    $KIND --loglevel debug create cluster --retain --name=${CLUSTER_NAME} --config=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml --image=$KIND_NODE_IMAGE --kubeconfig=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig \
         || ( $KIND --loglevel debug delete cluster --name=${CLUSTER_NAME} \
-        && $KIND --loglevel debug create cluster --retain --name=${CLUSTER_NAME} --config=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml --image=$KIND_NODE_IMAGE )
-
-    $KIND get kubeconfig --name=${CLUSTER_NAME} > ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig
+        && $KIND --loglevel debug create cluster --retain --name=${CLUSTER_NAME} --config=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml --image=$KIND_NODE_IMAGE --kubeconfig=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig )
 
     if ${CRI_BIN} exec ${CLUSTER_NAME}-control-plane ls /usr/bin/kubectl > /dev/null; then
         kubectl_path=/usr/bin/kubectl


cluster-up/cluster/kind/common.sh
```diff
@@ -193,11 +194,9 @@ function _fix_node_labels() {
 }
 
 function setup_kind() {
-    $KIND --loglevel debug create cluster --retain --name=${CLUSTER_NAME} --config=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml --image=$KIND_NODE_IMAGE \
+    $KIND --loglevel debug create cluster --retain --name=${CLUSTER_NAME} --config=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml --image=$KIND_NODE_IMAGE --kubeconfig=${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/.kubeconfig \
```
oshoval, line 197: Does it work also when `${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER` doesnt exists ?
i.e it creates nested dirs ?

cluster-up/cluster/kind/common.sh
```diff
@@ -3,7 +3,8 @@
 set -e
 
 function detect_cri() {
-    if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi
+    # Kind checks Docker first by default, so we follow the same order for consistency.
+    if docker ps >/dev/null 2>&1; then echo docker; elif podman ps >/dev/null 2>&1; then echo podman; fi
```
oshoval, line 7: Kubevirt checks podman first, so please check how it interacts with kubevirt once both docker and podman are installed

oshoval (commented):
Thanks for the PR

cluster-up/cluster/kind/common.sh
yaroslavborbat: I changed this behavior because it was causing issues in practice. On my machine both Docker and Podman are installed. Kind detected Docker first and created containers using Docker, but the script detected Podman and tried to perform subsequent tasks with it — leading to errors.

The root problem is a mismatch between which container runtime is used to create containers (Docker) and which one the script tries to use later (Podman). So I aligned the detection order with Kind's default behavior to avoid such conflicts.

https://github.com/kubernetes-sigs/kind/blob/180d624f741e3da5ceba52b643e29c4e64538537/pkg/cluster/provider.go#L114

However, after reverting the detection back to the original order (Podman first), I ran:
```bash
export KUBEVIRT_PROVIDER=kind-1.31
make cluster-up
```
And here's what I got:
```bash
Set kubectl context to "kind-kind-1.31"
You can now use your cluster with:

kubectl cluster-info --context kind-kind-1.31 --kubeconfig /home/yaroslavborbat/work/github.com/yaroslavborbat/kubevirt/kubevirtci/_ci-configs/kind-1.31/.kubeconfig

Have a nice day! 👋
++ podman exec kind-1.31-control-plane ls /usr/bin/kubectl
Error: no container with name or ID "kind-1.31-control-plane" found: no such container
++ podman exec kind-1.31-control-plane ls /bin/kubectl
Error: no container with name or ID "kind-1.31-control-plane" found: no such container
++ echo 'Error: kubectl not found on node, exiting'
Error: kubectl not found on node, exiting
++ exit 1
make: *** [Makefile:166: cluster-up] Error 1

```

oshoval: thanks, i understand but does it work with kubevirt/kubevirt cluster-sync ?
see here please
the non kubevirtci are using podman as default


https://github.com/search?q=repo%3Akubevirt%2Fkubevirt+echo+podman&type=code

personally, i am against this change because kubevirt still favor podman and it is the main consumer of kubevirtci
one can `export CRI_BIN=docker` if desired
i don't think we want to change kubevirt to match kubevirtci wrt to this order just because kind
we are using kind only for kind-sriov atm

the lanes on kubevirt wont detect it, but whoever does this change might need to maintain it, and people that have both docker and podman locally might have other mistakes

cluster-up/cluster/kind/common.sh
yaroslavborbat: Yes, it works fine. Kind does create all necessary nested directories to save the kubeconfig file.
However, in our case, we expect the directory ${KUBEVIRTCI_CONFIG_PATH}/${KUBEVIRT_PROVIDER} to already exist because it must contain the config file ${KUBEVIRTCI_CONFIG_PATH}/${KUBEVIRT_PROVIDER}/kind.yaml. Without this file, the cluster wouldn't be created in the first place.

cluster-up/cluster/kind/common.sh
yaroslavborbat: I checked `cluster-sync` and it works fine — probably because it doesn't involve CRI operations directly.
I understand your concern, and I also see the risk: if in the future `cluster-sync` or any related task starts relying on CRI, we might run into the same issue again.

After looking into this further, I found that Kind actually allows specifying the container runtime via an environment variable: `KIND_EXPERIMENTAL_PROVIDER`.
https://github.com/kubernetes-sigs/kind/blob/180d624f741e3da5ceba52b643e29c4e64538537/pkg/internal/runtime/runtime.go#L11

So instead of changing the detection order globally, I'll simply add:
```bash
export KIND_EXPERIMENTAL_PROVIDER="${CRI_BIN}"
```

This way, we let Kind and KubeVirt consistently use the desired runtime without introducing conflicts between Docker and Podman.

Thanks for pointing out the importance of keeping KubeVirt's behavior consistent

oshoval: Thank you,
the order is still changed,
can it be reverted please ? and use `KIND_EXPERIMENTAL_PROVIDER` where ever needed?

we also favor and encourage podman on docker, so this is another reason we put podman first

oshoval:
Thanks, please consider adding KIND_EXPERIMENTAL_PROVIDER also to kubevirt

oshoval:
Please squash the "fix" commit
and update PR desc according latest changes
(sorry if i am rushing, maybe you are in the middle of doing it already)

cluster-up/cluster/kind/common.sh
yaroslavborbat: I've pushed changes

oshoval:
Nice, thanks !
/lgtm

oshoval:
cc @brianmcarey @dhiller

yaroslavborbat:
@oshoval Could you help retrigger the tests? I can't restart them myself, it seems `ok-to-test` is required despite the `lgtm` label. Thanks

oshoval:
once there is lgtm approve it will retrigger them

this lane is optional, and this PR doesnt affect vm based providers, just kind so it doesnt affect
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1442/check-provision-k8s-1.31-s390x/1922659737250304000

oshoval:
this lane shouldn't even run
it seems there is some bug with provision manager
or just no new tagged lately of kubevirtci

https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1442/check-provision-manager/1922659737153835008/build-log.txt

dhiller (approved):
@yaroslavborbat thank you for the PR - looks great to me :tada: 

/approve

PR #1467, opened 2025-06-27
fix(check-pod-images.sh): use pre-pull-images, add --auto-update

dhiller:

* fix: Instead of calling fetch-images.sh we can use the pre-pull-images file which
is a result of the update-images logic and contains all the images.
* feat: adds --auto-update flag which instead of printing the missing images will directly append them to the `extra-pre-pull-images` file.

Fixes #

As discussed we want an auto bump mechanism. The `--auto-update` is the base for the periodic jobs that will create PRs.

/cc @brianmcarey

### DIFF
--- a/cluster-provision/k8s/check-pod-images.sh
+++ b/cluster-provision/k8s/check-pod-images.sh
@@ -20,6 +20,12 @@
 
 set -exuo pipefail
 
+auto_update=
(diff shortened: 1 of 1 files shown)


brianmcarey (approved):
/approve

thanks @dhiller !

PR #1475, opened 2025-07-03
Continue the fix of the issue with job names w/o k8s version

nunnatsa:

Follow up #1473

The previous try to fix the issue didn't help, because the code is failing when running inside the gocli container, where the `KUBEVIRT_PROVIDER` environment variable is not defined.

This PR adds the missing environment variable when running the container.

**Release note**:
```release-note
None
```


nunnatsa:
Thanks @sradco. Had to fix the syntax. Now it should work. Could you PTAL?

Issue #1501, opened 2025-07-29
s390x,centos: Periodic centos 9 bump is failing for s390x


brianmcarey:
The periodic job that bumps the centos 9 base image for s390x is failing. 

```
 [2/3] STEP 7/7: RUN if test "$BUILDARCH" != "s390x"; then       /download_box.sh https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-Vagrant-9-$centos_version.x86_64.vagrant-libvirt.box &&       curl -L -o /initramfs-amd64.img http://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/images/pxeboot/initrd.img &&       curl -L -o /vmlinuz-amd64 http://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz;     else       /download_box.sh https://cloud.centos.org/centos/9-stream/s390x/images/CentOS-Stream-GenericCloud-9-$centos_version.s390x.qcow2 &&       export LIBGUESTFS_BACKEND=direct &&       guestfish --ro --add box.qcow2 --mount /dev/sda1:/ ls /boot/ | grep -E '^vmlinuz-|^initramfs-' | xargs -I {} guestfish --ro --add box.qcow2 -i copy-out /boot/{} / ;     fi
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   507  100   507    0     0    613      0 --:--:-- --:--:-- --:--:--   613
100   507  100   507    0     0    613      0 --:--:-- --:--:-- --:--:--   613
libguestfs: error: mount_options: mount_options_stub: /dev/sda1: No such file or directory 
```

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirtci-bump-centos-base-s390x/1950087872417632256

nestoracunablanco:
@brianmcarey what do you think about my analysis? Do you need something else?

brianmcarey:
Hi @nestoracunablanco - thanks for looking into this. I'm afraid the job is still failing after the latest bump - https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirtci-bump-centos-base-s390x/1952580471313403904#1:build-log.txt%3A2191

[...]

PR #1503, opened 2025-07-30
fix: enable centos9 download curl command fail

nestoracunablanco:
The curl command does not fail when the download is unsuccesful, which causes the script to fail at a later point.

Error related to #1501

```release-note
NONE
```


dhiller (approved):
/approve

PR #1643, opened 2026-03-06
gocli: Increase default etcd in-memory tmpfs size from 512M to 1G

lyarwood:
## Summary

- Increases the default etcd in-memory tmpfs size from 512M to 1G to avoid etcd failures due to insufficient space.

Ref: https://github.com/kubevirt/kubevirt/issues/17019


dhiller (approved):
/approve

Thank you @lyarwood !

PR #1656, opened 2026-03-22
docs: rootless podman actually works

dankenigsberg:
Unlike what our docs say, I (with the help of Claude) was able to use rootless podman for `make cluster-up`.
Update the documentation with instructions on how to do that.

```release-note
Rootless podman `make cluster-up` documented
```

### DIFF
--- a/PODMAN.md
+++ b/PODMAN.md
@@ -4,22 +4,38 @@ Install podman 3.1+, then run it in docker compatible mode:
 
 ## Rootless podman
 
+### Setup (Fedora)
+
+Enable the user podman socket:
+
+```bash
+systemctl --user enable --now podman.socket
 ```
-systemctl start --user podman.socket
+
+Load required kernel modules now and during future boots:
+
+```bash
+sudo tee /etc/modules-load.d/kubevirtci.conf <<EOF
+# Kernel modules required for KubeVirt cluster-up
+ip_tables
+iptable_nat
+iptable_filter
+ip6_tables
+ip6table_nat
+ip6table_filter
+kvm
+EOF
+sudo systemctl restart systemd-modules-load
 ```
 
-Currently rootless podman is **not** working with the `make cluster-sync`
-command, essentially because incoming traffic is coming from the loopback device
-instead of eth0.
+Configure the podman socket path for your future shells:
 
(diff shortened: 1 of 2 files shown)


dankenigsberg:
/reopen

@dhiller @Whitedyl would you review this? I think it's cool not to require root.

dhiller (approved):

oshoval:
worked nicely, thanks

PODMAN.md
```diff
@@ -4,22 +4,38 @@ Install podman 3.1+, then run it in docker compatible mode:
...
+Configure the podman socket path for your future shells:
 
-This will use `fuse-overlayfs` as storage layer. If the performance is not
-satisfactory, consider running podman as root to use plain `overlayfs2`:
+```bash
+mkdir -p ~/.bashrc.d
+tee ~/.bashrc.d/kubevirtci <<EOF
+export KUBEVIRTCI_PODMAN_SOCKET=\$XDG_RUNTIME_DIR/podman/podman.sock
+EOF
+```
```
oshoval, line 38: On my laptop, `~/.bashrc.d` isn't auto loaded.
(seems this is the default behavior, not something i removed, Fedora CSB)

It caused an interesting state, since i have both docker and podman installed:
Kubevirtci failed to detect podman, fallback to docker on`make cluster-up`.
Kubevirt detects podman, because it has ligher condition
Due to that, `make cluster-sync` hangs on rsync (mix of cluster on docker, cluster-sync on podman)

Maybe worth to change this section to add the export to .bashrc directly ? (best solution imo)
(or on kubevirt to reuse the cluster that is already created, but this is overkill imo)

Another option is to fix `hack/dockerized` hang as well (but it can be considered secondary bug).

`make rpm-deps-cs10` is broken with podman atm,
possible due to having both installed, it creates a rsync conflict.

dankenigsberg: Ah, I assumed that your bashrc was created with a modern skeleton, with
```
# User specific aliases and functions
if [ -d ~/.bashrc.d ]; then
    for rc in ~/.bashrc.d/*; do
        if [ -f "$rc" ]; then
            . "$rc"
        fi
    done
fi
unset rc
```

oshoval:
```
# Source global definitions
if [ -f /etc/bashrc ]; then
	. /etc/bashrc
fi
```
i have just this at the very beginning, searching for `bashrc` in .bashrc

oshoval: some more bugs (some might be related to having both installed)

1. seems gocli doesn't detect current the CRI, it cause problems when trying to provision with podman (didn't have those problems at the past with rootful as far as i remember)
2. podman fails locally on `make cluster-sync` - fix https://github.com/kubevirt/kubevirt/pull/18268
3. seems we have lots of ways to pin, `CRI_BIN`, `KUBEVIRTCI_RUNTIME`, `KUBEVIRT_CRI` fun fun (each flow has different semantics)

PR #1657, opened 2026-03-23
fedora-with-test-tooling: Upgrade to fedora43

oshoval:
Upgrade the `fedora-with-test-tooling` container-disk image from Fedora 39 (EOL)
to Fedora 43, with all the compatibility fixes required to keep KubeVirt tests
working on the newer distribution.
Can be built with latest osinfo `fedora43`.

Additional motivation is the flakes, see please this [comment](https://github.com/kubevirt/kubevirt/issues/17085#issuecomment-4149870482).
Update the baseline, remove `pam_nologin`, so we can advance from here on fresh and updated baseline.

Tested on Kubevirt using https://github.com/kubevirt/kubevirt/pull/17257
Images are smaller, and hopefully some tests might be faster (at least versus raw F43) due to the changes.

Core upgrade
- Bump base image from F39 to F43

F43 compatibility fixes

- **chpasswd syntax** — `password`/`chpasswd` top-level keys removed in cloud-init 24.x; switch to `chpasswd.users` list format
- **Systemd shell integration** — systemd ≥258 injects OSC escape sequences via `PROMPT_COMMAND`/`PS0`, breaking `goexpect` regex matching in console tests; disable via `profile.d` drop-in
- **Classic NIC naming** — F43 enables predictable names (`enp1s0`) by default; restore `eth0` via `net.ifnames=0 biosdevname=0` through `grubby`
- **pam_nologin** — blocks non-root SSH logins during early boot, causing flaky test failures; remove from the SSH PAM stack (this is one of the reasons we have flakes since F39, see [issue](https://github.com/kubevirt/kubevirt/issues/17085)).
- **cloud-init datasource restriction** — without an explicit datasource_list, cloud-init 25.x probes all datasources including network-based ones (EC2, Azure, ...) each with their own timeout, delaying boot completion and blocking wait-for-cloud-init.service before qemu-guest-agent can start; restrict to [NoCloud, ConfigDrive, None] to complete instantly when no matching ISO is present
- **cloud-init target** — `wait-for-cloud-init.service` was `WantedBy=cloud-init.service`; correct to `WantedBy=cloud-init.target` for proper ordering against cloud-init completion
- **hostname -f** — the `resolve` NSS module returns the short hostname via systemd-resolved D-Bus; switch to `dns` so the stub resolver applies search domains and returns the FQDN
- **users_groups sed pattern** — fix match to use underscore as introduced on F43
- **nc setup** — replace ad-hoc `busybox.net` binary download with `dnf install nmap-ncat`; keep OpenBSD netcat as default `nc` via alternatives so cloud-init 25.x using `-U` (Unix sockets) keeps working, while tests can use `ncat` when needed
- **cloud-user definition** — add explicit `cloud-user` in cloud-config so modules referencing it do not fail when `users_groups` is removed
- **s390x boot artifacts** — regenerate boot artifacts (`dracut` + `zipl`) during provisioning to avoid bootmap/initramfs drift and early boot panics

Image build improvements
- **virt-sparsify** — replace `qemu-img convert` with `virt-sparsify --compress` for smaller images (zeroes unused blocks before compressing)

Fixes #

**Checklist**

**Release note**:

```release-note
None
```

### DIFF
--- a/cluster-provision/images/vm-image-builder/customize-image.sh
+++ b/cluster-provision/images/vm-image-builder/customize-image.sh
@@ -99,5 +99,5 @@ virt-sysprep -d $DOMAIN_NAME --operations machine-id,bash-history,logfiles,tmp-f
 # Remove VM
 undefine_vm "${DOMAIN_NAME}"
 
-# Convert image
-qemu-img convert -c -O qcow2 "${SOURCE_IMAGE_PATH}" "${CUSTOMIZE_IMAGE_PATH}"
+# Convert image, sparsify and compress
+virt-sparsify --compress "${SOURCE_IMAGE_PATH}" "${CUSTOMIZE_IMAGE_PATH}"
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -66,13 +66,12 @@ runcmd:
   - sudo grubby --update-kernel=ALL --args="net.ifnames=0 biosdevname=0"
   - sudo systemctl daemon-reload
   - sudo dnf install -y kernel-modules-$(uname -r) qemu-guest-agent stress-ng virt-what vm-dump-metrics tpm2-tools netcat sg3_utils iperf3 wget tcpdump
-  # Download the busybox netcat and relpace it with the OpenBSD netcat
-  # To align with the netcat tool used in the Alpine image for network e2e testing
-  - if [ $(uname -m) == "aarch64" ]; then export CPUARCH="armv8l"; elif [ $(uname -m) == "s390x" ]; then export CPUARCH="s390x"; else export CPUARCH="x86_64"; fi
-  - sudo wget https://busybox.net/downloads/binaries/1.31.0-defconfig-multiarch-musl/busybox-${CPUARCH}
-  - sudo chmod +x busybox-${CPUARCH}
-  - sudo mv /usr/bin/nc /usr/bin/netcat-openbsd
-  - sudo mv busybox-${CPUARCH} /usr/bin/nc
+  # Install nmap-ncat in addition to OpenBSD netcat.
+  # Use "nc" when OpenBSD behavior is needed (cloud-init uses "nc -U").
+  # Use "ncat" when nmap-ncat behavior is needed (--recv-only, --no-shutdown,
+  # and server commands that use --sh-exec in tests).
+  - sudo dnf install -y nmap-ncat
+  - sudo alternatives --set nc /usr/bin/netcat
   - sudo dnf clean all
   - sudo systemctl enable wait-for-cloud-init
   - sudo systemctl enable qemu-guest-agent.service
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -45,6 +45,12 @@ write_files:
     - path: /etc/profile.d/disable-bracketed-paste.sh
       content: |
         bind 'set enable-bracketed-paste off'
+    # Note: systemd>=258 (Fedora 43+) injects OSC 3008/8003 escape sequences
+    # via PROMPT_COMMAND and PS0 for shell integration. These break goexpect
+    # regex matching in console tests. Disable them at the profile level.
+    - path: /etc/profile.d/disable-systemd-shell-integration.sh
+      content: |
+        unset PROMPT_COMMAND PS0
 runcmd:
   - sudo systemctl daemon-reload
   - sudo dnf install -y kernel-modules-$(uname -r) qemu-guest-agent stress-ng virt-what vm-dump-metrics tpm2-tools netcat sg3_utils iperf3 wget tcpdump
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -51,6 +51,15 @@ write_files:
     - path: /etc/profile.d/disable-systemd-shell-integration.sh
       content: |
         unset PROMPT_COMMAND PS0
+    # Restrict cloud-init datasource search to NoCloud, ConfigDrive, and None
+    # so that VMs booted without a cloud-init ISO (e.g. Freeze/Unfreeze tests)
+    # complete datasource detection in seconds instead of timing out through
+    # network-based datasources (EC2, Azure, ...) which can take many minutes
+    # on arm64. With NoCloud or ConfigDrive data injected by KubeVirt the
+    # behaviour is unchanged; without data cloud-init falls through to None.
+    - path: /etc/cloud/cloud.cfg.d/99-kubevirt-datasource.cfg
+      content: |
+        datasource_list: [NoCloud, ConfigDrive, None]
 runcmd:
   # Fedora 43+ uses predictable network interface names (e.g. enp1s0).
   # Many KubeVirt tests assume "eth0". Restore classic naming.
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -88,4 +88,11 @@ runcmd:
   - |
     sudo sed -i 's/^hosts:.*/hosts:      files dns myhostname resolve [!UNAVAIL=return]/' /etc/nsswitch.conf
   - sudo sed -i '/pam_nologin/d' /etc/pam.d/sshd
+  # Ensure boot artifacts are refreshed on IBM Z. If cmdline/kernel artifacts drift,
+  # s390x can panic early with unknown-block and disabled-wait.
+  - |
+    if [ "$(uname -m)" = "s390x" ]; then
+      sudo dracut -f --regenerate-all
+      sudo zipl
+    fi
   - sudo shutdown
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -52,6 +52,9 @@ write_files:
       content: |
         unset PROMPT_COMMAND PS0
 runcmd:
+  # Fedora 43+ uses predictable network interface names (e.g. enp1s0).
+  # Many KubeVirt tests assume "eth0". Restore classic naming.
+  - sudo grubby --update-kernel=ALL --args="net.ifnames=0 biosdevname=0"
   - sudo systemctl daemon-reload
   - sudo dnf install -y kernel-modules-$(uname -r) qemu-guest-agent stress-ng virt-what vm-dump-metrics tpm2-tools netcat sg3_utils iperf3 wget tcpdump
   # Download the busybox netcat and relpace it with the OpenBSD netcat
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -71,9 +71,13 @@ runcmd:
   - sudo hostnamectl set-hostname "" --transient
   - sudo sed -i '/users_groups/d' /etc/cloud/cloud.cfg
   - sudo sed -i 's/^SELINUX=.*/SELINUX=permissive/' /etc/selinux/config
-  # Adjust /etc/nsswitch.conf to mitigate the bug with nss-myhostame which
-  # breaks FQDN: https://bugzilla.redhat.com/show_bug.cgi?id=2038634
+  # Prefer "dns" before "resolve" in nsswitch.conf hosts lookup order.
+  # In our test setup, the "resolve" NSS module (systemd-resolved D-Bus path)
+  # can return the local short hostname before DNS search domains are applied,
+  # so "hostname -f" may return only the short name.
+  # Keep "resolve [!UNAVAIL=return]" at the end to preserve fallback semantics.
+  # Note: use a literal block (|) to avoid YAML misinterpreting "hosts:" as a mapping key.
   - |
-    sudo sed -i 's/hosts:      files myhostname resolve \[\!UNAVAIL=return\] dns/hosts:      resolve \[\!UNAVAIL=return\] files myhostname dns/' /etc/nsswitch.conf
+    sudo sed -i 's/^hosts:.*/hosts:      files dns myhostname resolve [!UNAVAIL=return]/' /etc/nsswitch.conf
   - sudo sed -i '/pam_nologin/d' /etc/pam.d/sshd
   - sudo shutdown
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -75,4 +75,5 @@ runcmd:
   # breaks FQDN: https://bugzilla.redhat.com/show_bug.cgi?id=2038634
   - |
     sudo sed -i 's/hosts:      files myhostname resolve \[\!UNAVAIL=return\] dns/hosts:      resolve \[\!UNAVAIL=return\] files myhostname dns/' /etc/nsswitch.conf
+  - sudo sed -i '/pam_nologin/d' /etc/pam.d/sshd
   - sudo shutdown
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -69,7 +69,7 @@ runcmd:
   - sudo systemctl enable qemu-guest-agent.service
   - sudo hostnamectl set-hostname ""
   - sudo hostnamectl set-hostname "" --transient
-  - sudo sed -i /users-groups/d /etc/cloud/cloud.cfg
+  - sudo sed -i '/users_groups/d' /etc/cloud/cloud.cfg
   - sudo sed -i 's/^SELINUX=.*/SELINUX=permissive/' /etc/selinux/config
   # Adjust /etc/nsswitch.conf to mitigate the bug with nss-myhostame which
   # breaks FQDN: https://bugzilla.redhat.com/show_bug.cgi?id=2038634
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -1,7 +1,11 @@
 #cloud-config
-password: fedora
-chpasswd: { expire: False }
 ssh_pwauth: yes
+chpasswd:
+  users:
+    - name: fedora
+      password: fedora
+      type: text
+  expire: false
 write_files:
     - path: /etc/modules-load.d/mlx5.conf
       content: |
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -37,7 +37,7 @@ write_files:
           RemainAfterExit=true
 
           [Install]
-          WantedBy=cloud-init.service
+          WantedBy=cloud-init.target
     # Note: bash and libreadline in Fedora 35 come with 'bracketed-paste' mode
     # enabled by default. This results in extra escape sequences in the
     # console output. The mode is explicitly disabled here to make the existing
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -1,9 +1,16 @@
 #cloud-config
 ssh_pwauth: yes
+users:
+  - name: cloud-user
+    sudo: ALL=(ALL) NOPASSWD:ALL
+    shell: /bin/bash
 chpasswd:
   users:
     - name: fedora
       password: fedora
+      type: text
+    - name: cloud-user
+      password: cloud-user
       type: text
   expire: false
 write_files:
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/image-url
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/image-url
@@ -1,1 +1,1 @@
-https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/39/Cloud/x86_64/images/Fedora-Cloud-Base-39-1.5.x86_64.qcow2
+https://download.fedoraproject.org/pub/fedora/linux/releases/43/Cloud/x86_64/images/Fedora-Cloud-Base-Generic-43-1.6.x86_64.qcow2
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/image-url-arm64
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/image-url-arm64
@@ -1,1 +1,1 @@
-https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/39/Cloud/aarch64/images/Fedora-Cloud-Base-39-1.5.aarch64.qcow2
+https://download.fedoraproject.org/pub/fedora/linux/releases/43/Cloud/aarch64/images/Fedora-Cloud-Base-Generic-43-1.6.aarch64.qcow2
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/image-url-s390x
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/image-url-s390x
@@ -1,1 +1,1 @@
-https://archives.fedoraproject.org/pub/fedora-secondary/releases/39/Cloud/s390x/images/Fedora-Cloud-Base-39-1.5.s390x.qcow2
+https://download.fedoraproject.org/pub/fedora-secondary/releases/43/Cloud/s390x/images/Fedora-Cloud-Base-Generic-43-1.6.s390x.qcow2
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/os-variant
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/os-variant
@@ -1,1 +1,1 @@
-fedora39
+fedora43


cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -37,7 +37,7 @@ write_files:
           RemainAfterExit=true
 
           [Install]
-          WantedBy=cloud-init.service
+          WantedBy=cloud-init.target
```
ormergi: Is it a new target name comes with newer clould-init?
Or the WantedBy use the wrong target in the first place?

ormergi (commented):
Thank for the PR
Finally we can have newer fedora tests, nice!

There are few changes in the PR that were done to retain current behavior to avoid test failures.
IMO those changes should be addressed on kubevirt/kubevirt side, preferably on the PR that bring this PR changes, instead of addressed by workarounds for the image.

Having test being fixed instead of modifying the image, will make the image more portable and resilient to future changes making upgrades much easier:

- Disable systemd shell integration: tests console code adapt and use the new shell prefix.
- Classic interface naming (ethX): new default is predictable names (enp1s0..), tests should avoid hard-coding inface names as much as possible and relay on existing APIs instead (e.g.: status.ifaces..)
- BusyBox netcant wrapper: tests currently assume BusyBox version exist, tests should change to work with the netcat version that comes with fedora out-of-the-box (same goes for other tools).
  (To avoid that in the future we may need some wrapper for such tools.)
-  hostname -f to produce full FQDN hostnames: test expects full FQDN hostnames, not sure its a must. If not, test should use different tool or adapt.

oshoval:
> Thank for the PR Finally we can have newer fedora tests, nice!
> 
> There are few changes in the PR that were done to retain current behavior to avoid test failures. IMO those changes should be addressed on kubevirt/kubevirt side, preferably on the PR that bring this PR changes, instead of addressed by workarounds for the image.
> 
> Having test being fixed instead of modifying the image, will make the image more portable and resilient to future changes making upgrades much easier:
> 
> * Disable systemd shell integration: tests console code adapt and use the new shell prefix.
> * Classic interface naming (ethX): new default is predictable names (enp1s0..), tests should avoid hard-coding inface names as much as possible and relay on existing APIs instead (e.g.: status.ifaces..)
> * BusyBox netcant wrapper: tests currently assume BusyBox version exist, tests should change to work with the netcat version that comes with fedora out-of-the-box (same goes for other tools).
>   (To avoid that in the future we may need some wrapper for such tools.)
> * hostname -f to produce full FQDN hostnames: test expects full FQDN hostnames, not sure its a must. If not, test should use different tool or adapt.

Thanks,
I dont agree, having the image ready for our tests proved itself more bullet proof, for exampled when we fixed at the past the login by removing the user-groups.
Moreover, some of those you mentioned are real bugs, for example the FDQN, the fix was already there, i just adapted it to F43.
Same for shell itegration, it is not about prefix, it is because of noisy chars, same as there is few lines above it for other type of chars.
Lets consider one by one inline please, and see more opinions on which side should fix them, those are just examples.

EdDev (commented):
Thank you!

Most of the changes here looks great.
I had some comments, especially with the last one.

oshoval:
btw for visibility 

not sure if this one is a must
https://github.com/kubevirt/kubevirtci/pull/1657/commits/c048328e05960422c846485fef51671b994d704b
was done in the middle of researching, it is correct for kubevirt, but all might pass without it

cluster-provision/images/vm-image-builder/customize-image.sh
```diff
@@ -73,6 +73,51 @@ else
   consoleconfig=""
 fi
 
+# For Fedora variants: verify the requested version is known to the local
+# osinfo-db. os-variant is only a hint for virt-install driver/feature
+# selection and has no effect on the image itself.
+function check_remote_osinfo_variant() {
```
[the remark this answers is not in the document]
oshoval: dropped from this PR

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -71,9 +71,13 @@ runcmd:
...
-  # breaks FQDN: https://bugzilla.redhat.com/show_bug.cgi?id=2038634
+  # Use plain "dns" instead of "resolve" in nsswitch.conf hosts lookup order.
+  # The "resolve" NSS module uses systemd-resolved's D-Bus API which returns the
+  # local short hostname before applying DNS search domains, so "hostname -f" returns
+  # just the short name. The "dns" module queries the stub resolver (127.0.0.53)
+  # directly, which applies search domains and returns the correct FQDN.
+  # Note: use a literal block (|) to avoid YAML misinterpreting "hosts:" as a mapping key.
   - |
-    sudo sed -i 's/hosts:      files myhostname resolve \[\!UNAVAIL=return\] dns/hosts:      resolve \[\!UNAVAIL=return\] files myhostname dns/' /etc/nsswitch.conf
+    sudo sed -i 's/^hosts:.*/hosts:      files dns myhostname/' /etc/nsswitch.conf
```
EdDev, line 81: Do you know which tests are effected by this one?
I do not like that we mess up with the default behavior and find it hacky.

It is like declaring that Fedora is not behaving well for our e2e tests.
But then, we should probably have done tweaks per the are of tests which require it, so it will be clearer.

To be clear, our e2e tests are suppose to express user interaction with the system and it is an edge case that we need to adjust the OS to the user.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -51,6 +51,15 @@ write_files:
...
         unset PROMPT_COMMAND PS0
+    # Restrict cloud-init datasource search to NoCloud, ConfigDrive, and None
+    # so that VMs booted without a cloud-init ISO (e.g. Freeze/Unfreeze tests)
+    # complete datasource detection in seconds instead of timing out through
+    # network-based datasources (EC2, Azure, ...) which can take many minutes
+    # on arm64. With NoCloud or ConfigDrive data injected by KubeVirt the
+    # behaviour is unchanged; without data cloud-init falls through to None.
+    - path: /etc/cloud/cloud.cfg.d/99-kubevirt-datasource.cfg
+      content: |
+        datasource_list: [NoCloud, ConfigDrive, None]
```
EdDev, line 62: I noticed additional types are added and requested. Can you check that no new stuff are in the pipeline?

I guess that the e2e tests coverage should alarm us if new types are introduced, as those tests will probably fail and require adjusting the config.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: we already had this fix, see previous line, i just adapted it to F43
it affects the subdomain tests, that one that need FQDN `hostname -f` once subdomain is declared

see please commit message as well
https://github.com/kubevirt/kubevirtci/pull/1657/commits/ab1bc30a94bee183f6dc3117c64b58cd1a769a2d

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: > I noticed additional types are added and requested. Can you check that no new stuff are in the pipeline?

where ? 

> I guess that the e2e tests coverage should alarm us if new types are introduced, as those tests will probably fail and require adjusting the config.

indeed all passed, we are safe in that aspect, and if needed can add according needs
note please that this is the only commit i think we might able to drop, unless we find out its indeed required / saves boot time,
but it proved to be safe, without any flake about it, so i think we are good to keep

EDIT 
AI answer

```
> This change actually narrows datasource probing; it does not add new pipeline-facing datasource types.  
> By default cloud-init probes many datasources (including network ones like EC2/Azure) and can spend significant time timing out.
>
> Here we explicitly restrict to `[NoCloud, ConfigDrive, None]`, which are the only modes we use in KubeVirt tests:  
> - `NoCloud` / `ConfigDrive` when data is injected  
> - `None` when no datasource is provided
>
> So this reduces boot-time variability and keeps behavior explicit without introducing new external dependencies.
```

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
ormergi: +1 for retain default configurations, and change tests were needed, well.. as much possible.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -1,7 +1,11 @@
 #cloud-config
-password: fedora
-chpasswd: { expire: False }
 ssh_pwauth: yes
+chpasswd:
+  users:
+    - name: fedora
+      password: fedora
```
ormergi, line 6: Is'nt this user already exist out-of-the-box?

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -45,6 +45,12 @@ write_files:
     - path: /etc/profile.d/disable-bracketed-paste.sh
       content: |
         bind 'set enable-bracketed-paste off'
+    # Note: systemd>=258 (Fedora 43+) injects OSC 3008/8003 escape sequences
```
ormergi, line 48: How does the issue manifest in tests, how does goexpect breaks?
If it expect for some character to detect the guest prompt this could be easily addressed at the test suite code.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: it was already the state before
follow-ups can improve in case needed (imho it is wrong to move it, it will just complex things)
there is no reason to not have the order as the one we need

btw unless we drop the subdomain tests / find out we can move to alpine but there is no regression, just order change of the same bug, so follow-ups are fine imo

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: no, and it is not a regression, it was already the state, i just changed the format
take into account we even disable `users_groups` long ago, as it caused flakes on tests.
The wrong thing we did before that, was to reconfigure login, and retry (both on e2e), 
this is a classic example why it is better to bake the machine in the right method,
it saved lots of flakes and complexity back then

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: it adds noise characters that breaks the expected prompt,
it is wrong to address it in the tests, the console is already complex enough, we don't need those control characters that just do mess
we did same for other problem on L47
the current method is the most KIS and stable

EDIT
AI answer
```
> The issue manifests as extra OSC escape sequences injected into prompts via `PROMPT_COMMAND` / `PS0` (systemd shell integration on newer Fedora), and that adds noise into the console stream used by `goexpect` prompt matching.  
> This causes prompt detection flakiness/timeouts.
>
> We could try to absorb this in test code (e.g., sanitize escape sequences centrally), but that expands console parsing complexity in KubeVirt tests.  
> Since this is a dedicated e2e image, disabling shell integration in the image is the narrower and more deterministic fix for non-interactive automation.
```

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -75,4 +75,5 @@ runcmd:
   # breaks FQDN: https://bugzilla.redhat.com/show_bug.cgi?id=2038634
   - |
     sudo sed -i 's/hosts:      files myhostname resolve \[\!UNAVAIL=return\] dns/hosts:      resolve \[\!UNAVAIL=return\] files myhostname dns/' /etc/nsswitch.conf
+  - sudo sed -i '/pam_nologin/d' /etc/pam.d/sshd
```
ormergi: Asking to understand:
The image is configured with systemd rules to become ready once cloud-init is finished, and we dont expect logins for `fedora` user until then, right?
Does this change still necessary?
Nevertheless, removing this can be done later.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
ormergi: In this comment I was referring to the `fedora` user, the comment seem to points wrong diff due to pushed changes. It used to exist out-of-the-box in Fedora images, maybe it not longer true.

Regarding `cloud-user`, The workaround that disable creating it is still in place isnt it? (updated by the commit fedora-with-test-tooling: fix users-groups sed pattern)

In general I think we should revisit this workaround (rm users-groups from cloud config file), it was introduced long time ago and I can find the why.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: fwiw AI answer

> This impacts the subdomain/FQDN flows where tests rely on `hostname -f` returning a real FQDN.  
> With Fedora 43 defaults (`resolve` in `hosts:`), `hostname -f` can return only the short hostname, which breaks those scenarios.
>
> I agree we should avoid test-image hacks where possible, but this is a baseline DNS resolution behavior fix in the dedicated e2e image (not a per-test tweak), and it keeps behavior deterministic across all tests that depend on FQDN.  
> Doing this per test would spread resolver workarounds across the suite and increase complexity.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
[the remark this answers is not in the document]
ormergi: >because we don't need always to make sure everything is loaded perfect,

I dont agree with this statement, test assume the guest is ready to work with - all modules are loaded, expected OS services are up (e.g.: cloud-init ready network is ready).
Ensuring the guest is read and in valid state is the baseline even if it require more strict checks.

Thanks for the insights from CI.
Note that its impossible to tell whether the test failed to login due to the "pam no-login" setting or something else, unless I am missing somthign.
Such flakes occur when there is some general slowness in the environment / CI congestion.
I couldn't find the mentioned error in those jobs, is there some log that indicate login failed for this reason?

Anyhow please consider moving this change to separate PR.
It will enable measuring this setting impact and whether it stabilize tests or not. 
Having it as part of this PR would mask the results.

Just to make clear, I am not against this change.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -1,10 +1,17 @@
 #cloud-config
 ssh_pwauth: yes
+users:
+  - name: cloud-user
```
ormergi, line 4: Why do we need to create this user now, does the new cloud-init in F43 fail w/o it?
If not, please consider moving this change to different PR.

---

Regarding https://github.com/kubevirt/kubevirtci/pull/1657#issuecomment-4162361319
I think we should revisit this workaround, it was added long time ago and possibly not needed anymore.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -88,4 +88,11 @@ runcmd:
   - |
     sudo sed -i 's/^hosts:.*/hosts:      files dns myhostname/' /etc/nsswitch.conf
   - sudo sed -i '/pam_nologin/d' /etc/pam.d/sshd
+  # Ensure boot artifacts are refreshed on IBM Z. If cmdline/kernel artifacts drift,
+  # s390x can panic early with unknown-block and disabled-wait.
+  - |
+    if [ "$(uname -m)" = "s390x" ]; then
```
ormergi, line 94: I find this logic, based on the env (cpu arch), introduce additional complexity and inconsistent products of the vm-image-builder. This is a mere config-file lets keep it simple as possible.
The next dev will be tempted to continue with this pattern and make this config file worse.

Please consider replicating this file for s390x arch, and have the caller script decide  which config file to use based on the given arch, similar to what is done with the image URL file.
This should eliminate the need for this additional logic.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
ormergi: Following https://github.com/kubevirt/kubevirtci/pull/1657#issuecomment-4150314392, do we still need this change?

ormergi (commented):
Thanks for the changes

I listed below changes done to retain current behavior for kubevirt tests, these are tech-debt we should eventually resolve:
- Retain classic interface naming 
  Commit "fedora-with-test-tooling: restore classic network interface naming"
  Resolving this require changes in kubevirt tests, and kubevirtci network setup scripts.
- Retain workaround for hostname -f to produce FQDN over the default sorter form
  Commit "fedora-with-test-tooling: fix hostname -f to return FQDN"
  Kubevirt tests should change to query the FQDN in form it expects in a way we can leave default settings.
- Provide two flavors of `netcat`
  Commit: "fedora-with-test-tooling: keep OpenBSD nc default and add nmap-ncat"
  Kubevirt tests should change and align using single netcat flavor, the one that comes with the image.

The above were merely surfaced by this PR changes, there might be additional workarounds.
As agreed earlier those should be tracked and resolved on follow-up PRs.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
ormergi: Thanks you can resolve this

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
ormergi: The comment point the worng lines due to pushed changes, I was refering to `fedora` user.
I was under impression the fedora user comes out-of-the-box.

You can resolve this thread.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -69,7 +69,7 @@ runcmd:
   - sudo systemctl enable qemu-guest-agent.service
   - sudo hostnamectl set-hostname ""
   - sudo hostnamectl set-hostname "" --transient
-  - sudo sed -i /users-groups/d /etc/cloud/cloud.cfg
+  - sudo sed -i '/users[-_]groups/d' /etc/cloud/cloud.cfg
```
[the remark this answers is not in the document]
ormergi: Thanks, you can resolve this thread.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: it might save time, and might be required, i didn't run without it, but it doesn't hurt
all the tests pass with it, it narrow the support to just the ones we require

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: i think it wll be harder to maintain 2 files
for one simple change it doesnt worth it imo

EDIT 
btw it already had one that i removed and simplifed
https://github.com/kubevirt/kubevirtci/pull/1657/commits/7081fde6db59a2b2706855507a105d0bfeec023a#diff-60e708f2e40cf0f5d49420a079fdd43d2a2b2691cada7e745ee1033d195d8a80L71

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: > Why do we need to create this user now, does the new cloud-init in F43 fail w/o it?

the systemctl modules fail because they need cloud-user
like this (the 3rd was just because temporary mistake)
```
[systemd]
Failed Units: 3
  cloud-final.service
  cloud-init-main.service
  wait-for-cloud-init.service
```

> Regarding https://github.com/kubevirt/kubevirtci/pull/1657#issuecomment-4162361319
I think we should revisit this workaround, it was added long time ago and possibly not needed anymore.

not in the scope of this PR,
from quick glance with AI, it is still relevant, but just checking deeper will tell if it is

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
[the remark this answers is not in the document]
oshoval: i did see this error, and investigated deeper, F39 presented this regression
hence tests started to flake about console login, since some tests just need the login and not to wait for full boot with all the modules,
this way we can save hours for each PR update,
tests that need GA / cloud-init are waiting for it, tests which doesn't don't need to pay for everything to be loaded, its overkill that this `nologin` introduces, and we see the impact and the price

imho we don't need to split it, it just creates another precious cycle, lets see other opinions please
especially since we agree we aren't against the change itself

EDIT 
it also ran several time and was stable, on all sigs / archs

p.s
`- sudo sed -E -i '/\bpam_nologin\.so\b/d' /etc/pam.d/sshd`
is bit more accurate, and i validated it exists and work on both current F39 and F43

created https://github.com/kubevirt/kubevirtci/pull/1681 but imho we should save precious time,
there isn't a risk here and just benefit

oshoval:
> * Retain classic interface naming
>   Commit "fedora-with-test-tooling: restore classic network interface naming"
>   Resolving this require changes in kubevirt tests, and kubevirtci network setup scripts.

> * Retain workaround for hostname -f to produce FQDN over the default sorter form
> * Commit "fedora-with-test-tooling: fix hostname -f to return FQDN"
>   Kubevirt tests should change to query the FQDN in form it expects in a way we can leave default settings.

Those 2 bullets are one, and it was prior to this PR, also please lets hear other opinions first

> * Provide two flavors of `netcat`
>   Commit: "fedora-with-test-tooling: keep OpenBSD nc default and add nmap-ncat"
>   Kubevirt tests should change and align using single netcat flavor, the one that comes with the image.

Kubevirt itself already use the 2 variants, since each support different flag as far as i remember
Please see the Kubevirt PR mentioned on the PR desc, as both should be reviewed together
moreover the change is mandatory, since cloud-init need openbsd one, and Kubevirt need both
The PR even simplify the code as best it can do (it already had both with custom repo and one on the dnf)
https://github.com/kubevirt/kubevirtci/pull/1657/commits/7081fde6db59a2b2706855507a105d0bfeec023a#diff-60e708f2e40cf0f5d49420a079fdd43d2a2b2691cada7e745ee1033d195d8a80R68

anyhow i am glad we consider follow-ups, as it will allow to discuss without blocking deeper

oshoval:
Thanks, answered comments, hope i didnt miss anything

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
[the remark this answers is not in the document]
ormergi: @EdDev please vote.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
ormergi: Thanks for the ref.
Not in this PR yes.

ormergi:
> Those 2 bullets are one, and it was prior to this PR, also please lets hear other opinions first
> 

Bad formatting, fixed.

> > * Provide two flavors of `netcat`
> >   Commit: "fedora-with-test-tooling: keep OpenBSD nc default and add nmap-ncat"
> >   Kubevirt tests should change and align using single netcat flavor, the one that comes with the image.
> 
> Kubevirt itself already use the 2 variants, since each support different flag as far as i remember Please see the Kubevirt PR mentioned on the PR desc, as both should be reviewed together moreover the change is mandatory, since cloud-init need openbsd one, and Kubevirt need both The PR even simplify the code as best it can do (it already had both with custom repo and one on the dnf) [7081fde#diff-60e708f2e40cf0f5d49420a079fdd43d2a2b2691cada7e745ee1033d195d8a80R68](https://github.com/kubevirt/kubevirtci/commit/7081fde6db59a2b2706855507a105d0bfeec023a#diff-60e708f2e40cf0f5d49420a079fdd43d2a2b2691cada7e745ee1033d195d8a80R68)
> 
> anyhow i am glad we consider follow-ups, as it will allow to discuss without blocking deeper

I revisited the PR on k/k as well, unless I am missing something, it should work with single netcat flavor.

Yes these are for follow-ups, could you please post tracker issues?

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
ormergi: Having it structure like this could allow posting similar changes like this PR in iterations.
And allow s390x contributors / experts review and maintain this part in the project.

But yes, on the other hand it require additional changes in the callers scripts.

@EdDev WDYT? (tagging you because you are also reviewing this :))

oshoval:
> > Those 2 bullets are one, and it was prior to this PR, also please lets hear other opinions first
> 
> Bad formatting, fixed.
> 
> > > * Provide two flavors of `netcat`
> > >   Commit: "fedora-with-test-tooling: keep OpenBSD nc default and add nmap-ncat"
> > >   Kubevirt tests should change and align using single netcat flavor, the one that comes with the image.
> > 
> > 
> > Kubevirt itself already use the 2 variants, since each support different flag as far as i remember Please see the Kubevirt PR mentioned on the PR desc, as both should be reviewed together moreover the change is mandatory, since cloud-init need openbsd one, and Kubevirt need both The PR even simplify the code as best it can do (it already had both with custom repo and one on the dnf) [7081fde#diff-60e708f2e40cf0f5d49420a079fdd43d2a2b2691cada7e745ee1033d195d8a80R68](https://github.com/kubevirt/kubevirtci/commit/7081fde6db59a2b2706855507a105d0bfeec023a#diff-60e708f2e40cf0f5d49420a079fdd43d2a2b2691cada7e745ee1033d195d8a80R68)
> > anyhow i am glad we consider follow-ups, as it will allow to discuss without blocking deeper
> 
> I revisited the PR on k/k as well, unless I am missing something, it should work with single netcat flavor.
> 
> Yes these are for follow-ups, could you please post tracker issues?

I disagree about the netcat and the FQDN, so i prefer to not open an issue on them yet,
Moreover both prone to be prior to this PR.
lets first conclude about them and then we see.
(about `nc` can double check, EDIT - please see comment below)

about the first, its also not a regression in the tests themselves, but follow-up was agreed.
however opening an issue without someone that is dedicated to it, adds noise, so
imo i don't see rush to open now, if you want please go ahead.
Once someone is committed he can open the issue / fix etc, else it consumes people time

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: p.s
it is so small that in the future it might even went away if we find better way to fix it, or if its due to a bug that will be resolved on fedora itself or on our end, this is another reason why it is over end for it,
especially given that we had one already and replaced it with different one

it also requires double changes if there are 2 scripts, i am strongly against untangling

oshoval:
both ncat variants were installed and used before this PR,
and we do need them, as each one support different set of unique flags, beside the flags that both support.
This PR just simplified the install and adapted it to the cloud-init requirement of F43 (it needs openBSD for the `-U`)
https://github.com/canonical/cloud-init/pull/5933

# KubeVirt `nc` / `ncat` Usage Matrix

This document summarizes current command-level usages and whether they depend on OpenBSD netcat semantics, nmap-ncat semantics, or generic behavior.

## Current Usage Matrix

<table>
  <thead>
    <tr>
      <th>Path</th>
      <th>Command pattern</th>
      <th>Semantic dependency</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td><code>tests/network/services.go</code></td>
      <td><code>ncat ... --sh-exec</code>, <code>ncat ... --no-shutdown</code></td>
      <td>nmap-ncat required (explicit)</td>
    </tr>
    <tr>
      <td><code>tests/vmi_vsock_test.go</code></td>
      <td><code>nc -vl ...</code> (was <code>netcat-openbsd</code>)</td>
      <td>OpenBSD netcat behavior in Fedora test-tooling VM (<code>nc</code> alias)</td>
    </tr>
    <tr>
      <td><code>tests/libnet/job/job.go</code></td>
      <td><code>nc ... --no-shutdown</code></td>
      <td>nmap-like behavior (depends on <code>nc</code> alias in the utility image)</td>
    </tr>
    <tr>
      <td><code>tests/network/bindingplugin_slirp.go</code></td>
      <td><code>nc ... --recv-only</code></td>
      <td>nmap-like option</td>
    </tr>
    <tr>
      <td><code>tests/network/vmi_networking.go</code></td>
      <td><code>echo test | nc ... -i -w</code></td>
      <td>generic/simple <code>nc</code> usage</td>
    </tr>
    <tr>
      <td><code>tests/network/bindingplugin_passt.go</code></td>
      <td><code>echo test | nc ... -i -w</code></td>
      <td>generic/simple <code>nc</code> usage</td>
    </tr>
    <tr>
      <td><code>tests/libnet/vmnetserver/servers.go</code></td>
      <td><code>nc ... -klp ...</code></td>
      <td>listener usage; image-dependent semantics</td>
    </tr>
  </tbody>
</table>

## Before vs After the 6 Commits of https://github.com/kubevirt/kubevirt/pull/17257

<table>
  <thead>
    <tr>
      <th>Timeframe</th>
      <th>Uses OpenBSD side?</th>
      <th>Uses nmap side?</th>
      <th>Notes</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Before the 6 commits</td>
      <td>Yes</td>
      <td>Yes</td>
      <td><code>tests/vmi_vsock_test.go</code> used explicit <code>netcat-openbsd</code>; other paths already used nmap-style flags via <code>nc</code> in appropriate images.</td>
    </tr>
    <tr>
      <td>After the 6 commits</td>
      <td>Yes</td>
      <td>Yes</td>
      <td>vsock path uses <code>nc</code> alias; Fedora services path now explicitly uses <code>ncat</code> for <code>--sh-exec</code> / <code>--no-shutdown</code>.</td>
    </tr>
  </tbody>
</table>


btw even if we replace the openbsd one on kubevirt to nmap-ncat (didn't check if fully doable),
the image will still need both, 
since the cloud-init will require openbsd one, and kubevirt nmap-ncat,
in any case, the duality use was the case before the PR

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval:
```
-  # Ensure boot artifacts are refreshed on IBM Z. If cmdline/kernel artifacts drift,
-  # s390x can panic early with unknown-block and disabled-wait.
+  # Keep boot artifacts in sync on zipl-based systems. Bootmap entries can drift
+  # after kernel/initramfs updates, causing early unknown-block or disabled-wait panics.
   - |
-    if [ "$(uname -m)" = "s390x" ]; then
+    if command -v zipl >/dev/null 2>&1; then
       sudo dracut -f --regenerate-all
       sudo zipl
     fi
```
we can do this instead if desired

moreover i am researching a way to eliminate it as a whole, but would glad if we can postpone it for a follow-up

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
EdDev: I find improper to provide AI answers as if we are suppose to be convinced by that.
If *you* are convinced by this argument, *you* can take it on yourself to claim it.

This is *not* convincing me, for the same reasons I raised earlier. It is unclear why the tests are trying to do something using means which are not part of the default distribution.
And if it uses a tweak, then it should be visible when executed, making it a documentation of how to use something.

Please answer my original question in order to have progress here.
> Do you know which tests are effected by this one?

As we already perform the tweak, I am fine to follow up on this, given we know which are the effected tests and an issue is tracking it with the information.
Alternatively, convince me it makes sense we tweak the OS distribution to fit our tests.

oshoval: i already answered above, will post it again
"it affects the subdomain tests, that one that need FQDN `hostname -f` once subdomain is declared"

p.s
added AI, just to show with which direction it suggests as 2nd opinion,
ran it twice, with full comments context and without, both same answer,
moreover it was after i supplied personal answers, anyhow if you don't like it, won't add it

EDIT
do we really need to open an issue ? atm we all are full i believe, so it will just stand there most likely
moreover it is not a regression, and the imho the right place is the current one,
anyhow will open the issue so it will be a discussion playground

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
EdDev: > > I noticed additional types are added and requested. Can you check that no new stuff are in the pipeline?
> 
> where ?

https://github.com/kubevirt/enhancements/pull/101
https://github.com/kubevirt/enhancements/pull/151

But I think all are `NoCloud`.
We need to make this visible in a good fashion in order not to surprise users.
How about we summarize all the tweaks in an image documentation? Maybe a README under the folder?

oshoval:
Hi @dhiller @dollierp 

Once this PR is merged, the machine that provision it must have osinfo db entry for `fedora43`
`virt-install --osinfo list | grep -x fedora43`
can you please check if it is the state?
if needed please run `dnf update osinfo-db`

if this won't help, it means it depends on the OS (i.e F41 for example doesn't have F43 variants),
there are some solutions for such case as well, but lets do step by step please

P.S the requirement of Fx isn't a regression of this PR

Thanks

dollierp:
Hi @oshoval,

> The machine that provision it must have osinfo db entry for `fedora43` `virt-install --osinfo list | grep -x fedora43` can you please check if it is the state? 

I couldn't find any place where the `publish-multiarch-containerdisk.sh` is executed: https://github.com/search?q=org%3Akubevirt%20publish-multiarch-containerdisk&type=code

oshoval:
> Hi @oshoval,
> 
> > The machine that provision it must have osinfo db entry for `fedora43` `virt-install --osinfo list | grep -x fedora43` can you please check if it is the state?
> 
> I couldn't find any place where the `publish-multiarch-containerdisk.sh` is executed: https://github.com/search?q=org%3Akubevirt%20publish-multiarch-containerdisk&type=code

Hi Denis,
thanks, i believe it is trigger manually upon needs ?
maybe worth to ask please Daniel / Brian as they usually do it iiuc

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: Thanks for the links, indeed only `NoCloud`, `ConfigDrive` which are covered.

Opened a PR with all the tweaks (before this PR and including),
as seperate PR, so we can parallel the discussion please.
https://github.com/kubevirt/kubevirtci/pull/1683

EDIT 
double checked, having this change should save, because the default is to search for alternative clouds, and it consumes time, so this change should have benefit in such cases

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: created https://github.com/kubevirt/kubevirtci/issues/1686

amending this PR with missing fallback of the last 2 elements "resolve [!UNAVAIL=return]` for completeness

`sudo sed -i 's/^hosts:.*/hosts:      files dns myhostname resolve [!UNAVAIL=return]/' /etc/nsswitch.conf`

oshoval:
addressed @EdDev's comments
added missing fallbacks for NSS configuration (didn't run everything, tests passed even without fallback, so risk is low, if desired will rerun everything, beside s390x e2e where we need Vamsi to run it)

PR desc missing items like the s390x fix and adding the cloud-user, once we converge, will add those according needs - EDIT fixed

ormergi:
By tracking issues I mean Github issue in this repository for record, we can assign it later. 
Otherwise it will fall between chairs.

Although I dont like some additions of this PR and you insist keeping them, I understand each iteration takes lot of time so I dont have strong feeling about keeping them:
pam_no_login fix
Cloud sources defaults
Logic for s390x arch in cloud config file 

/lgtm

oshoval:
Thanks Or

cc @EdDev in case you want to finish the review please

Hi @dollierp @dhiller 
can you please review ?

Note please we will need https://github.com/kubevirt/kubevirt/pull/17257 to be reviewed as well,
unless we provision an image that doesn't override the current one meanwhile,
(but it means that the code will be different than the image that is used, until the new image is consumed)

Moreover we need to find out which machine provision this image, and if it is done manually
cc @brianmcarey maybe you remember please ?

brianmcarey:
> Moreover we need to find out which machine provision this image, and if it is done manually cc @brianmcarey maybe you remember please ?

It was manually done previously but that was due to how infrequently it was updated in the past. It's probably worth creating a postsubmit job to publish it.

oshoval:
Thanks @brianmcarey !
Great to hear from you

Lets consider that please as follow-up,
was it done on on of the CI machines manually, or on other dedicated machine ?

EdDev (approved):
Thank you.

oshoval:
Thanks @EdDev @ormergi 
can you please check https://github.com/kubevirt/kubevirt/pull/17257 as well ?
the counter part of Kubevirt

oshoval:
Hi @dhiller @dollierp 
would you able to review this one please?

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
dhiller, line 48: nit: not sure whether we need the comment also here, I believe the git commit message describes it well enough?

dhiller (approved):
/approve

One nit inline, but looks great enough that you can disregard it - your decision.

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: i think it can stay inline as well, it is easier than checking the commit message

oshoval:
Thanks Daniel
once it is merged + https://github.com/kubevirt/kubevirt/pull/17257 is merged
we will need to recreate the image please

the machine that provision it 
will need to have `virt-install --osinfo list | grep -x fedora43`
if it doesn't need to try `dnf update osinfo-db`
but if the OS is too old, let me know please and we will try to find a solution,
hopefully without updating the OS there

until Kubevirt side is merged, it means that the code will reflect the new yet to be provisioned image
so we better have the old one intact meanwhile

oshoval:
on some machines compiling arm with emulation hangs
usually i had at least one machine where it worked, but now it hanged on that one as well

this PR fix it at least on some of the OS (F42)
F41 has more problems, that it doesn't fix
https://github.com/kubevirt/kubevirtci/pull/1694

but i am checking there isn't other regression

PR #1665, opened 2026-04-01
cluster-provision: Move v1.37 to CentOS Stream 10

lyarwood:
## Summary

- Updates the `base` file for v1.36 to use CentOS Stream 10
- Removes the default `PROVISION_CENTOS_VERSION` value in `provision.sh` so the `base` file is respected when the variable is not explicitly set
- Fixes the gocli to read the `base` file when `PROVISION_CENTOS_VERSION` is not set, instead of hardcoding a default of CentOS Stream 9

## Related

- https://github.com/kubevirt/enhancements/issues/210

PR #1695, opened 2026-05-07
cluster-provision: Add eth0 NM connection for CentOS 10

lyarwood:
## Summary

- `NetworkManager-config-server` (installed during linux provisioning) sets `no-auto-default=*`, preventing auto-DHCP on unconfigured interfaces
- CentOS 9 works because cloud-init's `sysconfig` renderer creates `/etc/sysconfig/network-scripts/ifcfg-eth0` which persists across reboots
- CentOS 10 uses the `network-manager` renderer which does not create a persistent NM keyfile connection profile, so after virt-sysprep resets the machine-id the VM boots without any network configuration and never acquires an IP
- This causes SSH timeouts when using the pre-built centos10 base image with `PHASES=k8s` (e.g. `check-provision-k8s-1.36` in #1665)

The fix adds a persistent NM keyfile connection (`eth0.nmconnection`) for CentOS 10 after installing `NetworkManager-config-server`.

## Test plan

- [ ] `check-provision-k8s-1.34-centos10` still passes
- [ ] Once merged, periodic centos10 bump rebuilds the base image with the fix
- [ ] `check-provision-k8s-1.36` in #1665 passes with the rebuilt base image

/cc @dhiller

### DIFF
--- a/cluster-provision/k8s/1.36/provision.sh
+++ b/cluster-provision/k8s/1.36/provision.sh
@@ -99,5 +99,25 @@ fi
 
 dnf install -y NetworkManager NetworkManager-ovs NetworkManager-config-server
 
+# NetworkManager-config-server sets no-auto-default=* which prevents auto-DHCP
+# on unconfigured interfaces. CentOS 9 has ifcfg-eth0 from cloud-init but
+# CentOS 10 uses keyfile format and has no persistent connection profile.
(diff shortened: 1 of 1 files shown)


dhiller (approved):
/approve

/hold to be removed after check-provision-k8s-1.36 is successful.

PR #1704, opened 2026-05-14
cluster-provision: Switch CRI-O 1.36 to stable repository

lyarwood:
CRI-O 1.36 stable has been released. Switch the k8s 1.36
provisioning from the prerelease openSUSE OBS repository to the
stable GCS mirror, matching the pattern used by 1.34 and 1.35.

Depends on https://github.com/kubevirt/project-infra/pull/5062 to
populate the GCS mirror with the stable CRI-O 1.36 packages.

### DIFF
--- a/cluster-provision/k8s/1.36/k8s_provision.sh
+++ b/cluster-provision/k8s/1.36/k8s_provision.sh
@@ -68,10 +68,10 @@ function pull_container_retry() {
 
 export CRIO_VERSION=1.36
 cat << EOF >/etc/yum.repos.d/devel_kubic_libcontainers_stable_cri-o_${CRIO_VERSION}.repo
-[isv_cri-o_prerelease_v${CRIO_VERSION}]
-name=CRI-O v${CRIO_VERSION} (Prerelease) (rpm)
+[isv_cri-o_stable_v${CRIO_VERSION}]
+name=CRI-O v${CRIO_VERSION} (Stable) (rpm)
 type=rpm-md
-baseurl=https://download.opensuse.org/repositories/isv:/cri-o:/prerelease:/v${CRIO_VERSION}:/build/rpm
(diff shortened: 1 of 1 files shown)


cluster-provision/k8s/1.36/k8s_provision.sh
```diff
@@ -68,10 +68,10 @@ function pull_container_retry() {
 
 export CRIO_VERSION=1.36
 cat << EOF >/etc/yum.repos.d/devel_kubic_libcontainers_stable_cri-o_${CRIO_VERSION}.repo
-[isv_cri-o_prerelease_v${CRIO_VERSION}]
-name=CRI-O v${CRIO_VERSION} (Prerelease) (rpm)
+[isv_cri-o_stable_v${CRIO_VERSION}]
```
oshoval, line 71: Thanks,
might worth to have the 2 blocks, with a var that toggle between them 
stable/pre-release

since we need it every new release?

lyarwood: Thanks for the suggestion. I'd prefer to keep this implicit rather than adding a toggle — the switch from prerelease to stable is a one-time change per release and matches the pattern used by 1.34 and 1.35. The extra abstraction would add complexity for something that only happens once per k8s release.

Note this PR depends on https://github.com/kubevirt/project-infra/pull/5062 landing and the weekly mirror job running to populate the GCS mirror with stable CRI-O 1.36 packages. Marking as draft until then. The CI failures here are unrelated (VM boot flake on the required job, NBD issue on alpine).

oshoval: theoretically, it could check if a stable exists and use that, else use pre automatically
and then we never touch it 

but np

oshoval:
Thanks
/lgtm

oshoval:
btw how come we even need to mirror those CR-IO ?
it worked with the pre-release which wasn't mirrored,
i guess it is just a classic mirror in case the real is NA

brianmcarey (commented):
/approve

PR #1706, opened 2026-05-18
fix(alpine-cloud-init): enable optional network backend

nestoracunablanco:
Podman's rootful mode with systemd-resolved disables internal DNS.
Therefore, the user has the option to enable another network
backend if necessary.

Assisted-by: IBM Bob
Co-Authored-By: Or Shoval <oshoval@redhat.com>

### DIFF
--- a/cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
+++ b/cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
@@ -40,7 +40,14 @@ if [ ! -f alpine-make-vm-image ]; then
     chmod 755 alpine-make-vm-image
 fi
 
-podman run --rm -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
+# Users can set the network backend by setting the environment variable PODMAN_NETWORK_BACKEND
+# for example: export PODMAN_NETWORK_BACKEND=slirp4netns
+PODMAN_NETWORK_ARG=""
+if [ -n "${PODMAN_NETWORK_BACKEND}" ]; then
+    PODMAN_NETWORK_ARG="--network ${PODMAN_NETWORK_BACKEND}"
+fi
+
+podman run --rm ${PODMAN_NETWORK_ARG} -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
 ./alpine-make-vm-image \
     --image-format qcow2 \
     --image-size $IMAGE_SIZE \


cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
```diff
@@ -40,7 +40,7 @@ if [ ! -f alpine-make-vm-image ]; then
     chmod 755 alpine-make-vm-image
 fi
 
-podman run --rm -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
+podman run --rm --network=host -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
```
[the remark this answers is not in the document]
nestoracunablanco: Done.

nestoracunablanco:
@ormergi @oshoval what do you think about this change?

oshoval (approved):
Thanks

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
```diff
@@ -40,7 +40,7 @@ if [ ! -f alpine-make-vm-image ]; then
     chmod 755 alpine-make-vm-image
 fi
 
-podman run --rm -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
+podman run --rm --network=host -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v "$(pwd)":"$(pwd)":z -w "$(pwd)" alpine ash -c "
```
EdDev: AFAIU this breaks network isolation and can put the hosting OS at risk, in the sense that its network stack will be messed up by the container.

We need a very good reason with no alternatives in sight to hack this.

oshoval: ran it locally fwiw, (and used that trick on some other cases, but each script has its own story),
as long as the script doesn't have something intrusive (didn't check, Nestor might want to) i think it should be safe,
unless you still prefer to find a tighter solution

even if something bad happen, we will know and fix it, usually reboot fix it as well, 
unless it is persistent but i believe it is less probable

EdDev: If this script needs to run on a BM CI machine, locally on my machine or anything that should continue working after the script finishes, it is exposed to network side effects.

I do not think you can reason with "the script doesn't have something intrusive", as we do not know how the script will evolve.

oshoval: Please check this alternative
add exit 1 if `ALPINE_MIRROR_IP` is NA

CI might not required / forbid slirp4netns, if needed please make it configurable,
on my machine i needed it, but only because the whole script require sudo for nbd, otherwise it wouldn't,
so you might want to take this into account please

```
diff --git a/cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh b/cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
index 50b4b325..2917d6e2 100755
--- a/cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
+++ b/cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
@@ -40,7 +40,13 @@ if [ ! -f alpine-make-vm-image ]; then
     chmod 755 alpine-make-vm-image
 fi
 
-podman run --rm -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
+ALPINE_MIRROR_IP="$(getent ahostsv4 dl-cdn.alpinelinux.org | awk 'NR==1{print $1}')"
+PODMAN_ADD_HOST_ARG=""
+if [ -n "${ALPINE_MIRROR_IP}" ]; then
+    PODMAN_ADD_HOST_ARG="--add-host dl-cdn.alpinelinux.org:${ALPINE_MIRROR_IP}"
+fi
+
+podman run --rm --network slirp4netns ${PODMAN_ADD_HOST_ARG} -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
 ./alpine-make-vm-image \
     --image-format qcow2 \
     --image-size $IMAGE_SIZE \
```

nestoracunablanco: Thanks for the discussion. Implemented @oshoval suggestion. Please let me know what do you think.

nestoracunablanco: Ok, that was fast: slirp4netns is not available. Will take a look later.

EdDev: You should probably try [passta](https://passt.top/passt/about/#pasta-pack-a-subtle-tap-abstraction).

oshoval: on CI, podman network backend was updated to netavark from slirp4netns
i believe it is related

also please note, on CI you might can fully drop the `--network slirp4netns`,
it is required locally, only when i use sudo,
so to support local we can make it customizable with default NA, if CI doesn't need it,
or to fix deeper the root need to be scoped only to NBD related (not sure if possible)

nestoracunablanco: Ok, the new change makes it configurable via an environment variable. What do you think?

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
```diff
@@ -40,7 +40,16 @@ if [ ! -f alpine-make-vm-image ]; then
...
 
-podman run --rm -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
+
+# Network backend configuration for podman
+# Set PODMAN_NETWORK_BACKEND to customize (e.g., "host" or "slirp4netns" for local sudo usage)
+# Default: empty (uses podman's default)
+PODMAN_NETWORK_ARG=""
+if [ -n "${PODMAN_NETWORK_BACKEND}" ]; then
+    PODMAN_NETWORK_ARG="--network ${PODMAN_NETWORK_BACKEND}"
+fi
```
oshoval: please change to auto detect
if [[ $CI == false ]]: then 
    PODMAN_NETWORK_ARG=slirp4netns
fi

see it works locally, and then the comment can be dropped, as host is not encouraged

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
```diff
@@ -40,7 +40,16 @@ if [ ! -f alpine-make-vm-image ]; then
...
+
+podman run --rm ${PODMAN_NETWORK_ARG} -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v "$(pwd)":"$(pwd)":z -w "$(pwd)" alpine ash -c "
```
oshoval: please consider to do minimal changes, or to split the cosmetics to 2nd commit

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
EdDev: > PODMAN_NETWORK_ARG=slirp4netns

AFAIK slirp is supposed to be deprecated and replaced by passta. Why is it still mentioned?

> an optional PODMAN_NETWORK_BACKEND environmental
variable supports local sudo usage.

This implies that we do not have a good solution and pass it to the user that runs the script. While that is possible, it is not that nice.

Could you please summarize on why we had to reach this solution? And why other options are not a fit?

oshoval: on CI the backend is netavark, hence it works without explicit network
https://github.com/kubevirt/kubevirtci/pull/1706#discussion_r3317636515

locally worth to try --network netavark / passt/pasta as well, 
if it works it can be set as the default for both CI / local

nestoracunablanco: Hi, I implemented the way that @oshoval suggested. @EdDev Regarding pasta, I tried it locally, but it does not work in rootfull mode. And my assumption is that this script must run in privileged mode, even on a local workstation.
Let me know what do you think.

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
nestoracunablanco: Deleted, will be done in an extra PR.

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
```diff
@@ -40,7 +40,14 @@ if [ ! -f alpine-make-vm-image ]; then
     chmod 755 alpine-make-vm-image
 fi
 
-podman run --rm -v /lib/modules:/lib/modules -v /dev:/dev --privileged -v $(pwd):$(pwd):z alpine ash -c "cd $(pwd) &&
+
+# Sets the network backend to slirp4netns for local usage
+PODMAN_NETWORK_ARG=""
+if [ "${CI:-}" != "true" ]; then
```
oshoval: nits:

simply this ?
`if [ "${CI}" != "true" ]; then`

please please new line on L43

beside that lgtm imo

seems NBD requires sudo, i think the current fix is good enough

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
oshoval: lgtm from my side
see please https://github.com/kubevirt/kubevirtci/pull/1706#discussion_r3341829926
thanks

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
nestoracunablanco: Done. Thanks for the review.

oshoval:
Thanks

/lgtm

cluster-provision/images/vm-image-builder/alpine-cloud-init/create-image.sh
oshoval: Btw i wonder if having a container image will solve some of those problems (nbd, root, network),
anyhow imho, we don't need to block

oshoval:
Hi, suddenly it back to work without the PR locally, in case you want to try as well

nestoracunablanco:
Hi, thanks for the hint. Tried it locally, and I still need the PR to make it work. Will keep an eye on it.

oshoval:
> Hi, thanks for the hint. Tried it locally, and I still need the PR to make it work. Will keep an eye on it.

please consider then to make the field optional, whoever need it can export it locally

oshoval:
Thanks

/lgtm

nirdothan:
@EdDev, looking at the conversation it doesn't seem that your questions have been fully addressed. IIUC you are still concerned about network isolation.
If I'm wrong and you are satisfied with the solution, please LGTM to indicate that.

EdDev (commented):
It feels like the problem domain is very small and the solution is opening the door to nasty options.

We want to be as careful as we can, to keep safe and simple. I admit that the current solution seems simple, but the safety is questionable.
Safety is dependent on where it runs, i.e. we defer the responsibility on the one running it.

Lets start with first defining the problem in a more detailed manner, so future us will not strangle to understand why we need such an unsafe change.

Is the below description correct?

> On systems using systemd-resolved, the host's /etc/resolv.conf points to 127.0.0.53 (the stub listener). When rootful podman creates a container with bridge networking, the
> container gets its own network namespace. 127.0.0.53 is a loopback address — it only exists in the host's network namespace, not the container's. So DNS fails inside the container,
> and alpine-make-vm-image can't resolve dl-cdn.alpinelinux.org to fetch packages.

I have tried to process it through claude. Please let me know if this makes sense (or all lies, like many times happen to me). Sorry that I could not check it myself.

As I see it, there are 3 possible paths here:
1. Run podman as rootless.
2. Keep podman as rootful but fix the problem by using the right drivers/config.
3. Use the external network config as suggested so far (in order to use the host net namespace).

### Rootless

The root cause: the script calls modprobe nbd which requires root — but does the podman run itself need to run as root? If only the modprobe step needs sudo, we
could separate them:

```
sudo modprobe nbd
podman run --rm ...   # as the current non-root user
```

Rootless podman uses pasta/slirp4netns by default and handles DNS correctly without any extra configuration. If this works, the DNS problem disappears entirely and this PR becomes unnecessary.

### Rootful
If rootful is truly required for the podman run: the right fix is a dedicated network rather than making the network backend configurable. Since Podman 4.0,
creating a custom network enables aardvark-dns which injects a proper DNS server at the bridge gateway IP, so 127.0.0.53 is never seen by the container:

```
podman network create alpine-build
podman run --rm --network alpine-build ...
```

Could you try the rootless approach first and report back? And if that doesn't work, could you share your `podman --version, podman info | grep -A5 networkBackend`, and `ls -la /etc/resolv.conf`?
That would help us understand why the default netavark setup isn't handling DNS correctly on your machine.

nestoracunablanco:
Hi @EdDev, thanks for your comment. The script must be run in rootfull mode since, even if we sudo modprobe nbd, access to /dev/nbd[0-15] still requires root privileges—or we'd need to grant permissions to the devices via chmod, which I'd prefer to avoid.
Interestingly, I can no longer reproduce the error after switching to a new workstation, and everything works without issues. The problem was most likely a misconfiguration on my end.
As a result, I'm closing this PR. Thanks for your help!

PR #1710, opened 2026-05-20
gocli, net-res-injector: Set webhook fail policy to Ignore

nirdothan:

Currently various tests flake randomly with communication problems attempting to call the network-resources-injector webhook [1]. The vast majority of the failing scenarios do not even need to call it (only macvtap and bridge-marker).
The main consumer is SR-IOV however those tests run with the sriov-kind provider, which is not affected by this change.

This quick and dirty solution will spare flakes from unrelated tests. Should this problem occur in relevant tests they will most likely fail anyway, although the failure reason could be harder to identify. Look for scheduling issues with missing custom resource requests/limits in virt-launcher pod.
A proper solution could be introduced by manipulating the MutatingWebhookConfiguration manifest to increase timeout, however we currently do not explicitly deploy that manifest, and rather rely on the network-resources-injector's server to inject that object in its init container.

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/17805/pull-kubevirt-e2e-k8s-1.35-sig-network/2056297617125543936

Fixes #

**Checklist**

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-provision/gocli/opts/network_resources_injector/manifests/server.yaml
+++ b/cluster-provision/gocli/opts/network_resources_injector/manifests/server.yaml
@@ -45,6 +45,7 @@ spec:
         - -logtostderr
         - -v=2
         - -insecure
+        - -failure-policy=Ignore
         env:
         - name: NAMESPACE
           valueFrom:


orelmisan (commented):
Thank you for the PR @nirdothan.

Do we understand the root cause here?

nirdothan:
> Thank you for the PR @nirdothan.
> 
> Do we understand the root cause here?

No. Nothing wrong can be seen in the cluster. The service is running. It is assumed to be a communication failure.

oshoval:
what if you stop deploying it on lanes that not require it, is it possible?
(i guess not, due to bm, macvtap ?)
worth to consult Sebastian / Miguel

ignore isn't optimal
can you please try something easy instead ?
lets dive bit deeper if possible please to map what exactly happens, when, and the options and then we can choose imo,

note that soon we will move to emulated sriov on vm based providers, and we will need it when using DP + CNI,
but not when we move to DRA

oshoval:
```
Unexpected Warning event received: testvmi-kwwxn-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx,1907b66d-2a4d-4fae-9b78-855445507b4e: Error creating pod: Internal error occurred: failed calling webhook "network-resources-injector-mutating-config.k8s.cni.cncf.io": failed to call webhook: Post "https://network-resources-injector-service.kube-system.svc:443/mutate?timeout=10s": EOF
```
It is just a warning hence
just add this to warning ignore list, there is build in retry i think
the test is too strict, but we have mechanism to ignore specific warnings

orelmisan:
> > Thank you for the PR @nirdothan.
> > Do we understand the root cause here?
> 
> No. Nothing wrong can be seen in the cluster. The service is running. It is assumed to be a communication failure.

The network-resources-injector is a stable project.
I suggest investing some time looking into this.

oshoval:
> The network-resources-injector is a stable project. I suggest investing some time looking into this.

as Nir told offllne, 10s is too little, and aint straight forward to increase,
it is wrong to be strict for every warning, this why we have the mech for ignoring selected warnings, it proved itself

orelmisan:
Can we increase the timeout instead?

nirdothan:
> Can we increase the timeout instead?

The go code provides a flag to ignore errors but it does not expose a timeout flag. 
As I wrote in the description, editing the MutatingWebhookConfiguration is more work, because we currently don't specify it and rely on `network-resources-injector` init container to auto generate it with its built in logic to generate cert/ca. 
I can do that as a followup (but with static credentials - not with cert-manager).

oshoval:
imo it doesn't worth to effort, NRI here is a helper in the development, not the DUT,
simple timeout warning better to keep things simple,
or to open them an issue asking for an easy way to config it

orelmisan:
> imo it doesn't worth to effort, NRI here is a helper in the development, not the DUT, simple timeout warning better to keep things simple, or to open them an issue asking for an easy way to config it

The whole point of VEP 106 was to outsource work to the network-resources-injector so it is part of the system under test, not just a development helper.
Ignoring its failures or integration issues means we're closing our eyes to real problems in a path we chose to depend on.

oshoval:
> > imo it doesn't worth to effort, NRI here is a helper in the development, not the DUT, simple timeout warning better to keep things simple, or to open them an issue asking for an easy way to config it
> 
> The whole point of VEP 106 was to outsource work to the network-resources-injector so it is part of the system under test, not just a development helper. Ignoring its failures or integration issues means we're closing our eyes to real problems in a path we chose to depend on.

it is a timeout warning only, there is a build in retry mechanism if we ignore just this problem,
not saying to ignore the webhook errors
it is wrong to be that strict for every warning, especially without allowing breaking it to steps, and doing follow-ups,
which allows to process, think in parallel (which not once yield interesting findings), and prioritize

wrt VEP 106, if it didnt happen on D/S, fortify U/S doesn't have value
U/S CI is known to be congested, so it is sensitive to timeouts

orelmisan:
> > > imo it doesn't worth to effort, NRI here is a helper in the development, not the DUT, simple timeout warning better to keep things simple, or to open them an issue asking for an easy way to config it
> > 
> > 
> > The whole point of VEP 106 was to outsource work to the network-resources-injector so it is part of the system under test, not just a development helper. Ignoring its failures or integration issues means we're closing our eyes to real problems in a path we chose to depend on.
> 
> it is a timeout warning only, there is a build in retry mechanism if we ignore just this problem, not saying to ignore the webhook errors it is wrong to be that strict for every warning, especially without allowing breaking it to steps, and doing follow-ups, which allows to process, think in parallel (which not once yield interesting findings), and prioritize
> 
> wrt VEP 106, if it didnt happen on D/S, fortify U/S doesn't have value U/S CI is known to be congested, so it is sensitive to timeouts

Can we have a clear explanation of the issue? The scenario, what fails, why and why do you think its safe to do what you are suggesting?

oshoval:
> > > > imo it doesn't worth to effort, NRI here is a helper in the development, not the DUT, simple timeout warning better to keep things simple, or to open them an issue asking for an easy way to config it
> > > 
> > > 
> > > The whole point of VEP 106 was to outsource work to the network-resources-injector so it is part of the system under test, not just a development helper. Ignoring its failures or integration issues means we're closing our eyes to real problems in a path we chose to depend on.
> > 
> > 
> > it is a timeout warning only, there is a build in retry mechanism if we ignore just this problem, not saying to ignore the webhook errors it is wrong to be that strict for every warning, especially without allowing breaking it to steps, and doing follow-ups, which allows to process, think in parallel (which not once yield interesting findings), and prioritize
> > wrt VEP 106, if it didnt happen on D/S, fortify U/S doesn't have value U/S CI is known to be congested, so it is sensitive to timeouts
> 
> Can we have a clear explanation of the issue? The scenario, what fails, why and why do you think its safe to do what you are suggesting?

all i know is that there is a timeout warning, for accessing the webhook,
but this all what we need to know, we are very strict by default, this is an event that caught by VMI watcher,
if we just ignore this specific event, on the watcher, there is a natural reconcile, it will simply retry, since k8s
is eventually consistent,
even simple I/O congestion on CI, when a burst of jobs are released on the same time, can make some pods to starve,
it worth trying to just ignore this error, see if it self heal, asserting on each event, is noisy and costy,
therefore we implemented the selective ignore,
```
Unexpected Warning event received: testvmi-kwwxn-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx,1907b66d-2a4d-4fae-9b78-855445507b4e: Error creating pod: Internal error occurred: failed calling webhook "network-resources-injector-mutating-config.k8s.cni.cncf.io": failed to call webhook: Post "https://network-resources-injector-service.kube-system.svc:443/mutate?timeout=10s": EOF
```

you can even skip it only on the tests that need it, unless it is in general,
and then it is fine to ignore it all over imo for this specific webhook, webhooks add complexity
you can even do it just for upstream if you want to be strict (but imo it is wrong)

oshoval:
you might want to check on D/S what is the default timeout, if it is more there, it can sweeten the pill

nirdothan:
> you might want to check on D/S what is the default timeout, if it is more there, it can sweeten the pill

That's good advice and the result is interesting:
```
k get mutatingwebhookconfigurations.admissionregistration.k8s.io network-resources-injector-config -oyaml
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
  annotations:
    service.beta.openshift.io/inject-cabundle: "true"
  creationTimestamp: "2026-05-07T19:11:38Z"
  generation: 2
  name: network-resources-injector-config
  resourceVersion: "70859"
  uid: 95fb754e-e7f3-4f97-be40-dad3389a85df
webhooks:
- admissionReviewVersions:
  - v1
  - v1beta1
  clientConfig:
    caBundle: <REDUCTED>
    service:
      name: network-resources-injector-service
      namespace: openshift-sriov-network-operator
      path: /mutate
      port: 443
  failurePolicy: Ignore
  matchPolicy: Equivalent
  name: network-resources-injector-config.k8s.io
  namespaceSelector: {}
  objectSelector: {}
  reinvocationPolicy: Never
  rules:
  - apiGroups:
    - apps
    - ""
    apiVersions:
    - v1
    operations:
    - CREATE
    resources:
    - pods
    scope: '*'
  sideEffects: None
  timeoutSeconds: 10

```

oshoval:
`failurePolicy: Ignore` wow

nirdothan:
Adding more info:
[apiserver log](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/17805/pull-kubevirt-e2e-k8s-1.35-sig-network/2056297617125543936/artifacts/k8s-reporter/1/pods/1_kube-system_kube-apiserver-node01-kube-apiserver.log)  shows:
```
W0518 10:00:58.418340       1 dispatcher.go:225] Failed calling webhook, failing closed network-resources-injector-mutating-config.k8s.cni.cncf.io: failed calling webhook "network-resources-injector-mutating-config.k8s.cni.cncf.io": failed to call webhook: Post "https://network-resources-injector-service.kube-system.svc:443/mutate?timeout=10s": EOF
```

Does the deployment wait for completion?  [Yes](https://github.com/kubevirt/kubevirtci/blob/2605200707-a7bb5cdb/cluster-provision/gocli/opts/network_resources_injector/network_resources_injector.go#L49). however the process does not define a `readinessProbe`.

oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/17433/pull-kubevirt-e2e-k8s-1.36-sig-network/2071544341993623552#1:build-log.txt%3A7756

same problem right ? (well more or less)

PR #1712, opened 2026-05-24
fedora: Remove mlx5_ib due to kernel panic

oshoval:
**What this PR does / why we need it**:

F43 has a kernel panic about mlx5_ib [1],
since we don't actively use it, remove it.
It affects only ib, not SR-IOV.
In case someone needs it, he can modprobe it.

We are using `6.17.1-300.fc43`, which already has a fix [2] around it, but it doesn't seem enough.
If desired we can update the kernel in addition to `7.0.9-102.fc43`.

Tested on kubevirt via https://github.com/kubevirt/kubevirt/pull/17894

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/17832/pull-kubevirt-e2e-k8s-1.36-sig-network/2057043406697271296
[2] https://git.sceen.net/linux/linux-stable.git/commit/?id=2aebf5ee43bf0ed225a09a30cf515d9f2813b759

More details below.

<details>
  <summary>Click to expand details</summary>

<pre>
[   11.273632] ------------[ cut here ]------------
[   11.275942] kernel BUG at arch/x86/kernel/alternative.c:2456!
[   11.284910] Oops: invalid opcode: 0000 [#1] SMP NOPTI
[   11.287653] CPU: 0 UID: 0 PID: 508 Comm: systemd-modules Not tainted 6.17.1-300.fc43.x86_64 #1 PREEMPT(lazy) 
[   11.292010] Hardware name: KubeVirt None/RHEL, BIOS 1.16.3-4.el9 04/01/2014
[   11.295443] RIP: 0010:__text_poke+0x29e/0x400
[   11.297821] Code: 35 7f f2 ca 01 48 85 c0 74 20 48 01 f2 48 c1 ea 0c 48 c1 e2 06 48 8d 04 0a 48 89 44 24 28 48 83 7c 24 28 00 0f 85 28 fe ff ff <0f> 0b 48 c7 44 24 18 00 00 00 00 48 21 c7 e9 bf fe ff ff 48 8b 15
[   11.306451] RSP: 0018:ff7c6403c034f918 EFLAGS: 00010246
[   11.309110] RAX: 0000000000000000 RBX: ffffffffc05b64bd RCX: 0000000008433000
[   11.312578] RDX: 0000000000000043 RSI: ff1e426540000ff8 RDI: 0000000000000000
[   11.316034] RBP: 00000000000004bd R08: ffffffffbb62c000 R09: 0000000000000001
[...]

PR #1721, opened 2026-05-28
fedora: pin kernel install to fixed release across arches

oshoval:

In continuation of https://github.com/kubevirt/kubevirtci/pull/1712, 
we still see other modules fail due to the same underlying problem.
Which means it is not module specific.

For example, `zram`, which is installed by default, is also failing in CI:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/17922/pull-kubevirt-e2e-k8s-1.36-sig-compute/2059717554628202496

Since we are using a pretty old kernel, the first step is to update the kernel.
From `6.17.1-300.fc43` to `7.0.10-101.fc43`.

* Pin Fedora test-tooling image kernel packages to the latest validated release as of today while keeping architecture dynamic via uname -m, so image builds stay reproducible on x86_64, aarch64, and s390x.
* Fix the s390x flow to support the kernel bump and ensure the new kernel is active.
* Remove `mlx4_ib` which is not used, as clean-up, to keep symmetry of #1712.

Fixes #

Image was tested on all lanes of Kubevirt https://github.com/kubevirt/kubevirt/pull/17979.
Checked s390x boot and that it uses the new kernel.

**Checklist**

**Release note**:

```release-note
None
```

### DIFF
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -71,11 +71,18 @@ write_files:
       content: |
         datasource_list: [NoCloud, ConfigDrive, None]
 runcmd:
+  - |
+    set -e
+    KERNEL_VER_REL="7.0.10-101.fc43"
+    KERNEL_NEVRA="${KERNEL_VER_REL}.$(uname -m)"
+    echo "KERNEL_NEVRA=${KERNEL_NEVRA}" > /etc/kernel-nevra.env
+    sudo dnf -y install "kernel-${KERNEL_NEVRA}"
+    sudo grubby --set-default="/boot/vmlinuz-${KERNEL_NEVRA}"
+  - sudo systemctl daemon-reload
+  - sudo dnf install -y qemu-guest-agent stress-ng virt-what vm-dump-metrics tpm2-tools netcat sg3_utils iperf3 wget tcpdump
   # Fedora 43+ uses predictable network interface names (e.g. enp1s0).
   # Many KubeVirt tests assume "eth0". Restore classic naming.
   - sudo grubby --update-kernel=ALL --args="net.ifnames=0 biosdevname=0"
-  - sudo systemctl daemon-reload
-  - sudo dnf install -y kernel-modules-$(uname -r) qemu-guest-agent stress-ng virt-what vm-dump-metrics tpm2-tools netcat sg3_utils iperf3 wget tcpdump
   # Install nmap-ncat in addition to OpenBSD netcat.
   # Use "nc" when OpenBSD behavior is needed (cloud-init uses "nc -U").
   # Use "ncat" when nmap-ncat behavior is needed (--recv-only, --no-shutdown,
@@ -102,7 +109,13 @@ runcmd:
   # s390x can panic early with unknown-block and disabled-wait.
   - |
     if [ "$(uname -m)" = "s390x" ]; then
-      sudo dracut -f --regenerate-all
+      set -e
+      source /etc/kernel-nevra.env
+      echo -e '[defaultboot]\ntarget=/boot' > /etc/zipl.conf
+      sudo zipl-switch-to-blscfg
+      sudo dracut -f "/boot/initramfs-${KERNEL_NEVRA}.img" "${KERNEL_NEVRA}"
       sudo zipl
+      # Fail image build if bootmap was not refreshed to include the newly pinned kernel.
+      test /boot/bootmap -nt "/boot/vmlinuz-${KERNEL_NEVRA}"
     fi
   - sudo shutdown
--- a/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
+++ b/cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
@@ -25,7 +25,6 @@ write_files:
     - path: /etc/modules-load.d/mlx4.conf
       content: |
         mlx4_core
-        mlx4_ib
     - path: /etc/modules-load.d/i40e.conf
       content: |
         i40e


cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
```diff
@@ -72,11 +71,18 @@ write_files:
       content: |
         datasource_list: [NoCloud, ConfigDrive, None]
 runcmd:
+  - |
+    set -e
+    KERNEL_VER_REL="7.0.10-101.fc43"
+    KERNEL_NEVRA="${KERNEL_VER_REL}.$(uname -m)"
+    echo "KERNEL_NEVRA=${KERNEL_NEVRA}" > /etc/kernel-nevra.env
+    sudo dnf -y install "kernel-${KERNEL_NEVRA}"
```
dollierp, line 79: Are older packages still available in fedora repos when updated packages are published?

dhiller: I also fear we might get bitten by pinning if we don't have automation to regularly update here?

oshoval: > Are older packages still available in fedora repos when updated packages are published?

they age out after a while,
but we better fail and update manually than always take latest, and start flaking

if needed we can mirror maybe the one we use, but lets do it only if we see problem starts due to a too newish kernel, we can always bisect, and hopefully find one that works

we can also reuse current qcow in case needed btw,
if the changes are addon to existing fedora with specific kernel,
worst case we create an optional phase 2, which extract the qcow from the image,
and have thin layer of updates according needs (fast one),
this way the kernel we already used, never got lost

This shows if they are available for all archs
```
for arch in x86_64 aarch64 s390x; do
  pkg="kernel-7.0.10-101.fc43.${arch}"
  hit=$(dnf -q repoquery --forcearch "${arch}" --qf '%{name}-%{version}-%{release}.%{arch}' "${pkg}" 2>/dev/null)
  if [ -n "${hit}" ]; then
    echo "${arch}: AVAILABLE (${hit})"
  else
    echo "${arch}: MISSING"
  fi
done
```

and this to get latest
`dnf -q --refresh repoquery --latest-limit=1 --qf '%{version}-%{release}.%{arch}' kernel-core`

dhiller:
@jschintag @chandramerla are you able to take a quick look, since the s390x side is also touched here

cluster-provision/images/vm-image-builder/fedora-with-test-tooling/cloud-config
oshoval: > I also fear we might get bitten by pinning if we don't have automation to regularly update here?

we can just update once needed, better than having unpredicted update / changes every time we update the image,

we are doing the same on centos, and also there images gone every now and then,
for example the current image we reverted to on centos9 is gone, and we still find solution,
here we even update less

btw checked newer Fedora, base kernel, it is still older than what this PR suggest,
unless it will solve the modprobe, without other regressions, in case you prefer to try that first,
but imho we don't need to be worried from it, we can always try and adapt accordingly
```
Fedora Linux 44 (Cloud Edition)
Kernel found in image: 6.19.10-300.fc44.x86_64
```

oshoval:
> @jschintag @chandramerla are you able to take a quick look, since the s390x side is also touched here

fwiw i validated boot and new kernel is used, adding for context
note that the fedora bump itself (previous PR), was validated e2e by Vamsi, with selected kubevirt tests
so the diff is the kernel (had to fix the flow here, since zipl is broken)

asked Ashok to take a look as well

ashokpariya0:
> > @jschintag @chandramerla are you able to take a quick look, since the s390x side is also touched here
> 
> fwiw i validated boot and new kernel is used, adding for context note that the fedora bump itself (previous PR), was validated e2e by Vamsi, with selected kubevirt tests so the diff is the kernel (had to fix the flow here, since zipl is broken)
> 
> asked Ashok to take a look as well
>
LGTM for s390x related changes.

oshoval:
Thanks Ashok

oshoval:
discussed offline with @michalskrivanek 
since F44 is GA, maybe it will be better to start with it, and if it still flakes update kernel,
since if we find kernel bugs there, they are more interesting

for that created the tuple
https://github.com/kubevirt/kubevirt/pull/18039
https://github.com/kubevirt/kubevirtci/pull/1725
lets see test are passing there

EDIT - all works there now

chandramerla:
> > @jschintag @chandramerla are you able to take a quick look, since the s390x side is also touched here
> 
> fwiw i validated boot and new kernel is used, adding for context note that the fedora bump itself (previous PR), was validated e2e by Vamsi, with selected kubevirt tests so the diff is the kernel (had to fix the flow here, since zipl is broken)
> 
> asked Ashok to take a look as well

Asked @vamsikrishna-siddu to help in validating the changes for s390x

oshoval:
/hold

Thanks Chandra, Vamsi
lets take please first https://github.com/kubevirt/kubevirtci/pull/1725
and then if desired we can take this as well (after adapting to F44),
before we test it e2e with all the tests on s390x, lets see we agree to take it,
basically i tested boot of s390x and that it runs the new kernel, testing all the e2e might be overkill,
unless you insist to do it,
Ashok acked the changes themselves as well, this is the important part imo

oshoval:
Hi @dhiller @dollierp @Whitedyl @EdDev @michalskrivanek 

once F44 lands [1], do you want also this PR, that updates the kernel to latest ?
(if so will adapt it of course once F44 lands)

easier alternative is that we can also close it and focus on [1]
and reopen only if we see kernel bugs like the modprobe etc

[1] https://github.com/kubevirt/kubevirtci/pull/1725

michalskrivanek:
I would say we don't want to pin, but ia "global" `dnf update` to pick up updates since GA would be great (since you confirmed the rebuilds are manual it should be safe to do)

oshoval:
> I would say we don't want to pin, but ia "global" `dnf update` to pick up updates since GA would be great (since you confirmed the rebuilds are manual it should be safe to do)

but then it can also break things, since it will be less predictable
i prefer to not do it, unless we see problems it can solve
or you mean just because we updated kernel then we want them to be aligned ?
(it will still be newer than the pinned, and so far as so good without it)
anyhow lets please see first we want this PR, and then decide about it ?

michalskrivanek:
> > I would say we don't want to pin, but ia "global" `dnf update` to pick up updates since GA would be great (since you confirmed the rebuilds are manual it should be safe to do)
> 
> but then it can also break things, since it will be less predictable i prefer to not do it, unless we see problems it can solve or you mean just because we updated kernel then we want them to be aligned ? (it will still be newer than the pinned, and so far as so good without it) anyhow lets please see first we want this PR, and then decide about it ?

I thought when you only manually rebuild it's quite predictable - you update and test and if all looks good you merge it. No instabilities from nightly rebuilds that can potentially regress temporarily. 
But yes, as far as this PR is concerned then no, let's not pin unless we really really have to. We really should try hard to keep kernel and userspace aligns the same as in Fedora proper.

oshoval:
> > > I would say we don't want to pin, but ia "global" `dnf update` to pick up updates since GA would be great (since you confirmed the rebuilds are manual it should be safe to do)
> > 
> > 
> > but then it can also break things, since it will be less predictable i prefer to not do it, unless we see problems it can solve or you mean just because we updated kernel then we want them to be aligned ? (it will still be newer than the pinned, and so far as so good without it) anyhow lets please see first we want this PR, and then decide about it ?
> 
> I thought when you only manually rebuild it's quite predictable - you update and test and if all looks good you merge it. No instabilities from nightly rebuilds that can potentially regress temporarily. 
But yes, as far as this PR is concerned then no, let's not pin unless we really really have to. We really should try hard to keep kernel and userspace aligns the same as in Fedora proper.

do you mean also not pin kernel itself ?
this is wrong imo, because having implicit changing kernel between runs, can have impact hard to reverse / bisect

EDIT - btw about the userspace packages, even on laptop, we do update them and rarely update kernel, same state as this PR suggest atm, with and without dnf update as done on laptop locally usually

oshoval:
i will close this one, and reopen if needed, or in case you would like to have it
lets first see what https://github.com/kubevirt/kubevirtci/pull/1725 have in mind

cc @EdDev 

thanks

PR #1737, opened 2026-06-22
emulated SR-IOV: Adapt to enable running DRA in parallel mode

oshoval:
Do the changes required to run emulated SR-IOV, with DRA, in parallel mode,
folded into k8s-1.36.

1. Increase VFs to maximum 7,
it will allow to run 3 tests that have 2 VMs each + one that has one VM at the same time.
lucky we have only 3 tests that need 2 VMs, and igb supports max 7.
Default parallel is 4 processes, so it fits perfect.
It means we can't overcome this limit, but it is fine as not required to do more than that, maybe even less.
EDIT - there are new findings about max required, they will be handled on kubevirt side,
max VF is 7 anyhow.

2. Don't auto deploy standalone multus when CNAO + SRIOV are enabled.
If it is already enabled it will override the CNAO one, else just use the CNAO one,
but don't add it extra implicitly.
It allows the tests to keep their multus logic, without implicit suprises.

Takes around 8 minutes for 12 tests locally,
instead 20-30m for 13 tests (running) + 15-30 build and deploy.

Fixes #

**Checklist**

**Release note**:

```release-note
None
```

### DIFF
--- a/cluster-up/cluster/k8s-1.35/config_sriov_cluster.sh
+++ b/cluster-up/cluster/k8s-1.35/config_sriov_cluster.sh
@@ -20,7 +20,7 @@ SRIOVDP_RESOURCE_PREFIX="kubevirt.io"
 SRIOVDP_RESOURCE_NAME="sriov_net"
 VFS_DRIVER="vfio-pci"
 VFS_DRIVER_KMODULE="vfio_pci"
-VFS_COUNT="${VFS_COUNT:-6}"
+VFS_COUNT="${VFS_COUNT:-7}"
 KUBEVIRT_USE_DRA=${KUBEVIRT_USE_DRA:-false}
 
 # Source SR-IOV components deployment functions
--- a/cluster-up/cluster/k8s-1.35/provider.sh
+++ b/cluster-up/cluster/k8s-1.35/provider.sh
@@ -6,7 +6,9 @@ if [ "${KUBEVIRT_CGROUPV2}" == "false" ]; then
 fi
 
 if [ "${KUBEVIRT_WITH_SRIOV}" == "true" ]; then
-    export KUBEVIRT_WITH_MULTUS=true
+    if [ "${KUBEVIRT_WITH_CNAO}" != "true" ]; then
+        export KUBEVIRT_WITH_MULTUS=true
+    fi
     export KUBEVIRT_NUM_NUMA_NODES=2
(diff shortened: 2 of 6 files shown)


EdDev (approved):
Thank you.

oshoval:
Thanks Edy

oshoval:
Checked deeper

Emulated SR-IOV + DP + parallel (4) + Alpine with 512Mi and fixed igb.
Ran 13 of 1396 Specs in 584.351 seconds (instead 1200 - 1600 seconds on CI neto test time)

[1] should correctly plug all the interfaces... -> 1 VMI with 4 SR-IOV interfaces -> 4 VFs on one node
[2] should connect to another machine... -> 2 VMIs, 1 SR-IOV each -> 2 VFs on one node
[3] should be able to ping between two VMIs... -> 2 VMIs, 1 SR-IOV each -> 2 VFs on one node
[4] should NOT be able to ping between... -> 2 VMIs, 1 SR-IOV each -> 2 VFs on one node
That is 4 + 2 + 2 + 2 = 10 VFs worst case.

it means that if we fold it into current sig-network, which has 4 parallel jobs, it will have pending time (which is fragile if timeout)

There are 3 solutions:
1. hardest - move to `virtio-net-pci` instead igb, which allows more than 7 VFs, i don't think we need to rush about it.
2. Most of the tests pin already to the first worker, we can pin [1] to 2nd worked and it will solve the problem.
3. Mark [1] as serial, or just get rid of it / find other test to fold it into.

[2] is the best imo for now, it means the changes will be on kubevirt side, so this PR is fine as is.

Note that also migration require 2 VFs, but one on the other node, so it ain't count into max per node at the same time.

EdDev:
The solution is not to pin to a specific node. Affinity or anti-affinity should be used instead.
The scheduler should deploy the vm on the node that has the available resources.

oshoval:
> The solution is not to pin to a specific node. Affinity or anti-affinity should be used instead. The scheduler should deploy the vm on the node that has the available resources.

On it, need first https://github.com/kubevirt/kubevirt/pull/18261
and then to add anti affinity

Wanted first to have a naive solution, but anyhow done that

anyhow, shouldn't block this PR

dollierp (approved):
/approve

oshoval:
Thanks Denis

PR #1739, opened 2026-06-23
k8s: Enable cpuManagerPolicy static in kubelet drop-in config

### LINKED ISSUE #1648: CPUManager not configured on control-plane+worker nodes in single-node clusters
## Summary

On single-node clusters where the control-plane node also has the `worker` role, CPUManager (`cpuManagerPolicy: static`) is not configured on the kubelet. This causes the KubeVirt functional test suite to timeout during `SynchronizedBeforeSuite` waiting for all schedulable nodes to have the `cpumanager=true` label.

## Root Cause

The gocli provisioner sets `--cpu-manager-policy=static` on the kubelet only during the **worker node join** path in `cluster-provision/gocli/opts/nodes/nodes.go`. The control-plane node never goes through this join path, so it never gets CPUManager configured.

Previously, the `configure_cpu_manager` shell function in `cluster-up/cluster/k8s-provider-common.sh` handled this by SSH'ing into all nodes labeled `node-role.kubernetes.io/worker` post-provisioning and configuring CPUManager. This function was removed in #1289 (997cbff0) under the assumption that the Kubernetes CPUManager feature gate being GA since k8s 1.26 was sufficient. However, the k8s feature gate being GA only means the kubelet supports the feature — the `cpuManagerPolicy` still defaults to `none` and must be explicitly set to `static`.

## Reproduction

```bash
# Provision a single-node cluster
export KUBEVIRT_NUM_NODES=1
make cluster-up cluster-sync

# Check the node - cpumanager will be false
kubectl get nodes -L cpumanager
# NAME     STATUS   ROLES                  CPUMANAGER
# node01   Ready    control-plane,worker   false

# Check kubelet config confirms policy is "none"
kubectl proxy &
curl -s http://localhost:8123/api/v1/nodes/node01/proxy/configz | jq '.kubeletconfig.cpuManagerPolicy'
# "none"

# Run functional tests - will timeout in SynchronizedBeforeSuite
make functest
```

## Expected Behavior

The control-plane+worker node should have `cpuManagerPolicy: static` configured, and the `cpumanager=true` label should be set by virt-handler.

## Impact

All functional test runs on single-node clusters fail during test suite setup with a 360-second timeout.

lyarwood:
## Summary

- Enable `cpuManagerPolicy: static` in the kubelet drop-in config (`50-kubevirt.conf`) during base image provisioning for k8s 1.34, 1.35, and 1.36
- Add required `kubeReserved` and `systemReserved` CPU reservations (500m each) to match the existing gocli worker node configuration

## Problem

On single-node clusters where the control-plane node also has the `worker` role, CPUManager was not enabled. The gocli provisioner only sets `--cpu-manager-policy=static` during the worker node join path (`nodes.go`), which the control-plane node never executes. This was previously handled by the `configure_cpu_manager` shell function removed in #1289.

The KubeVirt functional test suite's `SynchronizedBeforeSuite` enables the CPUManager feature gate and then waits for all schedulable nodes to have the `cpumanager=true` label, causing a 360-second timeout on single-node clusters.

## Fix

Add `cpuManagerPolicy: static` and CPU reservations directly to the kubelet drop-in config file that is written during base image provisioning. This ensures CPUManager is configured on all nodes regardless of their role, including control-plane+worker nodes in single-node clusters.

Supersedes: #1649

Fixes: https://github.com/kubevirt/kubevirtci/issues/1648

### DIFF
--- a/cluster-provision/gocli/opts/node01/node01.go
+++ b/cluster-provision/gocli/opts/node01/node01.go
@@ -3,6 +3,7 @@ package node01
 import (
 	_ "embed"
 	"fmt"
+	"runtime"
 
 	"kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/libssh"
 )
@@ -68,6 +69,16 @@ func (n *node01Provisioner) Exec() error {
 
 	if n.secondaryNicBridges {
 		cmds = append(cmds, string(setupBridgesScript))
+	}
+
+	if runtime.GOARCH != "s390x" {
+		cmds = append(cmds, `cat >> /etc/kubernetes/kubelet.conf.d/50-kubevirt.conf <<'EOF'
+cpuManagerPolicy: static
+kubeReserved:
+  cpu: 500m
+systemReserved:
+  cpu: 500m
+EOF`)
 	}
 
 	cmds = append(cmds,
--- a/cluster-provision/gocli/opts/node01/testconfig.go
+++ b/cluster-provision/gocli/opts/node01/testconfig.go
@@ -2,6 +2,8 @@ package node01
 
 import (
 	"fmt"
+	"runtime"
+
 	"kubevirt.io/kubevirtci/cluster-provision/gocli/pkg/libssh"
 	kubevirtcimocks "kubevirt.io/kubevirtci/cluster-provision/gocli/utils/mock"
 )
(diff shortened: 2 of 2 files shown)


dhiller (commented):
/approve

Being curious: why is that change not done for s390x?

PR #1760, opened 2026-07-08
fix: validate minimum memory size before fake VFIO setup

nestoracunablanco:
Add a validate_memory() check to config_vfio_cluster.sh that exits
if KUBEVIRT_MEMORY_SIZE is below the 8G (8192M) minimum required by
some fake VFIO tests. The check handles G, GB, and GiB suffixes,
normalising everything to MB for comparison.

Assisted-by: IBM Bob

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-up/cluster/vfio-gpu/config_vfio_cluster.sh
+++ b/cluster-up/cluster/vfio-gpu/config_vfio_cluster.sh
@@ -40,7 +40,23 @@ encode_vfio_assets() {
     tar -C "${SCRIPT_PATH}" -cz fake-iommu fake-pci setup-fake-pci-host.sh | base64 | tr -d '\n'
 }
 
+# Some fake VFIO tests require at least 8G of memory. Output a warning if the cluster was started with less.
+validate_memory() {
+    local mem_value mem_unit mem_in_mb
+    mem_value=$(echo "${KUBEVIRT_MEMORY_SIZE:-0}" | sed 's/[^0-9]//g')
+    mem_unit=$(echo "${KUBEVIRT_MEMORY_SIZE:-0}" | sed 's/[^a-zA-Z]//g' | tr '[:lower:]' '[:upper:]')
+    case "${mem_unit}" in
+        G|GI|GB|GIB) mem_in_mb=$((mem_value * 1024)) ;;
+        T|TI|TB|TIB) mem_in_mb=$((mem_value * 1024 * 1024)) ;;
+        *) mem_in_mb=${mem_value} ;;
+    esac
+    if [ "${mem_in_mb}" -lt 8192 ]; then
+        echo "WARNING: some tests with KUBEVIRT_USE_FAKE_VFIO=true require KUBEVIRT_MEMORY_SIZE >= 8G (8192M). Current value: ${KUBEVIRT_MEMORY_SIZE:-unset}" >&2
+    fi
+}
+
 validate_inputs() {
+    validate_memory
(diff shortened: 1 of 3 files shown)


Sreeja1725:
@nestoracunablanco My concern is about the flexibility for contributors. If someone wants to add a test case that creates only a single VMI with vfio, they can reasonably run that on a host with less than 2 Gi of RAM allocated to the test environments right? In that case they can comment out the multiple VMI creations testcase locally.  In that scenario, they could locally disable or comment out the “multiple VMI” test case and still validate their single‑VMI testcase. If we add a lower bound, they will not be able to run even the single‑VMI test on the modest hardware.

Should we need a lower bound on the memory size or is it not sufficient to update the documentation and clearly state that one VM (virt‑launcher pod) with vfio requires approximately 1 Gi of memory? So that it will be upto the user to configure the memory.

nestoracunablanco:
Hi @Sreeja1725, you are making a good point here. I changed it to a warning. What do you think about?

nestoracunablanco:
/hold
until #1761 is merged

nestoracunablanco:
@alaypatel07 @Sreeja1725 @ormergi @qinqon what do you think about this change?

Sreeja1725:
@nestoracunablanco I think sometimes, the warning can be missed or ignored in the noisy cluster-up logs. Maybe it would be good to update the README as well, so that it’s easier to notice and understand

nestoracunablanco:
@Sreeja1725 I edited the Readme. Please let me know what do you think.

ormergi:
People tend to ignore warnings, this one may get lost in a long long stdout as a result of make cluster-up
Consider pivoting to make the process blow up if there is not enough memory for facilitating emulated SR-IOV.

nestoracunablanco:
@ormergi I implemented your suggestion, please let me know what you think.

cluster-up/cluster/vfio-gpu/config_vfio_cluster.sh
```diff
@@ -40,7 +40,24 @@ encode_vfio_assets() {
     tar -C "${SCRIPT_PATH}" -cz fake-iommu fake-pci setup-fake-pci-host.sh | base64 | tr -d '\n'
 }
 
+# Some fake VFIO tests require at least 8G of memory. Fail early if the cluster was started with less.
+validate_memory() {
```
ormergi, line 44: I know this change was done following previous review comment, and I am aware `KUBEVIRT_MEMORY_SIZE` is ambguent in regard to the quantity unit,
but please note that it seem non of this is necessary because the `KUBEVIRT_MEMORY_SIZE` eventually translated to the container-runtime `--memory` flag which support flexible quantity units (including G,g, GiB, GB, etc..)
So that for example following variation of Gi, GB, g, gi translated to gigabyte unit (1073741824 bytes). 
Its even worse because with the below translation one might think GB and GiB/Gi will get you different amount of memory (while its all translate to 1073741824b).
I verified this on both Docker and Podman on Fedora system.

FYI @Sreeja1725 

Let see what maintainer think.

ormergi:
Thanks for the changes, LGTM although I think we should re-consider the quantity unit change https://github.com/kubevirt/kubevirtci/pull/1760#discussion_r3913534451 
@dhiller WDYT?

/lgtm

nestoracunablanco:
/test check-up-kind-1.37-vgpu
@dhiller what do you think about @ormergi question?

PR #1761, opened 2026-07-09
fix: kernel-modules-extra for CentOs Stream 10 compatibility

nestoracunablanco:
On CentOS Stream 10, br_netfilter was moved from the kernel-modules package to the kernel-modules-extra.

Assisted-by: IBM Bob

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-provision/k8s/1.34/provision.sh
+++ b/cluster-provision/k8s/1.34/provision.sh
@@ -29,6 +29,9 @@ fi
 
 # Install modules of the initrd kernel
 dnf install -y "kernel-modules-$(uname -r)"
+if [ "$release" == "centos10" ]; then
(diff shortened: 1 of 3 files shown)


brianmcarey (commented):
/lgtm 

thanks @nestoracunablanco

dhiller (commented):
@nestoracunablanco dumb question: am I not getting something? The change in this PR looks exactly like the one that was merged here: https://github.com/kubevirt/kubevirtci/commit/f224bfccb94d88a8f5f82e157bee288fa33de7df

nestoracunablanco:
@dhiller point for you. I was not aware about the change of @lyarwood, probably we did the same fix at the same time.

PR #1767, opened 2026-07-14
feat: capture VM serial console output for CI debugging

### LINKED ISSUE #5098: Investigate root cause of Cent OS 9 Stream base image bump breaks kubevirtci VM boot
## Current state

The immediate problem is contained:
- **Revert merged**: [#1716](https://github.com/kubevirt/kubevirtci/pull/1716) reverts centos9 back to the known-good `20260413.0` snapshot.
- **Auto-bump disabled**: The periodic jobs that auto-bump the centos9 base have been removed.
- **Follow-up**: A PR in `kubevirt/kubevirt` will bump the kubevirtci tag to pick up the revert.

This issue tracks the root cause investigation so we can safely re-enable centos9 base image updates.

## Background

CentOS 9 Stream snapshots after `20260413.0` cause QEMU VMs in kubevirtci to fail to boot. The VM starts (QEMU monitor is ready) but sshd never becomes reachable, causing SSH timeout after ~50 minutes of retries. All k8s providers using the centos9 base (1.34, 1.35, 1.36) are affected.

This happened twice before being contained:

| Date | Event | PR |
|------|-------|----|
| May 12 | Auto-bump to `20260506.0` introduced failures | [#1698](https://github.com/kubevirt/kubevirtci/pull/1698) |
| May 15 | Reverted to `20260413.0` | [#1703](https://github.com/kubevirt/kubevirtci/pull/1703) |
| May 19 | Auto-bump to `20260513.0` re-introduced failures | [#1707](https://github.com/kubevirt/kubevirtci/pull/1707) |
| May 27 | Reverted to `20260413.0`, auto-bump jobs removed | [#1716](https://github.com/kubevirt/kubevirtci/pull/1716) |

## Failure mode

1. QEMU starts, monitor socket becomes ready.
2. `tap01` interface stays in `NO-CARRIER` / `state DOWN` — guest OS never brings up its virtio-net interface.
3. dnsmasq never receives a DHCP request.
4. SSH retries every 5 minutes for ~50 minutes, then times out.

When the broken image *does* boot, it takes 9–47 seconds (the good image consistently boots in 9–18 seconds), suggesting sensitivity to CPU contention on the Prow node.

## Impact (May 20–27)

15 SSH timeout failures across 7 job lanes. Pass rate on `pull-project-infra-prow-deploy-test` dropped from 88% to 40%.

## What changed between working and broken images

| Package | Working (`20260413.0`) | Broken (`20260513.0`) |
|---------|------------------------|-----------------------|
| **kernel** | `5.14.0-692`/`-693` | `5.14.0-704` |
| **systemd** | `252-67` | `252-69` |
| **openssh** | `9.9p1-7` | `9.9p1-8` |
| **dracut** | older | `057-110` |

The kernel jumped 10+ releases between the two snapshots. Prime suspect given the networking failure mode.

## Investigation plan

- [ ] **Binary search the compose snapshots** — Build kubevirtci centos9 images at mid-point composes between `20260413.0` (good) and `20260506.0` (first known broken) to pinpoint the first broken snapshot.
- [ ] **Isolate kernel vs userspace** — kubevirtci uses QEMU `-kernel`/`-initrd` direct boot. Build an image with the broken compose's qcow2 but the good compose's kernel/initramfs to determine if the kernel or userspace (systemd, dracut, NetworkManager) is responsible.
- [ ] **RPM diff** — Extract full RPM lists from the good and bad cloud images (via guestfish) and diff to identify all changed packages.
- [ ] **Report fix or workaround upstream** if needed.

dhiller:
## Summary

- Fix QEMU `-serial pty` boot hang that blocked CentOS 9 Stream image bumps since April 2026
- Use `-serial file:/dev/stderr` in CI (captures console output in build logs), keep `-serial pty` for local dev
- Forward `PROW_JOB_ID` and `CI` env vars into VM containers so `vm.sh` can detect CI
- Remove hardcoded `-serial pty` from `run.go` — vm.sh now manages the serial backend, eliminating duplicate serial ports
- Reduce `dockerize` SSH timeout from 300s to 120s and SSH retries from 10 to 5 (~10 min total instead of ~50 min)

Fixes: kubevirt/project-infra#5098 , kubevirt/kubevirtci#1770

## Root cause

When `-serial pty` is used and nobody reads from the PTY (always the case in CI), the QEMU PTY backend can stall guest execution under CPU contention. This is a [known class of QEMU bugs](https://bugzilla.redhat.com/show_bug.cgi?id=1455451). The kernel jump from `5.14.0-691` to `5.14.0-721` (~30 releases) produces more boot messages on `console=ttyS0`, crossing the threshold that triggers the stall.

With `-serial pty`: ~60% failure rate. With `-serial file:/dev/stderr`: 0% failures across 10+ runs.

Additionally, `run.go` hardcoded `-serial pty` in QEMU_ARGS, which created a second unread PTY serial port (COM2) alongside vm.sh's own serial backend — doubling the exposure to the bug.

See https://github.com/kubevirt/project-infra/issues/5098 for full investigation.

## Changes

- `cluster-provision/centos9/scripts/vm.sh`: conditional serial backend (CI vs local), 120s dockerize timeout
- `cluster-provision/centos10/scripts/vm.sh`: same
- `cluster-provision/gocli/cmd/utils/utils.go`: new `ForwardEnv()` helper
- `cluster-provision/gocli/cmd/provision.go`: forward `PROW_JOB_ID`/`CI` to VM containers
- `cluster-provision/gocli/cmd/run.go`: forward `PROW_JOB_ID`/`CI` to VM containers, reduce SSH retries from 10 to 5, remove hardcoded `-serial pty` (vm.sh handles it)

## Test plan

- [x] `check-provision-centos-base` passes with CentOS 9 compose `20260706.0`
- [x] 10+ consecutive successful boots confirm the serial backend fix
- [ ] Confirm `-serial null` also boots reliably (proves PTY backend is the cause independent of output)

## Related issues

- https://github.com/kubevirt/project-infra/issues/5098 — root cause investigation
- #1769 — CentOS 10 `br_netfilter` missing (unrelated, pre-existing)
- #1770 — s390x SSH timeout during check-cluster-up (partially addressed by removing duplicate serial port)

/kind bug

🤖 Assisted by Claude

### DIFF
--- a/cluster-provision/centos10/scripts/vm.sh
+++ b/cluster-provision/centos10/scripts/vm.sh
@@ -58,7 +58,7 @@ n="$(printf "%02d" $(( 10#${NODE_NUM} )))"
 cat >/usr/local/bin/ssh.sh <<EOL
 #!/bin/bash
 set -e
-dockerize -wait tcp://192.168.66.1${n}:22 -timeout 300s &>/dev/null
+dockerize -wait tcp://192.168.66.1${n}:22 -timeout 120s &>/dev/null
 ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ${VM_USER}@192.168.66.1${n} -i ${VM_USER_SSH_KEY} -p 22 -q \$@
 EOL
 chmod u+x /usr/local/bin/ssh.sh
@@ -216,6 +216,12 @@ if [ "${NUMA}" -gt 1 ]; then
     done
 fi
 
+if [ -n "${PROW_JOB_ID:-}" ] || [ "${CI:-}" = "true" ]; then
+  SERIAL_ARG="-serial file:/dev/stderr"
+else
+  SERIAL_ARG="-serial pty"
+fi
+
 if [ "$(uname -m)" == "s390x" ]; then
   # As per https://www.qemu.org/docs/master/system/s390x/bootdevices.html#booting-without-bootindex-parameter -drive if=virtio can't be specified with bootindex for s390x
   qemu_system_cmd="qemu-system-s390x \
@@ -233,7 +239,7 @@ if [ "$(uname -m)" == "s390x" ]; then
     -cpu host \
     -m ${MEMORY} \
     -smp ${CPU} ${numa_arg} \
-    -serial pty \
+    ${SERIAL_ARG} \
     -machine s390-ccw-virtio,accel=kvm \
     -uuid $(cat /proc/sys/kernel/random/uuid) \
     -monitor unix:/tmp/qemu-monitor.sock,server,nowait \
@@ -260,7 +266,7 @@ else
     -cpu host,migratable=no,+invtsc \
     -m ${MEMORY} \
     -smp ${CPU} ${numa_arg} \
-    -serial pty \
+    ${SERIAL_ARG} \
     -machine q35,accel=kvm,kernel_irqchip=split \
     -device intel-iommu,intremap=on,caching-mode=on \
     -device intel-hda,bus=pcie.0 -device hda-duplex -device AC97,bus=pcie.0 \
--- a/cluster-provision/centos9/scripts/vm.sh
+++ b/cluster-provision/centos9/scripts/vm.sh
@@ -58,7 +58,7 @@ n="$(printf "%02d" $(( 10#${NODE_NUM} )))"
 cat >/usr/local/bin/ssh.sh <<EOL
 #!/bin/bash
 set -e
-dockerize -wait tcp://192.168.66.1${n}:22 -timeout 300s &>/dev/null
+dockerize -wait tcp://192.168.66.1${n}:22 -timeout 120s &>/dev/null
 ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ${VM_USER}@192.168.66.1${n} -i ${VM_USER_SSH_KEY} -p 22 -q \$@
 EOL
 chmod u+x /usr/local/bin/ssh.sh
@@ -216,6 +216,12 @@ if [ "${NUMA}" -gt 1 ]; then
     done
 fi
 
+if [ -n "${PROW_JOB_ID:-}" ] || [ "${CI:-}" = "true" ]; then
+  SERIAL_ARG="-serial file:/dev/stderr"
+else
+  SERIAL_ARG="-serial pty"
+fi
+
 if [ "$(uname -m)" == "s390x" ]; then
   # As per https://www.qemu.org/docs/master/system/s390x/bootdevices.html#booting-without-bootindex-parameter -drive if=virtio can't be specified with bootindex for s390x
   qemu_system_cmd="qemu-system-s390x \
@@ -233,7 +239,7 @@ if [ "$(uname -m)" == "s390x" ]; then
     -cpu host \
     -m ${MEMORY} \
     -smp ${CPU} ${numa_arg} \
-    -serial pty \
+    ${SERIAL_ARG} \
     -machine s390-ccw-virtio,accel=kvm \
     -uuid $(cat /proc/sys/kernel/random/uuid) \
     -monitor unix:/tmp/qemu-monitor.sock,server,nowait \
@@ -260,7 +266,7 @@ else
     -cpu host,migratable=no,+invtsc \
     -m ${MEMORY} \
     -smp ${CPU} ${numa_arg} \
-    -serial pty \
+    ${SERIAL_ARG} \
     -machine q35,accel=kvm,kernel_irqchip=split \
     -device intel-iommu,intremap=on,caching-mode=on \
     -device intel-hda,bus=pcie.0 -device hda-duplex -device AC97,bus=pcie.0 \
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -787,9 +787,9 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 
 		vmContainerConfig := &container.Config{
 			Image: clusterImage,
-			Env: []string{
+			Env: append([]string{
 				fmt.Sprintf("NODE_NUM=%s", nodeNum),
-			},
+			}, utils.ForwardEnv("PROW_JOB_ID", "CI")...),
 			Cmd: []string{"/bin/bash", "-c", fmt.Sprintf("/vm.sh -n /var/run/disk/disk.qcow2 --memory %s --cpu %s --numa %s %s %s %s %s %s %s",
 				memory,
 				strconv.Itoa(int(cpu)),
@@ -1113,7 +1113,7 @@ func provisionNode(sshClient libssh.Client, n *nodesconfig.NodeLinuxConfig) erro
 func waitForVMToBeUp(cli *client.Client, prefix string, nodeName string) error {
 	var err error
 	// Wait for the VM to be up
-	for x := 0; x < 10; x++ {
+	for x := 0; x < 5; x++ {
 		err = _cmd(cli, nodeContainer(prefix, nodeName), "ssh.sh echo VM is up", "waiting for node to come up")
 		if err == nil {
 			break
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -610,10 +610,6 @@ func run(cmd *cobra.Command, args []string) (retErr error) {
 		volumes <- sharedVolume.Name
 	}
 
-	// Add serial pty so we can do stuff like 'screen /dev/pts0' to access
-	// the VM console from the container without ssh
-	qemuArgs += " -serial pty"
-
 	var qemuNetDevice = getNetDeviceByArch()
 	numaNodes := int(numa)
 	pcieBus := ""
--- a/cluster-provision/gocli/cmd/utils/utils.go
+++ b/cluster-provision/gocli/cmd/utils/utils.go
@@ -1,6 +1,8 @@
 package utils
 
 import (
+	"fmt"
+	"os"
 	"strconv"
 
 	"github.com/docker/go-connections/nat"
@@ -15,6 +17,17 @@ func AppendTCPIfExplicit(ports nat.PortMap, exposedPort int, flagSet *pflag.Flag
 // AppendUDPIfExplicit append UDP port to the portMap if the port flag exists
 func AppendUDPIfExplicit(ports nat.PortMap, exposedPort int, flagSet *pflag.FlagSet, flagName string) error {
 	return appendIfExplicit(ports, exposedPort, flagSet, flagName, UDPPortOrDie)
+}
+
+// ForwardEnv returns KEY=VALUE strings for env vars that are set in the current process.
+func ForwardEnv(names ...string) []string {
+	var result []string
+	for _, name := range names {
+		if val, ok := os.LookupEnv(name); ok {
+			result = append(result, fmt.Sprintf("%s=%s", name, val))
+		}
+	}
+	return result
 }
 
 func appendIfExplicit(ports nat.PortMap, exposedPort int, flagSet *pflag.FlagSet, flagName string, portFn func(port int) nat.Port) error {
--- a/cluster-provision/gocli/cmd/provision.go
+++ b/cluster-provision/gocli/cmd/provision.go
@@ -209,9 +209,9 @@ func provisionCluster(cmd *cobra.Command, args []string) (retErr error) {
 	}
 	node, err := cli.ContainerCreate(ctx, &container.Config{
 		Image: base,
-		Env: []string{
+		Env: append([]string{
 			fmt.Sprintf("NODE_NUM=%s", nodeNum),
-		},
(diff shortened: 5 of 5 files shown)


dhiller:
Hi! I looked into the CentOS 9 Stream boot issue in project-infra#5098 and analyzed the changes implemented in this PR to capture the serial logs and speed up VM failure detection.

While the PR successfully addresses the need for faster failure detection and ensures logs are captured in real-time when the container is force-killed (SIGKILL), the implementation has two significant flaws that affect developer experience and CI stability:

---

### 1. 🛑 **Disrupts Interactive Developer Console Access**
* **The Change:** Replaced `-serial pty` with `-serial file:/dev/stderr` in `cluster-provision/centos[9/10]/scripts/vm.sh`.
* **Why it's flawed:**
  * `-serial pty` dynamically allocates a pseudo-terminal (PTY) device on the host (such as `/dev/pts/0`) and connects the VM's serial console to it.
  * KubeVirtci's developer guide (`K8S.md`) explicitly details that developers can attach to the node's console interactively using:
    ```bash
    docker exec -it ${KUBEVIRT_PROVIDER}-node01 screen /dev/pts/0
    ```
  * By switching to `-serial file:/dev/stderr`, **no PTY device is allocated (`/dev/pts/X` doesn't exist)**. 
  * This completely breaks the interactive console connection. If a VM boot fails or has networking issues, developers can no longer use `screen` to log in, run diagnostics, or fix the VM manually.

### 2. ⏳ **Aggressive Boot Timings Risk CI Flakiness (False Negatives)**
* **The Change:** Reduced `dockerize` timeout waiting for SSH (port 22) from `300s` to `60s` in `ssh.sh`, and reduced Go-level retries in `waitForVMToBeUp` from `10` to `3`.
* **Why it's flawed:**
  * The total allowable boot-up time budget is reduced from `50 minutes` (which was indeed too long) to a mere **3 minutes (182 seconds)**.
  * In heavily loaded CI runner environments or nested virtualization environments, launching multiple QEMU VMs simultaneously causes high CPU/disk I/O contention. Under these conditions, healthy VMs can easily take longer than 3 minutes to boot and start the SSH daemon.
  * This aggressive 3-minute hard limit will introduce **false-negative failures and intermittent test flakes** in CI for slow but otherwise healthy VM boots.
  * In addition, the Go-level retry loop sleeps for only `1 * time.Second` before immediately spawning the next blocking 60-second `ssh.sh` wrapper, offering no backoff or breathing room under disk thrashing.

---

### 💡 **Proposed Solutions**

#### **Remedy for Serial Console Redirection (Conditional Serial Backend)**
We can detect if the container is running in CI (by checking standard env variables like `$CI` or `$PROW_JOB_ID`) and branch accordingly in `vm.sh`:

```bash
if [ -n "$CI" ]; then
  # CI mode: Redirect to stderr to ensure Prow captures unbuffered logs on hang/SIGKILL
  SERIAL_ARG="-serial file:/dev/stderr"
else
  # Local mode: Maintain interactive console debugging via PTY for developers
  SERIAL_ARG="-serial pty"
fi
```

#### **Remedy for Boot Timings (Balanced Timeout Budget)**
Set a more realistic timeout budget (e.g., **6 to 8 minutes** total instead of 3 minutes). This is long enough to tolerate heavy disk contention on busy CI runners but short enough to avoid 50-minute hangs:
* Set `dockerize -timeout 120s` (2 minutes).
* Set the retry loop in `gocli`'s `waitForVMToBeUp` to **3 or 4 attempts**, resulting in a robust 6–8 minute grace period.

dhiller:
## Analysis of PR changes

### Context (Issue #5098)

CentOS 9 Stream composes after `20260413.0` cause kubevirtci VMs to fail to boot — `tap01` stays `NO-CARRIER`, sshd never becomes reachable. The PR aims to capture serial console output to diagnose the hang, and intentionally bumps the version to the broken compose to reproduce it.

### Flaws identified

**1. Broken base image version shipped alongside diagnostic infrastructure**

The PR mixes two unrelated concerns: the serial console capture (legitimate infrastructure improvement) and the version bump to `20260706.0` (a known-broken compose used to reproduce the bug). If the serial capture changes are later cherry-picked or the PR is merged after the investigation, the broken version goes with it. These should be separate commits with the version bump clearly labeled as do-not-merge.

**2. Timeout/retry reduction is too aggressive and globally scoped**

The PR reduces the SSH wait from 300s × 10 retries (~50 min) to 60s × 3 retries (~3 min):
- `vm.sh`: `dockerize -timeout 300s` → `60s` (both centos9 AND centos10)
- `run.go`: retry loop `x < 10` → `x < 3` (affects ALL providers, all node types)

The issue itself notes the **good** image sometimes takes up to 18s under CPU contention, and the broken image takes 9–47s when it does boot. While 60s per attempt is enough headroom for the happy path, the `run.go` change is global — it affects every provider (1.34, 1.35, 1.36, centos10), every multi-node scenario where later nodes compete for CPU. The original generous timeout was a safety margin for CI node load variance.

**3. centos10 modified without justification**

The centos10 `vm.sh` gets the same serial-to-stderr and timeout changes "for consistency," but centos10 is unaffected by this issue. If `-serial file:/dev/stderr` produces unexpected behavior (see below), it would regress centos10 unnecessarily.

**4. Serial output interleaved with shell trace is hard to parse**

The script runs with `set -ex`, which echoes every command to stderr. The `nohup` wrapper preserves stderr inheritance in containers (where stderr is a pipe, not a terminal, so `nohup` doesn't redirect it). So QEMU's serial output and the shell's `set -x` trace both go to the same stderr stream, producing interleaved output that's difficult to read or grep for the actual boot failure.

A cleaner approach: write to a file (`-serial file:/tmp/serial.log`) and have the build infrastructure (gocli or the provision script) explicitly collect that file as a CI artifact on failure, similar to how junit results are collected.

**5. The "unbuffered stderr" claim in the commit message is misleading**

The final commit says: *"Writing to /dev/stderr bypasses this since stderr is unbuffered and captured by the container runtime in real-time."*

QEMU's `-serial file:` backend opens the path with `open()` and writes via its own I/O layer — the "unbuffered" property of C's `stderr` stream (FILE*) doesn't apply. What actually helps is that `/dev/stderr` resolves to the process's fd 2 (a pipe in container context), and pipe writes are immediately available to the reader. The improvement over a regular file is real (no page-cache-only data lost on SIGKILL), but the reasoning in the commit message is wrong.

**6. The EXIT trap → tail -f → stderr evolution reveals an unsolved problem**

The commit history shows three attempts to solve the same problem: gocli SIGKILLs the container, so you can't run cleanup code. The final approach (direct stderr) works around this but doesn't solve the underlying issue. A more robust solution would be to modify **gocli** to collect a serial log file from the container before removing it (via `docker cp` or similar), or to use QEMU's `-chardev file,id=serial0,path=... -serial chardev:serial0` with a `logappend=on` flag on a volume-mounted path that survives container removal.

**7. No isolation between "debugging this specific issue" and "permanent infrastructure changes"**

The PR makes four permanent changes (serial to stderr, timeout reduction, retry reduction) to chase one transient debugging goal (see the boot hang). After the investigation concludes and the version is reverted, these changes remain. The timeout/retry reduction in particular is a risk — it was introduced to fail faster during debugging, but it permanently narrows the safety margin for all future provisioning.

### Summary

The PR conflates debugging scaffolding with permanent infrastructure changes, applies aggressive timeout reductions globally, and modifies centos10 without cause. The serial capture approach works but produces noisy output and is based on an incorrect technical premise. The version bump to a known-broken compose is mixed with changes that should be independently mergeable.

🤖 Assisted by Claude

dhiller:
Proof that latest check-provision-centos-base succeeds: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1767/check-provision-centos-base/2077316723144396800

That commit contained a bump to the recent CentOS 9 Stream version.

cluster-provision/centos9/scripts/vm.sh
```diff
@@ -58,7 +58,7 @@ n="$(printf "%02d" $(( 10#${NODE_NUM} )))"
 cat >/usr/local/bin/ssh.sh <<EOL
 #!/bin/bash
 set -e
-dockerize -wait tcp://192.168.66.1${n}:22 -timeout 300s &>/dev/null
+dockerize -wait tcp://192.168.66.1${n}:22 -timeout 120s &>/dev/null
```
nestoracunablanco, line 61: What is the reasoning behind reducing the timeout to 120s? Do you think it is enough time?

cluster-provision/centos9/scripts/vm.sh
```diff
@@ -211,6 +211,12 @@ if [ "${NUMA}" -gt 1 ]; then
     sriov_pxb_numa_arg=",numa_node=0"
 fi
 
+if [ -n "${PROW_JOB_ID:-}" ] || [ "${CI:-}" = "true" ]; then
+  SERIAL_ARG="-serial file:/dev/stderr"
+else
+  SERIAL_ARG="-serial pty"
+fi
```
nestoracunablanco, line 223: +1

cluster-provision/gocli/cmd/utils/utils.go
```diff
@@ -17,6 +19,17 @@ func AppendUDPIfExplicit(ports nat.PortMap, exposedPort int, flagSet *pflag.Flag
...
+// ForwardEnv returns KEY=VALUE strings for env vars that are set in the current process.
+func ForwardEnv(names ...string) []string {
+	var result []string
+	for _, name := range names {
+		if val, ok := os.LookupEnv(name); ok {
+			result = append(result, fmt.Sprintf("%s=%s", name, val))
+		}
+	}
+	return result
+}
```
nestoracunablanco, line 31: +1

cluster-provision/gocli/cmd/run.go
```diff
@@ -1089,7 +1089,7 @@ func provisionNode(sshClient libssh.Client, n *nodesconfig.NodeLinuxConfig) erro
 func waitForVMToBeUp(cli *client.Client, prefix string, nodeName string) error {
 	var err error
 	// Wait for the VM to be up
-	for x := 0; x < 10; x++ {
+	for x := 0; x < 5; x++ {
```
nestoracunablanco, line 1112: If I am understanding the code correctly, in line 1098 a 1 second waiting time is specified. So we are expecting for the VM to be up in 5 seconds, instead of 10. Do you know how long does it usually take for a VM to start?

nestoracunablanco (commented):
Some questions regarding the new timeouts.

cluster-provision/centos9/scripts/vm.sh
dhiller: Since the code triggering this command is already retrying this (originally 10 times, now 5 times) I'd say 120s should be more than enough. Also it caps the feedback time from 50m (initially) to 10m (now), so it fails faster. But yes, whether that's enough we'll see.

cluster-provision/gocli/cmd/run.go
dhiller: no, since this is exactly the code calling`cluster-provision/centos9/scripts/vm.sh`, which has it's own timeout (now reduced to 120s)

dhiller: apologies: it's calling `ssh.sh` actually, which is generated by `vm.sh`

dhiller:
> Some questions regarding the new timeouts.

@nestoracunablanco thank you for the review, replied

cluster-provision/gocli/cmd/utils/utils.go
```diff
@@ -17,6 +19,17 @@ func AppendUDPIfExplicit(ports nat.PortMap, exposedPort int, flagSet *pflag.Flag
 	return appendIfExplicit(ports, exposedPort, flagSet, flagName, UDPPortOrDie)
 }
 
+// ForwardEnv returns KEY=VALUE strings for env vars that are set in the current process.
+func ForwardEnv(names ...string) []string {
```
nirdothan, line 23: Not critical, but the name is somewhat confusing. It does not really "forward" anything, does it?

nirdothan (approved):
Thank you!

cluster-provision/centos10/scripts/vm.sh
```diff
@@ -58,7 +58,7 @@ n="$(printf "%02d" $(( 10#${NODE_NUM} )))"
 cat >/usr/local/bin/ssh.sh <<EOL
 #!/bin/bash
 set -e
-dockerize -wait tcp://192.168.66.1${n}:22 -timeout 300s &>/dev/null
+dockerize -wait tcp://192.168.66.1${n}:22 -timeout 120s &>/dev/null
```
nirdothan, line 61: What exactly do we gain from this? Fail faster?

dhiller: Yes

EDIT: the timeout reduces the overall waiting in case of ssh.sh not establishing a connection from 10 * 5m to 5 * 2m

cluster-provision/gocli/cmd/utils/utils.go
dhiller: Well, you could say it forwards the env variables from the outer env to the vm.

dhiller:
/override check-provision-k8s-1.34

expected failure

PR #1772, opened 2026-07-15
k8s: Install kernel-modules-extra on CentOS 10 for br_netfilter

### LINKED ISSUE #1769: CentOS 10: k8s provisioning fails due to missing br_netfilter kernel module
## Problem

CentOS 10 Stream k8s provisioning fails because the `br_netfilter` kernel module is not available in the CentOS 10 kernel (`6.12.0-248.el10.x86_64`).

The `k8s_provision.sh` script runs `modprobe br_netfilter` which fails:
```
modprobe: FATAL: Module br_netfilter not found in directory /lib/modules/6.12.0-248.el10.x86_64
```

Since the script runs with `set -ex`, this kills the provisioning. The kernel also logs:
```
bridge: filtering via arp/ip/ip6tables is no longer available by default. Update your scripts to load br_netfilter if you need this.
```

## Affected jobs

All CentOS 10 check-provision lanes fail on #1767:
- [check-provision-k8s-1.34-centos10](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1767/check-provision-k8s-1.34-centos10/2077326522330386432)
- [check-provision-k8s-1.35-centos10](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1767/check-provision-k8s-1.35-centos10/2077326522519130112)
- [check-provision-k8s-1.36-centos10](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1767/check-provision-k8s-1.36-centos10/2077326522745622528)

## Notes

- The VM boots fine — the serial console fix from kubevirtci#1767 confirms this
- This is unrelated to the CentOS 9 `-serial pty` boot hang (project-infra#5098)
- The CentOS 10 kernel `6.12.0-248.el10` may have moved `br_netfilter` out of the default module set, or the kernel package installed during provisioning (`kernel-debug-core`) doesn't include it
- The provisioning script may need to install `kernel-modules-extra` or handle the missing module gracefully

/kind bug
/sig ci

🤖 Assisted by Claude

lyarwood:
## Summary

- The CentOS 10 kernel (`6.12.0-248.el10`) no longer ships `br_netfilter` in the base `kernel-modules` package, causing `modprobe br_netfilter` to fail during provisioning under `set -ex`
- Install `kernel-modules-extra` on CentOS 10 to provide the module, which is required for Kubernetes service routing (`net.bridge.bridge-nf-call-iptables`)
- Applied to all three k8s version directories (1.34, 1.35, 1.36)

Fixes: https://github.com/kubevirt/kubevirtci/issues/1769

## Test plan

- [ ] check-provision-k8s-1.34-centos10 passes
- [ ] check-provision-k8s-1.35-centos10 passes
- [ ] check-provision-k8s-1.36-centos10 passes

### DIFF
--- a/cluster-provision/k8s/1.34/provision.sh
+++ b/cluster-provision/k8s/1.34/provision.sh
@@ -29,6 +29,9 @@ fi
 
 # Install modules of the initrd kernel
 dnf install -y "kernel-modules-$(uname -r)"
+if [ "$release" == "centos10" ]; then
+  dnf install -y "kernel-modules-extra-$(uname -r)"
+fi
 
 # Resize root partition
 dnf install -y cloud-utils-growpart
--- a/cluster-provision/k8s/1.35/provision.sh
+++ b/cluster-provision/k8s/1.35/provision.sh
@@ -29,6 +29,9 @@ fi
 
 # Install modules of the initrd kernel
 dnf install -y "kernel-modules-$(uname -r)"
+if [ "$release" == "centos10" ]; then
+  dnf install -y "kernel-modules-extra-$(uname -r)"
+fi
 
 # Resize root partition
 dnf install -y cloud-utils-growpart
--- a/cluster-provision/k8s/1.36/provision.sh
+++ b/cluster-provision/k8s/1.36/provision.sh
@@ -29,6 +29,9 @@ fi
 
 # Install modules of the initrd kernel.
 dnf install -y "kernel-modules-$(uname -r)" "kernel-devel-$(uname -r)"
+if [ "$release" == "centos10" ]; then
+  dnf install -y "kernel-modules-extra-$(uname -r)"
+fi
 
 # Resize root partition
 dnf install -y cloud-utils-growpart


lyarwood:
Re: gemini-code-assist suggestions on combining `dnf install` calls and using `[[ ]]`:

- The existing script already has ~10 separate `dnf install -y` calls rather than building arrays. Adding one here would be inconsistent with the rest of the file. The dnf metadata is cached after the first invocation so the overhead of a second transaction is minimal.
- `[ "$release" == "centos10" ]` is already used elsewhere in the same file (e.g. the openvswitch conditional at line 93). Switching just this instance to `[[ ]]` would again be inconsistent.

Both changes would be fine as a broader cleanup across the script but don't belong in a targeted bug fix.

dhiller (approved):
/approve

PR #1780, opened 2026-07-20
chore(cluster-up): update default provider to k8s-1.36

dollierp:

Update the default provider to the latest k8s-1.36.


dhiller (approved):
/approve

PR #1794, opened 2026-07-30
fix(cluster-provision): use correct qemu path for s390x

dollierp:

It seems fedora builds of QEMU for s390x don't provide the symlink /usr/bin/qemu-kvm -> /usr/bin/qemu-system-$arch unlike x86_64.

/cc @chandramerla @dhiller


dhiller (commented):
/approve

Thank you @dollierp !

brianmcarey (commented):
/lgtm

PR #1796, opened 2026-07-31
fix: replace hardcoded podman with ${CRI_BIN} in publish.sh

nestoracunablanco:
All push, tag, and manifest operations were hardcoded to use podman, causing failures when the detected or user-specified CRI is docker. Replace every podman call with ${CRI_BIN} so the script consistently uses whichever runtime detect_cri.sh resolves.

Assisted-by: IBM Bob

**Release note**:

```release-note
NONE
```


dollierp (approved):
/approve

PR #1804, opened 2026-08-06
vm based: Support NUMA configuration also for NUMA=1

oshoval:
Update the VM NUMA and PCI bridge setup so it works with NUMA=1,
and set the required numa_node in such case.
It is a private case, where all the resources belong to the single NUMA.
s390x doesn't support NUMA, so exclude it.

In addition, remove the NUMA count setting to allow different consumers 
(such as emulated-igb and sig-network) to set the count according needs.
See 2nd commit for more info.

Fixes #

**Checklist**

**Release note**:

```release-note
None
```

### DIFF
--- a/cluster-provision/centos10/scripts/vm.sh
+++ b/cluster-provision/centos10/scripts/vm.sh
@@ -193,7 +193,8 @@ fi
 numa_arg=""
 sriov_pxb_numa_arg=""
 secondary_nic_pxb_args=""
-if [ "${NUMA}" -gt 1 ]; then
+host_arch="$(uname -m)"
+if [ "${host_arch}" != "s390x" ] && [ "${NUMA}" != 0 ]; then
     numa_mem_unit="${MEMORY//[[:digit:]]/}"
     numa_mem_value="${MEMORY//[!0-9]/}"
     if [ $((CPU % NUMA)) -gt 0 ] || [ $((numa_mem_value % NUMA)) -gt 0 ]; then
@@ -216,7 +217,7 @@ if [ "${NUMA}" -gt 1 ]; then
     done
 fi
 
-if [ "$(uname -m)" == "s390x" ]; then
+if [ "${host_arch}" == "s390x" ]; then
   # As per https://www.qemu.org/docs/master/system/s390x/bootdevices.html#booting-without-bootindex-parameter -drive if=virtio can't be specified with bootindex for s390x
   qemu_system_cmd="qemu-system-s390x \
     -enable-kvm \
--- a/cluster-provision/centos10/scripts/vm.sh
+++ b/cluster-provision/centos10/scripts/vm.sh
@@ -233,7 +233,7 @@ if [ "${host_arch}" == "s390x" ]; then
     -vnc :${n} \
     -cpu host \
     -m ${MEMORY} \
-    -smp ${CPU} ${numa_arg} \
+    -smp ${CPU} \
     -serial pty \
     -machine s390-ccw-virtio,accel=kvm \
     -uuid $(cat /proc/sys/kernel/random/uuid) \
--- a/cluster-provision/centos10/scripts/vm.sh
+++ b/cluster-provision/centos10/scripts/vm.sh
@@ -191,30 +191,27 @@ if [ "$n" = "01" ] ; then
 fi
 
 numa_arg=""
-sriov_pxb_numa_arg=""
-secondary_nic_pxb_args=""
-if [ "${NUMA}" -gt 1 ]; then
-    numa_mem_unit="${MEMORY//[[:digit:]]/}"
-    numa_mem_value="${MEMORY//[!0-9]/}"
-    if [ $((CPU % NUMA)) -gt 0 ] || [ $((numa_mem_value % NUMA)) -gt 0 ]; then
-        echo "unable to calculate symmetric NUMA topology with vCPUs:${CPU} Memory:${MEMORY} NUMA:${NUMA}"
-        exit 1
-    fi
-    node_mem="$((numa_mem_value / NUMA))${numa_mem_unit}"
-    node_first_cpu=0
-    node_cpu_step=$((CPU / NUMA - 1))
-    for node_id in $(seq 0 $((NUMA - 1))); do
-        node_last_cpu=$((node_first_cpu + node_cpu_step))
-        numa_arg+=" -object memory-backend-ram,size=${node_mem},id=m${node_id}"
-        numa_arg+=" -numa node,nodeid=${node_id},memdev=m${node_id},cpus=${node_first_cpu}-${node_last_cpu}"
-        node_first_cpu=$((node_last_cpu + 1))
-    done
-    sriov_pxb_numa_arg=",numa_node=0"
-    for node_id in $(seq 0 $((NUMA - 1))); do
-        # Leave enough room for the downstream buses allocated under each NUMA bridge.
-        secondary_nic_pxb_args+=" -device pxb-pcie,id=secondarypxb${node_id},bus=pcie.0,bus_nr=$((160 + (node_id * 32))),numa_node=${node_id}"
-    done
+sriov_pxb_numa_arg=",numa_node=0"
+numa_mem_unit="${MEMORY//[[:digit:]]/}"
+numa_mem_value="${MEMORY//[!0-9]/}"
+if [ $((CPU % NUMA)) -gt 0 ] || [ $((numa_mem_value % NUMA)) -gt 0 ]; then
+    echo "unable to calculate symmetric NUMA topology with vCPUs:${CPU} Memory:${MEMORY} NUMA:${NUMA}"
+    exit 1
 fi
+node_mem="$((numa_mem_value / NUMA))${numa_mem_unit}"
+node_first_cpu=0
+node_cpu_step=$((CPU / NUMA - 1))
+for node_id in $(seq 0 $((NUMA - 1))); do
+    node_last_cpu=$((node_first_cpu + node_cpu_step))
+    numa_arg+=" -object memory-backend-ram,size=${node_mem},id=m${node_id}"
+    numa_arg+=" -numa node,nodeid=${node_id},memdev=m${node_id},cpus=${node_first_cpu}-${node_last_cpu}"
+    node_first_cpu=$((node_last_cpu + 1))
+done
+secondary_nic_pxb_args=""
+for node_id in $(seq 0 $((NUMA - 1))); do
+    # Leave enough room for the downstream buses allocated under each NUMA bridge.
+    secondary_nic_pxb_args+=" -device pxb-pcie,id=secondarypxb${node_id},bus=pcie.0,bus_nr=$((160 + (node_id * 32))),numa_node=${node_id}"
+done
 
 if [ "$(uname -m)" == "s390x" ]; then
   # As per https://www.qemu.org/docs/master/system/s390x/bootdevices.html#booting-without-bootindex-parameter -drive if=virtio can't be specified with bootindex for s390x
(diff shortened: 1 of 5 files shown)


oshoval:
Hi @dollierp @dhiller 
can you please review (with whitespace disabled)

it just reorder the vars and removes the condition, tested manually e2e

oshoval:
Hi @dollierp 
the cluster-up is supposed to be fixed, or not because of the MAC issue ?
happens here
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirtci/1804/check-provision-k8s-1.34-s390x/2085335419787939840

oshoval:
Hi @brianmcarey  @dasionov 
can you please take a look ?

dollierp:
9 failures are expected with the job check-provision-k8s-1.34.

The s390x should pass successfully.

/test check-provision-k8s-1.34-s390x

dollierp (changes requested):
It seems this patch breaks s390x.

```
qemu-system-s390x: -numa node,nodeid=0,memdev=m0,cpus=0-1: NUMA is not supported by this machine-type
```

Please have a look.

/hold

cluster-up/cluster/k8s-1.35/provider.sh
```diff
@@ -9,7 +9,6 @@ if [ "${KUBEVIRT_WITH_SRIOV}" == "true" ]; then
     if [ "${KUBEVIRT_WITH_CNAO}" != "true" ]; then
         export KUBEVIRT_WITH_MULTUS=true
     fi
-    export KUBEVIRT_NUM_NUMA_NODES=2
```
akalenyu, line 12: isn't this backwards incompatible? shouldn't we keep defaulting to 2 for the sake of someone who relied on this for ages?

oshoval: i will add it instead on test.sh for emulated SR-IOV lane, which is the only one that needs it
and later on we can drop it, once we fold SR-IOV into sig-network lane, which then will need NUMA=1 (which is the default)

what i can do instead that will be safer, is to change it to set it to 2 here, unless it is already explicit set,
will test it out

anyhow it is not something that is dangerous, it just need to be set in the right order

akalenyu: what about developer using this/lanes that don't go through test.sh?

oshoval: for now he won't need to do anything
as i change to 
`export KUBEVIRT_NUM_NUMA_NODES="${KUBEVIRT_NUM_NUMA_NODES:-2}"`

once https://github.com/kubevirt/kubevirt/pull/18595 is merged,
he will simply need to set `export KUBEVIRT_NUM_NUMA_NODES=1`,
but only for a day or so, until we drop this line from kubevirtci

it is a non issue, there are 2 repos, it cant be atomic, but we are the only ones on network that use this lane
and it is for one specific test only

akalenyu: i am saying that some developers doing local work need soemthing like `export KUBEVIRT_NUM_NUMA_NODES="${KUBEVIRT_NUM_NUMA_NODES:-2}"` which doesn't exist in latest revision

oshoval: This will exist from now (gonna add it), 
and as long as https://github.com/kubevirt/kubevirt/pull/18595 isn't merged,
both `export KUBEVIRT_NUM_NUMA_NODES="${KUBEVIRT_NUM_NUMA_NODES:-2}"` and the current 
`export KUBEVIRT_NUM_NUMA_NODES=2` are the same, user won't need to do anything

there is only one test that need it, and i am the sole person that works on that test now, working on folding it into the sig-network lane,
i am doing it in steps, and will try to reduce friction

lets continue offline please if desired ?

oshoval: you can see here that i am checking it meanwhile
https://github.com/kubevirt/kubevirt/pull/18595
(want to make sure something)

and once i see it runs i rework this PR (that PR will stay draft for now)

EDIT
until that PR is merged, will need to add explicit `export KUBEVIRT_NUM_NUMA_NODES=2` on test.sh
because `cluster-up/hack/common.sh` is sourced first
so i think we don't need to change the current removal of this PR,
but just to add the explicit one for now, until https://github.com/kubevirt/kubevirt/pull/18595 is merged,

having to supply the env var when running manually, is not just for this var, also for example for 
`export KUBEVIRT_FUNC_TEST_SUITE_ARGS="${KUBEVIRT_FUNC_TEST_SUITE_ARGS} -emulated-sriov=true"`

oshoval: @akalenyu can you please check https://github.com/kubevirt/kubevirt/pull/18755 ?
again this is temporary, and just because kubevirtci and kubevirt are 2 repos, so PRs aren't atomic

oshoval:
Hi @dollierp can you please take another look ? fixed the s390x
tested it meanwhile - passed

anyhow we will need https://github.com/kubevirt/kubevirt/pull/18755 to be merged before this PR is bumped on kubevirt

cluster-provision/centos10/scripts/vm.sh
```diff
@@ -193,7 +193,7 @@ fi
 numa_arg=""
 sriov_pxb_numa_arg=""
 secondary_nic_pxb_args=""
-if [ "${NUMA}" -gt 1 ]; then
+if [ "$(uname -m)" != "s390x" ]; then
```
dollierp: Would be great to store the arch in a variable and reuse it instead of spawing uname multiple times.

oshoval: done

cluster-provision/centos10/scripts/vm.sh
dollierp: `${numa_arg}` should probably be removed from here as well.

oshoval: its not regression of this PR,
but anyhow done

dollierp (approved):
/lgtm

oshoval:
Thanks Denis
lets keep the hold until https://github.com/kubevirt/kubevirt/pull/18755 is merged

cluster-provision/centos10/scripts/vm.sh
```diff
@@ -193,7 +193,8 @@ fi
 numa_arg=""
 sriov_pxb_numa_arg=""
 secondary_nic_pxb_args=""
-if [ "${NUMA}" -gt 1 ]; then
+host_arch="$(uname -m)"
+if [ "${host_arch}" != "s390x" ]; then
```
EdDev: As we are providing a CI infrastructure for other projects, we may want to be less decisive and more conservative. At least, allow other projects to adjust and keep the NUMA-less option in case they need it.

How about we define `NUMA == 0` as a valid setting?

oshoval: no one needs NUMA == 0,
NUMA = 1 trivially adds all the resources to the sole NUMA, while having the numa_node configuration
i would strongly suggest to not add it, unless someone ever need it, which i doubt

as long as Topology manager and such aren't used (and they weren't, i added them 2 weeks ago),
the NUMA aint affecting,
moreover Kubevirt itself supports for example the cloud-init file only for SR-IOV, and even not for secondary interfaces

I think we should not add logic that isn't used before hand, lets keep it simple please

EdDev: > no one needs NUMA == 0, NUMA = 1 trivially adds all the resources to the sole NUMA, while having the numa_node configuration i would strongly suggest to not add it, unless someone ever need it, which i doubt

Until now "NUMA == 1" meant no NUMA configuration. You changed it.
You **assume** that no one is unterested to check NUMA-less setups. Try to explain why you are so sure that the assumption is valid.

> as long as Topology manager and such aren't used (and they weren't, i added them 2 weeks ago), the NUMA aint affecting, moreover Kubevirt itself supports for example the cloud-init file only for SR-IOV, and even not for secondary interfaces

I do not understand from where the assurance that a project that depends on kubevirtci is not aiming to test topologies that do not include NUMA. This is what existed before with NUMA==1.
The fact that kubevirt/kubevirt is not using it, says nothing about other consumers.

> I think we should not add logic that isn't used before hand, lets keep it simple please

I do not maintain the kubevirtci project, but I think we need to be careful not to break its consumers.
The change here seems right, but if consumers are depending on the previous NUMA-less setup, IMO we need to give them an answer somehow.

oshoval: I am familiar enough with kubevirtci, to say that no one uses it, and if someone does, i will directly solve it
now the default is that all the resources are allocated to a NUMA, thats all
i think that having defensive programming is wrong here

having 0 NUMA, or 1 NUMA is agostic for all the tests, and if not we will simply find out and fix it right away
it is more probable that it won't affect no one than it will
and keeping it simple gives less maintenance requirement from now on

oshoval: cc @dollierp wdyt ?
i am against this change, it is pure defensive programming that has cost without value most of the time,
and we will drag to maintain it forever for no reason

i think we should give it a chance simplified, and be proactive
if we look HW wise, HW with 1 NUMA, is the same logically as hardware without NUMA,
beside the numa_node config that was added

the only reason to add it, is that it is simple change and allow to advance faster,
but otoh it means that we can just be proactive and if someone need, only then i add,
this is healthier in this case imo

oshoval: anyhow added case of 0 to skip setting (as was before)
and fixed that secondary will also be placed on the sole numa if numa == 1

oshoval:
changes
https://github.com/kubevirt/kubevirtci/pull/1804#discussion_r3791679003

oshoval:
image NA suddenly
```
time="2026-08-16T11:50:49Z" level=info msg="Using remote image quay.io/kubevirtci/centos10:2608121207-883440a9"
2026-08-16T11:50:54.292921Z	info	failed to download quay.io/kubevirtci/centos10:2608121207-883440a9: Error response from daemon: no image found in image index for architecture "amd64", variant "", OS "linux"
2026-08-16T11:50:55.820105Z	info	failed to download quay.io/kubevirtci/centos10:2608121207-883440a9: Error response from daemon: no image found in image index for architecture "amd64", variant "", OS "linux"
2026-08-16T11:50:58.407505Z	info	failed to download quay.io/kubevirtci/centos10:2608121207-883440a9: Error response from daemon: no image found in image index for architecture "amd64", variant "", OS "linux"
2026-08-16T11:51:04.907282Z	info	failed to download quay.io/kubevirtci/centos10:2608121207-883440a9: Error response from daemon: no image found in image index for architecture "amd64", variant "", OS "linux"
panic: failed to download quay.io/kubevirtci/centos10:2608121207-883440a9 four times, giving up
```

not due to this PR

https://quay.io/repository/kubevirtci/centos10?tab=tags&tag=2608121207-883440a9

maybe just quay hiccup

EdDev (approved):
Thank you!

oshoval:
Thanks Edy

we will wait for https://github.com/kubevirt/kubevirt/pull/18755 to be merged, hence on hold
@dollierp once you can please take another look ?

dollierp (approved):
/approve

oshoval:
the PR is unhold for now, so no need to rush with retest,
thx

oshoval:
/hold cancel

/test check-provision-k8s-1.37-s390x 
/test check-provision-k8s-1.34-s390x

@dollierp 
we should drop 1.34 lanes, they don't exist on kubevirt right ?

ah anyhow all the failures aren't required

PR #1807, opened 2026-08-10
feat(cluster-provision): add k8s-1.37 provider

dollierp:

This new provider has been created from the existing k8s-1.36 provider.

It uses Kubernetes v1.37.0-beta.0 [1] and CentOS Stream 10 for the nodes.

CRI-O 1.37 stable is not yet available; the prerelease repo at openSUSE Build Service is used in the meantime. Once CRI-O 1.37 is released, a follow-up commit should switch back to the stable repository URL.

[1] https://github.com/kubernetes/kubernetes/releases/tag/v1.37.0-beta.0

/cc @lyarwood @Whitedyl


lyarwood:
@dollierp lgtm but how does this stage of standing up a new provider work, do you add optional 1.37 provision jobs now to exercise this pre-merge or just verify locally?

cluster-provision/k8s/1.37/k8s_provision.sh
```diff
@@ -0,0 +1,395 @@
...
+cgroupDriver: systemd
+failSwapOn: false
+kubeletCgroups: /systemd/system.slice
+EOF
+
+cat <<EOT >/etc/sysconfig/kubelet
+KUBELET_EXTRA_ARGS=--runtime-cgroups=/systemd/system.slice --config-dir=$kubelet_conf_d
+EOT
+
+# Enable userfaultfd for centos9 to support post-copy live migration.
```
lyarwood: nit: this provider runs on CentOS 10 not CentOS 9, so the centos9 qualifier here is inaccurate. The userfaultfd setting is relevant for post-copy live migration regardless of base OS, so something like "Enable userfaultfd to support post-copy live migration." would be more accurate.

cluster-provision/gocli/cmd/provision.go
```diff
@@ -56,8 +56,12 @@ func provisionCluster(cmd *cobra.Command, args []string) (retErr error) {
 	packagePath := args[0]
 	centosVersion := os.Getenv("PROVISION_CENTOS_VERSION")
 	if centosVersion == "" {
-		// default to Centos Stream 9
-		centosVersion = "9"
+		baseBytes, err := os.ReadFile(filepath.Join(packagePath, "base"))
```
lyarwood, line 59: The error from os.ReadFile is returned bare, so a missing "base" file produces a raw OS error with no indication of which provider path was being read. Wrapping would help:

```go
if err != nil {
    return fmt.Errorf("reading base file for provider %q: %w", packagePath, err)
}
```

dollierp:
> @dollierp lgtm but how does this stage of standing up a new provider work, do you add optional 1.37 provision jobs now to exercise this pre-merge or just verify locally?

Thanks for the review!

For now I just tested the new provider locally, I managed to provision a cluster with working SR-IOV.

I opened another PR against the project-infra repo to create the provision jobs for 1.37 which should be optional for now.

To be able to test the new jobs in the project-infra PR, I would like to merge this one first. It should be harmless because there is no jobs using this provider yet.

If there is any concern, I can also try to craft special jobs using mkpj to test both merge requests at once.

lyarwood (commented):
/lgtm

Cool sounds good to me!

Whitedyl (approved):
/lgtm

Ran it locally, provisioned, single node was ready on v1.37.9-beta.0 and all pods were healthy!

dollierp:
CentOS s390x failure is expected because the periodic job for the base image is failing: [periodic-kubevirtci-bump-centos10-base-s390x](https://prow.ci.kubevirt.io/job-history/gs/kubevirt-prow/logs/periodic-kubevirtci-bump-centos10-base-s390x).

/override rehearsal-check-provision-k8s-1.37-s390x

PR #1810, opened 2026-08-11
fix(cluster-provision): update clock-epoch to fix GPG signature verification on s390x

carterpewpew:

The `periodic-kubevirtci-bump-centos10-base-s390x` Prow job fails during VM provisioning with GPG signature verification errors:

```
error: Verifying a signature using certificate 99DB70FAE1D7CE227FB6488205B555B38483C65D
  (CentOS (CentOS Official Signing Key) <security@centos.org>):
  Signature 27a8 created at Thu Jul 30 16:57:51 2026 invalid: signature is not alive
  because: Not live until 2026-07-30T16:52:51Z
Error: GPG check FAILED
```

CentOS Stream 10 uses `rpm-sequoia` for GPG verification, which strictly validates signature timestamps. The "signature is not alive" error occurs when the system clock is behind the signature's "not before" time. CentOS 9 is unaffected because its older RPM does not perform this strict check.

systemd's `clock_apply_epoch` (present in CentOS 10's systemd 257) reads `/usr/lib/clock-epoch` at early boot (PID 1, before any services) and advances the system clock if it is behind the file's mtime. This file does not exist in the cloud image by default, so systemd falls back to the compile-time epoch which can be stale.

This PR adds `--run-command 'touch /usr/lib/clock-epoch'` to the `virt-sysprep` command that customizes the cloud image before QEMU boots. This sets the file's mtime to the current time, ensuring booting with an accurate minimum clock.

Failing run: https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/logs/periodic-kubevirtci-bump-centos10-base-s390x/2084852030840508416

**Tested**:

Tested end-to-end on an x86_64 CentOS Stream 10 QEMU/KVM VM with the hardware RTC set to Jan 2024 (`-rtc base=2024-01-15T12:00:00`).

**Without `/usr/lib/clock-epoch`** (current state): the kernel reads the backdated RTC, and `rpm --checksig` fails with the same error seen in CI:

```
Signature 30b1 created at Thu Jan 23 15:56:52 2025 invalid: signature is not alive
  because: Not live until 2025-01-23T15:51:52Z
```

**With `/usr/lib/clock-epoch`** (this fix): systemd detects the stale clock and advances it at PID 1 init, as shown in the boot journal:

```
kernel: rtc_cmos 00:05: setting system clock to 2024-01-15T12:00:12 UTC (1705320012)
systemd[1]: System time advanced to built-in epoch: Fri 2026-07-31 00:00:00 UTC
systemd[1]: System time advanced to timestamp on /usr/lib/clock-epoch: Tue 2026-08-11 13:10:41 EDT
```

After systemd corrects the clock, `rpm --checksig` returns `digests signatures OK` and `dnf install` succeeds.

Confirmed that CentOS 10's `libsystemd-shared-257-21.el10.so` contains `clock_apply_epoch`, the `/usr/lib/clock-epoch` path, and the associated log messages (`Current system time is before epoch`).

On x86_64, the RTC already provides correct time, so `clock_apply_epoch` sees the clock is already ahead of the epoch and does nothing. The only overhead is one extra `touch` during `virt-sysprep` and one `stat()` at boot.

The fix is applied to both `centos10/scripts/vm.sh` and `centos9/scripts/vm.sh` (they are identical files) for consistency.

This PR was created with AI assistance, but every change has been manually verified and tested.

**Checklist**

- [ ] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough). New features and bug fixes require at least one e2e test
- [ ] Documentation: A [user-guide update](https://github.com/kubevirt/user-guide/) was considered and is present (link) or not required. You want a user-guide update if it's a user-facing feature / API change.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-provision/gocli/cmd/run.go
+++ b/cluster-provision/gocli/cmd/run.go
@@ -1165,6 +1165,14 @@ func waitForVMToBeUp(cli *client.Client, prefix string, nodeName string) error {
 		return fmt.Errorf("could not establish a connection to the node after a generous timeout: %v", err)
 	}
 
+	// Sync VM clock from the container to fix s390x stale TOD clock
+	err = _cmd(cli, nodeContainer(prefix, nodeName),
+		`ssh.sh sudo date -s @$(date +%s)`,
+		"syncing VM clock from host")
+	if err != nil {
+		return err
+	}
+
 	logContainerDiagnostics(cli, prefix, nodeName, "ssh-ok")
 	return nil
 }


dollierp:
> **Without `/usr/lib/clock-epoch`** (current state): the kernel reads the backdated RTC, and `rpm --checksig` fails with the same error seen in CI:

Why is the RTC backdated in the first place?

From QEMU manual:
> By default the RTC is driven by the host system time. This allows using of the RTC as accurate reference clock  inside  the  guest

The host system time of the nodes of the `prow-s390x-workloads` cluster seems accurate so the RTC should also be correct.

carterpewpew:
Thanks for taking a look @dollierp, I investigated this but was unable to determine the exact root cause of the stale clock. Serial console output is not captured in the CI artifacts, so I can't see the kernel's initial clock value at boot.

Regarding the QEMU manual that describes the x86 MC146818 RTC (`-rtc` option). The s390x VMs here use `-machine s390-ccw-virtio`, which doesn't emulate an RTC. Timekeeping on s390x goes through the TOD (Time of Day) clock, which is a separate code path ([QEMU s390x docs](https://www.qemu.org/docs/master/system/target-s390x.html), [TOD device source](https://github.com/qemu/qemu/blob/master/hw/s390x/tod-kvm.c)). The QEMU s390x docs note that "QEMU will not emulate most of the traditional devices" for this machine type.

One data point: CentOS 9 s390x passes on the same infrastructure. CentOS 9 uses the older GnuPG-based RPM GPG backend which does not enforce the "not before" timestamp check, so the stale clock is likely present there too but doesn't cause a failure.

If you have suggestions on how to get better clock diagnostics from the s390x CI environment, I'm happy to dig further.

dollierp:
> I investigated this but was unable to determine the exact root cause of the stale clock. Serial console output is not captured in the CI artifacts, so I can't see the kernel's initial clock value at boot.
> 

For CentOS 9, it seems the clock is set to `Wed Aug  5 12:00:00 AM UTC 2026` at some point during boot.
```
> timedatectl status
               Local time: Tue 2026-08-04 20:00:05 EDT
           Universal time: Wed 2026-08-05 00:00:05 UTC
                 RTC time: n/a
                Time zone: America/New_York (EDT, -0400)
System clock synchronized: no
              NTP service: active
          RTC in local TZ: no
```

For CentOS 10, it seems the clock is set to `Fri Jul 31 12:00:00 AM UTC 2026` at some point during boot.
```
> timedatectl status
               Local time: Thu 2026-07-30 20:00:06 EDT
           Universal time: Fri 2026-07-31 00:00:06 UTC
                 RTC time: n/a
                Time zone: America/New_York (EDT, -0400)
System clock synchronized: no
              NTP service: active
          RTC in local TZ: no
```

I guess it is somehow related to the date the image was built.

> Regarding the QEMU manual that describes the x86 MC146818 RTC (`-rtc` option). The s390x VMs here use `-machine s390-ccw-virtio`, which doesn't emulate an RTC. Timekeeping on s390x goes through the TOD (Time of Day) clock, which is a separate code path ([QEMU s390x docs](https://www.qemu.org/docs/master/system/target-s390x.html), [TOD device source](https://github.com/qemu/qemu/blob/master/hw/s390x/tod-kvm.c)). The QEMU s390x docs note that "QEMU will not emulate most of the traditional devices" for this machine type.
> 

Indeed, I tried to add `-rtc base=utc,clock=host` explicitly to QEMU args and it didn't help for the s390x TOD clock.

> If you have suggestions on how to get better clock diagnostics from the s390x CI environment, I'm happy to dig further.
>

I'm afraid the rehearsal job for CentOS10 on s390x is failing with the same error despite the workaround of this PR.

Please have a look.

carterpewpew:
Thanks for the `timedatectl` data and for confirming the `-rtc` option doesn't help on s390x. You're right that the rehearsal is still failing. The `clock-epoch` approach doesn't work because `virt-sysprep` runs commands inside a libguestfs appliance that also has the stale TOD clock, so `touch` writes the file with a stale mtime. I'm replacing it with a direct clock sync in `provision.go`, after SSH is up but before any `dnf install`:

```
ssh.sh sudo date -s @$(date +%s)
```

The container has the correct host clock, so `$(date +%s)` gives the right epoch. Will push shortly.

cluster-provision/gocli/cmd/provision.go
```diff
@@ -269,6 +269,17 @@ func provisionCluster(cmd *cobra.Command, args []string) (retErr error) {
 		return err
 	}
 
+	// s390x VMs boot with a stale TOD clock (set to the cloud image build
+	// date rather than the current time). This causes rpm-sequoia in CentOS
+	// Stream 10 to reject GPG signatures as "not alive." Sync the VM clock
+	// from the container, which shares the host kernel's correct clock.
+	err = _cmd(cli, nodeContainer(prefix, nodeName),
```
dollierp: Looks like it did the trick.

I think this should be added to the function `waitForVMToBeUp` so that the clock fix is applied for both gocli `provision` and `run` commands.

carterpewpew: Addressed

dollierp (approved):
/approve

PR #1816, opened 2026-08-20
check: verify ip6_tables when IPv6 is enabled

nestoracunablanco:
If the module is not loaded, exit early with a non-zero status.
Without this pre-flight check, the fialure may only be reported
after the image has been downloaded and becomen difficult to
identify among the startup logs.

IPv4-only environments can set IPFAMILY=ipv4, so the ip6_tables
check is skipped when IPv6 is disabled.

Assisted-by: IBM Bob

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-up/check.sh
+++ b/cluster-up/check.sh
@@ -24,6 +24,20 @@ else
 	echo "[ OK ] found /dev/kvm"
 fi
 
+# IPv4-only clusters do not need the IPv6 ip6_tables module.
+if [[ "${IPFAMILY:-dual}" != "ipv4" ]]; then
+	if ! grep -q "^ip6_tables " /proc/modules; then
+		echo "[ERR ] kernel module ip6_tables is not loaded"
+		echo "[ERR ] run: sudo modprobe ip6_tables"
+		echo "[ERR ] see PODMAN.md for how to load required modules persistently"
+		exit 1
+	else
+		echo "[ OK ] kernel module ip6_tables is loaded"
+	fi
+else
+	echo "[ OK ] IPv6 is disabled; skipping ip6_tables check"
+fi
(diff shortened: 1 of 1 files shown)


nestoracunablanco:
@brianmcarey @ormergi what do you think?

ormergi:
@nestoracunablanco thanks for the PR
Could you please elaborate on the reasoning why this change is needed?
What if some enviroments cant support IPv6, are they going to break? (e.g.: single-stack IPv4 test environment).

nestoracunablanco:
@ormergi you are making a good point here. I added a verification for the IPFAMILY environment variable in the check and improved the description.
Taking a look at the code, I believe this is the only way to force IPv4 only, or do you know another methods?
I think `KUBEVIRT_SINGLE_STACK=true` enables a single-stack IPv6 cluster.

ormergi:
@nestoracunablanco thanks for the changes, I still dont get why this is needed. 
Could you elaborate on the use-case? CI env?
I trying to undersntad if this check is a must. The cluster-up workflow is already blotted.

Anyhow I dont have objections for this change.
/lgtm 

@dhiller WDYT?

nestoracunablanco:
@ormergi — The motivation for this change is to catch module-loading failures early in the process. Without this check, failures occur later during execution, requiring developers to dig through logs to identify the issue. By performing this validation before the container is downloaded, developers can immediately see what's wrong at a glance. These kinds of preflight checks significantly improve the developer experience, especially for newcomers to the project.

That said, your point about the cluster-up workflow is valid. However, addressing that concern would require a deeper discussion—one that is important but worth keeping separate from this PR.

Thanks for the lgtm!

PR #1818, opened 2026-08-21
fix(rook): update to version v1.19.10

dollierp:

KubeVirt CI rook deployment is failing to deploy Ceph since the release of version v19.2.6.

This change updates rook deployment to make it compatible with recent versions of Ceph and also pins Ceph to a specific version to prevent such incompatible changes in the future.

/cc @Whitedyl @brianmcarey


brianmcarey (approved):
/approve 

thanks @dollierp

PR #1824, opened 2026-08-25
fix(registry.k8s.io): directly pull images from Google registry without CDN

dollierp:
**What this PR does / why we need it**:

registry.k8s.io CDN switched to Fastly, which is causing CRI-O / Podman to fail to pull some container image blobs.

This change configure them to pull K8S images directly from Google Artifact Registry without going through the CDN, until the CDN-related issue is resolved.

**Special notes for your reviewer**:

/cc @Whitedyl

PR #1825, opened 2026-08-25
fix: pull kubeadm images individually to handle CDN connection failures

nestoracunablanco:
Replaxce `kubeadm config images pull` with a per-image loop using `pull_container_retry` across k8s 1.34-1.37 scripts.

A CDN connection failure on cdn.registry.k8s.io (e.g. HTTP 400 on a Range retry after unexpected EOF - see
https://github.com/kubernetes/registry.k8s.io/issues/333) would abort the entire bulk pull. Pulling each image individually gives every image its own fresh connection with independent retries, preventing a single transient CDN error from failing the whole provisioning run.

Assisted-by: IBM Bob

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster-provision/k8s/1.34/k8s_provision.sh
+++ b/cluster-provision/k8s/1.34/k8s_provision.sh
@@ -122,7 +122,13 @@ if [[ $version != $packages_version ]]; then
    replaceKubeBinaries
 fi
 
-kubeadm config images pull --kubernetes-version ${version}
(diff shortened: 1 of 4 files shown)


dollierp (commented):
Retrying doesn't help much.

Should be actually fixed by https://github.com/kubevirt/kubevirtci/pull/1824 instead.

/close

nestoracunablanco:
Thanks for the info @dollierp

PR #1845, opened 2026-09-16
fix(cri-o): update registries.conf to v2 format

dollierp:

The latest versions of cri-o fail to start because of the registries.conf file created by kubevirtci in the legacy format:

```
validating runtime config: invalid registries: loading registries
configuration "/etc/containers/registries.conf": registries.conf must be
in v2 format but is in v1.
```

This change converts the file from the old version (v1) to the latest version (v2).

/cc @Whitedyl


brianmcarey (commented):
/approve

thanks @dollierp

oshoval (approved):
thx

